PR #207, opened 2013-09-09
Avoid binding on public ip ports

jansabbe:
I was thinking it would maybe be safer to avoid binding the docker images on public ports. Instead let nginx proxy to the private docker ip addresses. This way you avoid people portscanning your host and directly connecting to a running docker image, bypassing any security you might have put up in the nginx configuration.


josegonzalez:
Closing as it's not currently mergeable, but opening an issue so that we can implement it before the next release.

Issue #82, opened 2013-07-07
First time deploys port mappings mess up after reboot


eugeneware:
When I deploy a new application and reboot the docker container restarts on another port so all the nginx mappings fail and the app can't be accessed.

Steps to reproduce:

1. Push a brand new repo to dokku
2. Reboot the machine
3. Try to access the app.

You'll either the a `502 bad gateway` error, or the wrong app (if the port got mapped to another app).

Here's an audit trail after I pushed a new app `nodetest2`. 

After I initially pushed it, here's what the `docker ps` returned:

```
ID                  IMAGE                       COMMAND                CREATED             STATUS              PORTS               SIZE
3c6858a2eab4        app/nodetest2:latest        /bin/bash -c /start    38 seconds ago      Up 23 seconds       49160->5000         16.39 kB (virtual 985.9 MB)
```

Notice that it was listening on port 49160.

And `cat`ing the CONTAINER and PORT entries in the /home/git/nodetest2 did:

```
root@euge:/home/git/nodetest2# cat CONTAINER
3c6858a2eab4
root@euge:/home/git/nodetest2# cat PORT
49160
root@euge:/home/git/nodetest2# cat nginx.conf
upstream nodetest2 { server 127.0.0.1:49160; }
server {
  listen      80;
  server_name nodetest2.euge.co;
  location    / {
    proxy_pass  http://nodetest2;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $http_host;
  }
}
```

Looking at the config.json for the docker container gives:

```
[...]

PR #218, opened 2013-09-18
Issue #82: Hack on hack :)

alessiodm:
There could be possible port conflicts between old dokku deploy and new dokku deploy (after reload) if a container of an app has taken a port already belonging to a different app. So it's better to stop all containers initially, and then proceed with all the deploys.
Finally, I found that calling too many dokku deploy one after the other causes the reload-nginx socket to be closed, so I wait 1s between deploys.
Hope this could be helpful before the final solution to issue #82.

### DIFF
--- a/dokku
+++ b/dokku
@@ -81,10 +81,18 @@ case "$1" in
 
   # temporary hack for https://github.com/progrium/dokku/issues/82
   deploy:all)
+    # Before stop all containers, in order to avoid port conflicts
+    for app in $(ls -d $HOME/*/); do
+       oldid=$(< "$app/CONTAINER")
+       docker kill $oldid > /dev/null
+    done
     for app in $(ls -d $HOME/*/); do
       APP=$(basename $app);
       IMAGE="app/$APP"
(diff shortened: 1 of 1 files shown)


josegonzalez:
@michaelshobbs thoughts on this PR? It seems like a useful addition. Not sure how we could test this though.

michaelshobbs:
Makes sense to me. A couple points of feedback...

I think we should call `dokku release and deploy` like we do in our other restart functions. (Which leads me to the growing importance of a central restart command. Coming soon!)

This will also need master rebased here and linted, of course.

josegonzalez:
What do you mean by `dokku release and deploy`?

michaelshobbs:
i.e. https://github.com/progrium/dokku/blob/2e220da491b2b5ba50171cbe80cac0ad7c8cdd51/plugins/config/commands#L38-L49

josegonzalez:
So lets wait until you get to ps:restart and then take a look at this one then?

michaelshobbs:
Sounds good

PR #296, opened 2013-11-07
Move deploy script into a plugin

leemhenson:
Hey. I've moved guts of the deploy command into a new plugin, ```01_dokku-deploy```. In doing this, it paves the way for other plugins to customise the deploy process a bit more. In my particular case, supporting the Shoreman plugin for greater process control:

https://github.com/musicglue/dokku-shoreman/blob/master/run-deploy

### DIFF
--- a/dokku
+++ b/dokku
@@ -46,21 +46,8 @@ case "$1" in
   deploy)
     APP="$2"; IMAGE="$3"
     pluginhook pre-deploy $APP $IMAGE
-
-    # kill the app when running
-    if [[ -f "$DOKKU_ROOT/$APP/PORT" ]]; then
-      oldid=$(< "$DOKKU_ROOT/$APP/CONTAINER")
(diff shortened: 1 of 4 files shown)


josegonzalez:
This isn't mergeable, though the `deploy` command is looking long in the tooth. @michaelshobbs Should we pull it out into a plugin, or just close this PR out? I'm fine either way.

michaelshobbs:
We've already started this move to plugins since the dockerfile support. Might as well finish the job. Currently things we getting moved into standard. Not sure we need an actual deploy plugin. Especially if we're going to rework plugins to require a file per command. In that case, its just an organizational choice.

josegonzalez:
Closing then. If we need further work, it'll be done in a separate pr. Thanks @leemhenson for the initial PR! Sorry for the long wait on a response :(

PR #298, opened 2013-11-07
dokku ps

slaskis:
dokku-ps
========

A process management plugin for [dokku](https://github.com/progrium/dokku).

Modeled after the process/dyno management of heroku.

Written to solve these problems:

1. Restarting crashed containers.

  Because sometimes the process exits because of something like a memory leak and, until we have it fixed, we just want it to keep going.

  And the idea is to use upstart to handle the containers, as described in [dockers host integration docs](http://docs.docker.io/en/latest/use/host_integration/).

2. Being able to start up non `web` processes.

Bonus problems to solve:

1. Monitor & notify.

  I'm guessing since upstart is monitoring the respawning of containers it should be able to notify me by email, irc or something even cooler using some custom events.

2. Process scaling.

  If we are able to add each process as a backend to the nginx config we should easily be able to load balance the requests over multiple processes. [This PR](https://github.com/progrium/dokku/pull/267) seems like it would help a lot as we would only have to generate `$APP_ROOT/nginx.conf` with the processes defined in `$APP_ROOT/PS`. Something like:

```
    CONF="$APP_ROOT/nginx.conf"
    echo "upstream $APP {" > $CONF
    grep "^web=" $APP_ROOT/PS | while read line; do
      NUM=${line#*=}
      for ((i=1; i<=$NUM; i++)); do
        PORT=$(docker port $APP.web.$i 5000 | sed 's/0.0.0.0://')
        echo "server 127.0.0.1:$PORT;" >> $CONF
      done
    done
    echo "}" >> $CONF
```

## Status

- [x] Install plugin
- [x] Deploy app
- [x] Restart a process `dokku ps:restart APP web`
- [x] Restart app `dokku ps:restart APP`
- [x] Scale processes `dokku ps:scale APP web=4 worker=1`
- [x] Stop a process (such as "web") `dokku ps:stop APP web`
- [x] Load balancing (of "web" processes only?)
- [ ] Logs (is it possible to `cat` multiple streams into one?)
- [ ] Cleanup and format the output/noise
- [ ] Tests

Thoughts?


josegonzalez:
I think we're not going to implement this as is, and instead provide a pluginhook. For instance, I can foresee a plugin that scales AWS instances with dokku containers, or one that modifies the number of procs running inside of a supervisor controlled container.

If anyone has any comments on that, feel free to pile them on, but doing anything more inside of dokku feels a bit weird (because there are so many corner cases).

josegonzalez:
We have an official `ps` plugin as of #897, and the items not implemented here are I think out of scope. Feel free to open a new PR to implement some of these in the new plugin as you see fit.

Issue #272, opened 2013-10-28
Running vagrant tests time out


slaskis:
I've been trying to run `tests/run_vagrant` and it's builds and starts beautifully but while running the actual tests it tends to time out. At first I thought I had missed some DNS configuration but after some probing from @asm89 on IRC I found that when it gets to this line in `./test_deploy` it seems like `dokku url` never exits:

```
URL=$(ssh dokku@$TARGET url $REPO)$FORWARDED_PORT
```

And I can confirm that when running it separately either like this (without an app name):

```
ssh dokku@dokku.me url
```

where with ssh -v the last line is:

```
debug1: Sending command: url
```

or like this (with a valid app name):

```
ssh dokku@dokku.me url test-nodejs-express-22080
```

it will never exit, or time out after like 5 minutes or something. The difference is that the one with a valid app name outputs the url as expected.

But running it like this (with an non-existent app-name) it exists properly with an error:

```
ssh dokku@dokku.me url foo
```

I'm thinking maybe it's really and issue in pluginhook and related to the comments in #236 but since it's easy to reproduce I figured it can be a separate issue.

PR #312, opened 2013-11-16
Plugin structure reorganization.

### LINKED ISSUE #236: Allow dokku run to support interactive commands
Currently the dokku command supports 'run' to allow executing a command within the docker container of an app; this is useful for things like Django's manage.py syncdb.... but it doesn't allow any interactivity, whereas Heroku allows interactive commands in its equivalent. This is useful for responding to prompts from a command, or launching a shell within the container.

plietar:
Was "Fix pluginhook problems by spliting commands into seperate hooks."
Fixes #236 and #272.

This PR used to be only meant to fix problems with pluginhook.
In the end, It is a quite deep change in how plugins are organized.

All commands implemented by plugins are now prefixed by the name of the plugin.
`00_standard-dokku` was renamed to have a simpler name, `core`.
So for example the existing `url` command is now called `core:url`.

Because it added more trouble with pty issues than it solved, pluginhook is not used anymore for commands, onyl for hooks.
Instead, each command is implemented in a seperate file, and is called directly, w/o using pluginhook.

The commands map as following :
* "Root commands", in the form `<plugin>` map to `$PLUGIN_PATH/<plugin>/command`
* Sub-commands, in the form `<plugin>:<name>` map to `$PLUGIN_PATH/<plugin>/commands/<name`

In order to avoid the need to always type in the full command name, command aliases have been implemented.
These are also required to support commands with fixed names, such as `git-*` commands.
Aliases also allow to override default commands, by creating or modifying an alias. Fixes #296

Aliases are defined in the `$DOKKU_ROOT/ALIAS` file in the simple

    <name>=<value>

A default set of aliases is included with dokku, and installed by `make install`. An `addons` plugin provides commands to manage them.

Additionally, when started with no parameters, dokku runs the `_` command. This can be aliased to any command.
See below about the shell command which was added as the default `_` alias.

The help command was also modified.
Instead of having each plugin define a help hook or command, every user-facing command's script includes the help message in it's header. See existing commands for the format.
The `core:help` command (aliased to `help`) parses this format, looks for a short alias in the `ALIAS` file, and pretty prints the message.

I also added a `dokku shell` command. Initially it was just a test for interactive commands, but I find it quite nice, so I've included it with two other tiny commands `ls` and `docker` which make working in the shell easier.
The shell should also make it easier to run tests with `expect` (Although I must admit I having trouble with it)
Basically the shell offers a command prompt, one can type in commands, and they will be run as `dokku <command>`.

It supports command history, and that's about it. I tried adding basic completion, but there doesn't seem to be an easy way which doesn't involve re-writing the whole readline in bash. Maybe this can be done in a improved shell plugin, written in eg python ruby, overriding the default one.

By default, the `_` is aliased to `shell`, so plain `dokku`, or `ssh dokku@dokku.me` starts the shell.


michaelshobbs:
@josegonzalez not sure this is necessary anymore. we've restructured the namespace through other PRs. i'm not a fan of supporting the alias bit. too much complexity for the slight benefit. The dokku shell seems interesting. I'm down to pull that in if you agree.

josegonzalez:
Pulling in the shell is fine by me.

Issue #457, opened 2014-02-18
Move all commands in dokku script to a plugin


stuartpb:
Is there anything wrong with putting the rest of Dokku's commands into plugins (like 00_dokku-standard kind-of does for some commands now)? For one thing, this would allow plugins to wrap the standard commands, allowing for things like multi-tenancy (a rudimentary "ownership" plugin could abort any command that operates on an an existing app with an "owner" different from the one connecting to the server).

PR #458, opened 2014-02-19
Move core into granular plugins

stuartpb:
This is my proposed implementation for moving Dokku's commands into plugins as described in #457.

While I moved most of the commands from the root `dokku` script and 00_dokku-standard out to their own plugins (and added individual documentation for each one), I moved the "version" command from 00_dokku-standard *into* the root script, as it seems wise to make that as plugin-independent as the "help" command (which I also left in the root script).

With this new layout, plugins can be placed in front of (almost) any command in Dokku, and individual commands can be overridden, replaced, or disabled entirely. Some things this enables:

- A plugin that requires "administrator access" before running certain commands, which can be enforced by moving the corresponding plugins into a directory within the access-control plugin.
- Similarly, a plugin that enforces ownership of an app by denying users who did not create the app from pushing / performing commands on it.
- By replacing the core "deploy" command with one that pushes the built container to a test / production cluster, Dokku could be used as the basis for a build / integration / deployment server.

Since execution order never really matters for any of the core plugins relative to each other, I'm ambivalent about the "XX_" prefixes. I mostly kept them to follow the precedent set by "00_dokku-standard" and other projects that use similar conventions like GRUB.


josegonzalez:
Succeeded by #592

PR #557, opened 2014-04-24
Expose less ports from the host

psm14:
Nginx running on the host is capable of forwarding directly into containers without needing port 5000 to be exposed from each one.

Say "SiteA" and "SiteB" are running in containers.  Currently, in master, port 5000 on both of them would be mapped to a port in the 49000-49900 range.  Let's say SiteA gets 49001 and SiteB gets 49002, for example.  Someone could connect to "http://sitea.example.com:49002/" to be connected to SiteB's container, and likewise "http://siteb.example.com:49001/" goes to SiteA.  In addition to being unwieldy, someone could enumerate every standard dokku container running on the machine by simply scanning the 49000-49900 range, which may or may not have security/architecture/whatever implications.

This PR changes this by configuring nginx to forward directly to the container's private 172.17.%.% address.  Unfortunately, it causes any of the nginx/vhost related plugins to break since they currently assume they can forward to "127.0.0.1:$PORT".  This is an easy fix; the changes made to the post-deploy hook of the default nginx-vhosts plugin in this PR can be used as an example.

### DIFF
--- a/dokku
+++ b/dokku
@@ -65,13 +65,15 @@ case "$1" in
 
     # start the app
     DOCKER_ARGS=$(: | pluginhook docker-args $APP)
-    id=$(docker run -d -p 5000 -e PORT=5000 $DOCKER_ARGS $IMAGE /bin/bash -c "/start web")
-    echo $id > "$DOKKU_ROOT/$APP/CONTAINER"
-    port=$(docker port $id 5000 | sed 's/0.0.0.0://')
+    port=5000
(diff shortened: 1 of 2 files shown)


josegonzalez:
@michaelshobbs Is this something you think you can re-implement, now that we have the fancy new nginx-vhosts plugin? It seems like a reasonable request.

josegonzalez:
Note, this is a dupe of #207. I'm going to close both and open an issue :)

PR #587, opened 2014-05-23
Default DOKKU_ROOT to ~dokku if unspecified

cconstantine:
Our system puts user's home directory in a non-standard place.  Instead of being forced to set the DOKKU_ROOT to this new place, default to trying to get the dokku user's home directory via ~dokku.

Depends on this pull request in sshcommand: https://github.com/progrium/sshcommand/pull/10


josegonzalez:
Closing as #861 is open (and progrium/sshcommand#10 has been merged). Thanks for the original PR!

PR #592, opened 2014-05-28
Refactor core into plugins with command and subcommand scripts

stuartpb:
This change makes the mapping of plugins to commands in Dokku significantly more consistent, giving each existing Dokku command its own plugin, and each plugin a dedicated location for a provided command and/or subcommands.

Beyond the navigational advantages, this allows individual commands / components of Dokku to be disabled or overridden at the plugin level, without having to muck around patching scripts that provide unrelated commands.

## Commands and subcommands

"command" scripts, like a hook, are executed when the name of a plugin is
called: for instance, "dokku build" calls the "build" plugin's "command"
script (plugins/build/command).

"subcommand" scripts are called for subcommands of plugins, as namespaced with
a colon: for example, "dokku backup:export" calls the "export" subcommand
script of the "backup" plugin (plugins/backup/subcommands/export).

Plugins that do not use their name for commands/subcommands (such as the git plugin, or plugins that wrap other commands) can continue to use the "commands" hook, which remains fully supported and backward-compatible.

## Updates to documentation

Each plugin / command gets a README describing its purpose and implementation.

Additionally, the "help" text describing its commands' usage is moved out to a separate "help.txt" file, which is then read and (potentially) columnized when running the "help" command.

This commit also removes the "warning" in the plugins directory - Dokku's stability, hooks and all, is
covered by the main Dokku README.

## Miscellany

This commit also removes the nginx-vhosts plugin's functionally-empty "commands" hook script, for clarity and cleanliness.

(This is effectively a resubmission of #458, as it was remade after taking #312 into account, squashed and rebased onto the current master.)


michaelshobbs:
@josegonzalez for whatever reason this seems cleaner I like the idea of moving commands into plugins but I'm not completely sold on the premise to restructure subcommands. Thoughts?

stuartpb:
For what it's worth, this subcommand structure is how [Plushu](https://github.com/plushu/plushu) has been working since the beginning, and it's worked just fine over the course of six months and ~50 plugins.

michaelshobbs:
@stuartpb the intention of my comment was not meant to judge the functionality of this structure. Instead, I'd like to understand the reasoning behind it. 

dokku currently has all commands tucked away in a single commands file per plugin. To me that seems, very neat, concise, and makes it dead simple to execute against.

i.e.
```
for script in $PLUGIN_PATH/*/commands; do
  set +e; $script "$@" ; exit_code=$? ; set -e
  if [ "$exit_code" -eq "$DOKKU_NOT_IMPLEMENTED_EXIT" ]; then
    continue
  fi

  implemented=1
  if [ "$exit_code" -ne "$DOKKU_VALID_EXIT" ]; then
    exit $exit_code
  fi
done
```

Rather than
```
if [ -z $DOKKU_NO_COMMAND ]; then
  if [ -x $PLUGIN_PATH/$1/command ]; then
    "$PLUGIN_PATH/$1/command" "$@"
  elif [ -x $PLUGIN_PATH/${1%%:*}/subcommands/${1#*:} ]; then
    "$PLUGIN_PATH/${1%%:*}/subcommands/${1#*:}" "$@"
  fi
fi
```

Since I lack the understanding of the additional benefits, its seems to just be personal preference and the former is easier to understand for me than the latter. 

Can you clarify why the subcommand pattern (i.e. `$PLUGIN_PATH/*/subcommands/$command`) is versus the current `commands` file pattern?

stuartpb:
The big win that I've seen in Plushu is that there's way less non-deterministic boilerplate in each plugin. On top of making it *way easier* to spin up new plugins, when you don't have *every plugin* sticking its thumbs in the command parser, there's way less opportunity for a bug in a one-off plugin to break every other plugin in the system.

michaelshobbs:
Can you give an example?

stuartpb:
An example of one plugin needlessly breaking command handling for every other plugin? Say `plugins/a-lousy-plugin/commands` has a bug that causes it to sporadically exit early, with a status code that is not `$DOKKU_NOT_IMPLEMENTED_EXIT`. Every time that happens, *every other plugin's commands* are skipped.

stuartpb:
Or, say that its name starts with `w`, and it makes this error *every time*. Now, plugins with names that sort alphabetically lower than `w` (anything starting with `x`, `y`, or `z`) will never be reached - but you never realize it during normal operation, until the day you install a plugin like `zero-conf`.

michaelshobbs:
Ah that completely makes sense and this implementation bypasses that because you are executing the plugin command directly instead of iterating and executing every plugin's commands file every time you attempt to run any command. Got it. Thanks!

josegonzalez:
@michaelshobbs thoughts on this PR, or ideas we can/should pull from it?

michaelshobbs:
Yeah I think we can/should use this pattern. I can take a swing at it. Let's leave this open for now.

josegonzalez:
Closing as #1039 is open. Thanks for your initial work @stuartpb !

PR #653, opened 2014-08-12
Added restart command.

cameron-martin:
This does a release, then a deploy. This pattern is used a few times throughout the code, and it is helpful to have an explicit command for it so that users can run it.


iamale:
+1

PR #702, opened 2014-10-02
Allows config:set to set parameters values with spaces

### LINKED ISSUE #482: Can't use config:set to set parameters values with spaces
There does not seem to be any combination of escaping which works with v0.2.2 for using config:set with a value with spaces, ex:

```
sudo dokku config:set myapp JAVA_OPTS="-Xmx1024m -Xms256m"
sudo dokku config:set myapp "JAVA_OPTS=-Xmx1024m -Xms256m"
sudo dokku config:set myapp 'JAVA_OPTS="-Xmx1024m -Xms256m"'
```

etc.

There's a bunch of logic in plugins/config/commands which uses whitespace or space characters as break points

jazzzz:
This fixes #482, and there is a test.

### DIFF
--- a/dokku
+++ b/dokku
@@ -12,6 +12,12 @@ if [[ $(id -un) != "dokku" && $1 != "plugins-install" ]]; then
   exit
 fi
 
+if [[ -n "$SSH_ORIGINAL_COMMAND" ]]; then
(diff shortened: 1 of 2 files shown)


kilianc:
+1 on this

PR #829, opened 2014-12-18
expose the port for `dokku run $APP /bin/bash`

MatthewMueller:
Super useful for debugging. So if you run `docker ps`:

```
CONTAINER ID        IMAGE                         COMMAND                CREATED             STATUS              PORTS                      NAMES
b55e977cbcd0        dokku/$APP:latest          "/exec /bin/bash"      6 seconds ago       Up 5 seconds        0.0.0.0:49176->5000/tcp    berserk_stallman
```

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -49,7 +49,7 @@ case "$1" in
     shift 2
 
     DOCKER_ARGS=$(: | pluginhook docker-args $APP)
-    docker run -i -t $DOCKER_ARGS $IMAGE /exec "$@"
+    docker run -i -t -p 5000 -e PORT=5000 $DOCKER_ARGS $IMAGE /exec "$@"
     ;;
 
   url | urls)


josegonzalez:
@MatthewMueller We can pass a second argument to the `docker-args` pluginhook with the command being run, which would allow you to do this. I'll open a PR for that.

josegonzalez:
@MatthewMueller Closing as there is an alternative PR that should allow you to easily build a plugin to do what you need to do, without potentially exposing this configuration to in production.

MatthewMueller:
much better! thanks!

MatthewMueller:
@josegonzalez can you provide a usage example?

josegonzalez:
Create a plugin with a `docker-args` command and have it output your args.

```shell
#!/usr/bin/env bash

STDIN=$(cat); APP="$1"; CMD="$2"
EXPOSE_PORT=$(dokku config:get $APP EXPOSE_PORT);

if [ "$CMD" == "run" ] && [ "$EXPOSE_PORT" == "1" ]; then 
  output="-p 5000 -e PORT=5000"
fi

echo "$STDIN $output "
```

PR #858, opened 2015-01-01
Disable server tokens in nginx. Closes #857

### LINKED ISSUE #857: VHOST Disable Nginx Server version information
Currently this is what we get on the response header
```
Status Code: 200
date: Thu, 01 Jan 2015 14:20:52 GMT
content-encoding: gzip
server: nginx/1.6.2
.....
```
Can we just have server: nginx without the version information as the default, without need to create nginx.conf.template on all applications. Am sure this is possible by setting
```
https://github.com/progrium/dokku/blob/master/plugins/nginx-vhosts/templates/nginx.conf
https://github.com/progrium/dokku/blob/master/plugins/nginx-vhosts/templates/nginx.ssl.conf
```
to have this value
```
server_tokens off;
```

josegonzalez:
Not merging since there are no tests. @michaelshobbs, do you mind adding tests for this? I'll use that as a guide going forward :)

closes #857

### DIFF
--- a/plugins/nginx-vhosts/install
+++ b/plugins/nginx-vhosts/install
@@ -44,6 +44,7 @@ events {
 }
 
 http {
+  server_tokens off;
   sendfile on;
   tcp_nopush on;
   tcp_nodelay on;
--- a/plugins/nginx-vhosts/templates/nginx.conf
+++ b/plugins/nginx-vhosts/templates/nginx.conf
@@ -2,7 +2,6 @@ server {
   listen      [::]:80;
   listen      80;
   server_name $NOSSL_SERVER_NAME;
-  server_tokens off;
   location    / {
     proxy_pass  http://$APP;
     proxy_http_version 1.1;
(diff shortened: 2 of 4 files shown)


michaelshobbs:
Sure thing. 

A point of clarification: This will need to be a deployed "unit" test similar to `nginx:build-config` as we'll need nginx running to validate. Point being we have two helpers for test app creation (`create_app` and `deploy_app`). The former just does a `dokku apps:create` and the latter deploys one of the test apps. Obviously the latter takes longer and should only be used when necessary.

michaelshobbs:
Thoughts on adding this to the top-level nginx config somewhere around [here](https://github.com/progrium/dokku/blob/8535900f9d256654ebf403a200b5c54547b1dc01/plugins/nginx-vhosts/install#L47) instead?

josegonzalez:
Oh thats fine with me. I can change the PR to do that

PR #864, opened 2015-01-03
bind docker container to internal port if using vhosts

### LINKED ISSUE #860: Bind nginx requests directly to docker internal IP
This would help avoid cases where users may not being using a firewall to block random ports *and* they are using VHOST-based deploys. Apps should only be routed to via the private docker ip address, not a public docker port.

Refs #207.
Refs #557.

@michaelshobbs I think you might be more familiar with how this works, so if you have time, would you mind taking a look at those PRs and modifying it to work in the new setup?

michaelshobbs:
Per IRC conversations, this was implemented as a (very simple) pluginhook. Includes tests.

close #860 
refs #207
refs #557

### DIFF
--- a/plugins/nginx-vhosts/bind-external-ip
+++ b/plugins/nginx-vhosts/bind-external-ip
@@ -2,8 +2,23 @@
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 APP="$1"
-
 set +e; NO_VHOST=$(dokku config:get $APP NO_VHOST); set -e
+
+RE_IPV4="([0-9]{1,3}[\.]){3}[0-9]{1,3}"
+
+RE_IPV6="([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|"                    # TEST: 1:2:3:4:5:6:7:8
+RE_IPV6="${RE_IPV6}([0-9a-fA-F]{1,4}:){1,7}:|"                         # TEST: 1::                              1:2:3:4:5:6:7::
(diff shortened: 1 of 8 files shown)


plugins/nginx-vhosts/bind-external-ip
```diff
@@ -0,0 +1,21 @@
...
+set +e; NO_VHOST=$(dokku config:get $APP NO_VHOST); set -e
+if [[ -f "$DOKKU_ROOT/VHOST" ]];then
+  GLOBAL_VHOST=$(< "$DOKKU_ROOT/VHOST")
+else
+  GLOBAL_VHOST=$(< "$DOKKU_ROOT/HOSTNAME")
+fi
+
+if [[ -n "$NO_VHOST" ]]; then
+  echo true # bind to external ip. VHOST is disabled for this app
+elif [[ "$GLOBAL_VHOST" =~ ([0-9]{1,3}[\.]){3}[0-9]{1,3} ]];then
```
josegonzalez: IPv6?

michaelshobbs:
Crap! I did not take that into account in this or the previous implementation. I'll take a look in the morning.

michaelshobbs:
Snagged an IPv6 regex from SO (http://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses) and put it into the plugin and the domains one as well.

josegonzalez:
Build failed. Womp.

michaelshobbs:
Docker is having issues according to their rss

josegonzalez:
Can you add some docs to this and then we can merge?

michaelshobbs:
Sure thing. Preference on the best place it should live? New md file or existing?

michaelshobbs:
Probably belongs in nginx, no?

josegonzalez:
In nginx probably. Feel free to reorder stuff as necessary.

josegonzalez:
Nice work. Merging and shipping a new release :)

PR #876, opened 2015-01-07
give CACHE_PATH env variable for forward compatibility with herokuish

vincentfretin:
herokuish defaults to /tmp/cache, so the build cache don't work without giving the CACHE_PATH environment variable. You can test it with buildstep herokuish branch. cc @progrium

### DIFF
--- a/dokku
+++ b/dokku
@@ -54,7 +54,7 @@ case "$1" in
     docker commit $id $IMAGE > /dev/null
     [[ -d $CACHE_DIR ]] || mkdir $CACHE_DIR
     pluginhook pre-build $APP
-    id=$(docker run -d -v $CACHE_DIR:/cache $IMAGE /build/builder)
+    id=$(docker run -d -v $CACHE_DIR:/cache -e CACHE_PATH=/cache $IMAGE /build/builder)
     docker attach $id
     test "$(docker wait $id)" -eq 0
     docker commit $id $IMAGE > /dev/null


michaelshobbs:
Perhaps this can go into the build-env plugin as a default env var?
Somewhere like here: https://github.com/progrium/dokku/blob/master/plugins/build-env/pre-build#L10

josegonzalez:
Why would it need to be in build-env ?

michaelshobbs:
This PR is just adding `-e CACHE_DIR=/cache` to the docker args of the build container. Figured it'd be cleaner to have all build environment variables in the same place.

josegonzalez:
Ah makes sense. Seems like it should be in the build-env bit then.

josegonzalez:
@vincentfretin mind updating your pr to put it in the right place? :)

vincentfretin:
It doesn't work if I put CACHE_PATH in build-env like CURL_TIMEOUT.
The /app/.env file is read in the main function via buildpack-build which calls buildpack-setup function. The CACHE_PATH is needed before the call to the main function, here: https://github.com/gliderlabs/herokuish/blob/master/include/herokuish.bash#L11

michaelshobbs:
Aw bummer. Welp I guess that settles that. :smile:

josegonzalez:
@michaelshobbs is there a test that might be added for this? I can't think of one, so feel free to merge unless you can.

michaelshobbs:
Doesn't look like there is a good place at this point.

PR #877, opened 2015-01-07
add MH to AUTHORS

michaelshobbs:
not sure if AUTHORS is kept up-to-date anymore but the instructions say I should add myself ;)


josegonzalez:
I like how CI failed on this.

michaelshobbs:
Wercker FTW!

michaelshobbs:
seems like the CircleCI integration isn't happening and wercker still is? @progrium thoughts?

josegonzalez:
He hasn't removed it from wercker yet so meh. We can merge.

PR #880, opened 2015-01-08
disable VHOST deployment if global VHOST file is missing and an app domain has not been added

### LINKED ISSUE #879: Wrong app URL when no VHOST file exists
**Problem:**

If you delete the VHOST file because you want to use port-based deployment (i.e. you're installing dokku in a VM), I expect the application to be deployed at http://`hostname`:`port` as per documentation.
But the current release deploys the app on http://`app-name`.`hostname`, ignoring the fact that no VHOST is set.

**Steps to reproduce:**

1) Install dokku
2) Remove /home/dokku/VHOST
3) Deploy sample node-js-app
4) In the git push output you'll see

    =====> Application deployed:
    http://node-js-app.<hostname>

**Expected outcome:**

    =====> Application deployed:
    http://<hostname>:<port>

**Possible fix:**

`dokku url node-js-app` returns the wrong hostname although the logic is correct.
The issue is in `dokku nginx:build-config node-js-app` as the logic to deploy on a port is executed only when the NO_VHOST variable is set (https://github.com/progrium/dokku/blob/master/plugins/nginx-vhosts/commands#L32)

Setting it manually works as expected:

    root@combotest:~# env NO_VHOST=1 dokku nginx:build-config node-js-app
    -----> NO_VHOST set, deleting node-js-app/VHOST
    -----> NO_VHOST set, deleting nginx.conf
    -----> NO_VHOST set, reloading nginx after nginx.conf deletion

I suspect this NO_VHOST variable should be set after deploy before calling nginx:build-config, but it isn't.

If you can point out which is the best place to set this variable, I can do a pull request myself.

michaelshobbs:
closes #879

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -141,7 +141,7 @@ dokku domains:remove myapp example.com
 
 > New as of 0.3.13
 
-The deployed docker container running your app's web process will bind to either the internal docker network interface (i.e. `docker inspect --format '{{ .NetworkSettings.IPAddress }}' $CONTAINER_ID`) or an external interface (i.e. 0.0.0.0) depending on dokku's VHOST configuration. Dokku will attempt to bind to the internal docker network interface unless you specifically set NO_VHOST for the given app or your dokku installation is not setup to use VHOSTS (i.e. $DOKKU_ROOT/VHOST or $DOKKU_ROOT/HOSTNAME is set to an IPv4 or IPv6 address)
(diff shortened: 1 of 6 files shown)


josegonzalez:
If a user manually set a domain via the domains plugin, do we want to disable vhosts then?

michaelshobbs:
Fair point. If the answer is no, then this whole feature need some refactoring I think...

josegonzalez:
If a user uses the domains plugin, that should mean that the app has a vhost. Is that the case with this plugin?

michaelshobbs:
I have something that I think works under that assumption. Writing tests now.

michaelshobbs:
So this update will use a port based deployment if there is no global VHOST and no app vhost but is less aggressive when turning off VHOST support. So you can add a domain even if you have no global VHOST

PR #884, opened 2015-01-09
IP and PORT are likely to get clobbered. rename them

michaelshobbs:
per our conversation on IRC. PORT was getting clobbered and its too generic IMO. Also renamed IP.


jomo:
woo thanks :beers:

PR #887, opened 2015-01-11
Debian packaging

josegonzalez:
This PR creates allows us to create debian packages for dokku via `make deb-all`. You can also use `make install-from-deb` to install debian packages from our custom packagecloud repository.

Note: This feature is in beta for now. If it eats your homework, sets your cat on fire, deletes your apps on your servers, etc., do not blame us. We want to test this thoroughly before promoting this as the way to install dokku going forward.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ endif
 .PHONY: all install copyfiles version plugins dependencies sshcommand pluginhook docker aufs stack count dokku-installer vagrant-acl-add vagrant-dokku
 
 include tests.mk
+include deb.mk
 
 all:
 	# Type "make install" to install.
--- a/Makefile
+++ b/Makefile
@@ -24,13 +24,16 @@ all:
 
 install: dependencies stack copyfiles plugin-dependencies plugins version
 
-release:
-	$(MAKE) deb-all
+release: deb-all package_cloud packer
+
+package_cloud:
 	package_cloud push dokku/dokku/ubuntu/trusty buildstep*.deb
 	package_cloud push dokku/dokku/ubuntu/trusty sshcommand*.deb
 	package_cloud push dokku/dokku/ubuntu/trusty pluginhook*.deb
 	package_cloud push dokku/dokku/ubuntu/trusty rubygem*.deb
 	package_cloud push dokku/dokku/ubuntu/trusty dokku*.deb
+
+packer:
 	packer build contrib/packer.json
 
 copyfiles:
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,15 @@ all:
 
 install: dependencies stack copyfiles plugin-dependencies plugins version
 
+release:
+	$(MAKE) deb-all
+	package_cloud push dokku/dokku/ubuntu/trusty buildstep*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty sshcommand*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty pluginhook*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty rubygem*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty dokku*.deb
+	packer build contrib/packer.json
+
 copyfiles:
 	cp dokku /usr/local/bin/dokku
 	mkdir -p ${PLUGINS_PATH}
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,12 +14,11 @@ if PREBUILT_STACK_URL
 end
 
 Vagrant::configure("2") do |config|
+  config.ssh.forward_agent = true
+
   config.vm.box = BOX_NAME
   config.vm.box_url = BOX_URI
   config.vm.synced_folder File.dirname(__FILE__), "/root/dokku"
-  config.vm.network :forwarded_port, guest: 80, host: 8080
-  config.vm.hostname = "#{DOKKU_DOMAIN}"
-  config.vm.network :private_network, ip: DOKKU_IP
 
   config.vm.provider :virtualbox do |vb|
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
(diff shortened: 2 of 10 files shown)


michaelshobbs:
`make install-from-deb` fails on new install

```
rm: cannot remove ‘/etc/nginx/sites-enabled/*’: No such file or directory
dpkg: error processing archive /var/cache/apt/archives/dokku_0.3.12_amd64.deb (--unpack):
 subprocess new pre-installation script returned error exit status 1
Errors were encountered while processing:
 /var/cache/apt/archives/dokku_0.3.12_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

josegonzalez:
Interesting. New install? I've definitely run `vagrant up dokku-deb` a few times and it seems to work fine...

michaelshobbs:
Ah I did a `vagrant up` then `make install-from-deb`

josegonzalez:
I'll retry with just a vagrant up.

josegonzalez:
Hah yeah `vagrant up` will only run `vagrant up dokku`, which is the old install method. If you want to use just the new version, you'll need to run `vagrant up dokku-deb`. Still, I guess it should work with the new method too.

Vagrantfile
```diff
@@ -7,6 +7,7 @@ BOX_MEMORY = ENV["BOX_MEMORY"] || "1024"
 DOKKU_DOMAIN = ENV["DOKKU_DOMAIN"] || "dokku.me"
 DOKKU_IP = ENV["DOKKU_IP"] || "10.0.0.2"
 PREBUILT_STACK_URL = File.exist?("#{File.dirname(__FILE__)}/stack.tgz") ? 'file:///root/dokku/stack.tgz' : nil
+PUBLIC_KEY_PATH = "#{Dir.home}/.ssh/id_rsa.pub"
```
michaelshobbs: What is this key used for? This path doesn't exist for me.

Vagrantfile
```diff
@@ -14,12 +15,11 @@ if PREBUILT_STACK_URL
 end
 
 Vagrant::configure("2") do |config|
+  config.ssh.forward_agent = true
```
michaelshobbs: I'm assuming this is required for github auth later? Anything else?

contrib/dokku-installer.rb
```diff
@@ -16,7 +16,7 @@
       f.puts "  }"
     f.puts "}"
   end
-  `rm /etc/nginx/sites-enabled/*`
+  `rm -f /etc/nginx/sites-enabled/*`
```
michaelshobbs: Yeah STFU `rm`!!! :smile:

Vagrantfile
josegonzalez: If the file exists, it's copied to `/root/.ssh/authorized_keys` so that the dokku-installer has a key listed automatically.

Vagrantfile
josegonzalez: Basically. Not for much else.

michaelshobbs:
`make deb-all` fails on github auth. (the ssh agent isn't available as the root user)

```
root@dokku:~/dokku# make deb-all
echo "-> Updating deb repository and installing build requirements"
-> Updating deb repository and installing build requirements
sudo apt-get update > /dev/null
sudo apt-get install -qq -y git ruby1.9.1-dev 2>&1 > /dev/null
command -v fpm > /dev/null || sudo gem install fpm --no-ri --no-rdoc
ssh -o StrictHostKeyChecking=no git@github.com || true
Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts.
Permission denied (publickey).
rm -rf /tmp/tmp /tmp/build buildstep_0.0.1_amd64.deb
mkdir -p /tmp/tmp /tmp/build
echo "-> Creating deb files"
-> Creating deb files
echo "#!/usr/bin/env bash" >> /tmp/tmp/post-install
echo "sleep 5" >> /tmp/tmp/post-install
echo "count=\`sudo docker images | grep progrium/buildstep | wc -l\`" >> /tmp/tmp/post-install
echo 'if [ "$count" -ne 0 ]; then' >> /tmp/tmp/post-install
echo "  echo 'Removing old buildstep image'" >> /tmp/tmp/post-install
echo "  sudo docker rmi progrium/buildstep" >> /tmp/tmp/post-install
echo "fi" >> /tmp/tmp/post-install
echo "echo 'Importing buildstep into docker (around 5 minutes)'" >> /tmp/tmp/post-install
echo "sudo docker build -t progrium/buildstep /var/lib/buildstep 1> /dev/null" >> /tmp/tmp/post-install
echo "-> Cloning repository"
-> Cloning repository
git clone -q "git@github.com:progrium/buildstep.git" /tmp/tmp/buildstep > /dev/null
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
make: *** [deb-buildstep] Error 128
```

josegonzalez:
It's meant to be run as non-root.

michaelshobbs:
But the shared folder is under /root, right?

josegonzalez:
`cd /vagrant` :)

michaelshobbs:
Derp!

josegonzalez:
I added a `packer.json` file which we can use to build packer images whenever we update the debian package. Seems to work well enough from my testing (which means our next release can finally have an updated DO image).

deb.mk
```diff
@@ -0,0 +1,159 @@
...
+	ssh -o StrictHostKeyChecking=no git@github.com || true
+
+deb-buildstep: deb-setup
+	rm -rf /tmp/tmp /tmp/build $(BUILDSTEP_PACKAGE_NAME)
+	mkdir -p /tmp/tmp /tmp/build
+
+	echo "-> Creating deb files"
+	echo "#!/usr/bin/env bash" >> /tmp/tmp/post-install
+	echo "sleep 5" >> /tmp/tmp/post-install
+	echo "count=\`sudo docker images | grep progrium/buildstep | wc -l\`" >> /tmp/tmp/post-install
```
michaelshobbs: I don't think you need to escape the backticks. It screwed with my syntax highlighting so I removed them and tested. Seems to work both ways.

michaelshobbs:
A `clean` target might be nice. Just an option to make sure you're starting from a clean slate.

michaelshobbs:
A `dist` target would be great as well. One that does deb, packer and possibly uploads to the appropriate repos.

michaelshobbs:
This is a separate issue, but it'd be nice if, in the vagrant env, we don't default to set the hostname as the response from `icanhazip.com`.

josegonzalez:
All the deb commands force a clean build state sans a `make clean`. As far as a dist command, packer will fail because we wouldn't be distributing our api key (same for package_cloud uploads).

michaelshobbs:
Failing by default is fine but you could source in an env var to be defined by valid parties. It's a convenience thing for us is all.

josegonzalez:
Oh when you build and you have `DIGITALOCEAN_API_TOKEN` set, it's picked up by packer automatically.

michaelshobbs:
Perfect! `make dist` or `make release` whatevs should do packer and deb uploads is all I'm sayin'

josegonzalez:
Added the release target. Anything else?

Makefile
```diff
@@ -17,12 +17,22 @@ endif
...
 
+release:
+	$(MAKE) deb-all
+	package_cloud push dokku/dokku/ubuntu/trusty buildstep*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty sshcommand*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty pluginhook*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty rubygem*.deb
+	package_cloud push dokku/dokku/ubuntu/trusty dokku*.deb
+	packer build contrib/packer.json
+
```
michaelshobbs: This may be pedantic but I like the idea of 1 target for `package_cloud` and 1 for `packer`. Then `release` just deps those 2. In case one fails and the other doesn't and you want to just re-run one. Just a thought.

michaelshobbs:
:+1:

metasansana:
Is this for creating custom debs with dokku or is there going to be an official debian package?

josegonzalez:
A bit of both. There actually are debian packages in a repo, and I am currently working on ensuring makefile-based installations upgrade properly to the debian packages before making that the default.

Issue #886, opened 2015-01-11
Modifying nginx.conf when installing hangs


MatthewMueller:
When upgrading Dokku and running `sudo make install`. This is consistently hanging for me

```
Configuration file '/etc/nginx/nginx.conf'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** nginx.conf (Y/I/N/O/D/Z) [default=N] ? Y
```

I'm not sure this is a Dokku thing, but it's weirdly consistent. If I say `N` for this option it's fine. Not really sure what's going on.

michaelshobbs:
Can you turn tracing on with `dokku trace on` and reproduce this? Would like to see which plugin is causing thing.

josegonzalez:
I think this is actually dokku core. We modify the file and *then* install plugins-install. Weird though, haven't seen this in Virtualbox on 1.4.04.

MatthewMueller:
it's a bit hard to reproduce, i've just noticed it every time i spin up a new digitalocean box and try to upgrade:

```
cd ~/dokku
git pull origin latest
rm -rf /var/lib/apt/lists/*
sudo make install
```

josegonzalez:
We do not recommend installing via the digitalocean dokku box. I'll make that explicitly clear in the docs.

MatthewMueller:
Honestly, I think that's how dokku is getting most of it's traffic. There's so many articles on how to get Dokku up and running on digital ocean. In general it's extremely easy, just when you are looking to upgrade.

If we're looking for as many people on the platform as possible (more plugins, more bugfixes, etc), I think making sure that pipeline runs smoothly is critical.

josegonzalez:
Looks like we might be able to avoid this by not templating out the `/etc/nginx/nginx.conf` and instead customizing everything in `/etc/nginx/conf.d/dokku.conf;` ? Haven't tested it to be honest.

josegonzalez:
We had to override the config file for #799. Not really sure what the correct choice is here - new nginx installs appear to have `server_names_hash_bucket_size` uncommented.

We'll be updating the default dokku image soon, but for now the best option is to simply choose `N` during an upgrade.

MatthewMueller:
[...]

PR #891, opened 2015-01-14
Keep existing configuration files when installing nginx. Refs #886

josegonzalez:


michaelshobbs:
While we do overwrite `/etc/nginx/nginx.conf`, this will cause us not to get any conf updates from the upstream. I think we need an nginx upgrade strategy at some point. Maybe pin this to a version for now?

josegonzalez:
We don't want or need upstream changes. Can you describe a time when that wouldn't be the case?

michaelshobbs:
Currently we install `nginx/1.6.2` because that's the latest in the repo. However, in this config, we could get `nginx/1.7.0` binaries but `nginx/1.6.2` config. This could be bad. I think the pedantically correct thing to do here is to pin to `nginx/1.6.2` *and* keep old configs. Right?

josegonzalez:
They won't upgrade to 1.7.0 in a trusty release, and thats what we promote to users. If a user installs on another system, thats outside of what we can logically support and thus don't need to worry. I don't foresee us moving to 14.10 for example, or 15.04 when that comes out - just the latest LTS.

michaelshobbs:
I was using `1.7.0` as an example. Substitute that for any release that causes a change in config that we may want. Perhaps a security item or something of that nature. 

I think pinning the version is the right move here. It also avoids any other oddities that might come up that break any of our config assumptions. 

My stance is probably a little on the conservative side of the spectrum but pinning as a principle generally saves you headaches in the future and allows/forces you to be deliberate in your infrastructure changes.

josegonzalez:
I don't think pinning will work unless we copy the nginx package to our deb repo. Otherwise, ubuntu can/will remove the package once they update their own version.

michaelshobbs:
I mean we would need to keep up-to-date within reason. Currently versions go back to 1.4.6 in the trusty repos. So I don't think we're in any danger of 1.6.2 disappearing any time soon. Besides its much easier to debug an installation issue where the nginx package can't be installed versus some weird config issue.  Ya know?

josegonzalez:
Gotcha, I'll update the PR to do that. Though it won't help for people updating from older versions of dokku...

michaelshobbs:
How do you mean it won't help?

josegonzalez:
Users coming from older versions will have their nginx upgraded and that is what is causing issues.

michaelshobbs:
Yeah that's fair. This would cause them to run 1.6.2 with 1.4.6 config other than our stuff we overwrite.

josegonzalez:
No, we update the nginx config to be what it should be, so they'd be running 1.6.2 with our approved config.

michaelshobbs:
There is other config that `/etc/nginx/nginx.conf`. We source in `/etc/nginx/conf.d/*` and /etc/nginx/sites-enabled/*`. It's probably fine though.... (famous last words)

josegonzalez:
The sourcing also happens in the normal config. The reason we even overwrite the config is because of the custom file in the conf.d. New versions make that setting default :|

josegonzalez:
Whats the verdict on this then?

michaelshobbs:
:+1:

PR #893, opened 2015-01-14
Add --rm=true to `docker run` command. Closes progrium/dokku#793

josegonzalez:


michaelshobbs:
I'm beginning to think I want to change my +1 vote here. It would be nice to have these around for a time in the crontab case, for example. These also get cleaned up on deployment via `dokku cleanup`. 

Perhaps this causes more pain than relief?

josegonzalez:
Fine with me :) . I'll close this and the issue.

PR #894, opened 2015-01-14
Dokku client

josegonzalez:
A slightly better local dokku client. Appears to work with multiple hosts, and can be set as an alias in your bashrc. Also enforces usage of bash.

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -1,13 +1,11 @@
-if [ -z "$DOKKU_HOST" ]
-then
+if [[ ! -z $DOKKU_HOST ]]; then
 	function dokku {
 		appname=$(git remote -v 2>/dev/null | grep dokku | head -n 1 | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
-		if [ "$?" != "0" ]
-		then
+		if [[ "$?" != "0" ]]; then
 			donotshift="YES"
 		fi
-		if [ "$1" = "create" ]
-		then
+
+		if [[ "$1" = "create" ]]; then
 			appname=$(echo "print(elfs.GenName())" | lua -l elfs)
 			if git remote add dokku dokku@$DOKKU_HOST:$appname
 			then
@@ -18,15 +16,16 @@ then
 				return
 			fi
 			git push dokku master
-		else
-			if [ -z "$donotshift" ]
-			then
-				ssh dokku@$DOKKU_HOST $*
-				exit
-			fi
-			verb=$1
-			shift
-			ssh dokku@$DOKKU_HOST "$verb" "$appname" $@
+      return $?
 		fi
+
+		if [[ -z "$donotshift" ]]; then
+			ssh dokku@$DOKKU_HOST $*
+			exit
+		fi
+
+		verb=$1
+		shift
+		ssh dokku@$DOKKU_HOST "$verb" "$appname" $@
 	}
 fi
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 if [[ ! -z $DOKKU_HOST ]]; then
 	function dokku {
 		appname=$(git remote -v 2>/dev/null | grep dokku | head -n 1 | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -19,7 +19,7 @@ if [[ ! -z $DOKKU_HOST ]]; then
 				return
 			fi
 			git push dokku master
-      return $?
+			return $?
 		fi
 
 		if [[ -z "$donotshift" ]]; then
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -2,7 +2,7 @@
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 if [[ ! -z $DOKKU_HOST ]]; then
-	function dokku {
+	function _dokku {
 		appname=""
 		if [ -d .git ] || git rev-parse --git-dir > /dev/null 2>&1; then
 			set +e
@@ -40,4 +40,9 @@ if [[ ! -z $DOKKU_HOST ]]; then
 		shift
 		ssh dokku@$DOKKU_HOST "$verb" "$appname" $@
 	}
+
+	if [[ "$0" == "dokku" ]] || [[ "$0" == *dokku_client.sh ]]; then
+		_dokku $@
+		exit $?
+	fi
 fi
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -2,47 +2,47 @@
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 if [[ ! -z $DOKKU_HOST ]]; then
-	function _dokku {
-		appname=""
-		if [ -d .git ] || git rev-parse --git-dir > /dev/null 2>&1; then
-			set +e
-			appname=$(git remote -v 2>/dev/null | grep -Ei "^dokku" | head -n 1 | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
-			set -e
-		else
-			echo "This is not a git repository"
-			exit 1
-		fi
+  function _dokku {
+    appname=""
+    if [ -d .git ] || git rev-parse --git-dir > /dev/null 2>&1; then
+      set +e
+      appname=$(git remote -v 2>/dev/null | grep -Ei "^dokku" | head -n 1 | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
+      set -e
+    else
+      echo "This is not a git repository"
+      exit 1
+    fi
 
-		if [[ "$appname" != "" ]]; then
-			donotshift="YES"
-		fi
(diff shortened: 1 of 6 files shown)


josegonzalez:
:|

plugins/apps/commands
```diff
@@ -20,17 +20,19 @@ case "$1" in
     [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
     [[ ! -d "$DOKKU_ROOT/$2" ]] && echo "App $2 does not exist" && exit 1
     [[ "$2" == "tls" ]] && echo "Unable to destroy tls directory" && exit 1
-    APP="$2"; IMAGE="dokku/$APP";
+    APP="$2"; IMAGE="dokku/$APP"; FORCE="$3"
```
michaelshobbs: Having to `echo app_name | dokku apps:destroy app_name` was annoying. This allows `dokku apps:destroy app_name force`

josegonzalez:
Fancy.

michaelshobbs:
I just realized this breaks the auto adding of the remote. I feex.

contrib/dokku_client.sh
```diff
@@ -1,32 +1,63 @@
...
+    else
+      echo "This is not a git repository"
+    fi
+
+    if [[ "$appname" != "" ]]; then
+      donotshift="YES"
+    fi
+
+    if [[ "$1" = "create" ]]; then
+      appname=$(echo "print(elfs.GenName())" | lua -l elfs)
```
michaelshobbs: This errors:
`dokku_client.sh: line 34: lua: command not found`

michaelshobbs:
Nevermind that's a lie. If you set DOKKU_HOST variable it will attempt to `git remote add`. However `lua` is not installed. Do we want to auto add a remote in the case of a DOKKU_HOST env var but no git remote?

josegonzalez:
We do not want to do that if we don't have a remote. I mentioned in irc, but here are a few ways we can look at it:

- https://gist.github.com/afriggeri/1266756
- http://stackoverflow.com/questions/7621341/how-can-i-programmatically-generate-heroku-like-subdomain-names/20059752#20059752
- https://github.com/Xe/Tetra/blob/master/lib/elfs.moon

Implement a solution in pure bash. OS X has Bash 3, so we might need to shell out to a temp file or something since you get arrays only in Bash 4 (or do something clever with sed/cut/counting spaces in a long string and getting the nth space and then the word before that...).

You'll probably also want to check that an app of that name doesn't collide with an existing app, but we can actually do that in a later PR, given that this is in contrib.

Either way :+1: thanks for the extra work :)

michaelshobbs:
Includes dup app check and random app generator all in bash (works in bash 3 and 4)!

contrib/dokku_client.sh
```diff
@@ -1,32 +1,111 @@
...
+}
+
+function random_name() {
+ NUM1=$(random_number ${#MOVES[@]})
+ NUM2=$(random_number ${#MOVES[@]})
+ NUM3=$(random_number ${#NAMES[@]})
+
+ UPPER_APPNAME="${MOVES[${NUM1}]}-${MOVES[${NUM2}]}-${NAMES[${NUM3}]}"
+
+ [[ "$BASH_VERSION" =~ 4.* ]] && lower_appname=${UPPER_APPNAME,,}
```
michaelshobbs: Because why `tr` if we don't have to :wink:

contrib/dokku_client.sh
```diff
@@ -1,32 +1,111 @@
...
... VIV ABOKE PIKAQ RYE SAN BREAD LIDEL LIDE PIP PIKEX ROK JUGEN PUD BUDE ZHIB GELU GRAS FLOW LAFUL ATH BALA CORN MOLUF DESP DAKED MIMI BOLUX KODA GELUD MONK SUMOY GEDI WENDI NILEM NILE NILEC KEZI YONGL HUDE WANLI GELI GUAIL MADAQ WUCI WUCI MUJEF JELLY SICIB GELU NELUO BOLI JIALE YED YEDE CLO SCARE AOCO DEDE DEDEI BAWU JIUG BADEB BADEB HOLE BALUX GES FANT QUAR YIHE SWAB SLIPP CLU DEPOS BILIY YUANO SOME NO YELA EMPT ZECUN XIAHE BOLEL DEJI MACID XIHON XITO LUCK MENJI GELU DECI XIDE DASAJ DONGN RICUL MINXI BALIY ZENDA LUZEL HELE5 0FENB KAIL JIAND CARP JINDE LAPU MUDE YIFU LINLI SANDI HUSI JINC OUMU OUMUX CAP KUIZA PUD TIAO FRMAN CLAU SPARK DRAGO BOLIU GUAIL MIYOU MIY QIAOK BEIL MUKEI RIDED MADAM BAGEP CROC ALIGE OUDAL OUD DADA HEHE YEDEA NUXI NUXIN ROUY ALIAD STICK QIANG LAAND PIQI PI PUPI DEKE DEKEJ NADI NADIO MALI PEA ELECT FLOWE MAL MALI HUSHU NILEE YUZI POPOZ DUZI HEBA XIAN SHAN YEYEA WUY LUO KEFE HULA CROW YADEH MOW ANNAN SUONI KYLI HULU HUDEL YEHE GULAE YEHE BLU GELAN BOAT NIP POIT HELAK XINL BEAR LINB MAGEH MAGEJ WULI YIDE RIVE FISH AOGU DELIE MANTE KONMU DELU HELU HUAN HUMA DONGF JINCA HEDE DEFU LIBY JIAPA MEJI HELE BUHU MILK HABI THUN GARD DON YANGQ SANAQ BANQ LUJ PHIX SIEI EGG)
```
josegonzalez: Do these lines work in bash3? I'll test in the morning, but OS X comes with Bash 3 and I'm like 90% sure arrays are Bash 4.

josegonzalez: Also trollerskares around the word-wrapping :P

contrib/dokku_client.sh
```diff
@@ -1,32 +1,111 @@
...
+      echo "This is not a git repository"
+    fi
+
+    if [[ "$appname" != "" ]] && [[ -n "$*" ]]; then
+      case "$1" in
+        apps*|backup*|help|plugins*|ps:restartall|trace|version)
+          true
+          ;;
+        delete)
+          verb="apps:destroy"
```
josegonzalez: We can omit this I think? No need to promote a deprecated app command.

contrib/dokku_client.sh
michaelshobbs: I tested this with 
```
# /bin/bash --version
GNU bash, version 3.2.53(1)-release (x86_64-apple-darwin14)
Copyright (C) 2007 Free Software Foundation, Inc.
```
and
```
# /usr/local/bin/bash --version
GNU bash, version 4.3.18(1)-release (x86_64-apple-darwin13.3.0)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

michaelshobbs: Yeah I was going to put the arrays in another file and source them but then that seemed to break the single-file nature of the client anyway...So, here we are? Open to suggestions.

contrib/dokku_client.sh
michaelshobbs: Well we support `create` similar to the `deis` client and it seems semantically "weird" to not support `delete` from the client. Btw, we do this in `dokku` core too.

https://github.com/progrium/dokku/blob/fbc96d92b5445324c6172010165277830764cc02/plugins/00_dokku-standard/commands#L23-L25

josegonzalez: Right, but thats just for BC compatibility. We can remove it going forward, but we don't need to continue promoting that method.

contrib/dokku_client.sh
josegonzalez: Nah, its all good :)

contrib/dokku_client.sh
michaelshobbs: Totally understand the BC point. However, in the client, we support create like the deis client. My understanding was that this was supposed to mimic that to a certain extent. Coupled with the fact that create and apps:destroy are not at all semantically similar and this is in contrib, I think the client can support delete, not for BC purposes but for semantic similarity. Know what I mean? It seems like a better user experience.

josegonzalez: Don't we already support `apps:create` and `apps:destroy`? What I'm saying is that we should avoid making aliases where possible, and the old aliases in `dokku` are there for BC, but that doesn't mean we need to continue with them.

josegonzalez: Like, rename `delete` in this context to `apps:destroy` and don't change the verb.

michaelshobbs: That makes sense but then we should drop create in favor of apps:create too. That probably makes more sense any since no args execution yields dokku help. I'll do that when I get back home.

AlJohri:
Does "enforces usage of bash" mean the script must be called from a bash shell or that the script needs to use bash specific syntax?

michaelshobbs:
I believe @josegonzalez is referring the shebang at the top of the file that was not there before. This also would mean that the script needs to use bash compatible syntax.

michaelshobbs:
@josegonzalez dropped `create` and `destroy`. 

This allows someone to pass `apps:create` and get an auto generated app name and then `apps:destroy` or `apps:destroy force` and operate on their auto generated app. If they pass in an app name it will act just like a local execution of `dokku apps:create` or `dokku apps:destroy`.

PR #895, opened 2015-01-14
document deploying private git submodules. closes #644

### LINKED ISSUE #644: private git submodule fails
I'm trying to deploy a repo with a private submodule. It fails of course.
```
git push dokku@localhost:dokku.example.com master
Counting objects: 8, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (5/5), done.
Writing objects: 100% (6/6), 655 bytes, done.
Total 6 (delta 3), reused 4 (delta 1)
-----> Cleaning up ...
-----> Building dokku.example.com ...
remote: HEAD is now at 2e28837... submodule for testing
remote: Host key verification failed.
remote: fatal: The remote end hung up unexpectedly
remote: Clone of 'git@github.com:example/nodecore' into submodule path 'nodecore' failed
To dokku@localhost:dokku.example.com
error: failed to push some refs to 'dokku@localhost:dokku.example.com'
```

I've read progrium/dokku#314. However, embedding the username and password in the repo is not an option. 

I'm running both [dokku-deployment-keys](https://github.com/cedricziel/dokku-deployment-keys) and [dokku-hostkeys-plugin](https://github.com/cedricziel/dokku-hostkeys-plugin). So, at first glance, I figured we just needed to run these plugins before the app gets built. Unfortunately there is currently no way of doing this. See [here](https://github.com/progrium/dokku/blob/master/dokku#L32-L37). 

So, one approach might be to break lines 33-35 into a new built-in plugin and subsequently move the aforementioned key plugins into a new phase that is called by the build-in.

Thoughts on this? Is this interesting to anyone else or am I off my rocker?

michaelshobbs:
closes #644

### DIFF
--- a/docs/application-deployment.md
+++ b/docs/application-deployment.md
@@ -28,6 +28,16 @@ Dokku only supports deploying from its master branch, so if you'd like to deploy
 Right now Buildstep supports buildpacks for Node.js, Ruby, Python, [and more](https://github.com/progrium/buildstep#supported-buildpacks). It's not hard to add more, [go add more](https://github.com/progrium/buildstep#adding-buildpacks)!
 Please check the documentation for your particular build pack as you may need to include configuration files (such as a Procfile) in your project root.
 
+## Deploying with private git submodules
+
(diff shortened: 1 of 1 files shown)


josegonzalez:
+1. Can you do a ci-skip and also ref the issue?

michaelshobbs:
its ref'd in the comments. yeah meant to skip ci.

josegonzalez:
It's not ref'd in the commit message...

michaelshobbs:
Ah haha! We've not been consistently doing that it seems. Perhaps a CONTRIBUTING.md readme should happen.

PR #898, opened 2015-01-15
Update man page for 0.3.13

joliv:
Some new commands, one deprecated, and a warning at the bottom that the page may not always be up-to-date.


josegonzalez:
We need a way to generate this.

joliv:
See #899 :D

PR #899, opened 2015-01-15
Use help2man to automatically generate man pages

joliv:
As much as I love updating the man pages for every new version, I guess writing the man pages automatically would be a bit easier :)

The only difference is that help2man won't underline/italicize the command args, but I never cared much for that anyhow.

New man pages!

EDIT: It uses `help2man`, documented [here](http://www.gnu.org/software/help2man/).


josegonzalez:
Fancy. Would also get the output of custom plugins people install. I'll test this out later today.

joliv:
Ah, it is currently only generated during bootstrap, so it won't list the help for custom plugins that are installed after that script is run. It could easily be generated each time `plugins-install` is run, however, with a little addition [here](https://github.com/progrium/dokku/blob/31c6444dae2a09adb17a7db2b00ca450d2d878be/dokku#L151).

joliv:
That header should read "**OPTIONS**" instead of "**DESCRIPTION**," so I've added an "Options:" header to the help text to get the correct header generated.

josegonzalez:
Fine with me. Merging. I'll be sure to run this before every release.

joliv:
@josegonzalez No need, it generates it anew with each installation during execution of `bootstrap.sh`.

michaelshobbs:
...and now on install/upgrade

https://github.com/progrium/dokku/blob/00babb5ff514a3bbbffdd06dbea95be33800c978/Makefile#L48

PR #900, opened 2015-01-15
support extracting SANs from SSL certificates and adding them to nginx config

michaelshobbs:
- includes tests for all SANs in test cert


josegonzalez:
Whats a SAN?

michaelshobbs:
Subject Alternative Name

https://www.digicert.com/subject-alternative-name.htm

josegonzalez:
:+1:

PR #901, opened 2015-01-16
Pull new tags when upgrading to update VERSION

misto:
During an upgrade, the output of `git describe --tags` is written to the VERSION file, so we need to pull the tags otherwise it looks like we're running an old version.


josegonzalez:
Good call. Is this currently done in the Makefile/bootstrap.sh?

michaelshobbs:
Looks like it. `git clone` on line 22.

PR #902, opened 2015-01-16
Install help2man

josegonzalez:


michaelshobbs:
:+1:

Makefile
```diff
@@ -34,6 +34,7 @@ copyfiles: addman
 		done
 
 addman:
+	apt-get install -qq -y help2man
```
josegonzalez: Should we add help2man target here as well?

Makefile
```diff
@@ -23,7 +23,7 @@ all:
 
 install: dependencies stack copyfiles plugin-dependencies plugins version
 
-copyfiles: addman
```
josegonzalez: Why?

michaelshobbs: This told make to run `addman` before `copyfiles`. `help2man` calls dokku and thus needs to be in the `$PATH` first

michaelshobbs:
Removed the redundant call to install help2man. I think help2man makes sense to be called in the addman task.

PR #903, opened 2015-01-16
move shorter running tests into pre test phase

michaelshobbs:
tells circle not to run deploy tests if lint or unit tests fail


josegonzalez:
Seems legit.

PR #907, opened 2015-01-18
Don't use -o pipefail for plugin

vincentfretin:
I don't think it's a good idea to document the use of -o pipefail by default.
Plugin author blindly follow this instruction and we have issue with the plugin after that.
A simple command like:
  elasticsearch:create)
    # Check if Elasticsearch container is installed
    IMAGE=$(docker images | grep "jezdez/elasticsearch" | awk '{print $3}')
    if [[ -z $IMAGE ]]; then
        echo "Elasticsearch image not found... Did you run 'dokku plugins-install' ?"
        exit 1
    fi

will completely stop dokku right before the if because grep returned exit code 1. (jezdez/elasticsearch image not build)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -70,6 +70,29 @@ A few notes:
 - A `help` command is required, though it is allowed to be empty.
 - Commands *should* be namespaced.
 - As of 0.3.3, a catch-all should be implemented which exits with a `DOKKU_NOT_IMPLEMENTED_EXIT` code. This allows dokku to output a `command not found` message.
+- Be sure you want the "set -eo pipefail" option. Look at the following example :
+
+    ```shell
+    IMAGE=$(docker images | grep "user/repo" | awk '{print $3}')
+    if [[ -z $IMAGE ]]; then
+        echo "user/repo image not found... Did you run 'dokku plugins-install'?"
+        exit 1
+    fi
+    ```
+
+  In the case where the "user/repo" is not installed, dokku exits just before the awk command,
+  you will never see the message printed with echo. You just want "set -e" in this case.
+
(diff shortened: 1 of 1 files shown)


josegonzalez:
@michaelshobbs thoughts on this? I feel as though if a plugin fails, then the deploy should also fail, but I'm not strong on that opinion.

michaelshobbs:
Seeing as core is fairly religious about using `pipefail` I would think others should as well. The way around this example would `|| true`. However, I think it ultimately up to the author of the plugin as to what makes sense in their use case. Augmenting the docs with pros/cons might be fine but I wouldn't remove the example altogether.

josegonzalez:
@vincentfretin do you mind updating the example and making it apparent as to what pipefail does?

vincentfretin:
@michaelshobbs 
IMAGE=$(docker images | grep "jezdez/elasticsearch" | awk '{print $3}' || true) doesn't work. The script exists before the awk command.

@josegonzalez I updated the PR.

docs/plugins.md
```diff
@@ -34,7 +34,7 @@ The below plugin is a dummy `dokku hello` plugin. If your plugin exposes command
 
 ```shell
 #!/usr/bin/env bash
-set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+set -eopipefail; [[ $DOKKU_TRACE ]] && set -x
```
josegonzalez: Should still have a space :)

vincentfretin:
Oups, fixed

michaelshobbs:
Yeah @vincentfretin you would need to do that for every command that could fail. I'm not saying its the right choice for your plugin. Just an option.

PR #914, opened 2015-01-21
conditionally set interactive and tty on dokku run

michaelshobbs:
closes #552
closes #913

PR #922, opened 2015-01-25
use tty cmd to detect if we have one. closes #921

### LINKED ISSUE #921: interactive shells don't work with dokku run
From @timbueno:

I believe this commit broke executing the TTY and Interactive functionality of my Flask-Script commands. I can no longer get a python prompt when executing dokku run my_app python manage.py shell or any other commands that require a prompt. Commands that do not require interactivity work fine.

It seems like [[ ! -z "$PS1" ]] doesn't evaluate when in a script - but in the prompt it works just fine.

michaelshobbs:
use output from `/usr/bin/tty`to detect if we need to run in interactive mode with a tty

closes #921
refs #914

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -49,8 +49,8 @@ case "$1" in
     shift 2
 
     DOCKER_ARGS=$(: | pluginhook docker-args $APP run)
-    # if we have a prompt, then we have a tty according to the LDP (http://www.tldp.org/LDP/abs/html/intandnonint.html)
-    [[ ! -z "$PS1" ]] && DOKKU_RUN_OPTS="-i -t"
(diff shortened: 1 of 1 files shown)


timbueno:
I haven't tested this myself - but this would seem to solve the problem.

michaelshobbs:
I have a django app that uses dokku run to easily access django's shell that was broken due to the previous change. This seemed to work for that. 

One todo: I don't know which package provides `/usr/bin/tty` but I think I want to ensure its there via the installer or plugin install. I'll research this when I'm back at a computer.

michaelshobbs:
So it turns out `/usr/bin/tty` is installed by the `coreutils` package and is of priority `required`. We can safely assume this will be on all Ubuntu systems.

michaelshobbs:
@josegonzalez you good with me merging this?

josegonzalez:
Good work!

timbueno:
Looks good to me! Everything seems to be working as expected now. Thank you for the quick response!

PR #925, opened 2015-01-27
implement rebuild command that reuses git_archive_all

michaelshobbs:
rebuilds an app using the same functionality as a regular deploy and thus supports git submodules. comments welcome!

### DIFF
--- a/plugins/git/commands
+++ b/plugins/git/commands
@@ -6,14 +6,13 @@ git_archive_all() {
   TMP_WORK_DIR=$(mktemp -d)
   chmod 755 $TMP_WORK_DIR
   unset GIT_DIR GIT_WORK_TREE
-  git clone -q $DOKKU_ROOT/$APP $TMP_WORK_DIR &> /dev/null
   pushd $TMP_WORK_DIR > /dev/null
+  git clone -q $DOKKU_ROOT/$APP $TMP_WORK_DIR &> /dev/null
   git config advice.detachedHead false
   git checkout $REV &> /dev/null
   git submodule update --init --recursive &> /dev/null
   find -name .git -prune -exec rm -rf {} \; > /dev/null
   tar c .
-  popd > /dev/null
   rm -rf $TMP_WORK_DIR > /dev/null
 }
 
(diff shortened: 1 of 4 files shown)


plugins/git/commands
```diff
@@ -6,14 +6,13 @@ git_archive_all() {
...
   unset GIT_DIR GIT_WORK_TREE
-  git clone -q $DOKKU_ROOT/$APP $TMP_WORK_DIR &> /dev/null
   pushd $TMP_WORK_DIR > /dev/null
+  git clone -q $DOKKU_ROOT/$APP $TMP_WORK_DIR &> /dev/null
   git config advice.detachedHead false
   git checkout $REV &> /dev/null
   git submodule update --init --recursive &> /dev/null
   find -name .git -prune -exec rm -rf {} \; > /dev/null
   tar c .
-  popd > /dev/null
```
josegonzalez: Why this removal?

michaelshobbs: When running locally from a directory that the `dokku` user didn't have access to, it would cause the deployment to fail because of `pipefail`. Additionally, it was unnecessary as it was being executed in a subshell anyway.

josegonzalez: Shouldn't the command be re-executed as the dokku user via the `dokku` command?

michaelshobbs: It is and that seems to be the issue. Try adding that line back without the `> /dev/null` and running `ps:rebuild <app>` from `~root/dokku`. It throws `/var/lib/dokku/plugins/git/commands: line 16: popd: /root/dokku: Permission denied`

Basically we `pushd $TMP_WORK_DIR` as the `dokku` user and that puts `~root/dokku` on the stack. Then `popd` as `dokku` fails to `cd` back to `~root/dokku`.

vincentfretin:
@michaelshobbs Did you look at the dokku-alt implementation? 
https://github.com/dokku-alt/dokku-alt/blob/master/plugins/dokku-rebuild/commands
There is a lock implemented there with flock to not rebuild an app if it is currently building. This is interesting.

michaelshobbs:
@vincentfretin :+1: I'll take a look at implementing something similar

michaelshobbs:
Ugh. I really don't like that locked/unlocked case pattern....I want to noodle on this a bit. Any additional thoughts are surely welcomed.

josegonzalez:
Why don't you like the locking pattern?

michaelshobbs:
To clarify, I like `flock`. I don't like the way they have the `function` and `function:locked` pattern in their case statements all over the place. Just looks messy and then exposes all of that to the end user. However, I don't have a better idea at the moment. 

I'm about to push up a *functional* implementation that uses a similar `function/function:locked` pattern but in a very limited scope. I wanted to try and put the locking in as little areas as possible to avoid unnecessary complexity.

Please bang on it and rip it to shreds if its crap.

josegonzalez:
:+1: feel free to merge when ready :)

PR #927, opened 2015-01-29
support both docker-args PHASE and docker-args-PHASE for the time being. closes #906

### LINKED ISSUE #906: dokku-persistent-storage plugin with latest buildstep fails in build phase
I believe our change to include docker-args in the build phase is causing some issues.

ref: https://github.com/progrium/dokku/blob/fb3492c8242521dec4ed8428983ea752db710f86/dokku#L58-L59

From `/build/builder` (using heroku-buildpack-python):
```
       =====> Downloading Buildpack: https://github.com/heroku/heroku-buildpack-python.git
       =====> Detected Framework: Python
remote: rm: cannot remove '/app/certs': Device or resource busy
To dokku@localhost:api-staging.xxx.com
 ! [remote rejected] deploy -> master (pre-receive hook declined)
error: failed to push some refs to 'dokku@localhost:api-staging.xxx.com'
```

From `/build/builder` (using heroku-buildpack-ruby):
```
-----> Discovering process types
       Procfile declares types -> web, db-migrate
       Default process types for Multipack -> rake, console, web
rm: cannot remove '/app/data': Device or resource busy
```

Same heroku-buildpack-ruby run but without the volume mounted:
```
+ /tmp/buildpacks/heroku-buildpack-multi/bin/release /tmp/build /cache
+ echo_title 'Discovering process types'
----->' Discovering process types
-----> Discovering process types
+ [[ -f /tmp/build/Procfile ]]
++ ruby -e 'require '\''yaml'\'';puts YAML.load_file('\''/tmp/build/Procfile'\'').keys().join('\'', '\'')'
+ types='web, db-migrate'
+ echo_normal 'Procfile declares types -> web, db-migrate'
      ' Procfile declares types '->' web, db-migrate
       Procfile declares types -> web, db-migrate
+ default_types=
+ [[ -f /tmp/build/.release ]]
++ ruby -e 'require '\''yaml'\'';puts (YAML.load_file('\''/tmp/build/.release'\'')['\''default_process_types'\''] || {}).keys().join('\'', '\'')'
+ default_types='rake, console, web'
+ [[ -n rake, console, web ]]
+ echo_normal 'Default process types for Multipack -> rake, console, web'
      ' Default process types for Multipack '->' rake, console, web
       Default process types for Multipack -> rake, console, web
+ [[ -f /tmp/build/.release ]]
+ ruby -e 'require '\''yaml'\'';(YAML.load_file('\''/tmp/build/.release'\'')['\''config_vars'\''] || {}).each{|k,v| puts "export #{k}='\''#{v}'\''"}'
+ cat
+ chmod +x /start
+ cp -fp start exec
+ rm -rf /app
+ mv /tmp/build /app
```

So the python buildpack is doing something with the `/app` dir and buildstep is as well. (https://github.com/progrium/buildstep/blob/091eafdad7ccf5f4584aa21704397fd1dc1fd208/builder/builder#L70)

I have a PR open against dokku-persistent-storage (dyson/dokku-persistent-storage#12) but I wonder if we're just asking for more problems? Perhaps documenting the pluginhook phase argument would be helpful as well? In any case, I think we may run into more of these issues. If nothing else, it's documented here.

EDIT: for clarification

michaelshobbs:
The idea here is to support docker-args more granularly by phase. See #896 and #906.

### DIFF
--- a/dokku
+++ b/dokku
@@ -55,7 +55,10 @@ case "$1" in
     [[ -d $CACHE_DIR ]] || mkdir $CACHE_DIR
     pluginhook pre-build $APP
 
+    # *DEPRECATED* in v0.4.0: `pluginhook docker-args` will be removed in future releases
+    # https://github.com/progrium/dokku/issues/896 & https://github.com/progrium/dokku/issues/906
     DOCKER_ARGS=$(: | pluginhook docker-args $APP build)
+    DOCKER_ARGS+=$(: | pluginhook docker-args-build $APP)
     id=$(docker run -d -v $CACHE_DIR:/cache -e CACHE_PATH=/cache $DOCKER_ARGS $IMAGE /build/builder)
     docker attach $id
     test "$(docker wait $id)" -eq 0
@@ -85,6 +88,7 @@ case "$1" in
 
     # start the app
     DOCKER_ARGS=$(: | pluginhook docker-args $APP deploy)
+    DOCKER_ARGS+=$(: | pluginhook docker-args-deploy $APP)
     BIND_EXTERNAL=$(pluginhook bind-external-ip $APP)
 
     if [[ "$BIND_EXTERNAL" = "false" ]];then
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -49,6 +49,7 @@ case "$1" in
     shift 2
 
     DOCKER_ARGS=$(: | pluginhook docker-args $APP run)
+    DOCKER_ARGS+=$(: | pluginhook docker-args-run $APP)
     [[ "$(/usr/bin/tty || true)" != "not a tty" ]] && DOKKU_RUN_OPTS="-i -t"
 
     docker run $DOKKU_RUN_OPTS $DOCKER_ARGS $IMAGE /exec "$@"


michaelshobbs:
refs: #896

michaelshobbs:
ping `docker-args` pluginhook authors
@ohardy 
@stuartpb
@sehrope
@dyson 
@alex-sherwin
@rlaneve

dyson:
If it wasn't clear, these are two mutually exclusive ways of handling the docker-args pluginhooks. Your plugin needs to move over to the new style docker-args-PHASE and drop the old style at the same time otherwise the pluginhook will be called twice as dokku is supporting both style for the interim.

michaelshobbs:
:+1:

josegonzalez:
Note: this will take effect once 0.4.0 comes out (which is only being released for this change).

michaelshobbs:
I'll update the code comment

josegonzalez:
So basically once this is merged, we need to do a 0.4.0 release. Do you want me to do a 0.3.14 release and then a 0.4.0 release with this change? We can back out the previous change to add `run` to docker args, release 0.3.14, un-back out and merge this, then release 0.4.0 as well.

michaelshobbs:
I think this is fine for a 0.3.14 release as it doesn't break anything. Then 0.4.0 would include the removal of `docker-args PHASE`. Or am I misunderstanding?

josegonzalez:
Ah gotcha, didn't think of that. I'll merge then and release later today.

PR #929, opened 2015-01-30
Skip building buildstep if SKIP_BUILD_STACK set. Useful for building dokku inside docker.

lsde:
When building dokku inside docker itself, the 'stack' step fails because of unaccessible /var/run/docker.sock, which cannot be bound via Dockerfile.
Thus, the only option is to skip that step and instead pull 'progium/buildstep' directly from docker Hub.
Something like 'ENV SKIP_BUILD_STACK=YES' in Dockerfile is assumed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,10 +96,14 @@ endif
 
 stack:
 	@echo "Start building buildstep"
+ifdef SKIP_BUILD_STACK
+	@echo "Skipping actual building"
+else
 ifdef BUILD_STACK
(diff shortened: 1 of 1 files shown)


lsde:
Here is my Dockerfile for reference
https://github.com/lsde/dokku-in-docker

michaelshobbs:
I get where you're coming from here. I'm on the fence though. @josegonzalez do you have any strong opinions here?

josegonzalez:
I'd prefer if we can detect whether the socket is accessible rather than have an env var...

michaelshobbs:
@lsde @josegonzalez please have a look at #936

lsde:
well i agree this is a better solution.
works for my workflow, closing the issue..

PR #932, opened 2015-01-31
Support --short flag for the "apps" command.

alessio:
Strip the header out the apps list to make the output machine-parseable.

### DIFF
--- a/plugins/apps/commands
+++ b/plugins/apps/commands
@@ -1,23 +1,9 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
-f_in_array() {
-  local \
-    l_array \
-    l_what
-
-  l_what="$1"
-  shift
-  for l_array ; do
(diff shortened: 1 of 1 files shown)


alessio:
A shorter one (e.g. -s) would work too as well as grepping $@ for such flag.
Thanks for considering though!

josegonzalez:
Well, should all output be machine parseable? I've been thinking about this for a long time, as my dashboard currently parses the dokku output into something it can use. The more I think about it, the less I like dokku's current bash implementation, as it makes argument parsing a bit weird, and can't be used as a library.

Not really strong on this PR, as I think there are other places where people will want this sort of functionality as well :|

alessio:
Users may appreciate it since AFAICS it's pretty common to use dokku as a component of complex deployment pipelines.

alessio:
I've simplified the command line parsing and replaced the long option with a shorter one (s/--short/-s/)

josegonzalez:
Closing in favor of #949

PR #938, opened 2015-02-03
enable ssl_prefer_server_ciphers

vincentfretin:
Since we create /etc/nginx/nginx.conf we don't have "ssl_prefer_server_ciphers on" option enabled anymore, enable it in dokku.conf


josegonzalez:
Fine by me :)

PR #941, opened 2015-02-03
source global env prior to app env. closes #931

### LINKED ISSUE #931: How to configure environment variables for multiple apps?
Hi, I have a group of related apps.

I keep an 'ops' repo separate from my code repos.  It has all my apps secrets.

Ideally, I'd like to set the config environments for all my apps to be the same, by loading one file.

The only way I see to do it is one app at a time, by scp'ing a file into /home/dokku/<app>/ENV, then restarting.

Setting them one-by-one would be kinda tedious.

What's the smartest way to set all these apps environment variables to be the same?

Also, how can I configure the environment *before* the app is run?

Thanks!

Mike

michaelshobbs:


josegonzalez:
Ship it!

michaelshobbs:
Ha! `glabal-env` :fatfingeremoticon:

PR #945, opened 2015-02-05
Implement a "run once" command to run containers with --rm option.

alessio:
- Helpful when running one-shot commands (e.g. cron jobs) that don't
  need to preserve the environment once exited. Also prevents the
  disk to filling up with dangling containers.

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -50,9 +50,14 @@ case "$1" in
 
     DOCKER_ARGS=$(: | pluginhook docker-args $APP run)
     DOCKER_ARGS+=$(: | pluginhook docker-args-run $APP)
-    [[ "$(/usr/bin/tty || true)" != "not a tty" ]] && DOKKU_RUN_OPTS="-i -t"
+    [[ "$(/usr/bin/tty || true)" != "not a tty" ]] && DOKKU_RUN_OPTS="$DOKKU_RUN_OPTS -i -t"
 
     docker run $DOKKU_RUN_OPTS $DOCKER_ARGS $IMAGE /exec "$@"
+    ;;
+
+  run:autorm)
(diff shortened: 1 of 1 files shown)


josegonzalez:
Why not just make --rm default?

josegonzalez:
Also, this could be implemented using the `DOCKER_ARGS` pluginhook.

alessio:
> Also, this could be implemented using the DOCKER_ARGS pluginhook.

True.

> Why not just make --rm default?

I'm fine with that, though if you set it as default then you'd need to provide the user with an interface to disable it.

josegonzalez:
Under what conditions wouldn't we want it default for the run command?

michaelshobbs:
This was brought up in another issue. Can't recall at the moment but some folks use `run` to execute cron jobs. I think it's useful to have the containers around for troubleshooting

alessio:
> I think it's useful to have the containers around for troubleshooting

I definitely agree. I didn't feel like adding --rm to DOKKU_RUN_OPTS by default as that would break the *dokku run*'s present interface.

josegonzalez:
I don't want to add a new option here. Perhaps we can add something to the .dokkurc directory? DOKKU_REMOVE_CONTAINER=1 or similar to how we do DOKKU_TRACE?

michaelshobbs:
`DOKKU_RM_CONTAINER` please. Reduce, reuse, recycle keystrokes. :smile:

michaelshobbs:
I have command line args working in a local branch that sets env variables like this one. I'll push it up in a bit. Let me know what you guys think.

alessio:
Cool, good one!

josegonzalez:
Closing this in favor of @michaelshobbs's PR. Thanks @alessio for the initial try!

PR #949, opened 2015-02-05
first stab at a common functions library. include argument parsing and a few definitions wired up. closes #932. closes #945

michaelshobbs:
To be clear, this is a *first* pass. Please have a look and beat it up.

Includes options to enable tracing, auto rm containers on `run`, and "quiet" output. The command line args just set env variables and thus can be used for multiple plugins/commands simultaneously.

This also sets DOKKU_QUIET_OUTPUT if we do not have a tty.

### DIFF
--- a/tests/unit/client.bats
+++ b/tests/unit/client.bats
@@ -21,7 +21,7 @@ teardown() {
 }
 
 @test "dokku client (no args should print help)" {
-  run /bin/bash -c "./contrib/dokku_client.sh | head -1 | grep -q 'dokku COMMAND <app>'"
+  run /bin/bash -c "./contrib/dokku_client.sh | head -1 | egrep -q '^Usage: dokku \[.+\] COMMAND <app>.*'"
   echo "output: "$output
   echo "status: "$status
   assert_success
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -35,11 +35,12 @@ The below plugin is a dummy `dokku hello` plugin. If your plugin exposes command
 ```shell
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
 
 case "$1" in
   hello)
     [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
-    [[ ! -d "$DOKKU_ROOT/$2" ]] && echo "App $2 does not exist" && exit 1
+    verify_app_name "$2"
     APP="$2";
 
     echo "Hello $APP"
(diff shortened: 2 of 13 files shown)


tests/unit/client.bats
```diff
@@ -21,7 +21,7 @@ teardown() {
 }
 
 @test "dokku client (no args should print help)" {
-  run /bin/bash -c "./contrib/dokku_client.sh | head -1 | grep -q 'dokku COMMAND <app>'"
+  run /bin/bash -c "./contrib/dokku_client.sh | head -1 | egrep -q '^Usage: dokku \[.+\] COMMAND <app>.*'"
```
josegonzalez: Why this change

michaelshobbs: The help message header changed. This allows for additional `dokku` args without changing the test.

josegonzalez:
Might make sense to have a shortcut for `DOKKU_QUIET_OUTPUT`?

michaelshobbs:
Shortcut? We expose `--quiet` or `-q` to set it. Is that what you mean?

josegonzalez:
No I mean like an `info2_quiet` function. Not sure how useful that would be in practice though.

michaelshobbs:
Ah! Yeah I was going to do something like that. Forgot between yesterday and today. :smile: I'll take a stab at it.

alessio:
Please use just "name() {}" instead of "function name() {}". The use of `function` is the ugliest POSIX-incompatible bashism :)

alessio:
I'd also suggest to prepend the logging functions (infoN, verbose, warn, etc) with a prefix (dokku_log_$funcname? whatever) to expose them as a sort of consistent public interface to external plugins. A namespace-style prefix would also avoid namespace clashes with functions that could be possibly defined by 3rd-party plugins.

Good job though, thanks!

josegonzalez:
This project is a bash project, and all the files reference the bash interpreter at the top, so I don't see any "bashisms" as bad :) Good call on the function prefixing though.

michaelshobbs:
From http://tldp.org/LDP/abs/html/functions.html:

> Wherever there is repetitive code, when a task repeats with only slight variations in procedure, then consider using a function.

> function function_name { 
> command... 
>} 
>
>or
>function_name () { 
>command... 
>} 
>
>This second form will cheer the hearts of C programmers (and is more portable).

As mentioned, this is distinctly a bash project with no desire to be portable to other shells. That being said if there is additional push back, I don't have a strong position on modifying the `function` syntax.

michaelshobbs:
@alessio i just came across this. https://github.com/progrium/bashstyle
@progrium backs you on not using `function` :smile:

I suppose that's additional push back :wink: I'll refactor it.

PR #954, opened 2015-02-07
dockerfile support

michaelshobbs:
This is a fairly large change to core functionality.
  - moves `build` and `release` into `00_dokku-standard` plugin
  - introduces `(pre/post)-build-dockerfile` and `(pre/post)-release-dockerfile` pluginhooks
  - implements `config` plugin for dockerfile apps with `docker-args-deploy` pluginhook
  - introduces `receive-app` pluginhook in an attempt to be git agnostic

Please pick this apart with extra vigor.


contrib/dokku_client.sh
```diff
@@ -4,7 +4,7 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
 
-function random_number() {
+random_number() {
```
josegonzalez: Why not do a separate PR with these changes first?

josegonzalez:
I'm +1 on the changes, -1 on the fact that it's a big PR. We should be moving towards smaller, more reviewable PRs going forward.

Assuming tests pass, I looked it over and can't think of anything glaring at the moment. I'll take a look again in a bit.

contrib/dokku_client.sh
michaelshobbs: we can. it was on my mind at the time.

josegonzalez: Sokay as is for now.

michaelshobbs:
That's reasonable. Let me split out the style stuff into another PR.

plugins/00_dokku-standard/commands
```diff
@@ -3,6 +3,76 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
+
+    case "$IMAGE_SOURCE_TYPE" in
+      buildstep)
+        id=$(tar -c . | docker run -i -a stdin $DOKKU_IMAGE /bin/bash -c "mkdir -p /app && tar -xC /app")
+        test "$(docker wait $id)" -eq 0
+        docker commit $id $IMAGE > /dev/null
+        [[ -d $CACHE_DIR ]] || mkdir $CACHE_DIR
+        pluginhook pre-build "$APP"
+
+        # *DEPRECATED* in v0.4.0: `pluginhook docker-args` will be removed in future releases
```
josegonzalez: Deprecation happened in 0.3.14, and it will be removed in 0.4.0.

plugins/00_dokku-standard/commands
```diff
@@ -3,6 +3,76 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
+        # https://github.com/progrium/dokku/issues/896 & https://github.com/progrium/dokku/issues/906
+        DOCKER_ARGS=$(: | pluginhook docker-args $APP build)
+        DOCKER_ARGS+=$(: | pluginhook docker-args-build $APP)
+        id=$(docker run -d -v $CACHE_DIR:/cache -e CACHE_PATH=/cache $DOCKER_ARGS $IMAGE /build/builder)
+        docker attach $id
+        test "$(docker wait $id)" -eq 0
+        docker commit $id $IMAGE > /dev/null
+
+        pluginhook post-build "$APP"
```
josegonzalez: Should we change this to post-build-buildstep? Same for pre-build.

michaelshobbs: Probably, yeah. I didn't want to break the current flow for buildstep plugins. I can call both for now.

plugins/00_dokku-standard/commands
michaelshobbs: Indeed.

plugins/00_dokku-standard/commands
josegonzalez: Well, we can call `post-build-buildstep` which calls `post-build` and then kill `post-build` in 0.4.0?

michaelshobbs: oh interesting....

michaelshobbs: So like this?

post-build-buildstep:
```
exec our stuff....
pluginhook post-build "$APP"
```
That's a different approach than our previous pluginhook deprecations. Should we implement yet another?

josegonzalez: Hmm you're right, thats a bit weird. I guess just calling both in order is fine.

michaelshobbs: I actually wish we would have thought of your approach the first time. :smile: I think its cleaner.

plugins/git/receive-app
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+APP="$1"; REV="$2"
+
+dokku git-build $APP $REV
```
michaelshobbs: I actually think this should probably be refactored out of commands directly in the plugin. I don't see the need now to be able to run git-build on it's own.

michaelshobbs: Oh except, derp!! We moved it to commands so the `ps:rebuild`, et al. would work. So nevermind, carry on, nothing to see here.

michaelshobbs:
I think we should let this simmer for a bit and see if it gets some more feedback.

dokku
```diff
@@ -104,12 +68,14 @@ case "$1" in
     DOCKER_ARGS+=$(: | pluginhook docker-args-deploy $APP)
     BIND_EXTERNAL=$(pluginhook bind-external-ip $APP)
 
+    is_image_buildstep_based "$IMAGE" && START_CMD="/start web"
+
     if [[ "$BIND_EXTERNAL" = "false" ]];then
       port=5000
-      id=$(docker run -d -e PORT=$port $DOCKER_ARGS $IMAGE /bin/bash -c "/start web")
```
josegonzalez: What about the `/bin/bash` part?

michaelshobbs: Superfluous, as far as I can tell.

michaelshobbs:
Notice to the following authors:
pre/post-build and pre/post-release pluginhooks are DEPRECATED and will be removed in a future version. (0.4.0 perhaps?)

@ohardy, @hughfletcher, @ohardy, @luxifer, @ohardy, @jeffutter, @stuartpb, @apmorton, @statianzo, @statianzo, @sehrope, @iskandar, @mikecsh, @nornagon, @pauldub, @cedricziel, @cedricziel, @cameron-martin, @ohardy, @pnegahdar, @ademuk, @alexbeletsky, @thrashr888, @gdi2290, @robv, @davidcollom, @baobaab, @Cactusbone, @jazzzz, @Volune, @mlebkowski, @darkpixel

gdi2290:
gg

plugins/00_dokku-standard/commands
```diff
@@ -3,6 +3,84 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
+        # *DEPRECATED* in v0.4.0: `pluginhook post-build` will be removed in future releases
+        pluginhook post-build "$APP"
+        pluginhook post-build-buildstep "$APP"
+        ;;
+
+      dockerfile)
+        # sticking with same pattern of building app image before pre-build for now.
+        docker build -t "$IMAGE" .
+        # buildstep pluginhooks don't necessarily make sense for dockerfiles. call the new breed!!!
+        pluginhook pre-build-dockerfile "$APP"
```
mlebkowski: shouldn’t this pre-hook be in fact before the build?

cedricziel:
@michaelshobbs thx for the mass ping ;) I havent been into dokku for a while now. I would like to continue to maintain my plugins.. So: Is it documented somewhere how exactly too hook in now? Read: Is there a replacement for the mentioned hooks?

mlebkowski:
@michaelshobbs, thanks for the ping, fixed in https://github.com/mlebkowski/dokku-acl/commit/a81aa53d0dadd10d68fb6990e9a799fb9379ecf8. 

But you should probably mention that those pluginhooks are just being renamed to `pre/post-build-buildstep` and `pre/post-release-buildstep`, so other devs have it easier without reading the code.

josegonzalez:
Yeah, we have pluginhook docs coming out pretty soon, which hopefully will have examples for all of them.

michaelshobbs:
@mlebkowski :+1: to reiterate, `pre/post-build` will be renamed to `pre/post-build-buildstep` and `pre/post-release` to `pre/post-release-buildstep`

PR #966, opened 2015-02-10
handle upgrade in debian/preinst script

djelic:
...otherwise, upgrade fails.


josegonzalez:
Ah, good catch. This'll go out soon!

PR #973, opened 2015-02-12
don't trigger pipefail on NO_VHOST check. refs #970

michaelshobbs:


josegonzalez:
Wouldn't the `set` calls avoid this?

michaelshobbs:
Yes, please ignore my idiocy..... :smile:

PR #974, opened 2015-02-12
don't use set to guard against pipefail

michaelshobbs:


josegonzalez:
Tests pass, seems legit.

PR #979, opened 2015-02-17
Progrium merging 1

leonardodino:


leonardodino:
sorry, i got confused in the github website interface.
(it was intended to merge into my branch)

josegonzalez:
Any reason why you're not using dokku as is? Just curious to see how we can improve dokku for our users :)

leonardodino:
#### Short version:
Mostly personal choices based for faster local `vagrant up`, multiple provision keys, as well as server parity...


#### Long version:
The differences now are mostly legacy code for setting an test vagrant faster. 
It includes copying a local cached version of the `stack.tgz` file. *slow connection issues... :snail:*  
And i set it up to use the same box as all my others vagrant boxes `ubuntu/trusty64`.

Now i use other vagrant file, just setting an empty server, and provisioning exactly like my deploy servers, with `ansible-playbook`. (the vagrant usually has a fixed mac address and my DHCP router makes it "fixed IP", for easier access)

In the future i plan to open-source my ansible playbook to help others. But right now it's pretty crude, and lacks some checks.

*This fork won't work properly without setting the keys, and downloading the `stack.tgz`*
*(Obviously they were not included in the repo for safety and repository size reasons, respectively.)*

and finally,
no longer using the bootstrap.sh (but it's a great idea, and it got me  hooked to this kind of dokku-magic)

thanks for this awesome project :smile:

PR #986, opened 2015-02-18
Switch to yabawock's static nginx buildpack

josegonzalez:
The new herokuish branch of buildstep requires changes to the buildpack, but this buildpack should be compatible with the old buildstep.


michaelshobbs:
:+1:

Issue #993, opened 2015-02-20
Allow to take internal port from a Dockerfile


mahnunchik:
Allow to take internal port from a Dockerfile.

if I'm not mistaken it should be here: https://github.com/progrium/dokku/blob/master/dokku#L73-L81

josegonzalez:
We've traditionally made it so that the port is *always* 5000, though I can see why you might want to customize that for a dockerfile.

mahnunchik:
Hi @josegonzalez 

It became necessary for me when I tried to run already dockerized application. It exposes not 5000 port.
Of course I could:
* to do some kind of proxy docker layer from 5000 to different port
* to change port of application

But it would be much easier if `dokku` could take already exposed port.

vbauer:
:+1:

michaelshobbs:
@mahnunchik @vbauer 

What are the expected outcomes here? When I deploy a Dockerfile with `EXPOSE 3000`, nginx still proxies correctly. 

Are you just looking for this when not deployed using nginx-vhosts?

michaelshobbs:
Scratch that. My test app has this:

```
app.set('port', (process.env.PORT || 5000))
```

michaelshobbs:
As a principle, I think the app should be modified to allow the deployment environment to tell it on what port to listen and include a sane default. This falls into line with the Heroku paas mentality.

michaelshobbs:
Also, what do we do if a Dockerfile exposes multiple ports? Which one do we pick to proxy to?

mahnunchik:
Hi @michaelshobbs 

My opinion:
* Heroku ideology is well suited for `normal` applications that are assembled from the source code on deploy.
* Dockerized applications are already built (image) or contain all the information to build themselves (Dockerfile)

We have two options:
* Specify in the documentation that all applications (including Dockerized) should take the PORT variable
* Make an assumption for Dockerized applications and read port from the image.

> Also, what do we do if a Dockerfile exposes multiple ports? 

I think we should pick first port (like it in Deis)

If we allow to take internal port from Dockerized app that we will give users ability to run already packed (maybe not their) applications without modification.

michaelshobbs:
Good points. 

@josegonzalez any additional thoughts here?

josegonzalez:
I would do the following:

- Highly suggest users take the PORT variable in all apps (buildstep/dockerfile)
[...]

PR #995, opened 2015-02-20
added test for Dockerfile deploy with not default exposed port

mahnunchik:
failed test for issue #993


josegonzalez:
@michaelshobbs should we merge this after #1009 as well?

michaelshobbs:
I don't think this needs to be merged. I included a bats test in #1009 that does this instead of including a new test app.

josegonzalez:
Cool, thanks @mahnunchik for verifying the issue! :)

PR #1004, opened 2015-02-24
remove quoting from dockerfile env. closes #1002

### LINKED ISSUE #1002: config:set encloses values with apostrophes when using a Dockerfile
All string config values get enclosed by apostrophes.

e.g. `dokku config:set NODE_ENV=production`
Generates the config `export NODE_ENV='production'`
in the files `/home/dokku/app/ENV` and `/dokku/app/DOCKERFILE_ENV_FILE`

The problem only occurs when using a custom Dockerfile in the root directory of the pushed repo.

I believe setting environment variables like that won't work in docker:
`docker run -e NODE_ENV='production'`

They needs to be written as:
`docker run -e 'NODE_ENV=production'`

michaelshobbs:
- removes single quotes from the generated `DOCKERFILE_ENV_FILE`
  - these are not needed by dockerfile deployments but are by buildpack deployments
- adds link to config/docker-args-run
- makes dockerfile test the same as buildpack test

### DIFF
--- a/plugins/config/docker-args-deploy
+++ b/plugins/config/docker-args-deploy
@@ -7,8 +7,8 @@ DOCKERFILE_ENV_FILE="$DOKKU_ROOT/$APP/DOCKERFILE_ENV_FILE"
 
 if ! is_image_buildstep_based "$IMAGE"; then
   > "$DOCKERFILE_ENV_FILE"
-  [[ -f "$DOKKU_ROOT/ENV" ]] && sed -e "s:^export ::g" "$DOKKU_ROOT/ENV" > "$DOCKERFILE_ENV_FILE"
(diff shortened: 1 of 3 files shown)


josegonzalez:
Why the link?

michaelshobbs:
This allows the app env to be included in containers started with `dokku run`.

marcbachmann:
@michaelshobbs Thanks for both fixes. That with the link would have been my next issue. :+1:
Already tapped into that.

michaelshobbs:
tests are failing. i'll need to dig into that

dubcanada:
Is this why my Dockerfile ENV variables have quotes around them, but the buildpack ones don't?

michaelshobbs:
Yes

dubcanada:
What is holding up the merge of this? Looks like tests pass according to CircleCI

michaelshobbs:
Was hoping for more comments I suppose. I'll merge now.

PR #1005, opened 2015-02-24
Support xip.io wildcard DNS as a VHOST

avidal:
http://xip.io

xip.io is a service by Basecamp that provides a wildcard DNS service on the public internet. Any \<ip address>.xip.io will return a DNS response for the IP address, same with \<subdomain>.\<ip address>.xip.io.

The domains plugin in dokku uses a regex to match IP4 and IP6 addresses in the VHOST file, and disables VHOST support if found. This PR changes those patterns to require that the VHOST entry **ends with** the IP address.

This is a gist where my VHOST setting is "127.0.0.1.xip.io" that demonstrates the original output, linked directly to the line from `plugins/domains/commands`: https://gist.github.com/anonymous/c529177f20b36beda80d#file-debug-log-L1373


michaelshobbs:
Ah makes sense now. A test would be great.

avidal:
Yeah, I'd love to add a test. Just gotta get a development copy running.

avidal:
I tried to get the test environment running, but `tests/run_vagrant` fails somewhere along the bootstrapping process and I end up not having a `/root/dokku`. I'll see if I can figure it out.

michaelshobbs:
Check out `make unit-tests` and `tests/unit/*.bats`

michaelshobbs:
@avidal ping when you're back on IRC. in the meantime, specifically check out `tests/unit/ports.bats`. That's where I'm testing the port exposure stuff with IP addresses and such.

josegonzalez:
@avidal ping :)

PR #1009, opened 2015-02-26
extract first port from Dockerfile and set config variable for use in deploy phase. closes #993

### LINKED ISSUE #993: Allow to take internal port from a Dockerfile

michaelshobbs:
uses `dokku config:set` to store first Dockerfile port for use in deploy phase

closes #993

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -34,6 +34,9 @@ case "$1" in
         ;;
 
       dockerfile)
+        # extract first port from Dockerfile
+        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }')
+        [[ -n "$DOCKERFILE_PORT" ]] && dokku config:set $APP DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT
         # sticking with same pattern of building app image before pre-build for now.
         docker build -t "$IMAGE" .
         # buildstep pluginhooks don't necessarily make sense for dockerfiles. call the new breed!!!
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -36,13 +36,8 @@ case "$1" in
       dockerfile)
         # extract first port from Dockerfile
         DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }')
-        if [[ -n "$DOCKERFILE_PORT" ]];then
-          if [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && [[ "$(grep DOKKU_DOCKERFILE_PORT $DOKKU_ROOT/$APP/ENV) > /dev/null 2>&1; echo $?" == "0" ]];then
-            sed  -i -e "s:DOKKU_DOCKERFILE_PORT=.*:DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT:g" "$DOKKU_ROOT/$APP/ENV"
-          else
-            echo "export DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT" >> "$DOKKU_ROOT/$APP/ENV"
-          fi
-        fi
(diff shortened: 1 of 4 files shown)


plugins/00_dokku-standard/commands
```diff
@@ -34,6 +34,9 @@ case "$1" in
         ;;
 
       dockerfile)
+        # extract first port from Dockerfile
+        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }')
+        [[ -n "$DOCKERFILE_PORT" ]] && dokku config:set $APP DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT
```
josegonzalez: Won't this trigger a `ps:restart`?

michaelshobbs: Good catch. I didn't see any issue because we guard against it on the first deploy.

https://github.com/progrium/dokku/blob/53f9974f6f905de5ec99163cc691f2493e6048e3/plugins/ps/commands#L86-L87

Let me rework the config portion.

plugins/00_dokku-standard/commands
```diff
@@ -34,6 +34,15 @@ case "$1" in
         ;;
 
       dockerfile)
+        # extract first port from Dockerfile
+        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }')
+        if [[ -n "$DOCKERFILE_PORT" ]];then
+          if [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && [[ "$(grep DOKKU_DOCKERFILE_PORT $DOKKU_ROOT/$APP/ENV) > /dev/null 2>&1; echo $?" == "0" ]];then
+            sed  -i -e "s:DOKKU_DOCKERFILE_PORT=.*:DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT:g" "$DOKKU_ROOT/$APP/ENV"
+          else
+            echo "export DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT" >> "$DOKKU_ROOT/$APP/ENV"
```
josegonzalez: This is the sort of thing I wanted to avoid. If we created a flag for config:set to not rebuild, that would be ideal, as there are other plugins (like databases) that could set variables during pre-deploy. Thoughts?

michaelshobbs:
I'd kind of rather have an internal-only command that does this instead exposing it to the user. For two reasons. 

1. The option is not Heroku compliant
2. Avoids more arg parsing code

josegonzalez:
Thats fine with me, something in common? That way plugins could still use it, but not users.

josegonzalez:
My concern is that adding something like this to common exposes configuration management outside of the config plugin, which I don't think is so great.

michaelshobbs:
How about `dokku config:nset` or something similar that wraps `config:set` but bypasses the restart?

josegonzalez:
Well it wouldn't be able to wrap `config:set` but would be wrapped. We can omit it from the help output so and document it in the development guide for plugins :)

michaelshobbs:
Works for me.

michaelshobbs:
ping @josegonzalez

plugins/00_dokku-standard/commands
josegonzalez: After we merge the config:set stuff, we should redo this using that, then merge this.

josegonzalez: This can be changed now to use the non-restart version, correct? Once that's done, merge away :)

plugins/00_dokku-standard/commands
```diff
@@ -34,6 +34,10 @@ case "$1" in
         ;;
 
       dockerfile)
+        # extract first port from Dockerfile
+        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }')
+        [[ -n "$DOCKERFILE_PORT" ]] && dokku config:set-norestart $APP DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT
```
josegonzalez: Much nicer! I wonder if there are other places we can use this pattern?

josegonzalez:
Merge once this passes :) Good work!

PR #1012, opened 2015-02-26
Added container ID to post-deploy as last arg.

dubcanada:
I needed container ID for my post-deploy hook so here is a pull that adds $id as the last arg (so as not to mess anyones current apps up) to post-deploy.

Thoughts?

### DIFF
--- a/dokku
+++ b/dokku
@@ -99,7 +99,7 @@ case "$1" in
     echo "http://$(< "$DOKKU_ROOT/HOSTNAME"):$port" > "$DOKKU_ROOT/$APP/URL"
 
     dokku_log_info1 "Running post-deploy"
-    pluginhook post-deploy  $APP $port $ipaddr
+    pluginhook post-deploy  $APP $port $ipaddr $id
     trap -        INT TERM EXIT
 
     # kill the old container


josegonzalez:
A few things:

- I hate that we have positional arguments to pluginhooks and not named ones :(
- This needs docs
- Can we maybe turn the retrieval of the ID into a helper method in common instead?

michaelshobbs:
:+1: on a common helper function

dubcanada:
What would we want to return if we are unable to get container id?

I'll deprecate the check-deploy $id arg?

michaelshobbs:
No docs yet. Check out `plugins/common/functions`.

With regards to not getting a container id, we do things like this now.
https://github.com/progrium/dokku/blob/53f9974f6f905de5ec99163cc691f2493e6048e3/plugins/ps/commands#L31-L32

However, we may just want to return nothing and let the caller handle the blank response?

josegonzalez:
@dubcanada bump :)

michaelshobbs:
@dubcanada: Why don't you just get the container id from `$DOKKU_ROOT/$APP/CONTAINER` instead of changing the pluginhook call?

josegonzalez:
Yeah that's probably a better call, given that by the time `post-deploy` runs, `$DOKKU_ROOT/$APP/CONTAINER` is up to date. Closing.

PR #1013, opened 2015-02-28
Fix URL schemes in `dokku urls` output

lmars:
Previously, the presence of TLS certificates would cause `dokku urls` to output all https URLs, even for domains which were not configured to use TLS.

This change introduces a URLS file in the app's directory which contains the URLs with schemes based on whether they have been configured to use TLS in the nginx config.


plugins/nginx-vhosts/commands
```diff
@@ -94,6 +95,11 @@ EOF
         echo "       Reloading nginx"
         restart_nginx
       fi
+
+      echo "-----> Generating app URLs..."
```
josegonzalez: Do we need the extra output? If so, you should use the helper in common for this :)

lmars: Not really needed no, dropped.

michaelshobbs:
In reference to my comment on #1008, perhaps we can right a test that ensures we don't return `https` for domains that only partially match the cert CN. (i.e. CN=*.dokku.me and domain is www.test.dokku.me)

lmars:
@michaelshobbs I have fixed the sub-subdomain match issue in #1008

PR #1018, opened 2015-03-06
confine arg shifting to between dokku and command. closes #1017

### LINKED ISSUE #1017: dokku run doesn't work with --options
```bash
$ dokku run new python --version
 !    `python --version` is not a dokku command.
 !    See `dokku help` for a list of available commands.
```

michaelshobbs:
This effectively enforces dokku "global" options position to exist between `dokku` and `command`


josegonzalez:
Seems good so long as quoted commands still work.

michaelshobbs:
Indeed it does

PR #1027, opened 2015-03-10
Add webhooks plugin to documentation

nickstenning:
I've written a small [`dokku-webhooks` plugin](https://github.com/nickstenning/dokku-webhooks) to ping remote endpoints when deployments happen. Added this to the plugins docs.


michaelshobbs:
Very cool!

josegonzalez:
:+1: nice!

PR #1029, opened 2015-03-14
ADDED several enhancements for CHECKS file:

assaf:
- Specify how long to wait before running first check
- Specify timeout for each check
- Check specific hosts, e.g. http://signin.example.com
- Check both HTTP and HTTPS resources


michaelshobbs:
Excellent! Thanks for PR'ing this work. Looks good to me. The CI platform upgraded docker and is interfering with the test suite. A patch is forthcoming and I'll merge this as soon as that is complete.

PR #1030, opened 2015-03-15
non ssl vhosts when using certs. fixes broken tests

michaelshobbs:


michaelshobbs:
I think this might have been an artifact of my build env. Closing.

PR #1036, opened 2015-03-17
create config set/unset without restart. closes #908

### LINKED ISSUE #908: dokku config:set but without restart (in a pre-release hook)
Hi,
I need to be able to call dokku config:set but without restarting the application, because I'm already in the process of restarting the application. I'm in a pre-release hook, see:
https://github.com/vincentfretin/dokku-elasticsearch-plugin/blob/master/pre-release#L37
How can we modify the config:set code to handle this case?
For now the application restarts twice if the elasticsearch container is not started.

michaelshobbs:
creates `config:set-norestart` and `config:unset-norestart`

I did not include these in the help output deliberately as this functionality is not compliant with Heroku functionality and is intended for internal use by things like plugins.

refs #1009

### DIFF
--- a/plugins/config/commands
+++ b/plugins/config/commands
@@ -42,7 +42,6 @@ config_write() {
   if ! cmp -s $ENV_FILE $ENV_FILE_TEMP; then
     cp -f $ENV_FILE_TEMP $ENV_FILE
     chmod 600 $ENV_FILE
-    dokku ps:restart $APP
   fi
   rm -f $ENV_FILE_TEMP
 }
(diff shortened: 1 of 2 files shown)


michaelshobbs:
ping @josegonzalez

plugins/config/commands
```diff
@@ -149,12 +165,29 @@ ${var}"
...
+    dokku ps:restart $APP
+  ;;
+
+  config:unset-norestart)
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    verify_app_name "$2"
+    APP="$2"
+
+    if [[ -z $3 ]]; then
+      echo "Usage: dokku config:unset APP KEY1 [KEY2 ...]"
```
josegonzalez: wrong usage

plugins/config/commands
```diff
@@ -101,6 +100,23 @@ case "$1" in
...
+    dokku ps:restart $APP
+   ;;
+
+  config:set-norestart)
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    verify_app_name "$2"
+    APP="$2"
+
+    if [[ -z "${*:3}" ]]; then
+      echo "Usage: dokku config:set APP KEY1=VALUE1 [KEY2=VALUE2 ...]"
```
josegonzalez: wrong usage

michaelshobbs:
@josegonzalez fixed usage message

josegonzalez:
I actually thought about it more, and since it's being called by the external command, we should just use the external names (set/unset) as opposed to the actual. Can you revert that change? As well, add some docs to the `plugin development` section regarding it's usage.

josegonzalez:
Once ^ is done, we can merge :)

michaelshobbs:
Done and done.

PR #1038, opened 2015-03-17
default container check. closes #1020

### LINKED ISSUE #1020: No errors reported on failed builds
When I push first time my app surely won't work cause I have not set configs yet (see #1019), still it says it deployed it and gives me an url. Url doesn't work, there is no even docker container running!

The other scenario - if I made a mistake like haven't checked in changed requirements file, so my app throws ImportErrors, web process doesn't start, still no reported errors. Worse is that when I log into server there is no place to look for error logs. I need to rerun parts of buildpack to get to the actual errors.

michaelshobbs:


plugins/checks/check-deploy
```diff
@@ -62,9 +62,17 @@ TIMEOUT="${DOKKU_CHECKS_TIMEOUT:-30}"
...
-    dokku_log_verbose "CHECKS file not found in container: skipping checks"
+    dokku_log_verbose "CHECKS file not found in container: running simple container check..."
     rm -rf $TMPDIR
-    exit 0
+
+    # simple default check to see if the container stuck around
+    # for more thorough checks, create a CHECKS file
+    DOKKU_DEFAULT_WAIT=$((WAIT+TIMEOUT))
+    dokku_log_info1 "Waiting for $DOKKU_DEFAULT_WAIT seconds ..."
+    sleep $DOKKU_DEFAULT_WAIT
```
josegonzalez: What would this sleep time be by default?

michaelshobbs: This mirrors the regular check wait + timeout as best we can and let's folks adjust it using the standard variables. I didn't want to introduce another variable and the default wait time of 5 seconds didn't seem reasonable.

PR #1039, opened 2015-03-19
refactor plugin structure. closes #592

michaelshobbs:
inspired by #592 

from @stuartpb:
This change makes the mapping of plugins to commands in Dokku significantly more consistent, giving each existing Dokku command its own plugin, and each plugin a dedicated location for a provided command and/or subcommands.

Beyond the navigational advantages, this allows individual commands / components of Dokku to be disabled or overridden at the plugin level, without having to muck around patching scripts that provide unrelated commands.

## Commands and subcommands

"command" scripts, like a hook, are executed when the name of a plugin is
called: for instance, "dokku build" calls the "build" plugin's "command"
script (plugins/build/command).

"subcommand" scripts are called for subcommands of plugins, as namespaced with
a colon: for example, "dokku backup:export" calls the "export" subcommand
script of the "backup" plugin (plugins/backup/subcommands/export).

Plugins that do not use their name for commands/subcommands (such as the git plugin, or plugins that wrap other commands) can continue to use the "commands" hook, which remains fully supported and backward-compatible.

## Updates to documentation

Each plugin / command gets a README describing its purpose and implementation.

Additionally, the "help" text describing its commands' usage is moved out to a separate "help.txt" file, which is then read and (potentially) columnized when running the "help" command.

This commit also removes the "warning" in the plugins directory - Dokku's stability, hooks and all, is
covered by the main Dokku README.


plugins/plugins/commands
```diff
@@ -0,0 +1,31 @@
...
+    ;;
+
+  plugins-update)
+    pluginhook update
+    ;;
+esac
+
+if [ -z $DOKKU_NO_COMMAND ]; then
+  case "$1" in
+    nginx:*)
```
michaelshobbs: I didn't want to rename `nginx-vhosts` to `nginx`. Seemed ambiguous....

michaelshobbs:
Do we really want to move `deploy`, `build`, and `cleanup` into plugins?

michaelshobbs:
Currently I don't have an answer for the 'is not a dokku command' scenario. Any suggestions are welcomed.

michaelshobbs:
Note: We'll need to merge the other pending plugin changes into the is branch before merging this into master.

plugins/plugins/commands
```diff
@@ -0,0 +1,40 @@
...
+      PLUGIN_NAME="nginx-vhosts"
+      ;;
+    *)
+      PLUGIN_NAME="$1"
+      ;;
+  esac
+  if [ -x $PLUGIN_PATH/$PLUGIN_NAME/command ]; then
+    "$PLUGIN_PATH/$PLUGIN_NAME/command" "$@"
```
josegonzalez: Does this handle legacy plugins? If so, do we have a test for that?

tests/unit/ps-dockerfile.bats
```diff
@@ -10,44 +10,36 @@ teardown() {
...
-  # CI support: 'Ah. I just spoke with our Docker expert --
-  # looks like docker exec is built to work with docker-under-libcontainer,
-  # but we're using docker-under-lxc. I don't have an estimated time for the fix, sorry
-  skip "circleci does not support docker exec at the moment."
-  run bash -c "dokku ps $TEST_APP | grep -q \"node web.js\""
-  echo "output: "$output
-  echo "status: "$status
-  assert_success
-}
+# @test "ps (dockerfile)" {
```
josegonzalez: Why is this commented out?

plugins/plugins/commands
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+case "$1" in
+  plugins-install)
```
josegonzalez: Why weren't these split out?

tests/unit/ps-dockerfile.bats
michaelshobbs: This is commented out similar to the other `dokku ps` test since circleCI doesn't support `docker exec`. Note that it is also "skipped" with the bats `skip` command. Additionally bats will run the `setup` and `teardown` functions even on skipped tests. Therefore commenting this out speeds up the test suite by not superfluously running these functions for test that will not actually be executed. I have an issue opened with the bats project for this as well.

plugins/plugins/commands
michaelshobbs: If we did split these out the command would be `dokku plugins:plugins-install` and I didn't necessarily want break out these simpler commands and potentially add more confusion amongst the user base? I don't know that I have a strong stance against breaking them out though.

plugins/plugins/commands
michaelshobbs: Yes and yes.

We still for loop here: https://github.com/progrium/dokku/blob/592_mh-plugin-structure-refactor/dokku#L157-L167

The `core`, `git`, and `plugins` plugins still have `commands` files.

tests/unit/ps-dockerfile.bats
josegonzalez: Gotcha.

plugins/plugins/commands
josegonzalez: Fair enough, sokay with me.

josegonzalez:
Did you add any new pluginhooks? If so, please doc them.

Unless I missed something, I think this is quite shippable :)

michaelshobbs:
No new hooks. This was only a `commands` refactor.

josegonzalez:
Ship it!

josegonzalez:
Actually let's hold off. I want to check that switching to this would be compatible with plugn, the new pluginhook @progrium is working on.

progrium:
The main incompatibility is one of the main changes this introduces. And it's actually more associated with the "new style" Bash approach. Namely that functions and sourcing is used over separate shell scripts. This is actually very different than pluginhook and plugn currently, but plugn will support a mechanism to help source files from a plugin as well as the traditional hook script style. 

As a simple example (that doesn't reflect how it might specifically work), instead of a command hook and many subcommand hooks, there would be a commands.bash file with each command defined in functions.

First issue you might see is that this means it can only be Bash. Yes, well it all is only Bash. But I'm looking at a way to make it try this first and then fall back to shell scripts. Even without fallback design, it would mean the plugin author would have to define wrapper functions to call their shell scripts. Not ideal, but just pointing out it doesn't preclude shell scripts. 

Arguably this is just another re-organization issue that's not fundamentally different. Normally we could just accept this PR and then change it again with plugn ... except we're changing the Dokku plugin API with both. And if we're going to do that, we should try to do it only once. 

My suggestion would be to keep these changes around as a branch for reference, since there a good structural ideas we want to move towards. The specifics of what it should be are just different. However, there's other good stuff in here that doesn't change the API, and I think we can merge those in pretty easily.

josegonzalez:
@progrium what would be the ideal plugin structure? If we're going to reorganize plugins, we may as well do it now - and the commands/subcommands are now separated so it would be maybe be easier to move to whatever structure we choose.

progrium:
Plugn uses the same directory structure as pluginhook. But my point is more
that the structural issue here is more than just directory structure. It's
very likely the canonical form would be to have commands exist as bash
functions in a single file. However, it's possible that the way this
refactoring is structured will also be allowed, just not preferred. So
there are a couple ways to proceed with different drawbacks.




On Thu, Mar 26, 2015 at 9:29 AM, Jose Diaz-Gonzalez <
notifications@github.com> wrote:

> @progrium <https://github.com/progrium> what would be the ideal plugin
> structure? If we're going to reorganize plugins, we may as well do it now -
> and the commands/subcommands are now separated so it would be maybe be
> easier to move to whatever structure we choose.
>
> —
> <https://github.com/progrium/dokku/pull/1039#issuecomment-86540921>.
>



-- 
Jeff Lindsay
http://progrium.com

josegonzalez:
@progrium well, given that redoing how we handle plugins was something you wanted to work on - and in order to decrease the amount of work we throw at you! - how do you recommend we proceed if someone else made the pluginhook/plugn changes? It seems as though `plugn` is meant to have 1 command per plugin, so that wouldn't work the same way that we currently do plugins.

michaelshobbs:
@progrium: can you enumerate the ways forward and their drawbacks? what bits don't change the api and should be merged?

josegonzalez:
@progrium bump :)

josegonzalez:
Supersceded by #1228.

PR #1054, opened 2015-03-24
Extract nginx.conf from application

assaf:
This change allows the application to ship with custom Nginx configuration, instead of having to separately install it in the application directory.

If the application includes a file called `nginx.conf`, that file will be used as template for creating the actual Nginx configuration file.

### DIFF
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -28,6 +28,7 @@ case "$1" in
     if [[ -z "$DOKKU_APP_LISTEN_IP" ]] && [[ -f "$DOKKU_ROOT/$APP/IP" ]]; then
       DOKKU_APP_LISTEN_IP=$(< "$DOKKU_ROOT/$APP/IP")
     fi
+    DOKKU_APP_CONTAINER_ID=$(< "$DOKKU_ROOT/$APP/CONTAINER")
 
     [[ -f "$DOKKU_ROOT/ENV" ]] && source $DOKKU_ROOT/ENV
     [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
@@ -76,6 +77,7 @@ EOF
       fi
 
       NOSSL_SERVER_NAME=$(echo $NONSSL_VHOSTS | tr '\n' ' ')
+      docker cp $DOKKU_APP_CONTAINER_ID:/app/nginx.conf $DOKKU_ROOT/$APP/nginx.conf.template 2> /dev/null || true
       APP_NGINX_TEMPLATE="$DOKKU_ROOT/$APP/nginx.conf.template"
       if [[ -f $APP_NGINX_TEMPLATE ]]; then
         dokku_log_info1 "Overriding default nginx.conf with detected nginx.conf.template"
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -47,9 +47,9 @@ This archive should is expanded via `tar xvf`. It should contain `server.crt` an
 
 > New as of 0.3.10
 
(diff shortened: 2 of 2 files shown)


plugins/nginx-vhosts/commands
```diff
@@ -76,6 +77,7 @@ EOF
       fi
 
       NOSSL_SERVER_NAME=$(echo $NONSSL_VHOSTS | tr '\n' ' ')
+      docker cp $DOKKU_APP_CONTAINER_ID:/app/nginx.conf $DOKKU_ROOT/$APP/nginx.conf.template 2> /dev/null || true
```
josegonzalez: What is the output of this during a deploy?

josegonzalez:
Can you add documentation for this?

michaelshobbs:
Seems this might be confusing to folks as elsewhere in the dokku world templates are named `nginx.tpl` or `nginx.conf.template`. Perhaps we should look for one of those names instead? Also there a plugin that handles this sort of stuff already. 

https://github.com/mikexstudios/dokku-app-configfiles

plugins/nginx-vhosts/commands
josegonzalez: Also, does this fail if an nginx.conf.template file already exists?

Should we also support the ssl template file?

assaf: > Also, does this fail if an nginx.conf.template file already exists?

No, it will copy over a new configuration file on every deploy, that way changes you make and `git push` will update the Nginx in sync with your code.

> What is the output of this during a deploy?

`Overriding default nginx.conf with detected nginx.conf.template`

assaf:
I edited the documentation to reflect that adding `nginx.conf` to your project and deploying, will replace the default `nginx.conf` from the plugin.

michaelshobbs:
I still find it odd to be looking for a *template* named `nginx.conf` in the container when, in the current code, we look for `nginx.conf.template` in `$DOKKU_ROOT/$APP`. 

https://github.com/progrium/dokku/blob/af626069dd5db32c9306549085571677eea3da6c/plugins/nginx-vhosts/commands#L79

I think we should be consistent in looking for the same file name in the app root and the container.

assaf:
As the documentation says, look for `nginx.conf` in the plugin directory, copy it to your project, make whatever edits are necessary, and deploy. So as a user of this feature, the abstraction you're dealing with is all about `nginx.conf`. One name, easy to remember, easy to work with.

There's an existing feature whereby you take `nginx.conf`, rename it `nginx.conf.template`, and place it in to the APP directory, so it can be used instead of the default `nginx.conf`. So this is more confusing with the file changing names, depending on what state it's in, as a result of how this feature was implemented.

I don't see the benefit of carrying over accidental complexity from the current implementation into this new feature. If this feature can have a simpler/cleaner abstraction, why not?

michaelshobbs:
To better clarify my point, dokku's templates should be renamed as well since they are templates and not useable configs.

assaf:
Right now Dokku's templates have the `.conf` extension, and this PR starts with what's currently in master.

michaelshobbs:
@assaf I'm not asking you to do it. Just adding to the overarching conversation. @josegonzalez any thoughts here?

assaf:
Then I kindly ask that you take this issue and discuss it elsewhere, I appreciate if this discussion stays on topic.

josegonzalez:
Yeah, our global ones should be renamed to `nginx.conf.template`. Didn't realize they weren't. I'll fix that.

@assaf Your pr will be merged once you change `nginx.conf` to `nginx.conf.template`. That is what they are overwriting, so therefore they should be named the same.

jakearchibald:
FWIW, I'd also love this feature

michaelshobbs:
@jakearchibald I'm not sure where @assaf ended up on this PR (needs a rebase) but you can also obtain this functionality with [dokku-app-configfiles](https://github.com/mikexstudios/dokku-app-configfiles)

jakearchibald:
Ohh, thanks!

PR #1055, opened 2015-03-24
Nginx configuration for running behind load balancer

assaf:
When running behind a load balancer that terminates SSL connections (e.g. AWS ELB), Nginx must pass the X-Forwarded headers from the load balancer to the application.

When Nginx is terminating SSL connections, it must not trust any headers sent from the client, and must set those directly.

This PR adds a new setting `DOKKU_SSL_TERMINATED` that, when set, will pick the Nginx configuration that supports a load balancer that terminates SSL connections.

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -43,6 +43,28 @@ dokku nginx:import-ssl myapp < archive-of-certs.tar
 
 This archive should is expanded via `tar xvf`. It should contain `server.crt` and `server.key`.
 
+
+## Running behind a load balancer
+
+> New as of 0.3.17
+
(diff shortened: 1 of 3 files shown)


josegonzalez:
Can you add documentation for this feature?

assaf:
Added documentation.

Right now I'm calling this flag `DOKKU_SSL_TERMINATED` since it evolved from our specific use case, but clearly not the best term, so ideas on what to call this?

josegonzalez:
Sorry for the wait, but can you rebase this? I think the env var naming is good as is, and docs make it clear as to how this is expected to work. I'll merge once rebased :)

assaf:
This will have to wait util next week, I'll pull what ever is in master, fix & rebase as necessary.

josegonzalez:
Closing in favor of #1470. Thanks for your work @assaf!

PR #1056, opened 2015-03-25
Review: Feature/retries

joshco:
This is for review and discussion.  This implements retries for the CHECKS.
I find it more efficient to set a low WAIT value, and have the system retry the checks until they succeed rather than guess a long WAIT time for a single check.  
This code adds a RETRIES variable that can be set in CHECKS (defaults to 5).
I have added an outer loop to the checks loop that simply retries until the checks succeed.  If after RETRIES times they do not, then it fails in a manner similar to the way it used to.


josegonzalez:
Useful diff [here](https://github.com/progrium/dokku/pull/1056/files?w=1).

@joshco can you add documentation and paste some sample log output for us?

joshco:
working on it

joshco:
I have pushed an update with documentation and examples
see updates to: https://github.com/joshco/dokku/blob/feature/retries/docs/application-deployment.md#zero-downtime-deploy
Which describes the feature and links to a new detail page on checks with examples and log output
https://github.com/joshco/dokku/blob/feature/retries/docs/checks-examples.md

josegonzalez:
@michaelshobbs what do you think about this?

joshco:
Do we want the checks to run when someone merely updates an environment variable?  This currently happens today.  If an admin is updating a series of variables in separate statements, this can take a long time if the checks have to run every time

josegonzalez:
I would think so. You want to verify your changes still work, so how would this be any different?

michaelshobbs:
@josegonzalez I'm not too excited about retries necessarily but I definitely understand the scenario where it would be helpful.

However, it would seem that in this implementation we will attempt all retries even in the case of an error response. What if we just retried on timeout or specific responses? Just spitballing here. I'd like to avoid retrying X times when the app may be clearly failing for a legitimate reason.

michaelshobbs:
Regarding the config changes, the only way for an app to "see" said changes, it to restart and thus employ the zero-downtime checks.

josegonzalez:
+1 @michaelshobbs .  Retry on timeout is what I'd prefer, rather than "always retry"

joshco:
OK.  I will work on making it only retry if it is a connect timeout and if an error happens then stop retrying

michaelshobbs:
@joshco in thinking further on this, i'm not sure we want to add more complexity here. perhaps we should just try it out as is.

specifically i was thinking about how sometimes the checks will fail because the instance/machine is overloaded and the `/exec` script is taking awhile to get to starting up the app. in this case, you'd get a `connection refused`. you would also get this result if the startup failed for a legitimate reason....so, its impossible to distinguish between the two. in the first case, i'd want a retry and in the second case, i'd want a failure....

michaelshobbs:
@joshco also, do we want to pull anything from #1072

joshco:
I will look into it.
I'll have an updated pull request by wednesday

joshco:
It's actually even more complicated than that.  Another ambiguous area is curl failure modes.
the existing (pre-my-changes) checks script uses the --fail flag for curl which tells curl to treat server errors (>=400 ) as failures.  Eg, curl exits with an error code.

This allows a check for '/' to work properly since a 500 error will no longer match an empty regex.  It would without the --fail flag. (I attempted to remove --fail, but then empty regex checks started passing when a 500 was received due to default curl behavior)

In the code:
````
 # Capture HTTP response or CURL error message
    if OUTPUT=$(curl -# $CURL_ARGS 2>&1) ; then
      # OUTPUT contains the HTTP response
      if [[ ! "$OUTPUT" =~ $EXPECTED ]] ; then
        dokku_log_warn "$LOG_URL: expected to but did not find: \"$EXPECTED\""
        #exit 1
        UNCHECKED=$(expr $UNCHECKED + 1 )
  
      fi
    else
      # Failed to connect/no response, OUTPUT contains error message
      dokku_log_warn "$OUTPUT"
      #exit 2
      UNCHECKED=$(expr $UNCHECKED + 1 )
    
    fi
  done
````
Consequently, HTTP server errors fall into the else case where curl connection timeout errors fall.  To distinguish between the two, the code has to have a nested if to determine which curl error occurred.
eg, 7 == couldnt connect, 22 == HTTP error >=400
> see all curl error codes: http://curl.haxx.se/libcurl/c/libcurl-errors.html

another complexity is that there are other codes like 6 == couldn't resolve host (dns) which are connection problems.  Are we clear on if a retry or fail is the right approach for everyone?  Or is the answer, "well, maybe"?

The latter was my initial thinking.  It's too hard to know a consistent answer for all apps so I did "retry all" as it is less costly to wait longer when needed and retry to see if the error clears itself up.
The alternative would be the dev needing to go edit/commit a new CHECKS file to increase the WAIT to a time long enough to avoid transient/temporary load based delays.  This is wasteful, because now we are always waiting a long time for errors that rarely occur.

As far as 	@MatthewMueller 's pull request, it's clever, nc can perhaps better deal with non-http containers, however, it's likely to cause errors in the app logs itself (connect/empty/disconnect)
And it only checks that the app answers the socket, not that it does anything conclusively operational.  Eg, the rails database error would not be caught.

What I'm working on now is a nested if block to make an intelligent guess.  I agree that it may be more complexity than we want, but I figured I'd put it together for us all to review and see if we want it or just stick with retry_all behavior for now.

I'm also going to explore allowing the CHECKS file to include a command line that points to a user provided script that can run their own app specific check.  This can return success or failure.  This would allow for non-http apps that are hard for a standardized dokku script to handle.

stay tuned...

**Comment by MatthewMueller** [comment]
My 2¢.

After playing with my PR more, it's not a great solution or default. I don't know that there's going to be a generalized solution without a `CHECKS` file.

**Comment by joshco** [comment]
while exploring the user provided SCRIPT...  
I was thinking that the smart thing to do is to run the user provided script *inside the new container*, however, I can't seem to make this work.  I'm trying to do a docker run with the container ID, but this isnt working.

Here is an example of just trying to run 'ls'
````
remote:        Running docker run 55acabf105ef ls
remote: Unable to find image '55acabf105ef:latest' locally
remote: Pulling repository 55acabf105ef
remote: time="2015-03-31T02:17:28Z" level="fatal" msg="Error: image library/55acabf105ef:latest not found"
````
I've tried using the full and concat ID.  Am I missing something, is this not possible given the state of a container?

As a test I'm running a docker ps just before the docker run, which shows the container
````
remote: CONTAINER ID        IMAGE                         COMMAND
 CREATED             STATUS              PORTS                    NAMES
remote: 55acabf105ef        dokku/tcfpledge2:latest       "/start web"
 6 seconds ago       Up 5 seconds                                 trusting_jang
````

**Comment by joshco** [comment]
I've built the selective retry function (only retry connection timeouts, fail 500 errors etc)

At this point, I'm inclined to suggest that we stick with the retry all behavior.  The more I think about it the more I'm worried we can't catch all of the cases where retry should happen.  

Another example would be a procfile that starts multiple processes such as rails web server, resque, job worker, or even an inline database.  In these environments it is possible that one will start before the other.  So you may have the web front end answering requests but getting errors trying to connect to resque, which is still in the process of starting.

In such a case, the right thing to do is to wait it out as the errors will stop once startup is complete.

It is here *on a separate branch outside this pull request*
https://github.com/joshco/dokku/blob/feature/retries_selective/plugins/checks/check-deploy
https://github.com/joshco/dokku/blob/feature/retries_selective/plugins/checks/check-deploy#L163-L182

Here is the example output:
`````
remote: -----> Deploying tcfpledge2...
remote: -----> Running pre-flight checks
remote: -----> Attempt 1/6 Waiting for 5 seconds ...
remote:        CHECKS expected result:
remote:        http://localhost/ => ""
remote:  !
remote: curl: (7) Failed to connect to 172.17.1.194 port 5000: Connection refused
remote:  !    Check attempt 1/6 failed.  Retrying...
remote: -----> Attempt 2/6 Waiting for 5 seconds ...
remote:        CHECKS expected result:
remote:        http://localhost/ => ""
remote:  !
remote: curl: (7) Failed to connect to 172.17.1.194 port 5000: Connection refused
remote:  !    Check attempt 2/6 failed.  Retrying...
remote: -----> Attempt 3/6 Waiting for 5 seconds ...
remote:        CHECKS expected result:
remote:        http://localhost/ => ""
remote: Non Retryable CURL error 22
remote:  !
remote: curl: (22) The requested URL returned error: 500 Internal Server Error
remote: =====> tcfpledge2 container output:
remote:        => Booting Thin
remote:        => Rails 4.2.0 application starting in production on http://0.0.0.0:5000
remote:        => Run `rails server -h` for more startup options
remote:        => Ctrl-C to shutdown server
remote:        Thin web server (v1.6.3 codename Protein Powder)
remote:        Maximum connections set to 1024
remote:        Listening on 0.0.0.0:5000, CTRL+C to stop
remote:        Started GET "/" for 172.17.42.1 at 2015-03-31 02:59:43 +0000
remote:        [Rollbar] Reporting exception: could not connect to server: Connection refused
remote:         Is the server running on host "172.17.42.1" and accepting
remote:         TCP/IP connections on port 5431?
remote:        [Rollbar] Exception not reported because Rollbar is disabled
remote:        PG::ConnectionBad (could not connect to server: Connection refused
remote:         Is the server running on host "172.17.42.1" and accepting
remote:         TCP/IP connections on port 5431?
remote:        ):
remote:          vendor/bundle/ruby/2.0.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/postgresql_adapter.rb:651:in `initialize'
[---snip---]
`````

**Comment by josegonzalez** [comment]
@joshco bump :)

**Comment by joshco** [comment]
I don't get it.  See my last comment.  I was waiting for feedback.
What are my next steps?

**Comment by michaelshobbs** [comment]
i'm fine with retry all to start. although what happens if retries is set to 0. do we get one run at least? basically i'd like a wait to disable the retry but still get the check run once.

**Comment by joshco** [comment]
I will have an update tuesday

**Comment by joshco** [comment]
Ready for review.
I addressed your point by moving from "RETRIES" to "ATTEMPTS.
````
ATTEMPTS=0 means don't run the checks.  Useful if you want to skip the checks and can be overridden by DOKKU_CHECKS_ATTEMPTS
ATTEMPTS=1 means run only once
ATTEMPTS=N means try N attempts
````

**Comment by josegonzalez** [comment]
@michaelshobbs this looks good to me :)

**Comment by michaelshobbs** [comment]
@joshco thanks for reworking this. 

I don't understand the naming of the `UNCHECKED` variable... It's tracking failures? Might be good to rename this. 

We'll want to make sure this passes linting.

```
In ./plugins/checks/check-deploy line 113:
  ATTEMPT=$(expr $ATTEMPT + 1 )
            ^-- SC2003: expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]].

In ./plugins/checks/check-deploy line 170:
        UNCHECKED=$(expr $UNCHECKED + 1 )
                    ^-- SC2003: expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]].

In ./plugins/checks/check-deploy line 176:
      UNCHECKED=$(expr $UNCHECKED + 1 )
                  ^-- SC2003: expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]].
```

File: plugins/checks/check-deploy [hunk 1]
```diff
@@ -95,73 +100,95 @@ while read LINE ; do
...
-  dokku_log_verbose "$LOG_URL => \"$EXPECTED\""
-  [[ $DOKKU_TRACE ]] && dokku_log_verbose "$ curl $CURL_ARGS"
-
-  # Capture HTTP response or CURL error message
-  if OUTPUT=$(curl -# $CURL_ARGS 2>&1) ; then
-    # OUTPUT contains the HTTP response
-    if [[ ! "$OUTPUT" =~ $EXPECTED ]] ; then
-      dokku_log_fail "$LOG_URL: expected to but did not find: \"$EXPECTED\""
-      exit 1
+UNCHECKED=0
```
michaelshobbs: No need to set this outside of the loop right?

plugins/checks/check-deploy
```diff
@@ -95,73 +100,95 @@ while read LINE ; do
...
+    dokku_log_verbose "$LOG_URL => \"$EXPECTED\""
+    [[ $DOKKU_TRACE ]] && dokku_log_verbose "$ curl $CURL_ARGS"
+  
+    # Capture HTTP response or CURL error message
+    if OUTPUT=$(curl -# $CURL_ARGS 2>&1) ; then
+      # OUTPUT contains the HTTP response
+      if [[ ! "$OUTPUT" =~ $EXPECTED ]] ; then
+        dokku_log_warn "$LOG_URL: expected to but did not find: \"$EXPECTED\""
+        UNCHECKED=$(expr $UNCHECKED + 1 )
+
```
michaelshobbs: Unnecessary newline?

plugins/checks/check-deploy
```diff
@@ -95,73 +100,95 @@ while read LINE ; do
...
+      if [[ ! "$OUTPUT" =~ $EXPECTED ]] ; then
+        dokku_log_warn "$LOG_URL: expected to but did not find: \"$EXPECTED\""
+        UNCHECKED=$(expr $UNCHECKED + 1 )
+  
+      fi
+    else
+      # Failed to connect/no response, OUTPUT contains error message
+      dokku_log_warn "$OUTPUT"
+      UNCHECKED=$(expr $UNCHECKED + 1 )
+
```
michaelshobbs: Unnecessary newline?

joshco:
Updated

michaelshobbs:
New lint errors:
```
linting...

In ./plugins/checks/check-deploy line 112:
  ATTEMPT=$(( $ATTEMPT + 1 ))
              ^-- SC2004: $ on variables in (( )) is unnecessary.

In ./plugins/checks/check-deploy line 169:
        FAILEDCHECKS=$(( $FAILEDCHECKS + 1 ))
                         ^-- SC2004: $ on variables in (( )) is unnecessary.

In ./plugins/checks/check-deploy line 174:
      FAILEDCHECKS=$(( $FAILEDCHECKS + 1 ))
                       ^-- SC2004: $ on variables in (( )) is unnecessary.
make: *** [lint] Error 123
```

joshco:
What is the command to run that lint so I can check it myself?

michaelshobbs:
Check out the docs here: http://progrium.viewdocs.io/dokku/development/testing

joshco:
updated

michaelshobbs:
:+1:

josegonzalez:
Good work @joshco! This is going out in the next release tonight!

matthewmueller:
<3!!!!

joshco:
Awesome. Glad to help out!

PR #1058, opened 2015-03-25
Packaging

josegonzalez:
This is the beginning of the work necessary to move us away from the perilous makefile installation. Don't merge, as I'll need to modify the bootstrap.sh to take into account older, non-deb released versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ ifeq (vagrant-dokku,$(firstword $(MAKECMDGOALS)))
   $(eval $(RUN_ARGS):;@:)
 endif
 
-.PHONY: all install copyfiles version plugins dependencies sshcommand pluginhook docker aufs stack count dokku-installer vagrant-acl-add vagrant-dokku
+.PHONY: all install copyfiles man-db version plugins dependencies sshcommand pluginhook docker aufs stack count dokku-installer vagrant-acl-add vagrant-dokku
 
 include tests.mk
 include deb.mk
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -61,11 +61,17 @@ plugin-dependencies: pluginhook
...
+
+apt-update:
+	apt-get update
 
 help2man:
 	apt-get install -qq -y help2man
 
+man-db:
+	apt-get install -qq -y man-db
+
```
michaelshobbs: Did we remove `man-db` at one time? Why is it back?

deb.mk
```diff
@@ -61,7 +61,7 @@ deb-all: deb-buildstep deb-dokku deb-gems deb-pluginhook deb-sshcommand
 deb-setup:
 	echo "-> Updating deb repository and installing build requirements"
 	sudo apt-get update > /dev/null
-	sudo apt-get install -qq -y git ruby1.9.1-dev 2>&1 > /dev/null
+	sudo apt-get install -qq -y gcc git ruby1.9.1-dev 2>&1 > /dev/null
```
michaelshobbs: Does gcc get installed only on the build host or will this also end up on a running web instance as well?

josegonzalez: Only on the build host.

michaelshobbs:
Copy that.

josegonzalez:
@michaelshobbs Can you do a once-over on this before I merge? I'm going to make a release after this.

josegonzalez:
I need to update the upgrade docs as well.

bootstrap.sh
```diff
@@ -1,37 +1,69 @@
...
+  sudo apt-get install -qq -y linux-image-extra-`uname -r` apt-transport-https
 
+  if [[ -n $DOKKU_CHECKOUT ]]; then
+    sudo apt-get install -qq -y dokku=$DOKKU_CHECKOUT
+  else
+    sudo apt-get install -qq -y dokku
+  fi
+}
+
+if [[ -n $DOKKU_BRANCH ]]; then
```
michaelshobbs: I assume this would be something a user would pass when calling bootstrap? Probably should be documented, ya? Or did I miss this from somewhere else?

josegonzalez: It's just what we allowed before afaik.

josegonzalez:
This is good to go. Our next release will have debian packaging as the default! :)

PR #1072, opened 2015-03-30
Better/check

MatthewMueller:
This PR solves: https://github.com/progrium/dokku/issues/1070. The deploy will wait for the container to start accepting TCP requests, effectively creating a zero-downtime dokku. It's based on: https://github.com/aanand/docker-wait/blob/master/wait

The `nc` dependency is available on ubuntu and inside of the buildpacks (well atleast the node one):

I do not think this PR is complete, but I wanted some feedback on the direction / implementation. Some improvements include:

- Testing (not sure how to test dokku atm)
- A timeout (so we don't end up in an infinite loop)
- Remove `nc` (is that possible?)
- Move to the checks plugin

I do think checks and this are not mutually exclusive, I think you'd want both to make sure your application has been deployed correctly.

### DIFF
--- a/dokku
+++ b/dokku
@@ -97,10 +97,9 @@ case "$1" in
     }
 
     # Wait until the port is bound on the docker host
-    echo -n "waiting for TCP connection to $ipaddr:$port..."
+    dokku_log_info1 "Waiting for [$ipaddr]:$port to come online"
     while ! nc -w 1 $ipaddr $port 2>/dev/null
     do
-      echo -n .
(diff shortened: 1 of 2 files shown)


MatthewMueller:
EDIT: doesn't work for deployed apps that do not listen on a port, like worker apps. Not sure if there's a way around this.

MatthewMueller:
going to close in favor of: https://github.com/progrium/dokku/pull/1056

PR #1075, opened 2015-03-30
TTY and Compression for SSH

Flink:
When using commands requiring a TTY (such as Rails console), this
script won’t work.
So enabling a TTY resolves the issue. Also compression should be
enabled by default I think.

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -94,7 +94,7 @@ if [[ ! -z $DOKKU_HOST ]]; then
     fi
 
     if [[ -z "$donotshift" ]]; then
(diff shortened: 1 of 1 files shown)


Flink:
@josegonzalez Could you merge this PR please or is there a problem with it?

josegonzalez:
@michaelshobbs any thoughts on this pr?

michaelshobbs:
From the ssh man page:
```
Compression is desirable on modem lines and other slow connections, but will only slow down things on fast networks.
```

I'm fine with `-t` though.

Flink:
@michaelshobbs I’ve removed the `-C` flag :)

PR #1077, opened 2015-03-30
Add dokku-psql-single-container to plugins

Flink:


josegonzalez:
Gosh we really need to just have an official plugin for postgres, it's getting ridiculous :P

Flink:
Yeah it’ll be awesome :)

PR #1084, opened 2015-04-02
Make port forwarding configurable

awendt:
My first attempt to build the Vagrant VM resulted in this error:

```
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 8080 is already in use
on the host machine.
```

because I already had another VM listening on 8080.

This change enables devs to avoid port collisions by specifying:

    FORWARDED_PORT=8091 vagrant up


josegonzalez:
Seems legit.

PR #1087, opened 2015-04-03
make docker-options adhere to DOKKU_NOT_IMPLEMENTED_EXIT pattern

michaelshobbs:
docker-options didn't exit with DOKKU_NOT_IMPLEMENTED_EXIT. now it does.


josegonzalez:
Maybe worth adding a test that a non-existent command returns properly?

michaelshobbs:
We did. Hence master failing :smile:

josegonzalez:
lolol

PR #1096, opened 2015-04-06
added two more cases to SSL template and gzip as default in templates.

krokhale:
There were two cases missing from the SSL template:

1) Ability to redirect from www to https://www
2) Ability to redirect from naked url to https://

And added gzip as default on both SSL and NOSSL templates.

### DIFF
--- a/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
+++ b/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
@@ -49,6 +49,3 @@ $SSL_DIRECTIVES
   }
   include $DOKKU_ROOT/$APP/nginx.conf.d/*.conf;
 }
-
-
-
--- a/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
+++ b/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
@@ -6,6 +6,21 @@ server {
 }
 
 server {
+  listen      [::]:80;
+  listen      80;
+  server_name $SSL_SERVER_NAME;
+  return 301 https://$SSL_SERVER_NAME$request_uri;
+}
+
+server {
+  listen      [::]:80;
+  listen      80;
+  server_name www.$SSL_SERVER_NAME;
+  return 301 https://www.$SSL_SERVER_NAME$request_uri;
+}
+
+
+server {
   listen      [::]:443 ssl spdy;
   listen      443 ssl spdy;
   server_name $SSL_SERVER_NAME;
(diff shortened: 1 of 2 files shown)


plugins/nginx-vhosts/templates/nginx.ssl.conf.template
```diff
@@ -26,3 +49,6 @@ $SSL_DIRECTIVES
   }
   include $DOKKU_ROOT/$APP/nginx.conf.d/*.conf;
 }
+
+
+
```
josegonzalez: Please remove these extra newlines

plugins/nginx-vhosts/templates/nginx.ssl.conf.template
```diff
@@ -6,6 +6,21 @@ server {
...
+  listen      80;
+  server_name $SSL_SERVER_NAME;
+  return 301 https://$SSL_SERVER_NAME$request_uri;
+}
+
+server {
+  listen      [::]:80;
+  listen      80;
+  server_name www.$SSL_SERVER_NAME;
+  return 301 https://www.$SSL_SERVER_NAME$request_uri;
```
michaelshobbs: this redirect is not desired in all deployments and i do not think it should be in the default template. in many of my deployments i want the opposite and add my own template to redirect from www.domain.com -> domain.com

krokhale: @michaelshobbs Ah, of course requirements would differ for people and that's fair enough. I actually like your way of the URL without the www better! Though what i did there just reflects the defaults dokku creates for a plain vanilla SSL deployment without the template.

For someone moving out of Heroku for the first time, I would't have bothered with the redirects myself if the default template was covering minimum base cases for an SSL deployment, this is what i observed after adding gzip to the default SSL template:

1) I discovered http://domain.com was not redirecting to http://domain.com at all, i would have expected a redirect and https.

2) Discovered http://www.domain.com was not working either. i would have expected a redirect and https.

3) Not sure if this qualifies as a point, but Heroku covers those cases in their basic SSL deployment. Just the least friction to going live.

To my mind, for a majority of users out there an SSL default should cover basic SSL routes, which is what someone would expect naturally to happen, just like gzip to just work without tinkering. Anyone is free to then edit or add a different route after the basics are covered. But those few will be a small minority.

But yeah i would agree with you if you want the default template to remain limited and focused and users like me coming along from Heroku and adding what they want themselves. But since dokku advertises to be a Heroku replacement, why not have intelligent defaults covered like Heroku.

josegonzalez: Adding this would maybe break what worked (or didn't) before, so I would remove these for now and we can revisit during a major release.

A better case would be to document the redirect situation so people know what works/doesn't and how to change it :)

krokhale: Cool. Sounds good.

josegonzalez: @krokhale bump :)

michaelshobbs:
Should the gzip bits perhaps go in the top-level nginx config? Like [here](https://github.com/progrium/dokku/blob/fc466e2c0541d5bad29e127df197f054e3e92e2e/plugins/nginx-vhosts/install#L57) maybe?

PR #1097, opened 2015-04-08
use config:set-norestart in domains plugin. config:get for dockerfile port. closes #1041

### LINKED ISSUE #1041: don't interact directly with app ENV
We now have `config:*-norestart`. We should use that instead of modifying app ENV directly. 

I can have a look at this after #1039 is in. 

refs #1009 #1036

michaelshobbs:
closes #1041

### DIFF
--- a/plugins/domains/commands
+++ b/plugins/domains/commands
@@ -46,7 +46,7 @@ case "$1" in
       fi
       if [[ "$VHOST" =~ $RE_IPV4 ]] || [[ "$VHOST" =~ $RE_IPV6 ]];then
         echo "unsupported vhost config found. disabling vhost support"
-        [[ ! $(grep -q NO_VHOST "$DOKKU_ROOT/$APP/ENV") ]] && echo "export NO_VHOST='1'" >> "$DOKKU_ROOT/$APP/ENV"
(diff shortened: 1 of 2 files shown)


plugins/domains/commands
```diff
@@ -46,7 +46,7 @@ case "$1" in
       fi
       if [[ "$VHOST" =~ $RE_IPV4 ]] || [[ "$VHOST" =~ $RE_IPV6 ]];then
         echo "unsupported vhost config found. disabling vhost support"
-        [[ ! $(grep -q NO_VHOST "$DOKKU_ROOT/$APP/ENV") ]] && echo "export NO_VHOST='1'" >> "$DOKKU_ROOT/$APP/ENV"
+        [[ ! $(grep -q NO_VHOST "$DOKKU_ROOT/$APP/ENV") ]] && dokku config:set-norestart $APP NO_VHOST=1
```
josegonzalez: Can we remove the grep? The `config:set-norestart` will force the correct value in all cases I would think.

michaelshobbs: Yes? :smile:

josegonzalez:
Seems legit :) We can merge and ship once tests pass :)

Issue #1101, opened 2015-04-09
Application-level configuration is not taken into use anymore


kblcuk:
I think [check-deploy](https://github.com/progrium/dokku/blob/master/plugins/checks/check-deploy#L49) misses the following at line 49:

`[[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV`

That line used to be there in 0.3.15, but went missing during some cleanup, I guess?

josegonzalez:
Good call. Mind making a pr? /cc @michaelshobbs

kblcuk:
Done.

PR #1102, opened 2015-04-09
Sourcing app-specific ENV during check-deploy

kblcuk:
Sourcing app-specific ENV during check-deploy, which gone missing in 052e3e88c8865ce04347b3dc72faa2ea2a1f94d8.


michaelshobbs:
closes #1101

PR #1106, opened 2015-04-11
Clarify the usage of system environment variables

davefp:
The section on zero downtime deploys discusses environment variables, but it's not clear whether it refers to application or system env vars. This PR clarifies this and links to the Ubuntu environment variable documentation for further reading.

### DIFF
--- a/docs/application-deployment.md
+++ b/docs/application-deployment.md
@@ -111,6 +111,8 @@ Dokku will wait `DOKKU_CHECKS_WAIT` seconds (default: `5`) before running the ch
 
 Dokku will wait `DOKKU_WAIT_TO_RETIRE` seconds (default: `60`) before stopping the old container such that no existing connections to it are dropped.
 
+Note that these environment variables need to be set on the host system, not in the app environment. [Here's a full guide to Ubuntu environment variables](https://help.ubuntu.com/community/EnvironmentVariables) that explains the various options for setting them up.
+
 # Removing a deployed app
 
 SSH onto the server, then execute:


michaelshobbs:
Is this actually true in master? I thought perhaps we sourced the app env early enough now.

davefp:
Hmm, I was trying to update the DOKKU_WAIT_TO_RETIRE variable and didn't have any luck. I'll try again and report back. I'm on the latest released version (0.3.16).

davefp:
Setting DOKKU_WAIT_TO_RETIRE in the application config (using `ssh dokku@host config:set app DOKKU_WAIT_TO_RETIRE=2`) Doesn't yield the expected result, I still see the 'Shutting down old container in 60 seconds' message during deploy.

Setting it in `/etc/environment` *does* affect subsequent deploys, and I see 'Shutting down old container in 2 seconds' as I expect.

@michaelshobbs question: How do you verify the value of DOKKU_CHECKS_WAIT during deploy? I don't see the default 5 second wait displayed in the deploy output.

michaelshobbs:
Oh yeah derp on me! This is related to #1102 and should be fixed in master and subsequently the next release.

josegonzalez:
Note: I'm waiting on the restart stuff to make a new release :)

michaelshobbs:
Restart stuff? or do you mean retry stuff?

josegonzalez:
Yeah sorry, retry.

davefp:
Excellent! I'll look out for the next release in that case. I'll close this
PR when I'm in front of a real computer.

PR #1111, opened 2015-04-14
Call pre-build-dockerfile before docker build

3onyc:


plugins/00_dokku-standard/commands
```diff
@@ -38,10 +38,12 @@ case "$1" in
         DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }' || true)
         [[ -n "$DOCKERFILE_PORT" ]] && dokku config:set-norestart $APP DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT
 
-        # sticking with same pattern of building app image before pre-build for now.
-        docker build -t "$IMAGE" .
         # buildstep pluginhooks don't necessarily make sense for dockerfiles. call the new breed!!!
         pluginhook pre-build-dockerfile "$APP"
+
+        # sticking with same pattern of building app image before pre-build for now.
```
michaelshobbs: This comment should probably be removed.

michaelshobbs:
This was (loosely) intentionally called in the same manner as buildstep deploys to maintain a similar execution path. Can you provide your use case that requires this change?

3onyc:
I don't have an example use-case right now as I noticed this while reading the code.

However, looking at the logic for buildstep, the [actual build](https://github.com/progrium/dokku/blob/c2053d532277eb00b1db074bebd203af12e0136e/plugins/00_dokku-standard/commands#L26) is triggered after the `pre-build*` hooks.

michaelshobbs:
Hm. I'm clearly not recalling correctly. Seems legit to me then.

plugins/00_dokku-standard/commands
3onyc: Done!

michaelshobbs:
@josegonzalez: what do you think here? 

Dockerfile deployments are brand new for dokku and I don't think there are any dockerfile-specific plugins yet. I was hoping to feedback on this workflow as folks started to use it. In any case, as this is early days making this particular change is probably semantically correct at the least. :smile:

josegonzalez:
Seems legit to me.

Issue #733, opened 2014-10-29
Provide process type as $DYNO to container


yabawock:
Heroku has a $DYNO environment variable in the container which shows which process from the Procfile is running in the container. It would be nice to have this variable available (even if hardcoded as „web“ like the start command) on Dokku too.

josegonzalez:
Seems like we could set this by default - check if set and set it in pre-deploy - and use it [here](https://github.com/progrium/dokku/blob/master/dokku#L71). Thoughts, @michaelshobbs ? I don't know if we have a `dokku config:set --norebuild` command yet, but that's the only blocker I can think of.

michaelshobbs:
To clarify, we'll modify `$START_CMD` if `$DYNO` is set in config? 

Do we want the default `config:set` action to rebuild or should that be explicit? I'm leaning towards `dokky config:set -r <app> <config>` to indicate rebuild and the default to not. Thoughts?

josegonzalez:
I think we should always have it set and default it to `web` if it's not set.

Does heroku rebuild or not? If not, we should not, if so, we should, but we should allow people to do either/or.

michaelshobbs:
According to https://devcenter.heroku.com/articles/config-vars a restart happens on set and unset.

josegonzalez:
So we should at least do a restart (with the env vars sourced in). Is that not what we do now?

michaelshobbs:
That is what we do now. Should we offer a flag to bypass the restart? If so, why?

josegonzalez:
I don't know what I was thinking. Ignore that.

joshco:
I'm bumping into this (or related) as well.  I use a Procfile in my app to specify multiple worker types.that all need to be running to service the app.

This is a rails app that has a web worker and a jobs worker (eg delayed_job)

The procfile is in the tree:
````
web: bundle exec  rails server -p $PORT
jobs: bundle exec rake jobs:work
````

For the moment, what I'm doing is manually starting a second docker for it:

> sudo docker run -d dokku/vwf /bin/bash -c "/start jobs"

But this doesn't seem optimal.  It seems like dokku should examine the procfile and do something useful.

I don't see the Heroku parallel metaphor here, which is their ps settings on each dyno/worker type.  In Heroku you specify the number of each dyno, and type/size you want to run.  The numbers arent part of the procfile, they are set via the heroku command.

> heroku ps:scale web=1, worker=2

That is what tells heroku what to do.

Is there a dokku idiom for this?

[...]

Issue #732, opened 2014-10-29
Enable configuring which process type (web, worker, etc.) to run instead of defaulting to "web"


motin:
Currently, [dokku is hard-coded to start the "web" process on deploy](https://github.com/progrium/dokku/blob/c092dd1719b77843cc1c565688de9482f3046ede/dokku#L66). 

Instead, this should be configurable so that any process type (web, worker, etc.) defined in the Procfile can be run. A config-directive `PROCESS_TYPE` that defaults to `web` if not set to determine this, maybe?

This would make it possible to start worker processes as different docker containers / deployments so that we don't have to resort to using [dokku-shoreman](https://github.com/statianzo/dokku-shoreman) or [dokku-supervisord](https://github.com/statianzo/dokku-supervisord), who both require manual efforts to avoid port collisions. 

This brings dokku closer to how Heroku and other PaaS handle process types - each process is started in it's own container/dyno.

josegonzalez:
As of #1118, we now have container-level scaling, which allows you to run whichever process in the Procfile you see fit :)

PR #1118, opened 2015-04-17
container-level scaling

### LINKED ISSUE #1117: ps:scale implementation
I believe this initially started with wanting to provide a core method of scaling processes on the same machine. #298 was the first attempt at this implementation but was not accepted as we let it sit around for too long. (I think)

Some questions come up in the new world of dokku that include things like zero-downtime checks and dockerfile deployments. 

Issue/Question 1: Currently the zero-downtime checks rely on a web service being accessible for dokku to 'ping'. Therefore, if we run each Procfile entry in its own container things like workers can't be checked in this same manner. 

There are two potential ways to go here. First would be to scale at the container level and somehow detect that a listener is running, then run checks. Otherwise, use the default container 'uptime' check we run when an app does not have a CHECKS file. The second method would be to use a similar pattern to the dokku-supervisor plugin and use a process manager in a single container to handle scaling.

Issue/Question 2: How should we scale dockerfile deployments.?
I don't use this method today and not sure I have a full grasp of implementation details. However, I think these can 'easily' be scaled using the multiple container approach.

refs: #733 

ping @progrium @josegonzalez @joshco

michaelshobbs:
This is a work in progress, DO NOT MERGE yet!

- [x] `dokku ps:scale <app> proc=count [proc=count]` (scales any process in Procfile. defaults to 1 if no DOKKU_SCALE file exists)
- [x] nginx load balancing
- [x] log aggregation
- [x] dockerfile scaling?
- [x] DOKKU_SCALE file precedence (app repo vs local)
- [x] handle CHECKS for non-web containers
- [x] fix old tests
- [x] new tests (scaling 1+ proc types)
- [x] docs
- [x] app w/o Procfile
- [x] dokku ls improvements
  - show docker state?
- [x] set $DYNO upon container start 

## Example `dokku ls` output:
```
root@dokku:~/dokku# dokku ls
-----> App Name           Container Type            Container Id              Status
mynewapp                  NOT_DEPLOYED              NOT_DEPLOYED              NOT_DEPLOYED
node-js-app               web                       5d3cfcc83cfb              running
nodejs-express-workers    web                       c284acaefe0b              running
nodejs-express-workers    worker                    c7df65e37899              running
nodejs-express-workers    worker                    90f7d101355d              stopped
```
closes #1117
closes #1126
refs: #732 #733 #298


josegonzalez:
On heroku, only the web process is scaled to 1 by default.

michaelshobbs:
Do you have a change proposal then? :smile:

michaelshobbs:
Per IRC convo, sticking with default of 1 since we don't have to pay by the container :smile:

joshco:
Sounds good. It should also support zero if it doesnt.

michaelshobbs:
Setting to zero will cause a process to not be started but I need to handle the transition from >1 -> 0. (Note to self)

michaelshobbs:
My initial thought on checks for different process types was to augment the CHECKS file format. Upon more thought, that seems overly complex. Instead, I decided to only parse the CHECKS file if the PROC_TYPE is `web`. This seems to line up with Heroku's special treatment of the `web` proctype.

michaelshobbs:
@josegonzalez @progrium @joshco I think this is ready for review. Please pick it apart.

josegonzalez:
Can an app have multiple web components?

michaelshobbs:
What do you mean?

michaelshobbs:
To clarify, an app can have as many components as the author wants. It's just that the `web` proctype is handled differently than others. So only `web` processes get exposed to nginx or the outside world and they are the only type that custom CHECKS run against.

josegonzalez:
Nevermind. I meant "can dokku load balance multiple processes in an app". The answer is no because of how the web process works, which I think we can deal with.

joshco:
What do you mean by 'load balance'?  Do you mean scale the number of web workers?

michaelshobbs:
Web workers are definitely load balanced with the `nginx-vhosts` plugin

josegonzalez:
No I meant having multiple web-like processes exposing ports. As in a `web` and `web-admin` in a single app.

joshco:
My initial test.  (this is a long comment)

Overall things work.  A few problems so far

# An app without a Procfile doesnt work right anymore

I pushed a rails app with no procfile, and it attempted to start 4 containers, web, worker, rake, console.

````
remote:        Default types for  -> rake, console, web, worker
remote: -----> Releasing foo...
remote: -----> Deploying foo...
remote: -----> New DOKKU_SCALE file generated
remote: =====> rake=1
remote: =====> console=1
remote: =====> web=1
remote: =====> worker=1
remote: -----> Running pre-flight checks
remote:        For more efficient zero downtime deployments, create a file CHECKS.
remote:        See http://progrium.viewdocs.io/dokku/checks-examples.md for examples
remote:        CHECKS file not found in container: Running simple container check...
remote: -----> Waiting for 10 seconds ...
remote: App container failed to start!!
remote: =====> foo container output:
remote:        Abort testing: Your Rails environment is running in production mode!
remote: =====> end foo container output
remote: /usr/local/bin/dokku: line 49: 19164 Killed                  dokku deploy "$APP"
To dokku@dokkudev.splitatom.com:foo
error: failed to push some refs to 'dokku@dokkudev.splitatom.com:foo'

````

# dokku ls strangeness

Now when you do a dokku ls, the output is somewhat confusing.  As you see below, 'vwf' is listed three times.  This makes sense because this app's procfile has two types, and I've got jobs scaled to 2 via the scale command.

````
web: bundle exec  rails server -p $PORT
jobs: bundle exec rake jobs:work
````

The output could be improved to indicate what is going on or what type of container it is

````
root@ip-10-123-195-142:~# dokku ls
=====> App Name                          Container id

foo                                      1ac4ffb5d56cfe73b375604c4e291e4857d7cbe0f5ab140e225291a65769d988
tmx                                      047e7d77a1ad5a71cafb8339d1fd3dcd54480c7e0d216c4955672779250dac4c
vwf                                      0a8ceeaa66d7f7fc15fdbf267ff2637047fc36b098cde8e045304bdd861fc13f
vwf                                      f687a8c04563a1e7f8194cfa257238b9e7b0953ca1cb31e0f36e2b3d162690da
vwf                                      ddc39318e766fcca3104673d4ec7cbaf7c363e32095f02661a9506fa3e7121ba
````

# supervisord-logger compat

Will this break compatibility with the logging supervisor plugin?  Wont it also try to create multiple processes inside each container if the Procfile has multiple types?  Seems like it's for them to change to accomodate.  I will still want to use the logging functionality, but leave the scaling to DOKKU.

**Comment by michaelshobbs** [comment]
I'll work on `dokku ls`. Regarding the other 2, I think you'll need to ship a `DOKKU_SCALE` file. In the first case, how do you handle this with the supervisor plugin? (Set proc scale to 0?) 

In the second case, I think it may be all or nothing. My initial thought for users that want to use another process manager plugin was that they would just set every proc except `web` to 0 with a `DOKKU_SCALE` file. I can doc this up. 

Thoughts on better implementation are very welcomed here. 

Note to self: write test for 10+ containers. Need to sort id files correctly.

**Comment by patrickjahns** [comment]
I am wondering if this would allow for dynamic scaling depending on the load/amount of requests?

Also I am curious if it would be possible to create containers that dynamically start on requests and stop after some idle time again. 
A possible solution for this could be to have a lightweight go server component, that listens on a certain port for requests. If nginx can`t reach any upstream source, this server would be available, the request would arrive. The go server would check if a container is already running, just drop the connection. If it is not running, check if it is being provisioned, if not, start provisioning and wait for X seconds before dropping the connection so that nginx would switch to the running container upstream.

Thoguhts/comments?

**Comment by progrium** [comment]
Definitely sounds out of scope.

On Wednesday, April 22, 2015, Patrick Jahns <notifications@github.com>
wrote:

> I am wondering if this would allow for dynamic scaling depending on the
> load/amount of requests?
>
> Also I am curious if it would be possible to create containers that
> dynamically start on requests and stop after some idle time again.
> A possible solution for this could be to have a lightweight go server
> component, that listens on a certain port for requests. If nginx can`t
> reach any upstream source, this server would be available, the request
> would arrive. The go server would check if a container is already running,
> just drop the connection. If it is not running, check if it is being
> provisioned, if not, start provisioning and wait for X seconds before
> dropping the connection so that nginx would switch to the running container
> upstream.
>
> Thoguhts/comments?
>
> —
> <https://github.com/progrium/dokku/pull/1118#issuecomment-94985659>.
>


-- 
Jeff Lindsay
http://progrium.com

**Comment by michaelshobbs** [comment]
I agree that it's out of scope. However, a plugin perhaps?

**Comment by joshco** [comment]
@michaelshobbs  
> I think you'll need to ship a DOKKU_SCALE file. In the first case, how do you handle this with the supervisor plugin? (Set proc scale to 0?)

I dont have to do that.  It just picked a sensible default and ran a single 'web'   That's how dokku does it even without logging-supervisord

@patrickjahns I concur that this is out of scope.   It's an interesting idea though. Sound like your next plugin project to me ;)

Another way it might be done is based on load.  That's the way I've done amazon autoscaling.  Keep metrics and increase/decrease on that

Your comments also made me think of something else I've been threatening to do:  Modify nginx to retry connection attempts that are refused by the container to get heroku like behavior.  Eg, accept the inbound connection,  hold on to that idle, then keep trying to connect upstream server until it answers.  This allow for dynamic provisioning, though not necessarily the best UX for the end user.

**Comment by michaelshobbs** [comment]
I meant how do you run this branch and that plugin at the same time?

**Comment by joshco** [comment]
@michaelshobbs  ah :)    I feared that having them both would cause a piano to fall on my head so I disabled the logging plugin before I pulled your branch.
I'll put my helmet on and see what happens with both.

**Comment by joshco** [comment]
I tried it with both your branch and the supervisord plugin.  I end up with two containers with both jobs and web running inside them.  I can't think of a way to make supervisord behavior as I wish.  From the supervisor point of view, I'd want different scale factors for each container.  Eg,

````
Container 1: web=1,jobs=0
Container 2: web=0, jobs=1
````

**Comment by joshco** [comment]
It looks like the web processes are fighting each other, where one keeps crashing and getting restarted.  Perhaps its a fight over the bind address
````
2015-04-22 02:59:36,570 INFO exited: web (exit status 1; not expected)
2015-04-22 02:59:37,572 INFO spawned: 'web' with pid 26384
2015-04-22 02:59:38,574 INFO success: web entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2015-04-22 02:59:45,878 INFO exited: web (exit status 1; not expected)
2015-04-22 02:59:46,880 INFO spawned: 'web' with pid 26598
2015-04-22 02:59:47,881 INFO success: web entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2015-04-22 02:59:55,275 INFO exited: web (exit status 1; not expected)
```
In my logging dir I see:
````
root@ip-10-123-195-142:/var/log/dokku/vwf# ls
jobs.00.log  supervisor.log  web.00.log
````

**Comment by michaelshobbs** [comment]
My expectation is that this is not a replacement for logging. So if you want that type of scaling, then you ship a `DOKKU_SCALE` file to override our behavior.

For example you might use a config like this
DOKKU_SCALE (dokku core):
```
web=1
```

SCALE (dokku-logging-supervisord):
```
web=1
jobs=1
```

**Comment by michaelshobbs** [comment]
@joshco can you confirm that your app without a Procfile works on this branch. We test with a nodejs-express app that has no Procfile. Curious, if rails behaves differently.

**Comment by joshco** [comment]
@michaelshobbs **Confirmed.**
The problem is that that we are creating a default DOKKU_SCALE that includes the 4 types worker, rake, console, web, which is wrong.  We should just have web=1.  
The particular failure is that we are trying to run a container with the 'rake' command which defaults to running the app's test suite, which fails because RAILS_ENV is production, as it should be.  (the problem is that we shouldn't be trying to run the test suite)

You can repro this yourself, by cloning the following repo and pushing it to your dokku.  No need to have ruby/rails on your local machine.
> https://github.com/joshco/static_rails_app_dokku

More detail, here's what happens when we push/build:

````
remote:        ###### WARNING:
remote:        No Procfile detected, using the default web server (webrick)
remote:        https://devcenter.heroku.com/articles/ruby-default-web-server
remote:
remote:        -----> Discovering process types
remote:        Default types for  -> worker, rake, console, web
````

Which creates:
````
root@ip-10-123-195-142:/home/dokku/foo# less DOKKU_SCALE
rake=1
console=1
web=1
worker=1
````

Which then leads to
````
remote: =====> foo container output:
remote: ++ docker logs 09f32befad3a9de2547a3f99eee6dfd824d7df96b6ede3f182527cbb7dea332a
remote: + for line in '$(docker logs $CID 2>&1)'
remote: + dokku_log_verbose_quiet 'Abort testing: Your Rails environment is running in production mode!'
remote: + [[ -z '' ]]
remote: + echo '       Abort testing: Your Rails environment is running in production mode!'
remote: + IFS='
remote: '
remote: + dokku_log_info2_quiet 'end foo container output'
remote: + [[ -z '' ]]
remote: + echo '=====> end foo container output'
**remote:        Abort testing: Your Rails environment is running in production mode!**
remote: =====> end foo container output
````

because we are trying to start a container that just runs 'rake'.  The default task for rake is to run the app's tests, which will fail because the environment is RAILS_ENV production, as it should be. 
The reality is that you don't want to be running tests.

The default case should be a DOKKU_SCALE with only 1 entry: web=1

**Comment by michaelshobbs** [comment]
Yeah I was going with the approach that we would attempt to scale everything unless the user specifies differently. You think we should scale web to 1 instead and let the user tell us differently?

@josegonzalez do you agree?

That route is much easier to code too.

**Comment by joshco** [comment]
Right.  A rails app is pretty much always going to have a web component.  Thats the rails MVC.
The standard way to launch that is rails s, though a user might choose to customize it or use different web engines (webrick, thin,puma etc)

It MAY have a job worker.  If it does, there really isn't one standard.  It could be a number of things that the developer will specify, Rails ActiveJob, Delayed_Job, Sidekiq, etc  
So a developer will specify it with something like rake jobs:work or something completely different.

console and rake tasks are run on-demand, not automatically launched or scaled. Rake is really a task runner for app management/maintenance/deployment 

Rake tasks used on production servers are typically things like

    rake db:migrate | db.setup | db:drop # Prepare/migrate the database
    rake assets:clean | assets:precompile # Precompile or clear static assets like css/js
    rake routes # show the available application routes/controllers etc

See https://gist.github.com/joshco/38a9eccd103213bf747d for a typical set of available rake tasks in a rails app

**Comment by josegonzalez** [comment]
As per our IRC conversation, we're going to go with just scaling `web=1` by default. It's simpler to deal with and users will expect this coming from heroku.

File: plugins/00_dokku-standard/commands [hunk 1]
```diff
@@ -109,15 +109,20 @@ case "$1" in
...
 
   ls)
-    apps=$(ls -d $DOKKU_ROOT/*/ 2>/dev/null) || (echo "You haven't depoyed any applications yet" && exit 1)
-
-    dokku_col_log_info2_quiet "App Name" "Container id"
-
-    for app in $apps; do
-      if [[ -f $app/CONTAINER ]]; then
-        dokku_col_log_msg "$(basename $app)" "$(< $app/CONTAINER)"
+    dokku_apps=$(ls -d $DOKKU_ROOT/*/ 2>/dev/null) || (echo "You haven't depoyed any applications yet" && exit 1)
```
josegonzalez: `depoyed`

joshco:
Is it possible to detect if logging-supervisor is present?
If so, when it does exist, we could only scale web->1 to avoid multiplication.
When it does not exist, then behavior normally and scale each of the types defined in the Procfile -> 1

josegonzalez:
No, I would rather not have a special case for a specific plugin.

michaelshobbs:
I concur

joshco:
I concur too. After thinking abiut it more i agree too. 

I was thinking that thay plugin is makimg us behave in an unusual way (compared to heroku) but i was wrong. 

Only automatically starting web, even if others are mentioned in the Procfile is what heroku does too.
(Due to cost reasons as the first web dyno is free, the worker dynos cost $$)

Logging-supervisor, which starts all=1 is the unudual bird. 

So the heroku compat pillar of my argument has disappeared like a dokku destroyed app. 

I concede.

michaelshobbs:
@joshco :+1: :smile: 

new `dokku ls` output posted above. docker inspect output for state is a little weird and looks like this:
```
"State": {
        "Dead": false,
        "Error": "",
        "ExitCode": 0,
        "FinishedAt": "0001-01-01T00:00:00Z",
        "OOMKilled": false,
        "Paused": false,
        "Pid": 19734,
        "Restarting": false,
        "Running": true,
        "StartedAt": "2015-04-23T01:31:22.034345704Z"
    }
```

So I'm just show the output of `docker inspect -f '{{ .State.Running }}' $CID`

patrickjahns:
What is the benefit of docker inspect over docker ps to get the status of the container? I proposed in #1126 a different solution for showing the state

michaelshobbs:
heyo! I completely missed that PR. My bad. Honestly I just wanted to avoid additional output parsing logic...but I don't have strong enough feelings either way if we really want to have a true "status" column...

patrickjahns:
no worries - was just wondering. I support your output and I like the status column. In my opinion adds a good overview of what is running and what is not. I don`t have all deployed projects running all the time and this way I can quickly see what is running and what not.

joshco:
+1 on having a status column. Thats what im looking for the majority of the time im doing a dokku ls. 
Of course then i realize it isnt there and have to type docker ps

josegonzalez:
@michaelshobbs This looks good so far, anything blocking it in your mind? I assume you're using this in prod yourself and have seen some edge-cases/things you want to improve if there are blockers.

josegonzalez:
This needs to be rebased before we can merge. I'll make an interim release now and this can be in it's own release if you're okay with that, @michaelshobbs.

3onyc:
If I understand correctly Dockerfile apps just scale the container based on the image up/down to X instances?

michaelshobbs:
@josegonzalez as long as we feel comfortable with the response thus far, I'm good. I'll rebase now.

michaelshobbs:
@3onyc that is correct

josegonzalez:
@michaelshobbs please rebase and merge :)

joshco:
Release Party!

PR #1121, opened 2015-04-18
Prompt users to run the web installer via MOTD. Closes #943

### LINKED ISSUE #943: MOTD to move users to use the web installer
As much as it is possible to use the non-web installer, we should push users to that to avoid installation snaffus. An MOTD should work here i think.

josegonzalez:
Not sure on the wording here, so feel free to chime in with an alternative.


michaelshobbs:
:+1:

PR #1126, opened 2015-04-20
add app status to dokku ls

patrickjahns:
show running/stopped status in dokku ls

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -108,19 +108,25 @@ case "$1" in
     dokku apps:destroy $2
     ;;
 
-  ls)
+ ls)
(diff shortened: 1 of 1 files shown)


josegonzalez:
I'd rather this at least be optional output via an arg, if not a separate command entirely. The point of `dokku ls` is to list apps, and this changes the output in such a way that people depending upon it for scripting will have their scripts break.

patrickjahns:
For just listing apps there is still `dokku apps` right? 

What do you propose as parameterized solution? And also what other outputs do you see as relevant to something like ls? (to my mind comes now: IP? )

josegonzalez:
Closing this in favor of #1118 since we'll have something similar (better?) there :)

PR #1128, opened 2015-04-20
switching to herokuish. closes #1091

### LINKED ISSUE #1091: switch to herokuish + docker hub
The original reason we started using "prebuilt" stack image of buildstep is because Docker Hub was slow and did not compress images. Now, however, it does. In some cases it can be even faster than downloading a tarball. So firstly, we should switch back to the model of a simple `docker pull` from Docker Hub.

Second, Buildstep has become a compatibility wrapper for Herokuish. But just the other day, I merged most of those compatibility bits into Herokuish. I also added a Dockerfile that is basically the same as Buildstep. As such, Herokuish can now be used instead of Buildstep.

I would recommend using Herokuish from Docker Hub now instead of Buildstep. This will do several things:

 1. Get people submitting issues to Herokuish instead of Buildstep (super frustrating)
 1. Eliminate the mostly useless and no doubt confusing middle project of Buildstep
 1. Allow Buildstep to freely be refactored into [Buildkit](https://github.com/progrium/dokku/wiki/Refactoring#buildstep-to-become-buildkit)

The switch also means Dokku will more immediately and more easily take advantage of security patches to upstream base layers. Something I've been hearing more and more about recently.

Other benefits of Herokuish:

 * Herokuish is set up for better releases. Anybody can just submit a PR to release branch.
 * Herokuish is close to having fully automatic testing of newly released buildpacks.
 * Herokuish, among other things, is great for [debugging Dokku pushes](https://github.com/gliderlabs/herokuish#using-herokuish-to-test-herokudokku-apps).

Later on, Dokku can come back to using what would then be Buildkit, which will orchestrate Herokuish builds (less code for Dokku), and support other builds via pluggable build strategies. For example, this would move Dockerfile building out of Dokku (even less code). But it would also open up more possibilities, such as what we're calling runtime images. Or other custom build strategies.

By the way "runtime images" are an exciting development providing new middleground between Dockerfile and buildpacks. They're lightweight prebuilt Docker images with ONBUILD hooks for a similar experience to Buildpacks except produces MUCH MUCH smaller images, while not being a fully custom Dockerfile. So they're as easy as Buildpacks (no special code, auto-detected, etc) with most benefits of Dockefile builds (Docker native, much more efficient).

michaelshobbs:
naive switch to herokuish. @josegonzalez need help with the packaging stuff. maybe just drop it since its a simple `docker pull` now?


josegonzalez:
What do you mean by that? I'm on irc if it helps :)

michaelshobbs:
@progrium any reason you don't tag versions in docker hub? looks like the latest tag is the release branch but it would be nice to be able to pin to a specific version.

https://registry.hub.docker.com/u/gliderlabs/herokuish/builds_history/172630/

patrickjahns:
I think that we can drop the deb package for herokuish

michaelshobbs:
tests are failing on building herokuish from master.
```
Start building herokuish
Cloning into '/tmp/herokuish'...
remote: Counting objects: 905, done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 905 (delta 0), reused 0 (delta 0), pack-reused 900
Receiving objects: 100% (905/905), 990.77 KiB | 0 bytes/s, done.
Resolving deltas: 100% (334/334), done.
Checking connectivity... done.
Sending build context to Docker daemon 44.54 kB


Sending build context to Docker daemon 

Step 0 : FROM heroku/cedar:14
Pulling repository heroku/cedar


INFO[0000] Error pulling image (14) from heroku/cedar, Driver btrfs failed to create image rootfs 511136ea3c5a64f264b78b5433614aec563103b4d4702f3ba7d4d2698e22c158: Failed to create btrfs subvolume: file exists 

make: *** [stack] Error 1 ./tests/ci/parallel_runner.sh returned exit code 2
```

Awaiting a response from CircleCI.

progrium:
It can't be automated. What if we tried not pinning? Heroku doesn't pin
their environment.

On Tuesday, April 21, 2015, Michael Hobbs <notifications@github.com> wrote:

> @progrium <https://github.com/progrium> any reason you don't tag versions
> in docker hub? looks like the latest tag is the release branch but it would
> be nice to be able to pin to a specific version.
>
>
> https://registry.hub.docker.com/u/gliderlabs/herokuish/builds_history/172630/
>
> —
> <https://github.com/progrium/dokku/pull/1128#issuecomment-94594704>.
>


-- 
Jeff Lindsay
http://progrium.com

patrickjahns:
I am wondering if we should really rename the plugin hooks to include herokuish ??
This will force many users to update the plugins ...

Should we either keep the naming of buildstep and just change the image? Or if we already rename the hooks, can we rename them so they are not dependant on the docker image name ??

michaelshobbs:
I'm fairly ambivalent on the hook name. @josegonzalez thoughts?

josegonzalez:
@progrium It would make debugging issues much easier - oh you got this working in release 0.0.5 and 0.0.6 doesn't work? - if we pinned. Even if it's just "version 1" versus "version 2".

@michaelshobbs changing hook names here doesn't make sense, unless our next release is going to be 0.4.0. Which I'm fine with, just saying. We may want to hold off if we're going to move to buildkit though, since thats yet another name change :)

michaelshobbs:
I do think dokku would probably be best served to push this change in the next major release. However, if we're going to rename the hook, we should probably only do it once and name it something like `-buildpack` instead as that doesn't attach it to an underlying image name and is more reflective of what we're actually doing. i.e. `buildpack` vs. `Dockerfile`

patrickjahns:
I do agree - when it comes to plugins and hooks let`s create a feature outline for the major release and merge everything related to that together. So we break it only once

michaelshobbs:
@progrium ping on tagging releases?

michaelshobbs:
@progrium how would i go about building the herokish docker image from master as opposed to the release tgz referenced in the Dockerfile?

progrium:
I have a tool now that programmatically adds tags for versions on Docker
Hub for Automated Builds from CI, but I haven't had time to add it.

On Tue, Jun 2, 2015 at 7:30 PM, Michael Hobbs <notifications@github.com>
wrote:

> @progrium <https://github.com/progrium> how would i go about building the
> herokish docker image from master as opposed to the release tgz referenced
> in the Dockerfile?
>
> —
> <https://github.com/progrium/dokku/pull/1128#issuecomment-108140114>.
>



-- 
Jeff Lindsay
http://progrium.com

michaelshobbs:
@progrium it would also be cool to have a nightly build (based on master) for testing with dokku; similar to how we were testing with buildstep.

progrium:
Yeah, I'm going to make sure herokuish has `master`, `latest`, and specific version tags.

josegonzalez:
@michaelshobbs can you rebase this and also change the hooks to say `buildpack` instead of `herokuish`?

deb.mk
```diff
@@ -1,8 +1,8 @@
-BUILDSTEP_DESCRIPTION = 'Buildstep uses Docker and Buildpacks to build applications like Heroku'
-BUILDSTEP_REPO_NAME ?= progrium/buildstep
-BUILDSTEP_VERSION ?= 0.0.1
-BUILDSTEP_ARCHITECTURE = amd64
-BUILDSTEP_PACKAGE_NAME = buildstep_$(BUILDSTEP_VERSION)_$(BUILDSTEP_ARCHITECTURE).deb
+HEROKUISH_DESCRIPTION = 'Herokish uses Docker and Buildpacks to build applications like Heroku'
```
vincentfretin: Herokuish instead of Herokish :)

michaelshobbs:
@josegonzalez where do you want that changed?

plugins/00_dokku-standard/commands
```diff
@@ -10,27 +10,27 @@ case "$1" in
...
-      buildstep)
+      herokuish)
         id=$(tar -c . | docker run -i -a stdin $DOKKU_IMAGE /bin/bash -c "mkdir -p /app && tar -xC /app")
         test "$(docker wait $id)" -eq 0
         docker commit $id $IMAGE > /dev/null
         [[ -d $CACHE_DIR ]] || mkdir $CACHE_DIR
         # *DEPRECATED* in v0.4.0: `pluginhook pre-build` will be removed in future releases
         pluginhook pre-build "$APP"
-        pluginhook pre-build-buildstep "$APP"
+        pluginhook pre-build-herokuish "$APP"
```
josegonzalez: For instance, this hook :)

josegonzalez: I can go through the pr if you'd like and point out all the other hooks. TBH I was being lazy and should have done it myself, but given the rebase i figured you'd just handle it all in one :)

michaelshobbs: No worries. I'll get to it this week.

michaelshobbs:
currently there is no way to *actually* build herokuish from master because the dockefile is pinned to a release. i'm going to take a stab at a PR for that now.

michaelshobbs:
references https://github.com/gliderlabs/herokuish#35

michaelshobbs:
refs: https://github.com/gliderlabs/herokuish/pull/35

michaelshobbs:
Per IRC convo, we'll call the old pluginhooks as well for a patch release and then deprecate in 0.4.0.

josegonzalez:
What else does this depend upon for being released? Do I need to update our release process to do herokuish now?

michaelshobbs:
What are you referring to wrt the release process?

josegonzalez:
Well, do I need to make a deb package of herokuish now?

michaelshobbs:
Yeah I think so

michaelshobbs:
I think this is ready to go pending tests.

michaelshobbs:
I don't think this is bc-break anymore

michaelshobbs:
@josegonzalez ping. is the debian side of this ok? i don't know what i'm doing there. can you double check it before we merge?

michaelshobbs:
@josegonzalez the release branch of gliderlabs/herokuish needs to be merged to master to make the tests pass

josegonzalez:
You mean we need a new release? [The release branch has already been merge](https://github.com/gliderlabs/herokuish/branches).

michaelshobbs:
Hrm. We did something wrong then. 

https://github.com/gliderlabs/herokuish/pull/52

michaelshobbs:
@progrium we'll need to get the docker hub image built. can i get contributor access to gliderlabs/herokuish and docker hub access as well? (would be good if @josegonzalez had access as well)

progrium:
what is @josegonzalez's docker hub user?

josegonzalez:
`catmilla` is my docker hub user (maybe `josegonzalez`, but lord knows docker makes it difficult to figure that out).

josegonzalez:
Seems legit with ^ fix :) . I'll make one more small release and then merge this in for a new release!

PR #1137, opened 2015-04-22
Safer installation method

YellowApple:
Piping the output of `curl` or `wget -O-` to a shell is a very bad idea for multiple reasons; security concerns aside, doing so provides no opportunity for the script's download to finish before execution starts, so if the download is interrupted for some reason (perhaps because of an intermittent network failure), the incompletely-downloaded script will still run with potentially disastrous results.  See [here](http://www.seancassidy.me/dont-pipe-to-your-shell.html) for a more in-depth description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,10 @@ Docker powered mini-Heroku. The smallest PaaS implementation you've ever seen. S
 
 ## Installing
 
-To install the latest stable release, you can run the following command as a user that has access to `sudo`:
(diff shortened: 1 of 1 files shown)


josegonzalez:
FWIW, that command would run the following:

- https://github.com/progrium/dokku/blob/master/bootstrap.sh#L33

Which is what is in our [debian installation notes](http://progrium.viewdocs.io/dokku/getting-started/install/debian).

PR #1138, opened 2015-04-22
Include tip about using sshcommand acl-add

chrisbutcher:
Playing with dokku for the first time in a while, I found that this was a step we could better assist new users through.


josegonzalez:
Did you use the web installer to add the key? It prompts you for it.

chrisbutcher:
I may have missed something else. I was only seeing 'Welcome to nginx' after install. There was no url provided to get to the installer page.

josegonzalez:
How did you install it?

chrisbutcher:
```
wget https://raw.github.com/progrium/dokku/v0.3.17/bootstrap.sh
sudo DOKKU_TAG=v0.3.17 bash bootstrap.sh
```

josegonzalez:
Gotcha, what host was this on?

chrisbutcher:
Digital Ocean, Ubuntu 14.04x64

josegonzalez:
On digital ocean, there is a ready-made image that should work.

PR #1144, opened 2015-04-24
Allow git-remote with different port

LAndreas:
If you want to push data to dokku via an ssh-daemon running on a different port (eg. 22333) you need to add something like ```ssh://<ip or domain>:22333/<appname>``` as a git-remote.

The problem now is: dokku will handle ```/<appname>``` as the appname. This works fine while initializing a new repository but fails after the appname is passed to ```git-receive-pack```. (Cause there is no valid git-repository at ```/<appname>```)

This PR removes a leading '/' from the appname and pass modified arguments to ```git-shell``` if the first argument was ```git-receive-pack```.


ze-phyr-us:
Just hit the same problem. Hope this gets merged soon!

michaelshobbs:
@LAndreas I'm going to submit a PR to your fork that adds testing to this.

michaelshobbs:
https://github.com/protonet/dokku/pull/1

LAndreas:
@michaelshobbs thank you! I've merged the PR :-)

PR #1148, opened 2015-04-27
Add information about 444 for nginx in default_sever.

TheEmpty:
Something I would've found helpful. Also I'm not a fan of using 410 over 404. Not sure how that handles with SEO if I were to decide to use a site on that domain. 404 is the default behavior of NGINX.


josegonzalez:
:+1:

PR #1149, opened 2015-04-28
pluginhook to receive branches different than master

mlebkowski:
I need an option to receive multiple versions of the same application in an easy for the developer way (so branches instead of remotes). This patch allows me to do this:

```bash
#!/bin/bash

reference_app=$1
refname=$3
newrev=$2
APP=${refname/*\//}.$reference_app

if [[ ! -d "$DOKKU_ROOT/$APP" ]]; then
  REFERENCE_REPO="$DOKKU_ROOT/$reference_app
  git clone --bare --shared --reference "$REFERENCE_REPO" "$REFERENCE_REPO" "$DOKKU_ROOT/$APP" > /dev/null
fi
pluginhook receive-app $APP $newrev
```

I would leave the warning in the code but I don’t know how to check if any plugin handled the event. I’d use the same approach as with `DOKKU_NOT_IMPLEMENTED_EXIT`, but it’s too much hassle.

### DIFF
--- a/plugins/git/commands
+++ b/plugins/git/commands
@@ -39,8 +39,12 @@ case "$1" in
         # broken out into pluginhook so we might support other methods to receive an app
         pluginhook receive-app $APP $newrev
       else
-        echo $'\e[1G\e[K'"-----> WARNING: deploy did not complete, you must push to master."
-        echo $'\e[1G\e[K'"-----> for example, try 'git push <dokku> ${refname/refs\/heads\/}:master'"
+        if test -f "$PLUGIN_PATH"/enabled/*/receive-branch; then
+          pluginhook receive-branch $APP $newrev $refname
+        else
+          echo $'\e[1G\e[K'"-----> WARNING: deploy did not complete, you must push to master."
+          echo $'\e[1G\e[K'"-----> for example, try 'git push <dokku> ${refname/refs\/heads\/}:master'"
+        fi
       fi
     done
     ;;


alessio:
FWIW -1 on this
This represents a serious breakage of the contract provided by the deployment interface to me.

mlebkowski:
@alessio, it doesn’t change anything by default. My script is just an example. If it doesn’t fit your point of view, you don’t have to use it, that’s the beauty of pluginhooks :)

josegonzalez:
I wonder, can we provide a fallback pluginhook that provides the current output?

michaelshobbs:
you could do something like
```
if test -f /var/lib/dokku/plugins/enabled/*/receive-branch; then
  pluginhook receive-branch $APP $newrev $refname
else
  echo $'\e[1G\e[K'"-----> WARNING: deploy did not complete, you must push to master."
  echo $'\e[1G\e[K'"-----> for example, try 'git push <dokku> ${refname/refs\/heads\/}:master'"
fi
```

josegonzalez:
@mlebkowski mind updating your PR with @michaelshobbs suggestion? I'm a fan of that :)

mlebkowski:
I’ve updated the patch with @michaelshobbs suggestion

josegonzalez:
Can you restart your tests?

michaelshobbs:
This might need to be rebased to get tests to pass....

mlebkowski:
@michaelshobbs, yeah, I just realised that…

plugins/git/commands
```diff
@@ -39,8 +39,12 @@ case "$1" in
         # broken out into pluginhook so we might support other methods to receive an app
         pluginhook receive-app $APP $newrev
       else
-        echo $'\e[1G\e[K'"-----> WARNING: deploy did not complete, you must push to master."
-        echo $'\e[1G\e[K'"-----> for example, try 'git push <dokku> ${refname/refs\/heads\/}:master'"
+        if test -f /var/lib/dokku/plugins/enabled/*/receive-branch; then
```
michaelshobbs: This should probably be `$PLUGIN_PATH` and not `/var/lib/dokku/plugins`.

michaelshobbs:
@mlebkowski will you add this plugin hook to the docs?

PR #1151, opened 2015-04-28
Do not kill docker container with SIGKILL

LTe:
When we will use `kill` command on `docker` main process will receive
`SIGKILL` signal. This will result in immediate termination of
processes. We need to send a signal `SIGTERM` to main process and give
the opportunity to close application in correct way.

### DIFF
--- a/dokku
+++ b/dokku
@@ -91,7 +91,7 @@ case "$1" in
 
     # if we can't post-deploy successfully, kill new container
     kill_new() {
-      docker inspect $id &> /dev/null && docker kill $id > /dev/null
+      docker inspect $id &> /dev/null && docker stop $id > /dev/null && docker kill $id > /dev/null
       trap - INT TERM EXIT
       kill -9 $$
     }
@@ -120,6 +120,7 @@ case "$1" in
         exec >/dev/null 2>/dev/null </dev/null
         trap '' INT HUP
         sleep $WAIT
+        docker stop $oldid
         docker kill $oldid
       ) & disown -a
       # Use trap since disown/nohup don't seem to keep child alive


dokku
```diff
@@ -120,7 +120,7 @@ case "$1" in
         exec >/dev/null 2>/dev/null </dev/null
         trap '' INT HUP
         sleep $WAIT
-        docker kill $oldid
+        docker stop $oldid
```
michaelshobbs: In practice this does not work consistently. See the open PR for scaling. (I'm on mobile. Can't remember the number) 

Instead I think we want to attempt `stop` (default timeout is 10 sec) then `kill`.

LTe: @michaelshobbs make sense. So:

```bash
docker stop $oldid
sleep $WAIT_FOR_KILL
docker kill $oldid
```

michaelshobbs: Just the second one. 

From https://docs.docker.com/reference/commandline/cli/#stop
```
Usage: docker stop [OPTIONS] CONTAINER [CONTAINER...]

Stop a running container by sending SIGTERM and then SIGKILL after a
grace period

  -t, --time=10      Seconds to wait for stop before killing it
```

LTe: @michaelshobbs updated

dokku
```diff
@@ -91,7 +91,7 @@ case "$1" in
 
     # if we can't post-deploy successfully, kill new container
     kill_new() {
-      docker inspect $id &> /dev/null && docker kill $id > /dev/null
+      docker inspect $id &> /dev/null && docker stop $id > /dev/null && docker kill $id > /dev/null
```
3onyc: You don't need `docker kill` as `docker stop` already sends a SIGKILL after 10 seconds<sup>1</sup>.

1. https://docs.docker.com/reference/commandline/cli/#stop

michaelshobbs:
I jumped the gun here and missed the change to `kill_new()`. There is no reason (that I can see) to `docker stop` a new container that failed CHECKS. @LTe would you mind following up with reverting that particular bit?

LTe:
@michaelshobbs sure, I will create pull request for that.

PR #1155, opened 2015-04-29
Better, actually working, git-rev plugin

cjblomqvist:
The previous one broke deployment on first deployment and was always one (relative deploys) REV behind when "working".


michaelshobbs:
@cjblomqvist need rebase

cjblomqvist:
Created a new PR instead: #1168

PR #1160, opened 2015-05-01
Allow to kill new container that failed CHECKS

LTe:
When the check does not pass there is no reason to close the container
in the correct way. We can do this by using SIGKILL signal.

Related: #1151

### DIFF
--- a/dokku
+++ b/dokku
@@ -91,7 +91,7 @@ case "$1" in
 
     # if we can't post-deploy successfully, kill new container
     kill_new() {
-      docker inspect $id &> /dev/null && docker stop $id > /dev/null && docker kill $id > /dev/null
+      docker inspect $id &> /dev/null && docker kill $id > /dev/null
       trap - INT TERM EXIT
       kill -9 $$
     }


3onyc:
The checks failing only indicate that the service is not behaving as expected, it may still have things running that need to clean up before exiting.

LTe:
@3onyc So the current implementation is correct?

3onyc:
@LTe In my opinion, yes.

michaelshobbs:
@3onyc This PR restores the previous functionality due to the assumption that the if the new container fails checks it's better to fail quickly as we no something went wrong. Can you provide more information as to why you think the current `stop` first method is correct?

michaelshobbs:
@3onyc bump

3onyc:
Stop allows a container to properly clean up, CHECKS failing only indicates that something is wrong, and does not indicate the app failed to start completely.

I can't currently think of a specific example, but giving containers a chance to clean up doesn't sound like a bad thing.

josegonzalez:
I think that's fair. If we have background containers, they may work while the web version won't. Therefore, I'm :-1: on this PR.

josegonzalez:
@LTe thanks for being patient with us :)

joshco:
+1 on the decision to keep the stop behavior.

A rails example would be a job processor, such as sideqik.  Upon crash, the OSS version of sidekiq can lose jobs that have been loaded into memory.  So in a CHECKS failing container, the job processor could have started up and begun loading jobs, only to be forcibly terminated (crash) thus losing those jobs

> https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#my-sidekiq-process-is-crashing-what-do-i-do

PR #1166, opened 2015-05-03
Correct "upgrade to debian package-based installation"

Benjamin-Dobell:
Fix documentation error and work-around nginx issue identified in https://github.com/progrium/dokku/issues/1165

Complete solution for the nginx issue is out-of-scope.


josegonzalez:
Don't merge this just yet.

josegonzalez:
This was fixed in 626cd958a51a89e37432b9e9632e90a174ee71ab. Thanks for the PR though!

PR #1167, opened 2015-05-03
Add curl to install-from-deb make command

3onyc:
install-from-deb doesn't install curl, while it's needed to add the Docker GPG keys.

This also fixes provisioning on the Vagrant dokku-deb box, as that uses install-from-deb.


josegonzalez:
Don't merge this just yet.

3onyc:
Implemented by #1190

PR #1168, opened 2015-05-04
Better, actually working, git-rev plugin

cjblomqvist:
The previous one broke deployment on first deployment and was always one (relative deploys) REV behind when "working".


josegonzalez:
Did you fork the other one? If so, maybe make a PR to it?

cjblomqvist:
I actually did, see:
https://github.com/nornagon/dokku-git-rev/pull/2
https://github.com/nornagon/dokku-git-rev

I have no personal preference which repo is pointed to, no more than that I think it'd be better if the repo it points to works (so that others don't have to go through the same hassle I did with it).

Btw. really appreciate the help in the IRC channel the other day!

Issue #1161, opened 2015-05-01
Install deb package without using web installer?


3onyc:
I'm installing Dokku using the deb package, but looking at the source it seems I can't bypass the web installer (Have it not initialise/run). Am I overlooking something or is this not yet implemented?

josegonzalez:
The [web installer](https://github.com/progrium/dokku/blob/master/contrib/dokku-installer.rb#L38):

- Writes out or removes a `VHOST` file
- Writes out a `HOSTNAME` file
- Calls `sshcommand acl-add dokku admin` with your key on stdin
- Removes the dokku-installer nginx conf and removes its init file.

You could do all of that to bypass a web installer. If you want to disable it straightaway, thats not implemented, though I would consider a pull request that created `/etc/default/dokku` with some env var to disable it.

3onyc:
Is there any reason why this is done using a web interface? as the same could be accomplished through a prompt during installation, which would also remove the dependency on ruby.

Not sure what we could do if we want to bypass the prompts, I don't know what the idiomatic Debian packaging way of handling that is.

josegonzalez:
It's done using the web interface as its more user friendly for the common user of dokku (seriously).

If you'd like to modify the debian package to prompt for this information, sure, but doing so probably means you want an automated way to do this, in which case a defaults file is the way to go...

3onyc:
It would be for just the Debian package yes, and a config option could be added to disable the web-based config.

As far as I can tell [debconf](http://www.fifi.org/doc/debconf-doc/tutorial.html) would be the idiomatic way of handling this.

Values can be set before install, and can be prompted for during install.

What do you think? If it sounds like the right way to go, I'll look into it.

josegonzalez:
Go for it

josegonzalez:
Closed by #1171.

PR #1171, opened 2015-05-05
[RFC] Use debconf for package configuration

3onyc:
This PR implements debconf to allow for command-line or unattended installation.

### Config variables

| Name | Type | Default | Description |
| ------- | ------ | ---------- | -------------- |
| dokku/web_config | boolean | true | Use web-based config, if yes other options will be skipped |
| dokku/vhost_enable | boolean | false | Use vhost based deployments or not |
| dokku/hostname | string | dokku.me | Hostname that Dokku should use |
| dokku/key_file | string | /root/.ssh/id_rsa.pub | Key to add to the Dokku user (Will not be asked on `dpkg-reconfigure` |

### How to pre-seed variables

```shell
debconf-set-selections <<< "dokku dokku/web_config boolean false"
```

### Installation workflow
*Note: this uses my own PackageCloud repository, as that has the updated packages*

```shell
#!/bin/bash

# Do your debconf-set-selections stuff here
# debconf-set-selections <<< "dokku dokku/web_config boolean false"

apt-get update -qqy
apt-get install -qqy curl

curl --silent https://get.docker.io/gpg | apt-key add - >/dev/null
curl --silent https://packagecloud.io/gpg.key | apt-key add - >/dev/null

echo "deb http://get.docker.io/ubuntu docker main" > /etc/apt/sources.list.d/docker.list
echo "deb https://packagecloud.io/x3tech/dokku/ubuntu/ trusty main" > /etc/apt/sources.list.d/dokku.list

apt-get update -qqy
apt-get install -qqy linux-image-extra-`uname -r` apt-transport-https dokku
```

Related Issues:
* #1161 

Progress:

- [x] Configuration through debconf is implemented
- [x] Part of logic to handle configuration
- [x] Logic for switching from vhost to non-vhost config 
- [x] Don't ask for ssh key to import on reconfigure

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 # If you're submitting a patch, please add your name here in alphabetical order as part of the patch.
 # This file lists all individuals having contributed content to the repository.
 #
+3onyc <3onyc@x3tech.com>
 Alexander <iam.asm89@gmail.com>
 Alexander Beletsky <alexander.beletsky@gmail.com>
 Alexis Gavoty <kload@kload.fr>
--- a/debian/config
+++ b/debian/config
@@ -1,0 +1,21 @@
+#!/bin/sh -e
+
+. /usr/share/debconf/confmodule
+
+readonly ACTION="${1:-configure}"
+readonly VERSION="${2:-dev}"
+
+db_input "high" "dokku/web_config" || true
+db_go || true
+
+db_get "dokku/web_config"
+if [ $RET = "true" ]; then
+    exit 0
+fi
+
+db_input "high" "dokku/hostname" || true
+db_input "high" "dokku/vhost_enable" || true
+if [ "$ACTION" != "reconfigure" ]; then
+  db_input "high" "dokku/key_file" || true
+fi
+db_go || true
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,6 +1,10 @@
 #!/bin/sh
 set -e
 
+. /usr/share/debconf/confmodule
+
+readonly DOKKU_ROOT="${DOKKU_ROOT:-/home/dokku}"
+
 case "$1" in
   abort-upgrade|abort-remove|abort-deconfigure)
     ;;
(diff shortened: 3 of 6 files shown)


josegonzalez:
What if I want an unattended install? How would I go about doing that?

3onyc:
`debconf-set-selections` can be used to set the configuration values before the package is installed, e.g.

```shell
debconf-set-selections <<< "dokku dokku/web_config boolean false"
```

Note that I haven't tested/implemented handling the config vars yet, so currently there's not anything to test.

3onyc:
@josegonzalez Considering that #1167 is not to be merged just yet, and that I'm extensively modifying the makefiles for this, are you working on anything makefile related I should keep in mind?

Vagrantfile
```diff
@@ -42,7 +42,7 @@ Vagrant::configure("2") do |config|
     vm.vm.network :forwarded_port, guest: 80, host: ENV['FORWARDED_PORT']
     vm.vm.hostname = "#{DOKKU_DOMAIN}"
     vm.vm.network :private_network, ip: DOKKU_IP
-    vm.vm.provision :shell, :inline => "cd /root/dokku && make install-from-deb"
+    vm.vm.provision :shell, :inline => "cd /root/dokku && make install-from-deb && usermod -a -G docker vagrant"
```
josegonzalez: Why are we creating these users?

josegonzalez:
It looks like you've expanded the scope of this issue significantly. While I applaud that, maybe the makefile changes can be moved out?

Note: [Here](http://progrium.viewdocs.io/dokku/getting-started/install/debian) is the workflow I'd want for installing via debian packages.

3onyc:
I got carried away in modifying the Makefiles to speed up building the packages, for this PR only look at the `[WIP] debconf` commit as I'll remove the rest for the final pull.

I'll clean-up and finish the other changes and submit separate pull requests for them then, following that workflow example.

3onyc:
Status so far: 

* DONE: Configuration through debconf is implemented
* DONE: Part of logic to handle configuration
* TODO: Logic for switching from vhost to non-vhost config 
* TODO: Further testing

3onyc:
ping @josegonzalez

josegonzalez:
Can you update the debian install docs?

3onyc:
Updated, I'm not a good writer so let me know if there's anything I should change.

josegonzalez:
With the new deb package, should we also remove any reference to the web installer? I think so.

3onyc:
Currently the default is to still run the web installer, as in #1161 you mentioned that being the preferred way to do things.

josegonzalez:
So this won't prompt during installation?

3onyc:
During normal installation it'll ask whether you want to use the web installer, if the answer to that is yes, no further questions are asked and the other debconf values are ignored.

When the user chooses not to use the web installer, the questions for vhost/hostname/key are asked through debconf and handled in the preinst and postinst scripts.

When running non-interactively, the values configured with `debconf-set-selections`, or the defaults as listed in the main post are used instead.

josegonzalez:
I'm starting to think that having multiple ways of configuring is actually bad. I go back on my previous statement. Can you just make it cli-only to install? And remove the web installer?

3onyc:
Whoops

PR #1178, opened 2015-05-11
Update plugins.md

jagandecapri:
Update plugins.md with another Bower/Gulp plugin.


josegonzalez:
Why this new one?

jagandecapri:
The current one does not work well. I referred to [thrashr888/dokku-bower-grunt-build-plugin](https://github.com/thrashr888/dokku-bower-grunt-build-plugin) and made some small modifications as it is updated to current plugin hooks and has more complete commands to build.

PR #1180, opened 2015-05-12
Update docker-options.md

sherbondy:
My changes may not be quite correct, but as far as I can tell, DOCKER_OPTIONS now requires specifying a phase explicitly.
Also updated references to `--link` since `-link` has been deprecated in Docker.


josegonzalez:
It doesn't require, it lets you optionally specify one.

vincentfretin:
@josegonzalez No, the phase is definitely required, see the code.
+1 to merge.

josegonzalez:
Oh crap, I was reading the `docker-options` command, completely missed the others.

I'd definitely be +1 on adding the ability to ignore the phase, but yeah, good catch.

PR #1202, opened 2015-05-20
filter out literal wildcard when deploying an unrelated domain. closes #1185

### LINKED ISSUE #1185: Wildcard domain always added to app
I have this issue where *.domain.com always get added to a newly pushed app when using a VHOST, why is this happening?

Have tried multiple ways of removing it, manually removing it from VHOST doesn't seem to work, using the dokku command doesn't do anything. How do I get rid of it?

There is an SSL certificate for *.domain.com in /home/dokku/tls, may this be causing the issue?

michaelshobbs:
Filtering out the literal strings that start with `*.` seemed to be the simplest fix here.

@lmars @josegonzalez @vincentfretin


josegonzalez:
Seems legit to me. Does this work for you @lmars ?

lmars:
LGTM

PR #1206, opened 2015-05-22
Handle for installs in /usr/local/bin and the like.

rockymadden:
I wish to install the client as minimally as possible, without cloning the entire repo, via:

```bash
curl -o /usr/local/bin/dokku https://raw.githubusercontent.com/progrium/dokku/master/contrib/dokku_client.sh && chmod 0755 /usr/local/bin/dokku
```

However, the commit made is needed for this approach to work. If the client is ever included in something like homebrew, this should work there as well.


josegonzalez:
Can you fix the lint issue?

rockymadden:
Yep, one second.

michaelshobbs:
@rockymadden can you rebase this with master for us? CircleCI made some changes that broke our previous test config.

PR #1207, opened 2015-05-23
Fixed bug with client commands taking verb, appname, and also arguments.

rockymadden:
When parsing args when handling verbs, appname can be included as an argument causing bugs with commands that accept a verb, appname, and also arguments.


josegonzalez:
General question: are there cases where this might be desired?

PR #1217, opened 2015-06-05
Fix broken ssl wildcard redirect

kirushanth-sakthivetpillai:
Example Actual:
http://example.com  >  https://example.com  (works as it should)
http://test.example.com  >  https://\*.example.com  (BROKEN)
http://api.example.com  >  https://\*.example.com  (BROKEN)
http://app.example.com  >  https://\*.example.com  (BROKEN)
( https://*.example.com may appear as https://%2A.example.com depending on browser)

Example Expected:
http://example.com  >  https://example.com  (works as it should)
http://test.example.com  >  https://test.example.com  (Expected)
http://api.example.com  >  https://api.example.com  (Expected)
http://app.example.com  >  https://app.example.com  (Expected)

nginx redirects to https://\*.example.com/ which is a bogus url due to the $SSL_SERVER_NAME in the nginx template
Use nginx variable $host to do proper redirect

### DIFF
--- a/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
+++ b/plugins/nginx-vhosts/templates/nginx.ssl.conf.template
@@ -2,7 +2,7 @@ server {
   listen      [::]:80;
   listen      80;
   server_name $NOSSL_SERVER_NAME;
-  return 301 https://$SSL_SERVER_NAME\$request_uri;
+  return 301 https://\$host\$request_uri;
 }
 
 server {


michaelshobbs:
Are you sure this is still an issue in master? Ostensibly this was fixed via #1202

kirushanth-sakthivetpillai:
Im using Dokku v0.3.17
Still seems to be an issue. My use case might be special since I have one dokku instance that accepts multiple sub domain urls, so I need the wildcard ssl redirect to maintain the original url.

Even then, using nginx $host is a better solution than just dumping the domain variable from the bash script into the nginx config

michaelshobbs:
As a side note, you could also provide your own template for these apps instead. 

Your explanation makes sense. I'm wondering if we're breaking any assumptions or expected behaviors with this change. 

@lmars @josegonzalez thoughts on this?

josegonzalez:
I'm okay with this change, makes sense to me, so long as tests pass :)

michaelshobbs:
:+1:

kirushanth-sakthivetpillai:
:+1:

PR #1218, opened 2015-06-05
add dokku-logspout plugin

michaelshobbs:


joshco:
Neat.  Can this do file based stuff like logging supervisord?  eg, just have everything show up in /var/log/dokku/appname/web.log

michaelshobbs:
Currently it's just a manager for logspout and logspout is merely a container output router. I'm not particularly in the know of it's entire feature set but its main function is to redirect container output to a syslog server endpoint.

PR #1219, opened 2015-06-05
adding checks for curl and sudo.

cellularmitosis:
Hi,

I just tried installing dokku using Debian Jessie i386.

bootstrap.sh failed because curl and sudo aren't part of the default installation.

I've added a check for them.  Can you pull this in?

Thanks,
Jason

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -43,11 +43,14 @@ dokku_install_package() {
   curl -sSL https://get.docker.io/gpg | apt-key add -
   curl -sSL https://packagecloud.io/gpg.key | apt-key add -
 
+  sudo apt-get update > /dev/null
+  sudo apt-get install -qq -y apt-transport-https
+  
   echo "deb http://get.docker.io/ubuntu docker main" > /etc/apt/sources.list.d/docker.list
   echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ trusty main" > /etc/apt/sources.list.d/dokku.list
 
   sudo apt-get update > /dev/null
-  sudo apt-get install -qq -y "linux-image-extra-$(uname -r)" apt-transport-https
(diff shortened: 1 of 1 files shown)


cellularmitosis:
Also, apt-transport-https was being installed after an https source was added.  That has to happen the other way around.

bootstrap.sh
```diff
@@ -14,6 +14,14 @@ if ! command -v apt-get &>/dev/null; then
...
+for CMD in curl sudo
+do
+  if ! command -v ${CMD} &>/dev/null; then
+    echo "Error: ${CMD} not found." >&2
+    exit 1
+  fi
+done
+
 apt-get update
 which curl > /dev/null || apt-get install -qq -y curl
```
josegonzalez: Curl is installed here though. And why wouldn't you have sudo?

bootstrap.sh
```diff
@@ -35,11 +43,14 @@ dokku_install_package() {
   curl -sSL https://get.docker.io/gpg | apt-key add -
   curl -sSL https://packagecloud.io/gpg.key | apt-key add -
 
+  sudo apt-get update > /dev/null
+  sudo apt-get install -qq -y apt-transport-https
+
```
josegonzalez: Good catch.

bootstrap.sh
3onyc: There's already a check for curl so this one isn't needed.

@josegonzalez Best to assume as little as possible, getting an early error message when sudo is missing is better than some command failing.

josegonzalez: We should instead `apt-get install sudo` then, no?

josegonzalez:
We've since update the bootstrap script to avoid calling sudo within the script. #1401 takes care of part of this pr, and we ensure curl is available (or install it) early on. Thanks for the pr though!

PR #1225, opened 2015-06-10
allow tagging and deployment of tagged app images

michaelshobbs:
as with most "simple" things, this was not as straightforward as we've been fairly lax about how $IMAGE is set.

- add tag plugin, documentation & tests
    - tag:add
    - tag:deploy
    - tag:list
    - tag:rm
- consolidate setting `$IMAGE` to `verify_app_name()`
- extend `verify_app_name()` to accept `$TAG` (image tag)
- created `get_running_image_tag()`
- consolidate `release`/`deploy` calls to `release_and_deploy()`
- add `$TAG` to appropriate pluginhook calls
    - `pre-deploy`
    - `docker-args-deploy`
    - `docker-args-run`
    - `pre-release`
    - `pre-release-buildstep`
    - `pre-release-dockerfile`
    - `post-release`
    - `post-release-buildstep`
    - `post-release-dockerfile`
- make apps:destroy container/image cleanup more robust (read: actually destroy an app and *all* its artifcats)


josegonzalez:
Do images get tagged on deploy atm? Would be nice to do that automatically.

michaelshobbs:
Tagging is not automatic at the moment. I wanted to provide the method to do it first before automating it.

I am also not sure what makes sense as far as naming and retention. Automatic tagging could definitely eat up disk space quickly.

josegonzalez:
Ah in that you'd have things to rollback to etc.

Are there new pluginhooks for when an image is tagged?

michaelshobbs:
No new pluginhooks here

josegonzalez:
GIVE ME PLUGINHOOKS

josegonzalez:
plz

michaelshobbs:
Can you be more specific about what functionality you'd like? Example use-case?

josegonzalez:
I'd like to be able to:

- push a tag after it has been created to a docker hub
- remove tags from docker hub if they have been removed on a server

Basically it.

michaelshobbs:
Done.

michaelshobbs:
@josegonzalez thoughts on where to communicate to plugin devs that they need to get the image name + tag from get_running_image() when developing non *-build plugins that require the image name?

josegonzalez:
Uh what? I'm not sure I understand your question.

michaelshobbs:
Yeah sorry I was poorly multi-tasking when I wrote that. 

This PR exposes to the user control over the `$IMAGE` variable that was previously hard-coded (in multiple places) to `dokku/$APP`. Currently plugins that need to modify the app image also hard-code this variable. This practice will need to change for those plugins in order to modify the correct image when tags are in use. So my question is, how should we communicate this change in behavior?

Does that make more sense?

josegonzalez:
That sounds like a breaking change, and we should make a 0.4.0 release to handle it imo.

michaelshobbs:
Totally agree with you there. We should probably call it out in the release notes for 0.4.0 and update the plugin dev docs to call it out specifically. I'll do the latter now and update this PR.

josegonzalez:
Anything else we want to do for 0.4.0? We may as well do that now-ish.

michaelshobbs:
what about buildkit and plugn?

https://github.com/progrium/dokku/wiki/Refactoring

michaelshobbs:
@josegonzalez this needs another code review pass given its age

docs/development/pluginhooks.md
```diff
@@ -447,13 +456,16 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
+- Arguments: `$APP $TAG`
 - Example:
 
 ```shell
 #!/usr/bin/env bash
 
+source "$PLUGIN_PATH/common/functions"
+APP="$1"; TAG=$(get_running_image_tag $APP)
+verify_app_name "$APP" "$TAG"
```
josegonzalez: Do all our built-in hooks follow this pattern?

plugins/common/functions
```diff
@@ -114,23 +114,38 @@ dokku_container_log_verbose_quiet() {
 }
 
 verify_app_name() {
-  local APP="$1"
+  local APP="$1"; local TAG="$2"
   [[ ! -n "$APP" ]] && dokku_log_fail "(verify_app_name) APP must not be null"
   [[ ! -d "$DOKKU_ROOT/$APP" ]] && dokku_log_fail "App $APP does not exist"
-  IMAGE="dokku/$APP"
+
+  if [[ -n "$TAG" ]]; then
```
josegonzalez: Should we verify the format of a tag?

plugins/common/functions
```diff
@@ -114,23 +114,38 @@ dokku_container_log_verbose_quiet() {
...
     return 1
   fi
 }
 
+get_running_image_tag() {
+  local APP="$1"
+  [[ ! -n "$APP" ]] && dokku_log_fail "(get_running_image_tag) APP must not be null"
+  verify_app_name "$APP"
+
+  CIDS=( $(get_container_ids $APP) )
```
josegonzalez: What happens if `CIDS` is empty?

plugins/common/functions
```diff
@@ -267,3 +282,23 @@ is_app_running() {
     return 1
   fi
 }
+
+release_and_deploy() {
+  local APP="$1"; local TAG="$2"; verify_app_name "$APP" "$TAG"
```
josegonzalez: Newline the method call?

plugins/common/functions
```diff
@@ -267,3 +282,23 @@ is_app_running() {
...
+    else
+      IMAGE_SOURCE_TYPE="dockerfile"
+    fi
+
+    dokku_log_info1 "Releasing $APP ($IMAGE)..."
+    dokku release "$APP" "$IMAGE_SOURCE_TYPE" "$TAG"
+    dokku_log_info1 "Deploying $APP ($IMAGE)..."
+    dokku deploy "$APP" "$TAG"
+    dokku_log_info2 "Application deployed:"
+         dokku urls $APP | sed "s/^/       /"
```
josegonzalez: Does the `dokku urls` need to be indented?

plugins/config/docker-args-deploy
```diff
@@ -3,8 +3,9 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_PATH/common/functions"
 source "$PLUGIN_PATH/config/functions"
 
-STDIN=$(cat); APP="$1"; IMAGE="dokku/$APP"
+STDIN=$(cat); APP="$1"; TAG="$2"
 DOCKERFILE_ENV_FILE="$DOKKU_ROOT/$APP/DOCKERFILE_ENV_FILE"
+verify_app_name "$APP" "$TAG"
 
 if ! is_image_herokuish_based "$IMAGE"; then
```
josegonzalez: `IMAGE` doesn't appear to be set?

josegonzalez: Oh nvm, it's exported from `verify_app_name`. That seems weird and not something I'd expect...

plugins/tag/functions
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+tag_image() {
```
josegonzalez: No test for this?

plugins/common/functions
michaelshobbs: Since a tag will be applied to a docker image, I'm letting docker do that for us.

plugins/common/functions
michaelshobbs: This was cut straight from `dokku`.

josegonzalez: Lets de-dent? I'm assuming thats fine with this, right?

plugins/config/docker-args-deploy
michaelshobbs: Yeah this requires a central place for setting `$IMAGE`. I suppose we can break it up into another call.

something like
```
STDIN=$(cat); APP="$1"; TAG="$2"
verify_app_name "$APP"
IMAGE=$(get_app_image "$APP" "$TAG")
```

plugins/tag/functions
michaelshobbs: What do you mean? We test in `tests/unit/tag.bats`/

michaelshobbs: Still I'll move it to the commands file since it's not shared.

docs/development/pluginhooks.md
michaelshobbs: Just ones that use an image but depending on your thoughts on how `$IMAGE` is set we can revert this and make a new function call to get the app image name and tag.

plugins/config/docker-args-deploy
josegonzalez: I'd prefer this, since it's more "intuitive" as to what is going on.

plugins/common/functions
josegonzalez: Gotcha, :+1:

docs/development/pluginhooks.md
josegonzalez: Yeah I'd rather it be clear as to how variables are set than placing magic in random places like this ;)

plugins/common/functions
michaelshobbs: Then we return an empty string

michaelshobbs:
@josegonzalez lemme know what you think

plugins/tag/commands
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
```
josegonzalez: We should use `$PLUGIN_PATH` here instead of `$(dirname $0)/..`...

plugins/apps/pre-delete
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
```
josegonzalez: Same here re: `$PLUGIN_PATH`

plugins/enter/commands
```diff
@@ -4,7 +4,7 @@ source "$(dirname "$0")/../common/functions"
```
josegonzalez: Mind also modifying this one as well?

josegonzalez:
A few more comments. Once they are addressed, I think this is good to merge :)

plugins/tag/commands
```diff
@@ -0,0 +1,69 @@
...
+        *)
+            docker rmi "$IMAGE:$IMAGE_TAG"
+            ;;
+    esac
+    pluginhook tag-remove $APP $IMAGE_TAG
+    ;;
+
+  help | tag:help)
+    cat && cat<<EOF
+    tag:add <app> <tag>                             Add tag to latest running app image
```
u2mejc: Can you update the with the new ", " convention instead of the hard white spacing? I'm guessing these lines might have been committed before that convention was widely used.

josegonzalez: Good catch!

dokku
```diff
@@ -51,15 +51,12 @@ fi
...
+    APP="$2"; IMAGE=$(get_app_image_name $APP); IMAGE_SOURCE_TYPE="$3"; TMP_WORK_DIR="$4"
     dokku_log_info1 "Cleaning up..."
     dokku cleanup
     dokku_log_info1 "Building $APP from $IMAGE_SOURCE_TYPE..."
     dokku build "$APP" "$IMAGE_SOURCE_TYPE" "$TMP_WORK_DIR"
-    dokku_log_info1 "Releasing $APP..."
-    dokku release "$APP" "$IMAGE_SOURCE_TYPE"
-    dokku_log_info1 "Deploying $APP..."
-    dokku deploy "$APP"
+    release_and_deploy "$APP"
```
josegonzalez: The output after this is duplicated in `release_and_deploy` I think.

plugins/apps/post-delete
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_PATH/common/functions"
+APP="$1"; IMAGE_REPO=$(get_app_image_repo $APP)
 
-[[ -n $1 ]] && rm -r "$DOKKU_ROOT/$1" > /dev/null
+[[ -n $APP ]] && rm -r "$DOKKU_ROOT/$APP" > /dev/null
+
+# remove all application containers & images
+# shellcheck disable=SC2046
```
josegonzalez: Why disable the shellcheck?

plugins/common/functions
```diff
@@ -117,20 +117,51 @@ verify_app_name() {
...
-  DOKKU_APP_IMAGES=$(docker images -q $IMAGE)
-  if [[ -n "$DOKKU_APP_IMAGES" ]]; then
+  if (docker inspect "$IMAGE" &>/dev/null); then
     return 0
   else
     return 1
   fi
 }
 
+get_app_image_repo(){
```
josegonzalez: Needs space between `()` and `{`.

plugins/common/functions
```diff
@@ -117,20 +117,51 @@ verify_app_name() {
...
+  # central definition of our image repo pattern
+  local APP="$1"; IMAGE_REPO="dokku/$APP"
+  [[ -z "$APP" ]] && dokku_log_fail "(get_app_image_repo) APP must not be null"
+  echo $IMAGE_REPO
+}
+
+get_app_image_name(){
```
josegonzalez: Ditto on spacing here.

plugins/apps/post-delete
michaelshobbs: without shellcheck spits out:
```
In ./plugins/apps/post-delete line 9:
docker rm -f $(docker ps -a --no-trunc | egrep "dokku/${APP}:" | awk '{ print $1 }' | xargs) &>/dev/null || true
             ^-- SC2046: Quote this to prevent word splitting.


In ./plugins/apps/post-delete line 10:
docker rmi $(docker images -q $IMAGE_REPO | xargs) &>/dev/null || true
           ^-- SC2046: Quote this to prevent word splitting.
```

I remember there being a bug at some point if we quoted it.

plugins/tag/commands
```diff
@@ -0,0 +1,64 @@
...
+    release_and_deploy $APP $IMAGE_TAG
+    ;;
+
+  tag:list)
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    APP="$2"; IMAGE_REPO=$(get_app_image_repo $APP)
+    verify_app_name "$APP"
+
+    dokku_log_info2_quiet "Image tags for IMAGE_REPO"
+    docker images "IMAGE_REPO"
```
u2mejc: @michaelshobbs $ make you holler! :money_with_wings:

josegonzalez: https://gist.github.com/michaelshobbs/90b2445e4d2c8da5ff40

plugins/tag/commands
```diff
@@ -0,0 +1,64 @@
...
+    ;;
+
+  tag:rm)
```
josegonzalez: Should we name this `tag:destroy`, to follow the pattern of naming we do elsewhere?

plugins/tag/commands
```diff
@@ -0,0 +1,64 @@
...
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    APP="$2"; IMAGE_TAG="$3"; IMAGE_REPO=$(get_app_image_repo $APP)
+    verify_app_name "$2"
+
+    case "$IMAGE_TAG" in
+        latest)
```
josegonzalez: Should this be 2 spaces, not four?

plugins/tag/commands
michaelshobbs: points! well played

tests/unit/tag.bats
```diff
@@ -0,0 +1,47 @@
...
+  run /bin/bash -c "dokku tag:rm $TEST_APP v0.9.0"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run /bin/bash -c "dokku tag:list $TEST_APP | egrep -q 'v0.9.0'"
+  echo "output: "$output
+  echo "status: "$status
+  assert_failure
+
+}
```
josegonzalez: Remove the extra newline?

plugins/tag/commands
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_PATH/common/functions"
+source "$(dirname $0)/functions"
+
+case "$1" in
+  tag:add)
```
josegonzalez: Maybe `tag:create` would be a better name (same as `apps:create` and all the other services).

plugins/tag/commands
u2mejc: Is there a pattern? I also see ":remove" used twice and destroy once (also unset, but that makes sense in env context). I'm keen to keeping to the 'nix style shorthand.

josegonzalez: If we use `:create`, we should also use `:destroy`. All the services use `:destroy`.

For reference, `domains:remove` and `docker-options:remove` are things. Destroy implies something more destructive, hence why I want to use that here.

docs/process-management.md
```diff
@@ -42,3 +42,58 @@ dokku ps:scale app_name web=1 worker=2
 ## The web proctype
 
 Like Heroku, we handle the `web` proctype differently from others. The `web` proctype is the only proctype that will invoke custom checks as defined by a CHECKS file. It is also the only proctype that will be launched in a container that is either proxied via nginx or bound to an external port.
+
+## Image tagging
```
josegonzalez: Should this go in the deployment doc instead?

plugins/common/functions
```diff
@@ -267,3 +298,24 @@ is_app_running() {
     return 1
   fi
 }
+
+release_and_deploy() {
+  local APP="$1"; local IMAGE_TAG="$2"; IMAGE=$(get_app_image_name $APP $IMAGE_TAG)
```
josegonzalez: add a `local` to `$IMAGE` here?

plugins/tags/functions
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_PATH/common/functions"
+
+tag_image() {
+  docker tag -f "$1" "$2"
```
josegonzalez: I like that this came back. Whatever :)

michaelshobbs: yeah see my previous comment about it.

josegonzalez: gotcha.

docs/process-management.md
michaelshobbs: yes. moving.

plugins/common/functions
michaelshobbs: fixing

josegonzalez:
Image tagging == crushed.

josegonzalez:
@michaelshobbs A few questions now that I'm using this (ffs I should have tested this before):

- Why are we using `$IMAGE_REPO:$IMAGE_TAG` as the format for a tag?
- Is it possible to override the `$IMAGE_REPO` prefix? For instance, `quay.io` requires that you specify your tag prefix as `quay.io/username/repository`.
- What happens if we have an app with multiple containers (`web`, `worker`, `cron`)? Which one gets tagged? Can we specify a container type to get tagged?

michaelshobbs:
To close the loop here
- `$IMAGE_REPO:$IMAGE_TAG` is the string used to reference an image
- `$IMAGE_REPO` is hardcoded to be `dokku/$APP`
- containers != images

If you want to push to a docker registry then use a plugin that implements the `tags-create`/`tags-destroy` trigger as in the documentation example.  http://progrium.viewdocs.io/dokku/development/pluginhooks/

josegonzalez:
To setup a mobius loop:

- I am not a smart man and didn't realize how tags work...
- I'll be creating a quay.io plugin integration for dokku (pushing/pulling image tags and deploying them to applications).

Thanks @michaelshobbs for setting me straight :)

PR #1227, opened 2015-06-11
Use --no-cache when building Dockerfile

Flink:
Since `docker build` is only invoked when actually building a new image
from a Dockerfile, using `--no-cache` allows to have an updated image
even if the Dockerfile hasn't changed.
This works well with Dockerfiles using `git` or packages referencing
latest versions instead of "static" ones.


michaelshobbs:
Any negative impacts that would need to be documented here?

Flink:
For a given Dockerfile, if you only update the last RUN command for example, then you could lose a little time since the cache won’t be used anymore. But I can’t see a real use case where this is a problem. Maybe there is one though :)

michaelshobbs:
:+1:

PR #1228, opened 2015-06-13
switching from pluginhook to plugn

michaelshobbs:
This is an initial pass at implementing [plugn](https://github.com/progrium/plugn)

TODOS
- [x] setup plugn `PLUGIN_PATH`
- [x] convert core dokku plugins to plugn format
- [x] enable core dokku plugins
- [x] convert `pluginhook` calls to `plugin trigger`
- [x] update .deb builds @josegonzalez :smile:
- [x] update documentation
- [x] move plugins-* to plugin: namespace
- [x] introduce core plugin management
- [x] introduce DOKKU_API_VERSION
```
plugin:disable <name>                                              Disable an installed plugin (third-party only)
plugin:enable <name>                                               Enable a previously disabled plugin
plugin:install [--core|git-url]                                    Optionally download git-url & run install trigger for active plugins (or only core ones)
plugin:install-dependencies [--core]                               Run install-dependencies trigger for active plugins (or only core ones)
plugin                                                             Print active plugins
plugin:uninstall <name>                                            Uninstall a plugin (third-party only)
plugin:update [git-url]                                            Optionally download git-url & run update trigger for active plugins```

### CODE REVIEW

**Comment by joshco** [comment]
Does this break compatibility with the existing plugins? would they all need to be rewritten?

**Comment by michaelshobbs** [comment]
Not rewritten completely but updated for sure. I'll update the docs but in short, each plugin will need a `plugin.toml` file, and their `help` command augmented a little to account for the difference in the way output is handled. The installation instructions will need updating as well. All plugins will need to be installed in `$PLUGIN_PATH/available` and then enabled with `plugn enable <plugin_name>`.

**Comment by joshco** [comment]
that makes me very nervous.  I think if you break compat with existing plugins, that's going to be an adoption blocker. It's also going to detach you from the existing plugin ecosystem, make people reluctant to upgrade until various plugins are updated...

Can dokku support both types of plugins?

**Comment by michaelshobbs** [comment]
This is part of the refactoring that's been discussed on IRC captured in the [wiki doc](https://github.com/progrium/dokku/wiki/Refactoring) @progrium put together.  It will also the next major release and not in `0.3.x`.

**Comment by michaelshobbs** [comment]
To directly answer your question, it doesn't look like plugn and pluginhook can coexist due to their output handling differences. (Specifically around help output)

I also don't think we need to maintain the API contract between major releases as this is not necessarily expected and adds significant complexity even if we really wanted to do it. 

That being said, this is open source software and I think an open discussion needs to be maintained. So all opinions are welcomed.

**Comment by progrium** [comment]
I was not happy with how help output was done, and we need to be
comfortable with breaking the APIs between now and 1.0

On Sat, Jun 13, 2015 at 11:47 AM, Michael Hobbs <notifications@github.com>
wrote:

> To directly answer your question, it doesn't look like plugn and
> pluginhook can coexist due to their output handling differences.
> (Specifically around help output)
>
> I also don't think we need to maintain the API contract between major
> releases as this is not necessarily expected and adds significant
> complexity even if we really wanted to do it.
>
> That being said, this is open source software and I think an open
> discussion needs to be maintained. So all opinions are welcomed.
>
> —
> <https://github.com/progrium/dokku/pull/1228#issuecomment-111728795>.
>



-- 
Jeff Lindsay
http://progrium.com

**Comment by michaelshobbs** [comment]
Yay for multiple plugin output streams!!!!

```
Pseudo-terminal will not be allocated because stdin is not a terminal.
Counting objects: 418, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (335/335), done.
Writing objects: 100% (418/418), 215.01 KiB | 0 bytes/s, done.
Total 418 (delta 65), reused 392 (delta 51)
-----> Cleaning up...
-----> Building node-js-app from buildstep...
       ***/var/lib/dokku/plugins/enabled/hello/pre-build-buildstep hello!!!***
-----> Fetching custom buildpack
-----> Node.js app detected

-----> Reading application state
       package.json...
       build directory...
       cache directory...
       environment variables...

       Node engine:         ^0.10.0
       Npm engine:          2.x
       Start mechanism:     Procfile
       node_modules source: package.json
       node_modules cached: false

       NPM_CONFIG_PRODUCTION=true
       NODE_MODULES_CACHE=true

-----> Installing binaries
       Resolving node version ^0.10.0 via semver.io...
       Downloading and installing node 0.10.38...
       Resolving npm version 2.x via semver.io...
       Downloading and installing npm 2.11.2 (replacing version 1.4.28)...

-----> Building dependencies
       Installing node modules
       express@4.12.4 node_modules/express
       ├── merge-descriptors@1.0.0
       ├── utils-merge@1.0.0
       ├── cookie-signature@1.0.6
       ├── methods@1.1.1
       ├── cookie@0.1.2
       ├── fresh@0.2.4
       ├── escape-html@1.0.1
       ├── range-parser@1.0.2
       ├── content-type@1.0.1
       ├── finalhandler@0.3.6
       ├── vary@1.0.0
       ├── parseurl@1.3.0
       ├── serve-static@1.9.3
       ├── content-disposition@0.5.0
       ├── path-to-regexp@0.1.3
       ├── depd@1.0.1
       ├── qs@2.4.2
       ├── on-finished@2.2.1 (ee-first@1.1.0)
       ├── debug@2.2.0 (ms@0.7.1)
       ├── etag@1.6.0 (crc@3.2.1)
       ├── proxy-addr@1.0.8 (forwarded@0.1.0, ipaddr.js@1.0.1)
       ├── send@0.12.3 (destroy@1.0.3, ms@0.7.1, mime@1.3.4)
       ├── accepts@1.2.9 (negotiator@0.5.3, mime-types@2.1.1)
       └── type-is@1.6.3 (media-typer@0.3.0, mime-types@2.1.1)

-----> Checking startup method
       Found Procfile

-----> Finalizing build
       Creating runtime environment
       Exporting binary paths
       Cleaning npm artifacts
       Cleaning previous cache
       Caching results for future builds

-----> Build succeeded!

       node-js-sample@0.1.0 /tmp/build
       └── express@4.12.4

-----> Discovering process types
       Procfile declares types -> web
       ***/var/lib/dokku/plugins/enabled/hello/post-build-buildstep hello!!!***
-----> Releasing node-js-app...
       ***/var/lib/dokku/plugins/enabled/hello/pre-release-buildstep hello!!!***
       ***/var/lib/dokku/plugins/enabled/hello/post-release-buildstep hello!!!***
-----> Deploying node-js-app...
       ***/var/lib/dokku/plugins/enabled/hello/pre-deploy hello!!!***
-----> DOKKU_SCALE not found in app image. Defaulting to a single web process
-----> New DOKKU_SCALE file generated
=====> web=1
-----> Running pre-flight checks
       For more efficient zero downtime deployments, create a file CHECKS.
       See http://progrium.viewdocs.io/dokku/checks-examples.md for examples
       CHECKS file not found in container: Running simple container check...
-----> Waiting for 10 seconds ...
-----> Default container check successful!
=====> node-js-app container output:
       Detected 512 MB available memory, 512 MB limit per process (WEB_MEMORY)
       Recommending WEB_CONCURRENCY=1
       > node-js-sample@0.1.0 start /app
       > node index.js
       Node app is running at localhost:5000
=====> end node-js-app container output
-----> Running post-deploy
       ***/var/lib/dokku/plugins/enabled/hello/post-deploy hello!!!***
-----> Creating new /home/dokku/node-js-app/VHOST...
-----> Configuring node-js-app.dokku.me...
       Reloading nginx
=====> Application deployed:
       http://node-js-app.dokku.me

To dokku@dokku.me:node-js-app
 * [new branch]      master -> master
```

File: docs/development/plugn-creation.md [hunk 1]
```diff
@@ -1,14 +1,14 @@
-# Plugin Creation
+# Plugn Creation
```
josegonzalez: We probably don't need to change the docs here, the change makes it sounds weird.

tests.mk
```diff
@@ -65,7 +65,7 @@ lint:
 	# SC2001: See if you can use ${variable//search/replace} instead. - https://github.com/koalaman/shellcheck/wiki/SC2001
 	@echo linting...
 	@$(QUIET) shellcheck -e SC2029 ./contrib/dokku_client.sh
-	@$(QUIET) find . -not -path '*/\.*' | xargs file | egrep "shell|bash" | grep -v directory | awk '{ print $$1 }' | sed 's/://g' | grep -v dokku_client.sh | xargs shellcheck -e SC2034,SC2086,SC2143,SC2001
+	@$(QUIET) find . -not -path '*/\.*' | xargs file | egrep "shell|bash" | egrep -v "directory|toml" | awk '{ print $$1 }' | sed 's/://g' | grep -v dokku_client.sh | xargs shellcheck -e SC2034,SC2086,SC2143,SC2001
```
josegonzalez: Does a `toml` binary need to be installed?

michaelshobbs: No, it's included in the built plugin binary.

docs/development/plugin-creation.md
```diff
@@ -3,12 +3,12 @@
...
 
 1. Take a look at the plugins shipped with dokku and hack away!
-2. Check out the [list of hooks](http://progrium.viewdocs.io/dokku/development/pluginhooks) your plugin can implement.
+2. Check out the [list of hooks](http://progrium.viewdocs.io/dokku/development/plugn-triggers) your plugin can implement.
 3. Upload your plugin to github with a repository name in form of `dokku-<name>` (e.g. `dokku-mariadb`)
 4. Edit [this page](http://progrium.viewdocs.io/dokku/plugins) and add a link to it.
 5. Subscribe to the [dokku development blog](http://progrium.com) to be notified about API changes and releases
 
-### Sample plugin
+### Sample plugn
```
josegonzalez: Same here re: `plugn` naming.

docs/development/plugn-triggers.md
```diff
@@ -1,15 +1,15 @@
-# Pluginhooks
+# Plugn triggers
```
josegonzalez: We should provide the old `pluginhooks` file and link to the new content.

michaelshobbs: What do you mean?

josegonzalez: Basically a page saying the old `pluginhooks.md` docs can now be found at `plugn-triggers.md`. To be honest, I feel as though we should keep the old filename with the new content.

michaelshobbs: Makes sense. Let's keep the old filename then and I'll include a note about pluginhooks moving to plugn triggers.

stuartpb:
Guys, I would **really really really** appreciate it if you took a look at how Plushu has been designed for the last eleven months (you can built its Dokku-comparable setup with https://meta.sh/setup/plusku). We (I) went through all these same design considerations almost a year ago, and I've really done something with it.

I know @progrium was thrown by the way I'm not focusing on "Fully containerized!!" the way Dokku 1.0 is, but that's solveable with a `docker commit $(curl meta.sh/setup/plusku | docker run -d ubuntu bash) plusku` that you run with `--net=host` or bind the Docker, HTTP(S), and SSH ports to (and frankly I can't see how any app aiming to be a Docker hypervisor could be anything else), so I don't get why that was ever a concern.

josegonzalez:
@stuartpb I'm not opposed to it, but are there any cliffnotes on how plushu is better? :)

stuartpb:
https://github.com/plushu/plushu/wiki/Differences-from-Dokku

josegonzalez:
Reading through that, what in this PR would be affected by that change?

I'm very willing to break BC in the 0.4.0 release, just want to be sure we're changing the right things.

stuartpb:
The **really key** ones:

- Plushu calls hooks for functionality implemented by plugins.
  - The wiki page doesn't list this, but Plushu is also built of **way, way more** plugins and hooks to handle its operation (like, I would roughly estimate Plusku is split into five times more plugins with ten times as many hook interfaces - and I think I'm lowballing the hook estimate). This lets Plusku be configured and extended **way** more easily than I ever was able to with Dokku - it also lets me do useful things like [disable app deployment when we build its sandbox server](https://github.com/plushu/build-sandbox/blob/master/build.sh#L173) so the app won't be running until *after* the VM has been shut down and imaged.
- Plugins' hook scripts in Plushu are granularly ordered by placing numbered files in a subdirectory with the hook's name.
  - This has been, if you'll pardon the pun, *innumerably* useful in Plushu's design, both from an operational standpoint and just from a clarity standpoint. At first, I had this as an optional thing, where scripts could be either the name of the hook *or* a numbered file under a directory for that hook, but I changed it so all the hooks are numbered (and named!) scripts in `.d`-style directories. It's difficult to elucidate just *how much simpler* things are with this structure, but suffice it to say it is **essential** once you start designing against hooks of any degree of complexity.

stuartpb:
What I'm asking, vis. this PR, is that Dokku 1.0 adopt the *minimal* Plushu core as the basis of its architecture rather than trying to solve these design problems again in its own, very-slightly-incompatible way.

There are really very few differences between the systems Dokku uses and the systems Plusku (again - the Plushu core is distinct from the Plusku 12-factor app plugin collection) uses - I've already written an example of how Dokku's differences of opinion from Plusku could be handled with its own Plusku setup with https://github.com/plushu/plushu-dokku. If Dokku could share these components I've built for Plusku (ie. opting to use its own Buildkit or whatever over Plusku's build-cedarish scripts), **everybody** stands to benefit, from an ecosystem point of view.

josegonzalez:
@michaelshobbs as the implementor of the pr, thoughts on his comments?

stuartpb:
Also, if you'd like to take a tour of Plushu, you can add an SSH public key via https://enter.sandbox.plushu.org, then log in with that keypair as `root@sandbox.plushu.org` (the `plushu` user can be used to execute commands the same as Dokku's `dokku` user).

josegonzalez:
@stuartpb care to hop onto irc for a bit? I'm `savant` there :)

progrium:
Report back what you guys think here. Adopting Plushu will prevent Dokku
from getting where I want it to be, so I'd probably totally abandon the
project and drop the potential for Glider Labs to support it. Maybe that's
fine, I don't know.

On Friday, July 3, 2015, Jose Diaz-Gonzalez <notifications@github.com>
wrote:

> @stuartpb <https://github.com/stuartpb> care to hop onto irc for a bit?
> I'm savant there :)
>
> —
> <https://github.com/progrium/dokku/pull/1228#issuecomment-118407054>.
>


-- 
Jeff Lindsay
http://progrium.com

josegonzalez:
@progrium Do you mind hopping on irc when you get a chance? I'd like to see where you want dokku to go, which would definitely clear up what we should do with plugn/pluginhook/any other plugin methodology.

It's been a bit unclear to me other than "we should use plugn and buildkit", both of which seem like good ideas.

stuartpb:
> Adopting Plushu will prevent Dokku from getting where I want it to be, so I'd probably totally abandon the project and drop the potential for Glider Labs to support it. Maybe that's fine, I don't know.

I honestly think this is fine, and furthermore, I think it is right. Glider Labs already *has* its framework (Flynn) for pushing out large-scale apps: trying to *simultaneously push Dokku* to use components from Flynn (other than Buildkit which is legitimately a shared thing), in the service of transforming Dokku into some kind of [nonesuch beast](https://sites.google.com/site/steveyegge2/nonesuch-beast) is to muddle this project's purpose.

josegonzalez:
As far as I'm aware, Glider Labs doesn't use flynn at all, which is a completely separate project. Whether or not they support dokku - or even if @progrium abandons dokku completely - is irrelevant. I just want to ensure we have the best possible outcome for all of our users. Doesn't make sense to me that we make these decisions in a vacuum.

Regarding BC breakage, I'm fine with releasing a 0.4.0 tomorrow and then a 0.5.0 Sunday. We'll move as fast as we can and as fast as we need to. Dokku's had a good release cadence over the past year, and I've been moving PRs along as quickly as possible, so if there is a large amount of refactoring to be done to get Dokku to perfection, let's do it. But less discuss those changes on technical merit, not potential help from third parties or gut feelings of hate for where the project is. :)

I like the directory structure setup that plushu has, though I want to talk it over with @michaelshobbs before deciding what to do. This would be his second attempt at redoing how we handle plugins and he'd probably take the lead on a third attempt (I sure don't want to :P). As well, since he actually uses dokku in production, he might have a better grip as to how we proceed going forward. 

There are a few things I like about the `plugn` project, though it needs better documentation before I can say "yey this is going to maybe work". Looking through the codebase more thoroughly, I personally don't see too much reason why the `plugn` and `plushu` model of triggering hooks can't coexist nicely.

progrium:
Glider Labs started 9 months after I stopped working with the Flynn team. I want what's best for Dokku users as determined by the community. If that means going different directions, I'm fine with that. I play the long game, so it can be hard to wait for everything to come together. And if I just describe the end state, it won't be as compelling as seeing it. So I can understand doing what's best in the short term, it's just unfortunate if it's in a different direction.

stuartpb:
> Glider Labs started 9 months after I stopped working with the Flynn team.

I didn't know this, I was under the impression Glider Labs *was* the Flynn team.

> I want what's best for Dokku users as determined by the community. If that means going different directions, I'm fine with that. I play the long game, so it can be hard to wait for everything to come together. And if I just describe the end state, it won't be as compelling as seeing it. So I can understand doing what's best in the short term, it's just unfortunate if it's in a different direction.

Same here. What makes the Dokku/Plushu divergence so frustrating to me is that I feel we have the same end-game, yet we're taking barely-different parallel routes to get there (and, with Dokku having 1000+ issues with nearly 10,000 stars, while Plushu has had about 4 users total even *look* at it, I'm on the losing side of that split, from a contributor point of view).

What I want - and what I think @progrium wants, too - is to provide developers with a simple-to-spin-up system that they completely control (and can customize without locking themselves out of updates), that will allow them to write 12-factor apps that will allow them to scale up with as little ops friction as possible (and no dev friction).

Where I feel there's *any* disagreement about the path to reach this is that @progrium feels this should be done by putting Dokku *into* your eventually-scaling system, with scaling involving Dokku eventually being replaced/thrown out for something else, whereas I feel this should be done by putting Plushu *in front of* your cluster, as essentially a network-accessible configured binary (which I think probably comes from my thin-client development focus, where I make it a point to keep *nothing* important to development on machines that can be killed by a swimming pool).

I only really care about those things I mentioned in the last three run-on sentences (specifically the contributor disparity, which is essentially *the* thing that sticks in my craw about all this) - other than that, I have no issue with Dokku continuing to go its own way. I've done fine this far without inflow from Dokku (although Plusku still doesn't have a CI, either [in the codebase sense because I don't have the time to write it][plushu#64] or [in the app-deployment sense because I don't have time to extensibly factor it][pdalc#5]), I expect I'll essentially be fine without it going forward (if the Dokku  maintainers ever encounter an issue that might affect the wider platform or involve really arcane Bash edges, they can ping @plushu about it and I'll take a look).

[plushu#64]: https://github.com/plushu/plushu/issues/64
[pdalc#5]: https://github.com/plushu/plushu-deploy-app-local-container/issues/5

Besides that, I'm moving toward a less complex-backend-cluster design for the majority of my new projects, using [static files][igh#408] and PouchDB, with a CouchDB server (or servers) to replicate to for cloud sync, so my diminishing dependence on complex-server clusters is diminishing the amount that I end up fussing over this.

[igh#408]: https://github.com/isaacs/github/issues/408

progrium:
I'm not sure I entirely agree with @stuartpb's idea of what my feelings are. All I can say is I believe the paths of plugn et al vs Plushu to be fundamentally different and I'm not sure what I can do to communicate that without more pieces coming together. Again, I don't expect anybody to care that I have some grand master plan except that without going the path I've already suggested in the refactoring wiki page, we won't get there and I'll end up making yet another Dokku-like project in the future. I'd rather not do that since I'd prefer to get there together.

michaelshobbs:
My thoughts:

I need a platform to deploy multiple applications in multiple languages on my own AWS infrastructure. Dokku does that. Hence me being here and wanting to help. I don't care particularly about the implementation details so much as they are easy to understand and I can make my own modifications in true OSS-style. I liked the simplicity (in concept) of the original plugin refactor (specifically around calling plugin commands directly vs hunting through a directory structure + case statements) and would really like to get dokku to that spot sooner rather than later. 

Again, don't really care whether that's via plugn or some other method. Either way, glad to be of help. :smile:

josegonzalez:
I rebased the hell out of this. Still need to update the packaging stuff (that's on me) but otherwise if tests pass, we're in a good place.

Note: once this is merged, master will be broken for anyone using custom plugins. What do we need to do before a 0.4.0 to get any plugins in good shape?

- add a `plugin.toml` file
- remove the extra `cat` from the help subcommands (is this necessary?)
- anything else?

josegonzalez:
For example, if we can just send an empty string as input via `plugn` for the help hook, then maybe that would let us use all existing plugins as is? Not sure how much of a debbie-downer that would be. Up to @progrium and @michaelshobbs I guess.

josegonzalez:
Not sure why it's hanging, as it completes successfully locally. @michaelshobbs thoughts?

michaelshobbs:
I'm back today. I'll have a look.

michaelshobbs:
The build timeout seems to be caused by the `cat || true` in the git and tar plugins. I'll get this fixed ASAP.

josegonzalez:
Why does that appear to succeed locally?

michaelshobbs:
Try running a rebuild. Hangs on me everytime. However, now the tar plugin doesn't work...

michaelshobbs:
plugn doesn't have a `-p`

josegonzalez:
Woops :|

michaelshobbs:
First pass at a rebase with #1452. Don't review yet. Need to sort out any bugs.

u2mejc:
TL;DR but what was the communication choice for this change to the plugn model? Not just plugin developers, but just as important, plugin consumers? Seems to make old plugins very nosy in my testing.

josegonzalez:
@u2mejc what do you mean by that question?

michaelshobbs:
Noisy = if a plugin is missing a plugin.toml file, then plugn barfs all over the screen

josegonzalez:
>  but what was the communication choice for this change to the plugn model?

Not sure what that question means.

michaelshobbs:
I think he's asking how are we letting folks know that current plugins may break (at minimum barf out crap about a missing plugin.toml file) and letting plugin authors know the change is coming?

josegonzalez:
There are a few things we can do:

- on upgrade, spit out broken plugins (this can happen in the bootstrap.sh)
- when we run `dokku`, we can check plugin status and exit immediately so users can fix everything
- we can have our trusty @Flink issue PRs to every plugin to fix the derps!
- Add it to our release notes
- Blog about whats going on
- Add a banner to the dokku docs homepage (and readme) stating about potential incompatibilities.

Flink:
Every plugin seems a lot of plugins :grin:

josegonzalez:
Once rebased, should I review, or is there anything else this PR is missing?

plugins/plugin/functions
```diff
@@ -0,0 +1,41 @@
...
+  plugn enable $PLUGIN
+  dokku_log_info1_quiet "Plugin $PLUGIN enabled"
+}
+
+download_plugin() {
+  local PLUGIN_GIT_URL="$1"
+  plugn install $PLUGIN_GIT_URL
+}
+
+download_and_enable_plugin() {
```
josegonzalez: Should this support plugin versioning? Since it's git, basically anything tree-ish should work I think?

josegonzalez:
I rebased the hell out of this, so it needs a review for sure.

plugins/00_dokku-standard/commands
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
-source "$PLUGIN_PATH/common/functions"
-source "$PLUGIN_PATH/config/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/config/functions"
```
josegonzalez: If I change this to `source "$PLUGIN_AVAILABLE_PATH/config/functions`, will that pick up whatever config plugin is available? Basically, I'd like to be able to override every core plugin except common.

michaelshobbs: By default `$PLUGIN_AVAILABLE_PATH/config/function` and `$PLUGIN_CORE_AVAILABLE_PATH/config/function` are the same file because of symlinks. However `$PLUGIN_AVAILABLE_PATH` and `$PLUGIN_CORE_AVAILABLE_PATH` are different. So if you modify the directory contents, then yeah, you could override everything by the `common` plugin.

josegonzalez: So should we be sourcing from `$PLUGIN_CORE_AVAILABLE_PATH` when sourcing anything not in common?

plugins/plugin/functions
michaelshobbs: Yes. It's just doing a git clone. See here. https://github.com/progrium/plugn/blob/2a2dc1c3d8862e5611186ed0d6ce19ee8ab133bf/bashenv/plugn.bash#L2-L8

plugins/00_dokku-standard/commands
michaelshobbs: If you want to guarantee you're getting a core plugin regardless of it being enabled or not, yes. (side note: while dokku won't let you disable a core plugin, plugn is like a honeybadger, it don't care)

michaelshobbs:
I want to see tests pass one final time so I triggered a run.

josegonzalez:
lolrus stupid merge conflict.

plugins/00_dokku-standard/commands
josegonzalez: I just want to use a custom plugin in some cases but fall back to core if they dont exist. I'll update this pr to ensure if it's not common, then we use `$PLUGIN_AVAILABLE_PATH`.

michaelshobbs:
Looks like this is ready?

PR #1254, opened 2015-06-25
added DOKKU_DOCKERFILE_START_CMD support

### LINKED ISSUE #1249: Is it possible to set custom docker run command?
Is it possible to override the `CMD` directive in the `Dockerfile` with a per app conf? With base docker I do this just doing `docker run myimage echo hello` and it will ignore the `CMD`

This will be very useful when deploying the same image for multiple envs or just the same app but with different parameters

kilianc:
closes #1249

I need little help with the test suite here, I tested it on my server it worked smoothly.

### DIFF
--- a/dokku
+++ b/dokku
@@ -92,7 +92,7 @@ case "$1" in
 
         if [[ -z "$DOKKU_BUILDSTEP" ]];then
           DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
-          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
+          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
         fi
 
         if [[ "$BIND_EXTERNAL" = "false" ]] && [[ "$PROC_TYPE" == "web" ]];then
--- a/dokku
+++ b/dokku
@@ -89,7 +89,11 @@ case "$1" in
         BIND_EXTERNAL=$(pluginhook bind-external-ip $APP)
 
         [[ -n "$DOKKU_BUILDSTEP" ]] && START_CMD="/start $PROC_TYPE"
-        [[ -z "$DOKKU_BUILDSTEP" ]] && DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
+
+        if [[ -z "$DOKKU_BUILDSTEP" ]];then
+          DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
+          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
+        fi
 
         if [[ "$BIND_EXTERNAL" = "false" ]] && [[ "$PROC_TYPE" == "web" ]];then
           port=${DOKKU_DOCKERFILE_PORT:=5000}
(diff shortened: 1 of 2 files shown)


dokku
```diff
@@ -89,7 +89,11 @@ case "$1" in
         BIND_EXTERNAL=$(pluginhook bind-external-ip $APP)
 
         [[ -n "$DOKKU_BUILDSTEP" ]] && START_CMD="/start $PROC_TYPE"
-        [[ -z "$DOKKU_BUILDSTEP" ]] && DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
+
+        if [[ -z "$DOKKU_BUILDSTEP" ]];then
+          DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
+          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
```
michaelshobbs: This will break if DOKKU_DOCKERFILE_START_CMD is not set. Example:
```
root@dokku:~/dokku# dokku config:get node-js-app DOKKU_DOCKERFILE_START_CMD ; echo $?
1
```

kilianc: would a `|| ""` make it ?

michaelshobbs: Might be best to get it with `|| true` and then if it's not null `-n` then set `START_CMD`

michaelshobbs: Sorry just saw your question. That would make `$START_CMD` literally quotes in the non-set case and `docker run` will error out.

kilianc: then `|| $START_CMD`  ?

michaelshobbs:
Sorry I'm on mobile right now. What are you asking?

michaelshobbs:
Looks legit to me. @josegonzalez thoughts?

kilianc:
@michaelshobbs I was commuting, thank for being patient on this, I am a noob with bash!

michaelshobbs:
No worries at all. Thanks for putting this together.

josegonzalez:
Looks good to me, does it work?

kilianc:
it does on my DO machine! So happy that I can now deploy same app multiple times and just change the args :)

michaelshobbs:
Thoughts on documenting this?

kilianc:
@michaelshobbs yeah, I am Italian so I would probably do a shitty job at it but I can give it a try and maybe you can review it ?

michaelshobbs:
sounds good to me :smile:

kilianc:
@michaelshobbs hey, I did my best!

I actually have a question: there are inconsistencies on names I think and I am not sure `DOKKU_DOCKERFILE_START_CMD` is the correct name, we have:

* `DOCKERFILE_PORT`
* `DOCKER_OPTIONS_DEPLOY`
* `DOCKER_OPTIONS_RUN`
* `DOCKER_OPTIONS_BUILD`
* `DOKKU_DOCKER_BUILD_OPTS`
* `DOKKU_DOCKERFILE_CACHE_BUILD`
* `DOKKU_DOCKERFILE_START_CMD`

Thoughts?

kilianc:
I think that it would be more appropriate to call it `DOCKER_OPTIONS_RUN_CMD`

https://docs.docker.com/reference/run/

And I am not mistaken we can remove `DOCKERFILE_PORT`, `DOKKU_DOCKER_BUILD_OPTS`, `DOKKU_DOCKERFILE_CACHE_BUILD` which are just shortcut for `DOCKER_OPTIONS_*` ?

michaelshobbs:
I'm sure we could do better on naming here. However these are all used in their own way without duplication. `DOCKER_OPTIONS_*` variables are specifically used in the docker-options plugin and are named that way only because that's how the original author of the once external plugin named them.

Let's not remove or change any names as part of this PR. If you want to take a stab at making these conform to a standard in another PR, that would definitely be welcomed.

kilianc:
@michaelshobbs I didn't mean to change names in this PR I just wanted to think forward and use a different name for this option.

I can bake another PR. You said those options work their own way but it seems to me that `DOKKU_DOCKER_BUILD_OPTS ` and `DOCKER_OPTIONS_BUILD` are the same, all we need is the docker plugin ones, there nothing you can't so passing params to docker, included disabling cache and exposing the port. I'll open a new ticket to iterate on this a bit more?

michaelshobbs:
Sounds good to me

PR #1256, opened 2015-06-26
Implement events logging feature

alessio:
Record events (i.e. pluginhook's calls) as syslog entries and provide a shortcut command to display the last part of the log file.

A brief list of changes follows:

`dokku`:
 - New __DOKKU_EVENTS_LOGFILE__ envvar to hold logfile location.

`plugins/20_events/commands`:
 - __events__ output log entries, a tail's follow-like mode is provided too.
 - __events:[on|off]__ enables/disables logging.
 - __events:list__ lists events that are logged.

`plugins/20_events/hook`:
 - Generic hook that writes log entries. All the events are registered via symlink to this script.

`plugins/common/functions`:
 - _dokku_log_event()_ writes log entries via logger.
 - _dokku_log_pluginhook_call()_ formats log entries and writes them via dokku_log_event().


michaelshobbs:
I have a few questions about this feature in general.
 - I'm not clear on the benefit of this plugin. Can you include the intended use-case? 
 - Why do we think this belongs in the core plugin set?

michaelshobbs:
Can you include an excerpt from the logs to show what this looks like?

plugins/20_events/commands
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+find "$l_plugin_dir" -type l -printf "%f " | sort
```
michaelshobbs: What does this do and why?

michaelshobbs: In fact this happens on deployment as well
```
remote: find: cannot search `': No such file or directory
remote: find: cannot search `': No such file or directory
remote: find: cannot search `': No such file or directory
```

michaelshobbs:
Just tested this. I don't see anything in `/var/log/dokku.log` but I do see this in `/var/log/syslog`:

```
Jun 26 16:41:03 dokku dokku[12857]: INVOKED: pre-delete( node-js-app )
Jun 26 16:41:13 dokku dokku[12996]: INVOKED: post-delete( node-js-app )
Jun 26 16:41:15 dokku dokku[13126]: INVOKED: receive-app( node-js-app a836400724f1c8aa34e6f873039621053ad6a639 )
Jun 26 16:41:17 dokku dokku[13478]: INVOKED: pre-build-buildstep( node-js-app )
Jun 26 16:41:17 dokku dokku[13491]: INVOKED: docker-args-build( node-js-app )
Jun 26 16:41:33 dokku dokku[13772]: INVOKED: post-build-buildstep( node-js-app )
Jun 26 16:41:33 dokku dokku[13811]: INVOKED: pre-release-buildstep( node-js-app )
Jun 26 16:41:33 dokku dokku[13887]: INVOKED: post-release-buildstep( node-js-app )
Jun 26 16:41:33 dokku dokku[13922]: INVOKED: pre-deploy( node-js-app )
Jun 26 16:41:33 dokku dokku[14046]: INVOKED: docker-args-deploy( node-js-app )
Jun 26 16:41:34 dokku dokku[14100]: INVOKED: bind-external-ip( node-js-app )
Jun 26 16:41:34 dokku dokku[14185]: INVOKED: check-deploy( node-js-app 9050d85117a547bce73b5ac91dac075029ece28355a68f80315587581af650cd web 5000 172.17.0.43 )
Jun 26 16:41:44 dokku dokku[14253]: INVOKED: post-deploy( node-js-app 5000 172.17.0.43 )
Jun 26 16:41:44 dokku dokku[14316]: INVOKED: nginx-hostname( node-js-app node-js-app dokku.me )
Jun 26 16:41:44 dokku dokku[14373]: INVOKED: nginx-pre-reload( node-js-app )
```

I did run `dokku plugins-install` before testing.

plugins/20_events/commands
```diff
@@ -0,0 +1,55 @@
...
+    ;;
+
+  events:off)
+    echo "Disabling dokku events logger"
+    rm -f $DOKKU_ROOT/.dokkurc/DOKKU_EVENTS
+    ;;
+
+  events:list)
+    if [[ "$DOKKU_EVENTS" ]]; then
+      logged="$(find $(dirname $0) -type l -printf '%f ' | sort)"
```
michaelshobbs: lint fails.
```
In ./plugins/20_events/commands line 31:
      logged="$(find $(dirname $0) -type l -printf '%f ' | sort)"
                     ^-- SC2046: Quote this to prevent word splitting.
```

alessio:
(The discussion begun on IRC, so I'm mostly copying here what I said in #dokku)

We experienced kernel hangs on containers removal and this has proven to be very helpful to understand which were the events that triggered the issue.

To answer both questions, the rationale is to mitigate the lack of logging functionalities in dokku by providing a not too fine-grained view of what's going on behind the scenes.

alessio:
@michaelshobbs can you please check whether `/etc/rsyslog.d/99-dokku.conf` was created?

michaelshobbs:
@alessio indeed
```
root@dokku:~/dokku# cat /etc/rsyslog.d/99-dokku.conf
:syslogtag, contains, "dokku" /var/log/dokku.log
```

alessio:
@michaelshobbs run:
```shell
ls -l /var/log/dokku.log
```

Please confirm that your result matches the following
```shell
$ ls -l /var/log/dokku.log 
-rw-rw-r-- 1 syslog dokku 171986 Jun 26 20:30 /var/log/dokku.log
```

It's crucial that the owner is `syslog`, otherwise rsyslog can't write to it.
It should be enough to reload rsyslog (and the install should be doing that), though I'd suggest to stop the service and start it again:
```shell
sudo service rsyslog restart
```

michaelshobbs:
Restarting rsyslog seems to work. Can we add that to the install script?

alessio:
Done.

josegonzalez:
Can you add logrotation for this?

alessio:
Definitely we can. I'll work on it.

sent via mobile

alessio:
I was wondering whether it would be a good idea to create a /var/log/dokku/ sub-dir and redirect the plugin's logs to its own /var/log/dokku/events.log as we may want to log other things in the future. Comments are much welcome!

michaelshobbs:
@alessio looks like there's an issue with your test
```
not ok 25 (events) check conffiles
# (in test file tests/unit/events.bats, line 22)
#   `stats -c '%U:%G:%a' /var/log/dokku.log' failed with status 127
# Creating my-cool-guy-test-app... done
# output: 
# status: 
# /home/ubuntu/dokku/tests/unit/test_helper.bash: line 17: [: : integer expression expected
# output: 
# status: 
# /home/ubuntu/dokku/tests/unit/test_helper.bash: line 17: [: : integer expression expected
# /tmp/bats.53387.src: line 22: stats: command not found
```

michaelshobbs:
@alessio since we're now log rotating, i think a `/var/log/dokku/` directory makes sense.

alessio:
Indeed it does

plugins/20_events/hook
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+[[ ! "$DOKKU_EVENTS" ]] || dokku_log_pluginhook_call "$(basename $0)" "$@"
```
michaelshobbs: This question may be naive, but any reason we're doing a double negative here. More directly why not do this?
```
[[ "$DOKKU_EVENTS" ]] && dokku_log_pluginhook_call "$(basename $0)" "$@"
```

alessio: On Wed, Jul 1, 2015 at 6:50 PM, Michael Hobbs <notifications@github.com> wrote:
> This question may be naive, but any reason we're doing a double negative here. More directly why not do this?

It's just my personal preference, but if that breaks the code style
I'm more than happy to fix it then.


-- 
Alessio Treglia          | www.alessiotreglia.com
Debian Developer         |     alessio@debian.org
Ubuntu Core Developer    |  quadrispro@ubuntu.com
0416 0004 A827 6E40 BB98 90FB E8A4 8AE5 311D 765A

alessio: Well, actually due ```set -e``` if $DOKKU_EVENTS is unset the hook would return 1, and that would break the plugins chain, wouldn't it?

michaelshobbs:
I was wrong, you were right here. `pipefail` strikes again!! Can we make them both double negative for consistency?

plugins/20_events/hook
michaelshobbs: Yes :smile:

alessio:
Done, it's all yours now!

michaelshobbs:
:+1:

michaelshobbs:
@alessio I hit you up on IRC but just in case, would you be willing to add some documentation around this?

alessio:
:+1: sure, I will write some docs ASAP

PR #1261, opened 2015-06-29
Allows to skip checks (all or default)

Flink:
Setting `DOKKU_SKIP_ALL_CHECKS` globally or for a given app will skip
all checks.
Setting `DOKKU_SKIP_DEFAULT_CHECKS` globally or for a given app will skip
default checks only.

### DIFF
--- a/dokku
+++ b/dokku
@@ -70,6 +70,9 @@ case "$1" in
     DOKKU_SCALE_FILE="$DOKKU_ROOT/$APP/DOKKU_SCALE"
     oldids=$(get_container_ids $APP)
 
+    [[ -f "$DOKKU_ROOT/ENV" ]] && source $DOKKU_ROOT/ENV
+    [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
+
     while read line || [ -n "$line" ]
     do
       PROC_TYPE=${line%%=*}
@@ -111,10 +114,14 @@ case "$1" in
         }
 
         # run checks first, then post-deploy hooks, which switches Nginx traffic
-        trap kill_new INT TERM EXIT
-        dokku_log_info1 "Running pre-flight checks"
-        pluginhook check-deploy  $APP $id $PROC_TYPE $port $ipaddr
-        trap -        INT TERM EXIT
+        if [[ "$DOKKU_SKIP_ALL_CHECKS" = "true" ]]; then
+          dokku_log_info1 "Skipping pre-flight checks"
+        else
+          trap kill_new INT TERM EXIT
+          dokku_log_info1 "Running pre-flight checks"
+          pluginhook check-deploy  $APP $id $PROC_TYPE $port $ipaddr
+          trap -        INT TERM EXIT
+        fi
 
         # now using the new container
         [[ -n "$id" ]] && echo $id > "$DOKKU_CONTAINER_ID_FILE"
--- a/docs/application-deployment.md
+++ b/docs/application-deployment.md
@@ -129,6 +129,8 @@ Dokku will wait `DOKKU_WAIT_TO_RETIRE` seconds (default: `60`) before stopping t
 
 Dokku will retry the checks DOKKU_CHECKS_ATTEMPTS times until the checks are successful or DOKKU_CHECKS_ATTEMPTS is exceeded.  In the latter case, the deployment is considered failed. This can be overridden in the CHECKS file by setting ATTEMPTS=nn.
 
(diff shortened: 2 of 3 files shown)


alessio:
:+1:

dokku
```diff
@@ -70,6 +70,9 @@ case "$1" in
     DOKKU_SCALE_FILE="$DOKKU_ROOT/$APP/DOKKU_SCALE"
     oldids=$(get_container_ids $APP)
 
+    [[ -f "$DOKKU_ROOT/ENV" ]] && source $DOKKU_ROOT/ENV
+    [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
```
michaelshobbs: these lines may have wider consequences than intended. we've seen in the past where we've added sourcing of app envs cause some issues. that isn't to say these line should be changed but that we may want to remember them specifically when troubleshooting after this gets merged. cc: @josegonzalez

michaelshobbs:
:+1:

dokku
josegonzalez: Why not just retrieve the value via config:get ?

Flink: Because if I’m not mistaken, you can’t get global config with `config:get`

michaelshobbs:
@josegonzalez @Flink anything else to do here or shall I merge?

Flink:
@michaelshobbs it’s good for me, I don’t know if @josegonzalez saw my answer the other day ;)

dokku
josegonzalez: Can you do this some other way? I'm mostly afraid someone will define something like `PROC_TYPE` and break all the things.

Perhaps it's time to have a `globalconfig:get` method?

Flink: I was thinking we could have a special app named `dokku` to set and get global config?

michaelshobbs:
I'm :-1: on a special app name. I thinking keeping this stuff in the config plugin makes sense. Perhaps just `config:globalset` or `config:set --global`. I think I like the UX of the second more even though I'm not a fan of the additional arg parsing but whatevs.

Flink:
The `--global` flag seems nice but I think it lacks consistency with the existing UX. If we use it, we would have to check that an application name is not given.
Maybe we could use `config:set-global` without an app name, since there is already a `config:set-norestart`, the pattern will remain the same.
What do you think?

josegonzalez:
@Flink +1

michaelshobbs:
@Flink yeah my point there was to make it a bit inconsistent because it's not setting and app's config. If we parse for `--` args up front then we just bypass the whole $APP check anyway.

Either way, looking at the code now, it'll be a significant (but probably necessary) change to the config plugin.

Flink:
So do I add a `config:set-global` action or not? :)

michaelshobbs:
I think we should overload `config:set` and `config:get` with `--global`. I'd like to hear @josegonzalez weigh in with his thoughts (more than a :+1:) :smile:

Flink:
Ok then :wink:

josegonzalez:
@michaelshobbs I'd rather what @Flink proposed, assuming we're going to keep `config:set-norestart`. Seems like we should standardize on one method of `config` setting/getting. Though I'm happy if the flag-based method is implemented, and we deprecate `config:set-norestart` :)

michaelshobbs:
@josegonzalez Upon much reflection after implementing the `config:set-norestart` syntax, I think I'd prefer the flag-based method. @Flink let me know if you want to run with that. otherwise, i can.

josegonzalez:
@michaelshobbs makes sense to me, just want to expose standard interfaces where possible. If we have a new standard, good :).

Flink:
@michaelshobbs ok for the new syntax so ;) I think you should give it a try since it seems you already have something in mind about implementation :grin:

michaelshobbs:
@Flink we can also work together on it. i can submit a PR against your branch if you'd like.

josegonzalez:
@michaelshobbs Maybe you work on your pr, merge to master, and @Flink can rebase?

Flink:
Yes I agree with @josegonzalez, it seems more natural to me this way. But thanks @michaelshobbs :)

michaelshobbs:
sounds good

josegonzalez:
Also makes it easy for me to make release notes :)

michaelshobbs:
`dokku config:get --global DOKKU_SKIP_ALL_CHECKS` will work now

josegonzalez:
Okay, *now* `dokku config:get --global DOKKU_SKIP_ALL_CHECKS` will work.

josegonzalez:
I think I'l going to merge this for now. @michaelshobbs Mind making a separate pr that updates all places where we retrieve global config manually?

PR #1278, opened 2015-07-03
Nginx template from app

josegonzalez:
Refs #1054. Thanks @assaf :)

### DIFF
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -49,13 +49,8 @@ case "$1" in
       shopt -u nullglob
     fi
 
-    DOKKU_APP_CIDS=$(get_container_ids $APP)
-    if [[ -n $DOKKU_APP_CIDS ]]; then
-      for DOKKU_APP_CID in $DOKKU_APP_CIDS; do
-        docker cp "$DOKKU_APP_CID:/app/nginx.conf" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
-        break
-      done
-    fi
+    DOKKU_APP_CIDS=($(get_container_ids $APP))
+    docker cp "${DOKKU_APP_CID[0]}:/app/nginx.conf" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
 
     [[ -f "$DOKKU_ROOT/ENV" ]] && source $DOKKU_ROOT/ENV
     [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -49,6 +49,9 @@ case "$1" in
       shopt -u nullglob
     fi
 
+    DOKKU_APP_CONTAINER_ID=$(< "$DOKKU_ROOT/$APP/CONTAINER")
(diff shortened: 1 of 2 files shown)


plugins/nginx-vhosts/commands
```diff
@@ -49,6 +49,9 @@ case "$1" in
       shopt -u nullglob
     fi
 
+    DOKKU_APP_CONTAINER_ID=$(< "$DOKKU_ROOT/$APP/CONTAINER")
+    docker cp "$DOKKU_APP_CONTAINER_ID:/app/nginx.conf" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
```
josegonzalez: @michaelshobbs This is the correct place to put this command, right?

michaelshobbs: Yes but that's not how to get the container id anymore since we did scaling. You'll want something like [this](https://github.com/progrium/dokku/blob/483ca9627f390a2b5c7e0382488cad95fc62b032/plugins/00_dokku-standard/commands#L120). Then grab the first one or just use the image and start a container with `--rm` maybe?

josegonzalez: Gotcha. I'll update the pr tomorrow to do so.

josegonzalez:
@michaelshobbs this look good to go?

plugins/nginx-vhosts/commands
```diff
@@ -49,6 +49,14 @@ case "$1" in
       shopt -u nullglob
     fi
 
+    DOKKU_APP_CIDS=$(get_container_ids $APP)
+    if [[ -n $DOKKU_APP_CIDS ]]; then
```
michaelshobbs: Hrm. I may have led you astray. Might be best to spin up a new container with the APP image and cp from just one container. Alternatively, maybe just `docker cp` from the first CID in the list.

plugins/nginx-vhosts/commands
```diff
@@ -49,6 +49,14 @@ case "$1" in
...
+      for DOKKU_APP_CID in $DOKKU_APP_CIDS; do
+        docker cp "$DOKKU_APP_CID:/app/nginx.conf" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
```
josegonzalez: I call `docker cp` from the first in the list here :)

michaelshobbs: Oh and then break. Boooooo. :smile: 

This may be personal preference but maybe something like this?

```
DOKKU_APP_CIDS=($(get_container_ids $APP))
```

josegonzalez: Why a second set of parens on that first line?

michaelshobbs: Attempting to make `$DOKKU_APP_CIDS` an array so that `${DOKKU_APP_CID[0]}` is valid syntax.

plugins/nginx-vhosts/commands
```diff
@@ -49,6 +49,9 @@ case "$1" in
       shopt -u nullglob
     fi
 
+    DOKKU_APP_CIDS=($(get_container_ids $APP))
+    docker cp "${DOKKU_APP_CID[0]}:/app/nginx.conf" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
```
josegonzalez: Should we use nginx.conf.template here?

michaelshobbs: I really think we should to be consistent and to call out that the file is actually a template and not a raw conf file.

PR #1280, opened 2015-07-04
fixes #1259 now ignores protocol of Dockerfile EXPOSE

### LINKED ISSUE #1259: EXPOSE parser doesn't understand port protocol
something like `EXPOSE 8888/tcp` is not supported, if you are using some else's image that was not tailored for dokku this can easily happen.

kilianc:
updating tests now

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -36,7 +36,7 @@ case "$1" in
 
       dockerfile)
         # extract first port from Dockerfile
-        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }' || true)
+        DOCKERFILE_PORT=$(grep "^EXPOSE \+[[:digit:]]\+\(\/tcp\)\? *$" Dockerfile | head -1 | sed 's/EXPOSE \+\([[:digit:]]\+\)\(\/tcp\)\?.*/\1/' || true)
         [[ -n "$DOCKERFILE_PORT" ]] && dokku config:set-norestart $APP DOKKU_DOCKERFILE_PORT=$DOCKERFILE_PORT
 
         # buildstep pluginhooks don't necessarily make sense for dockerfiles. call the new breed!!!
--- a/tests/apps/dockerfile/Dockerfile
+++ b/tests/apps/dockerfile/Dockerfile
@@ -2,7 +2,9 @@ FROM ubuntu:trusty
 ENV LC_ALL C
 ENV DEBIAN_FRONTEND noninteractive
 ENV DEBCONF_NONINTERACTIVE_SEEN true
-EXPOSE 3000
+EXPOSE 3001/udp
+EXPOSE  3000/tcp
+EXPOSE 3003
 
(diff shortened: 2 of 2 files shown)


plugins/00_dokku-standard/commands
```diff
@@ -36,7 +36,7 @@ case "$1" in
 
       dockerfile)
         # extract first port from Dockerfile
-        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }' || true)
+        DOCKERFILE_PORT=$(grep EXPOSE Dockerfile | head -1 | awk '{ print $2 }' | sed 's/\/.*//' || true)
```
josegonzalez: Should we ensure this is a tcp port and not a udp one?

kilianc: I managed to do that let's wait for the test suite result

kilianc:
btw, I feel like we're doing it wrong, we should probably inspect the container and look for first open tcp port

josegonzalez:
The container isn't running yet though, is it? And I think we need this information beforehand (though maybe I'm wrong).

kilianc:
@josegonzalez this is how it works now, but it's doable because we turn it on -> grab port -> wait for checks -> reload nginx -> kill old container

It may require some work, so for the time being I would just merge this if it works and iterate again

josegonzalez:
And that regex works if you don't use `/tcp`, right?

kilianc:
:+1:

josegonzalez:
Seems like tests fail.

kilianc:
Seems like it, I will double check in the next days

kilianc:
@josegonzalez it appears that grep has different regexp implementations in different platforms, so I rewrote it using basic POSIX. it's ugly but it works.

Any advice on how to test both `EXPOSE 3000` and `EXPOSE 3000/tcp` ? I don't want to have 2 docker apps and the test suite is insanely slow already... I wish we had some unit testing in place for this

josegonzalez:
@michaelshobbs given that you implemented dockerfile support, thoughts on his last question?

michaelshobbs:
eh well, if the point of the test is to have a running dockerfile container with two different formats, I'm not sure we have a good way right now. Stubbing out the deployment functions for this type of stuff that doesn't actually require a `curl`-style request at the end would be nice.

kilianc:
> EDIT: maybe make it a function, source the functions file in the test and then use bats from there?

yep this is what I was looking for, I will add a `get_exposed_port()` in this PR

michaelshobbs:
how's this looking @kilianc?

kilianc:
@michaelshobbs no time this week, bee busy on another project :/ feel free to take over

josegonzalez:
@michaelshobbs want to take a crack at this? Just this and the other pr before I can make a release.

michaelshobbs:
@josegonzalez yes. #1311

josegonzalez:
Closing as #1311 is now merged. Thanks @kilianc for reporting the issue and providing an initial fix :)

PR #1299, opened 2015-07-09
Set X-Forwarded-Proto header if TLS is enabled when running checks

SonicHedgehog:
If I have an app which only responds to HTTPS requests (or redirects non-secure requests to HTTPS) any checks defined in the `CHECKS` file will fail. This isn’t a problem for “real requests” as nginx sets the `X-Forwarded-Proto` header accordingly.

I therefore suggest setting the `X-Forwarded-Proto` header to `https` when running checks against a TLS-enabled app.

### DIFF
--- a/plugins/checks/check-deploy
+++ b/plugins/checks/check-deploy
@@ -125,6 +125,11 @@ do
   # --location   Follow redirects
   CURL_OPTIONS="-q --compressed --fail --location --max-time $TIMEOUT"
 
+  # Set X-Forwarded-Proto header if TLS is enabled.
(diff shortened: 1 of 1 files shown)


michaelshobbs:
Seems reasonable. Have you used this in a live environment?

SonicHedgehog:
I have (only) tested the change with this simple Express app:

```node
app.enable('trust proxy')

app.get('/', function(request, response) {
  if(!request.secure) {
    return response.redirect(301, 'https://' + request.get('Host') + request.url)
  }

  response.send('Hello World!')
})
```

michaelshobbs:
#1300 comes with some test fixes. can you rebase once I merge that?

SonicHedgehog:
No problem.

michaelshobbs:
g2g

PR #1300, opened 2015-07-09
fix urls command when NO_VHOST=1 + tests

michaelshobbs:


michaelshobbs:
blocking #1299 sort of...

PR #1310, opened 2015-07-12
use config:get for checks skipping variables

michaelshobbs:
this uses `config:get` to set `$DOKKU_SKIP_ALL_CHECKS` and `$DOKKU_SKIP_DEFAULT_CHECKS`

### DIFF
--- a/dokku
+++ b/dokku
@@ -73,9 +73,6 @@ case "$1" in
     DOKKU_SCALE_FILE="$DOKKU_ROOT/$APP/DOKKU_SCALE"
     oldids=$(get_container_ids $APP)
 
-    [[ -f "$DOKKU_ROOT/ENV" ]] && source $DOKKU_ROOT/ENV
-    [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
-
     while read line || [ -n "$line" ]
     do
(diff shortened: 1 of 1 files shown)


josegonzalez:
We still use global sourcing in `check-deploy` and `nginx-vhosts`. Mind fixing those up as well? :D

michaelshobbs:
In those cases, I would generally consider it to be safe to source those files. 

More specifically though:
In `check-deploy` the additional lines of code to extract 3 more variables seems to not outweigh that additional cost in code complexity.

In `nginx-vhosts` I thought we wanted to source the entire thing there in case folks had custom template-y stuff they wanted to do.

josegonzalez:
I guess what I'm trying to do is contain all manipulation of the env files to the config plugin where possible.

michaelshobbs:
I would contend that sourcing is not manipulation but that may just be me being pedantic. 

I think changes and/or inspection (i.e. does this env var exist) should only be handled in the config plugin. However activation of global/app env is fine to be done where sane and necessary. Just my 2 cents.

josegonzalez:
This is re: me trying to replace the config plugin with something that can be distributed. Is there maybe a way we can expose the data we need to from within the config plugin as opposed to direct inspection of the files on disk?

michaelshobbs:
I mean we could wrap `source ....` in a config plugin command...

Issue #1303, opened 2015-07-10
Documentation: Improvements about explanation about custom nginx-conf


fwolfst:
Hey there,

I struggle(d) a bit with nginx-configuration per app, and still have a confusing issue. I prematurely issued a PR, which I will improve upon and bump here.

Have fun
Felix

josegonzalez:
Thanks for the docs change.

fwolfst:
Hey,
I wasn't finished yet, its not correct as-is. Still have to spend some minutes understanding what happens. Could you re-open?
Thanks,
Felix

josegonzalez:
You can open pull requests with your changes, no need to have this issue open to do that.

fwolfst:
There is something in https://github.com/progrium/dokku/issues/1309 that I have not yet understood, will update the docs when I understood (or fixed) what is happening

PR #1321, opened 2015-07-16
In doc: mention alternative to nginx.conf templates: include-dir.

fwolfst:
In the line of #1303 , mention that parts of nginx-configuration can also be put in `$APP/nginx.conf.d/` if default templates are used.

Reference to comment in other issue (does not look very nice in the documentation, can probably be removed).

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -102,14 +102,12 @@ The default nginx.conf- templates will include everything from your apps `nginx.
 
     include $DOKKU_ROOT/$APP/nginx.conf.d/*.conf;
 
-. That means you can put additional configuration in separate files, for example to limit the uploaded body size to 50 megabyte, do
+. That means you can put additional configuration in separate files, for example to limit the uploaded body size to 50 megabytes, do
 
     mkdir /home/dokku/myapp/nginx.conf.d/
     echo 'client_max_body_size 50M;' > /home/dokku/myapp/nginx.conf.d/upload.conf
     chown dokku:dokku /home/dokku/myapp/nginx.conf.d/upload.conf
     service nginx reload
-
-(taken from https://github.com/dokku-alt/dokku-alt/issues/195#issuecomment-75092205).
 
 ## Customizing hostnames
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
This isn't dokku-alt,  so this actually isn't true.

fwolfst:
As far as I can tell from the distance (I do not use dokku-alt) the templates are the same (`include **conf.d`) and it works like a charm if I didn't mess my test up.

michaelshobbs:
@josegonzalez thoughts on adding this to the default template?

fwolfst:
@michaelshobbs @josegonzalez It IS part of the default templates, its just not mentioned in the docs.

docs/nginx.md
```diff
@@ -96,6 +96,21 @@ A few tips for custom nginx templates:
 
 After your changes a `dokku deploy myapp` will regenerate the `/home/dokku/myapp/nginx.conf` file which is then used.
 
+### Customizing via configuration files included by the default templates
+
+The default nginx.conf- templates will include everything from your apps `nginx.conf.d/` subdirectory in the main `server {}` block (see above):
+
+    include $DOKKU_ROOT/$APP/nginx.conf.d/*.conf;
+
+. That means you can put additional configuration in separate files, for example to limit the uploaded body size to 50 megabyte, do
```
michaelshobbs: I think you mean megabytes?

docs/nginx.md
```diff
@@ -96,6 +96,21 @@ A few tips for custom nginx templates:
...
+
+    mkdir /home/dokku/myapp/nginx.conf.d/
+    echo 'client_max_body_size 50M;' > /home/dokku/myapp/nginx.conf.d/upload.conf
+    chown dokku:dokku /home/dokku/myapp/nginx.conf.d/upload.conf
+    service nginx reload
+
+(taken from https://github.com/dokku-alt/dokku-alt/issues/195#issuecomment-75092205).
```
michaelshobbs: Not sure why we would mention a different project here. Can you remove this line?

josegonzalez:
This can be merged once the two comments are fixed.

PR #1331, opened 2015-07-20
[bugfix] in nginx.conf.template getting from /app directory

alexkruegger:
There is bugfix for ability to get 'nginx.conf.template' from original repository. The bug is small and fix is just syntax correction.

### DIFF
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -50,7 +50,7 @@ case "$1" in
     fi
 
     DOKKU_APP_CIDS=($(get_app_container_ids $APP))
-    docker cp "${DOKKU_APP_CID[0]}:/app/nginx.conf.template" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
+    docker cp "${DOKKU_APP_CIDS[0]}:/app/nginx.conf.template" "$DOKKU_ROOT/$APP/" 2> /dev/null || true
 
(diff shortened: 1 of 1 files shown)


plugins/nginx-vhosts/commands
```diff
@@ -50,7 +50,7 @@ case "$1" in
     fi
 
     DOKKU_APP_CIDS=($(get_app_container_ids $APP))
-    docker cp "${DOKKU_APP_CID[0]}:/app/nginx.conf.template" "$APP_NGINX_TEMPLATE" 2> /dev/null || true
+    docker cp "${DOKKU_APP_CIDS[0]}:/app/nginx.conf.template" "$DOKKU_ROOT/$APP/" 2> /dev/null || true
```
josegonzalez: Doesn't `$APP_NGINX_TEMPLATE ` need to be the full path?

alexkruegger: As I tested it - no. You should define only directory where template should be stored. If you use $APP_NGINX_TEMPLATE as the full path, docker will create directory with this name, not the file.

michaelshobbs: :+1: on fixing `$DOKKU_APP_CIDS`. I missed that one in review. Any reason to change from using `$APP_NGINX_TEMPLATE`? I like variable re-use. :smile:

@josegonzalez to your question, in docker 1.6.2 at least, the destination does not *have* to be a path to a file.

alexkruegger: :) Thank you for nice tool and fast response! Happy coding! )

Issue #1336, opened 2015-07-22
Usage Question


zimt28:
Hey there! I'm quite new to Docker and was looking for a way to easily deploy my apps - seems like dokku is a good fit :) I now want to deploy a really basic app, but can't get it running ..

Here's the `Dockerfile`:
```Dockerfile
FROM ubuntu:14.04

RUN apt-get update
RUN apt-get install -y nginx
RUN echo 'Hello from your container' \
  >/usr/share/nginx/html/index.html

CMD ["nginx",  "-g", "daemon off;"]

EXPOSE 80
```

I can build the image `docker build -t="zimt28/static_web" .` and run a container `docker run -d -p 80 zimt28/static_web`. Locally, everything works fine; I can now access the container using the port from `docker ps`.

I've now created a git repo, added the Dockerfile and the dokku remote and pushed the app. My app gets listed using `dokku apps` on the server and I can get a url using `dokku url appname`.  However, when I open that url, I can't connect to the server. I've then added the `-d` and `-p 80` options using `dokku docker-options:add appname run` and restarted the server using `dokku ps:restart appname`. However, I still can't access the page.

Can you tell me, what I'm doing wrong here?
Thank you :)

josegonzalez:
What is the output of `docker ps`?

zimt28:
```
root@minon1:~# docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS                             NAMES
284a3d44cdd6        dokku/master        "nginx -g 'daemon of   12 minutes ago      Up 12 minutes       0.0.0.0:32775->80/tcp             boring_almeida
9a87301abcaa        dokku/master        "nginx -g 'daemon of   18 minutes ago      Up 18 minutes       80/tcp, 0.0.0.0:32774->5000/tcp   sharp_tesla
```

zimt28:
Ok, so now when I open the app on port `32775`, everything works fins. Looks like `dokku url` is just giving me the port of the second container

michaelshobbs:
what is the output of `dokku url` and `docker ps` from the same deployment?

zimt28:
Outputs are `http://{serverIP}:32774` and 
```
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS                             NAMES
284a3d44cdd6        dokku/master        "nginx -g 'daemon of   15 hours ago        Up 15 hours         0.0.0.0:32775->80/tcp             boring_almeida
[...]

PR #1341, opened 2015-07-24
set and use $port whenever $PROC_TYPE == 'web'

michaelshobbs:
This honors `$DOKKU_DOCKERFILE_PORT` when binding the docker container to an external IP.

refs #1336

### DIFF
--- a/dokku
+++ b/dokku
@@ -101,15 +101,17 @@ case "$1" in
           START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
         fi
 
-        [[ "$PROC_TYPE" == "web" ]] && port=${DOKKU_DOCKERFILE_PORT:=5000}
-        if [[ "$BIND_EXTERNAL" = "false" ]] && [[ "$PROC_TYPE" == "web" ]];then
-          id=$(docker run -d -e PORT=$port $DOCKER_ARGS $IMAGE $START_CMD)
-          ipaddr=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' $id)
-        elif [[ "$BIND_EXTERNAL" = "true" ]] && [[ "$PROC_TYPE" == "web" ]];then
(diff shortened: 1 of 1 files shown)


michaelshobbs:
@josegonzalez I think we just missed this use case. Not sure we want to add a deploy test specifically for this though. Thoughts?

dokku
```diff
@@ -101,13 +101,13 @@ case "$1" in
           START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
         fi
 
+        [[ "$PROC_TYPE" == "web" ]] && port=${DOKKU_DOCKERFILE_PORT:=5000}
```
josegonzalez: The first if/elif stanza can probably be simplified a bit as well.

josegonzalez:
No test needed imo.

dokku
michaelshobbs: How do you mean?

josegonzalez: Like, we check several times what the PROC_TYPE is, as well as two checks on BIND_EXTERNAL.

michaelshobbs: I guess I meant, do you have a recommendation?

josegonzalez: Something like:

```
if [[ "$PROC_TYPE" == "web" ]]; then
    port=${DOKKU_DOCKERFILE_PORT:=5000}
    if [[ "$BIND_EXTERNAL" = "true" ]]; then
        id=$(docker run -d -p $port -e PORT=$port $DOCKER_ARGS $IMAGE $START_CMD)
        port=$(docker port $id $port | sed 's/[0-9.]*://')
        ipaddr=127.0.0.1
    else
        id=$(docker run -d -e PORT=$port $DOCKER_ARGS $IMAGE $START_CMD)
        ipaddr=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' $id)
    fi
fi
```

*shrug*

michaelshobbs: oh oh oh yeah sorry I was overthinking it. this makes sense.

PR #1344, opened 2015-07-27
Add better error checking on nginx:import-ssl

AdamVig:
The script would only accept files named exactly `server.crt` and `server.key`. Some certificate authorities distribute these files with names according to the domain they apply to, so this approach on dokku's end is rather limited.

I added error handling that ensures there is only one `.crt` file and one `.key` file and then modified the `mv` command so it renames the files from whatever to `server.crt` and `server.key`.

An error case not handled by my addition (feel free to tack on!) is if the `tar`ed files have a path in front of them, for example `~/ssl/server.crt`. EDIT: This error case is now handled! Files will be found if they are extracted to a subdirectory.

Let me know if my code does not follow standards/best practices, this is my first time contributing to dokku and to a bash project! 😄

### DIFF
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -162,17 +162,17 @@ EOF
     cd $TEMP_DIR
     tar xvf - <&0
 
-    crt_count = $(ls -1 *.crt | wc -l)
-    if [[ $crt_count -lt 1 ]]; then
+    CRT_FILE_COUNT=$(ls -1 ./*.crt | wc -l)
+    if [[ $CRT_FILE_COUNT -lt 1 ]]; then
       echo "Tar archive is missing .crt file" && exit 1
-    elif [[ $crt_count -gt 1 ]]; then
+    elif [[ $CRT_FILE_COUNT -gt 1 ]]; then
       echo "Tar archive contains more than one .crt file" && exit 1
     fi
 
-    key_count = $(ls -1 *.key | wc -l)
-    if [[ $key_count -lt 1 ]]; then
+    KEY_FILE_COUNT=$(ls -1 ./*.key | wc -l)
+    if [[ $KEY_FILE_COUNT -lt 1 ]]; then
       echo "Tar archive is missing .key file" && exit 1
-    elif [[ $key_count -gt 1 ]]; then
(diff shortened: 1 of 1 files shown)


plugins/nginx-vhosts/commands
```diff
@@ -161,12 +161,24 @@ EOF
...
+      echo "Tar archive is missing .key file" && exit 1
+    elif [[ $KEY_FILE_COUNT -gt 1 ]]; then
+      echo "Tar archive contains more than one .key file" && exit 1
+    fi
 
     mkdir -p "$DOKKU_ROOT/$APP/tls"
-    mv "$TEMP_DIR/server.crt" "$DOKKU_ROOT/$APP/tls/server.crt"
-    mv "$TEMP_DIR/server.key" "$DOKKU_ROOT/$APP/tls/server.key"
+    mv "$TEMP_DIR/*.crt" "$DOKKU_ROOT/$APP/tls/server.crt"
+    mv "$TEMP_DIR/*.key" "$DOKKU_ROOT/$APP/tls/server.key"
```
josegonzalez: Maybe use `find` to find the full path to the cert files within the tar?

plugins/nginx-vhosts/commands
```diff
@@ -161,12 +161,24 @@ EOF
     TEMP_DIR=$(mktemp -d)
     cd $TEMP_DIR
     tar xvf - <&0
-    [[ ! -f "$TEMP_DIR/server.crt" ]] && echo "Tar archive missing server.crt" && exit 1
-    [[ ! -f "$TEMP_DIR/server.key" ]] && echo "Tar archive missing server.key" && exit 1
+
+    CRT_FILE_COUNT=$(ls -1 ./*.crt | wc -l)
```
michaelshobbs: Probably better to glob it vs using find.

plugins/nginx-vhosts/commands
AdamVig: That's a great idea! `shellcheck` is telling me to use `find` in place of `ls` as well. I will implement both changes tomorrow.

AdamVig:
@josegonzalez Updated this PR to incorporate your suggestions and other fixes/improvements. Please revisit.

josegonzalez:
Thanks for the PR! This must have gotten lost in the shuffle :)

PR #1346, opened 2015-07-27
document dokku cleanup and the potential of compat issues

michaelshobbs:


michaelshobbs:
@josegonzalez let me know if you think this is clear enough or not

PR #1349, opened 2015-07-28
add plugin dokku-app-configfiles

alexkruegger:
If I correctly understand I should edit this file to add link to the new plugin and then create PR to commit it. Am I right?

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -16,9 +16,6 @@ cd /var/lib/dokku/plugins
 git clone <git url>
 dokku plugins-install
 ```
-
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -17,6 +17,8 @@ git clone <git url>
 dokku plugins-install
 ```
 
+> todo: add a command to dokku to install a plugin, given a git repository `dokku plugin:install <git url>`?
```
michaelshobbs: We probably don't want this line here. However it's worth opening an enhancement issue to request this feature or a PR implementing it.

alexkruegger: ok. Thank you.

PR #1352, opened 2015-07-30
Add nginx:access-logs and nginx:error-logs.

### LINKED ISSUE #1159: dokkus nginx-vhost plugin should make nginx access.log and error.log available via a command
Hey guys,
really happy with my dokku experience, but I'd like an addition to the nginx(-vhost) plugin, building on https://github.com/progrium/dokku/issues/1158:

Provide access to the respective logs, like

    $ dokku nginx:access-log myapp
    $ dokku nginx:access-log myapp -t
    $ dokku nginx:error-log -t

I hope its relatively clear what I mean :)

isundaylee:
Add two nginx subcommands to display nginx error/access logs for a particular app.

closes #1159

### DIFF
--- a/plugins/nginx-vhosts/commands
+++ b/plugins/nginx-vhosts/commands
@@ -172,10 +172,30 @@ EOF
     dokku nginx:build-config $APP
     ;;
 
+  nginx:access-logs|nginx:error-logs)
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    verify_app_name "$2"
+    APP="$2"
+
+    NGINX_LOGS_TYPE=${1#nginx:}
+    NGINX_LOGS_TYPE=${NGINX_LOGS_TYPE%-logs}
+    NGINX_LOGS_PATH="/var/log/nginx/$APP-$NGINX_LOGS_TYPE.log"
+
+    if [[ $3 == "-t" ]]; then
+      NGINX_LOGS_ARGS="-F"
+    else
+      NGINX_LOGS_ARGS="-n 20"
+    fi
+
+    sudo tail "$NGINX_LOGS_ARGS" "$NGINX_LOGS_PATH"
+    ;;
+
   help | nginx:help)
     cat && cat<<EOF
     nginx:import-ssl <app>, Imports a tarball from stdin; should contain server.crt and server.key
     nginx:build-config <app>, (Re)builds nginx config for given app
+    nginx:access-logs <app> [-t], Show the nginx access logs for an application (-t follows)
+    nginx:error-logs <app> [-t], Show the nginx error logs for an application (-t follows)
 EOF
     ;;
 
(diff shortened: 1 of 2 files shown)


plugins/nginx-vhosts/install
```diff
@@ -4,10 +4,12 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 case "$DOKKU_DISTRO" in
   ubuntu)
     echo "%dokku ALL=(ALL) NOPASSWD:/etc/init.d/nginx reload, /usr/sbin/nginx -t" > /etc/sudoers.d/dokku-nginx
+    echo "%dokku ALL=(ALL) NOPASSWD:/usr/bin/tail" >> /etc/sudoers.d/dokku-nginx
     ;;
 
   opensuse)
     echo "%dokku ALL=(ALL) NOPASSWD:/sbin/service nginx reload, /usr/sbin/nginx -t" > /etc/sudoers.d/dokku-nginx
+    echo "%dokku ALL=(ALL) NOPASSWD:/usr/bin/tail" >> /etc/sudoers.d/dokku-nginx
```
michaelshobbs: I think this can be pared down to `/usr/bin/tail /var/log/nginx/*` and `/usr/bin/tail -f /var/log/nginx/*`. Test this of course. :smile:

plugins/nginx-vhosts/commands
```diff
@@ -172,10 +172,30 @@ EOF
...
+    NGINX_LOGS_TYPE=${NGINX_LOGS_TYPE%-logs}
+    NGINX_LOGS_PATH="/var/log/nginx/$APP-$NGINX_LOGS_TYPE.log"
+
+    if [[ $3 == "-t" ]]; then
+      NGINX_LOGS_ARGS="-F"
+    else
+      NGINX_LOGS_ARGS="-n 20"
+    fi
+
+    sudo tail "$NGINX_LOGS_ARGS" "$NGINX_LOGS_PATH"
```
michaelshobbs: Might be good to line this call up with what's in the sudoers file?

plugins/nginx-vhosts/install
michaelshobbs: The more I think about it, allowing any user with access to the dokku account to tail any file, is a security risk I don't think is worth introducing.

josegonzalez:
@isundaylee mind updating the PR to match expectations?

plugins/nginx-vhosts/install
michaelshobbs: After doing the `plugin` plugin, I remembered that we assume `dokku` will be run by root and thus we `sudo su - dokku` here:  https://github.com/progrium/dokku/blob/63d89942438e2806c5322c2ecddbe2d008e03f42/dokku#L44-L47

Therefore I think we can just filter these commands as well to make them not sudo. I'll create a new PR based on this work.

josegonzalez:
Closing in favor of #1469. Thanks for your work @isundaylee!

PR #1356, opened 2015-07-30
Allow stop/kill to continue in the event that it cannot stop or kill one of the old containers

akatrevorjay:
In the event that we cannot stop a container, we should try to kill it, and if that then fails, we should continue to the next $oldid , otherwise one container not actually being there or not able to be stopped will prevent the rest from being cleaned up.

The way this is, if a container ever fails to stop, it will never kill it anyway, due to being executed with -e.
This fixes that as well with a simple or.

The bigger issue at hand here is that if you change your container process types around it doesn't really cleanup after itself, leaving CONTAINER.$type.$nid and PORT.$type.$nid files around, but I didn't get around to addressing that.

### DIFF
--- a/dokku
+++ b/dokku
@@ -178,8 +178,12 @@ case "$1" in
         trap '' INT HUP
         sleep $WAIT
         for oldid in $oldids; do
-          docker stop $oldid &> /dev/null
-          docker kill $oldid &> /dev/null  # force a kill as docker seems to not send SIGKILL as the docs would indicate
+          # Attempt to stop, if that fails, then force a kill as docker seems
+          # to not send SIGKILL as the docs would indicate. If that fails, move
+          # on to the next.
+          docker stop $oldid \
+          || docker kill $oldid \
+          || :  # Always continue in case we have multiple to kill
         done
       ) & disown -a
       # Use trap since disown/nohup don't seem to keep child alive


michaelshobbs:
Thanks for the PR. :+1: for me. @josegonzalez thoughts here?

josegonzalez:
Reminds me of that `docker_sleep` thing I wrote.

@akatrevorjay is there a way we can log a message when the kill fails? Maybe a hook call...

michaelshobbs:
Oh yeah. Where is that thing? Was it just in a gist somewhere or do you have it in a repo?

The docker stop/kill loop gets `disown`'d. I'm not sure we want to introduce a pluginhook call "out-of-band" like that. Sounds like a race-condition waiting to happen. Just my thoughts though.

josegonzalez:
Is there a way we can introduce general logging of this?

michaelshobbs:
We could trigger an event

josegonzalez:
Lets do that then. I want to have *something* there for our users so that at least it can be logged/detected and taken care of, even if out of band.

michaelshobbs:
:+1: 
Implementation thoughts
- add a new events plugin pluginhook 
- call it after every kill/stop

michaelshobbs:
@akatrevorjay let me know if you're comfortable adding the logging or if you'd prefer assistance

josegonzalez:
@michaelshobbs pretty sure @akatrevorjay won't get to the logging bit, so if you want to take the reins and merge this, go ahead.

PR #1357, opened 2015-07-30
only run domains and nginx config if we have a port and ip

### LINKED ISSUE #1355: Cannot deploy without web process
Hi awesome people!
I recently set up a new machine with the latest dokku and noticed I was having problems deploying.
my scale settings are configured for ``worker=4 web=0``.

Here's the output:

```
-----> Running post-deploy
-----> Configuring SSL for example.com...
-----> Configuring SSL for www.example.com...
       Reloading nginx
remote: nginx: [emerg] host not found in upstream "maggie" in /home/dokku/maggie/nginx.conf:30
remote: nginx: configuration file /etc/nginx/nginx.conf test failed
To dokku@web2.ext.example.com:maggie
error: failed to push some refs to 'dokku@web2.ext.example.com:maggie'
```

Any hints?
It might be worth noting that my workers are actually running fine.

michaelshobbs:
- only run domains:setup and nginx:build-conifg if IP.web.1 and PORT.web.1 files exist
- added test app that just contains a worker process

fixes #1355

### DIFF
--- a/plugins/nginx-vhosts/post-deploy
+++ b/plugins/nginx-vhosts/post-deploy
@@ -3,12 +3,14 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$(dirname $0)/../common/functions"
 
 APP="$1"
-NO_VHOST=$(dokku config:get $APP NO_VHOST || true)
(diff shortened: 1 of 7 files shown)


josegonzalez:
Does this work for apps that have multiple web procs?

michaelshobbs:
Yes. We test scaling and those tests pass.

PR #1369, opened 2015-08-05
From /dev/null to -qq, from --silent to -sSSL, and sudo

SamuelMarks:


SamuelMarks:
[closing for now, working on a better patch]

PR #1370, opened 2015-08-05
From /dev/null to -qq, from --silent to -sL, and sudo

SamuelMarks:


docs/getting-started/install/debian.md
```diff
@@ -3,13 +3,13 @@
...
 
 ```shell
-curl --silent https://get.docker.io/gpg 2> /dev/null | apt-key add - 2>&1 >/dev/null
-curl --silent https://packagecloud.io/gpg.key 2> /dev/null | apt-key add - 2>&1 >/dev/null
+curl -sL https://get.docker.io/gpg 2> /dev/null | sudo apt-key add - 2>&1 >/dev/null
+curl -sL https://packagecloud.io/gpg.key 2> /dev/null | sudo apt-key add - 2>&1 >/dev/null
 
-echo "deb http://get.docker.io/ubuntu docker main" > /etc/apt/sources.list.d/docker.list
-echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ trusty main" > /etc/apt/sources.list.d/dokku.list
+sudo echo "deb http://get.docker.io/ubuntu docker main" > /etc/apt/sources.list.d/docker.list
```
epheph: This specific line doesn't work, since only echo is being sudo'd. Needs to be wrapped in "sh -c" so it is all sudo, or piped to `sudo tee /etc/apt/sources.list.d/docker.list`

Also, i'm not sure I'm not sure this is the behavior one might expect, a script escalating itself via sudo.

SamuelMarks: Ah, true that. I was running this example in a slightly modified context—using fabric—so didn't realise. Thanks.

josegonzalez:
Thanks for the pr!

SamuelMarks:
nps :)

PR #1372, opened 2015-08-05
Do no longer force-install a default nginx.conf

SonicHedgehog:
As far as I can tell, a default `nginx.conf` does no longer need to be force-installed. Let me know if I forgot something.

### DIFF
--- a/plugins/nginx-vhosts/install
+++ b/plugins/nginx-vhosts/install
@@ -22,7 +22,6 @@ ssl_session_cache shared:SSL:20m;
 ssl_session_timeout 10m;
 
 ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5;
-ssl_protocols TLSv1.2 TLSv1.1 TLSv1;
 ssl_prefer_server_ciphers on;
 
 # ssl_certificate $DOKKU_ROOT/tls/server.crt;
--- a/plugins/nginx-vhosts/install
+++ b/plugins/nginx-vhosts/install
@@ -18,6 +18,8 @@ if [[ ! -f /etc/nginx/conf.d/dokku.conf ]]; then
   cat<<EOF > /etc/nginx/conf.d/dokku.conf
 include $DOKKU_ROOT/*/nginx.conf;
 
+server_tokens off;
+
 ssl_session_cache shared:SSL:20m;
 ssl_session_timeout 10m;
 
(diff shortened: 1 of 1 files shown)


plugins/nginx-vhosts/install
```diff
@@ -22,45 +22,12 @@ ssl_session_cache shared:SSL:20m;
 ssl_session_timeout 10m;
 
 ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5;
-ssl_protocols TLSv1.2 TLSv1.1 TLSv1;
-ssl_prefer_server_ciphers on;
```
josegonzalez: Why did you remove these ciphers?

josegonzalez:
The force install was introduced [here](https://github.com/progrium/dokku/issues/799).

plugins/nginx-vhosts/install
SonicHedgehog: I have removed the `ssl_protocols` and `ssl_prefer_server_ciphers` directives because they are already included in the default `nginx.conf` of nginx 1.8.0 (the current stable version). Otherwise, nginx would throw an `[emerg] directive is duplicate` error.

josegonzalez: Is that version installed with ubuntu 14.04?

SonicHedgehog: No, but [it is installed when installing dokku](https://github.com/progrium/dokku/blob/v0.3.22/plugins/nginx-vhosts/dependencies#L9) as it uses the [PPA nginx/stable](https://launchpad.net/~nginx/+archive/ubuntu/stable).

josegonzalez: Gotcha, makes sense to me.

josegonzalez:
Seems like tests fail?

SonicHedgehog:
Fixed.

plugins/nginx-vhosts/install
josegonzalez: If you install dokku from source, it won't use the PPA. Hmmm...

josegonzalez:
Going to close this. A real fix needs to take into account the fact that we don't use the PPA for nginx with the debian solution :( I'll take care of this in #1467.

josegonzalez:
Scratch that. Seems this can go in after all :)

josegonzalez:
Closing in favor of #1468. Thanks for your work @SonicHedgehog!

michaelshobbs:
Whoa! We want both I think, yea?

josegonzalez:
Right I'm a nub. Reopened.

PR #1379, opened 2015-08-10
Deleted my old plugins since there is no maintenance happening.

jezdez:


josegonzalez:
Thanks for the cleanup, much appreciated :)

PR #1380, opened 2015-08-11
Removed uses of `sudo` from `bootstrap.sh`

emdantrim:
`apt` package installs in `bootstrap.sh` fails to acknowledge the `DEBIAN_FRONTEND=noninteractive` envvar because of the use of `sudo`, because the -E flag is not set. However, the script also assumes that it will be run with `sudo` in the first place, so it shouldn't be needed. In the event that it is needed, it can be run with -E instead.


josegonzalez:
Good catch, thanks for the pr.

PR #1399, opened 2015-08-20
Make dokku play nice when we have multiple receive-app hooks 

basicer:


michaelshobbs:
What is the use case of multiple `receive-app` hooks?

josegonzalez:
He wants to use both mercurial and git for deployments. As is, the output of one of them is eaten up, so deploy output isn't very useful :(

plugins/git/receive-app
```diff
@@ -4,4 +4,8 @@ source "$(dirname $0)/../common/functions"
 
 APP="$1"; REV="$2"
 
-dokku git-build $APP $REV
+if [ ! -d "$DOKKU_ROOT/$APP/refs" ]; then
+  cat
+else
+  dokku git-build $APP $REV
+fi
```
michaelshobbs: Can you add context here?

basicer: If you run ps:rebuild, but the application is not backed by git, you get an ugly error message to that effect from git.  Also its my understanding that if your plugin doesn't wish to act, that it should exec cat so that the messages on std out by other processes get passed though.

Basically we only want to call git-build if we are building a git based app.

michaelshobbs: :+1:

PR #1402, opened 2015-08-21
Consolidate config

josegonzalez:
This PR is an attempt at consolidating configuration retrieval and setting within the `config` plugin. This would enable developers to replace their `config` plugin with another plugin, one potentially backed by a database, and therefore would allow users to deploy a cluster of dokku instances backed by a single configuration database.

- Added the `--export` option to `dokku config`.
- Switch to config:set where possible.
- Use `eval` (oooo evil) to set config env vars in the current context.

Todo:

- Remove all remaining checks for `ENV` files.
- Add `dokku config:has ENV_VAR` method.
- Fix cases where we are concatenating the raw `ENV` files for docker container runs.

### DIFF
--- a/docs/development/pluginhooks.md
+++ b/docs/development/pluginhooks.md
@@ -338,7 +338,7 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 CONTAINERID="$1"; APP="$2"; PORT="$3" ; HOSTNAME="${4:-localhost}"
 
-eval $(dokku config $APP --export)
+eval "$(dokku config $APP --export)"
 DOKKU_DISABLE_DEPLOY="${DOKKU_DISABLE_DEPLOY:-false}"
 
 if [[ "$DOKKU_DISABLE_DEPLOY" = "true" ]]; then
--- a/docs/development/pluginhooks.md
+++ b/docs/development/pluginhooks.md
@@ -338,7 +338,7 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 CONTAINERID="$1"; APP="$2"; PORT="$3" ; HOSTNAME="${4:-localhost}"
 
-[[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
+eval $(dokku config $APP --export)
 DOKKU_DISABLE_DEPLOY="${DOKKU_DISABLE_DEPLOY:-false}"
 
 if [[ "$DOKKU_DISABLE_DEPLOY" = "true" ]]; then
--- a/docs/development/pluginhooks.md
+++ b/docs/development/pluginhooks.md
@@ -335,6 +335,7 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 # `DOKKU_DISABLE_DEPLOY` env var is set to `true` for an app
 
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_PATH/config/functions"
 
 CONTAINERID="$1"; APP="$2"; PORT="$3" ; HOSTNAME="${4:-localhost}"
 
--- a/docs/development/pluginhooks.md
+++ b/docs/development/pluginhooks.md
@@ -339,7 +339,7 @@ source "$PLUGIN_PATH/config/functions"
 
 CONTAINERID="$1"; APP="$2"; PORT="$3" ; HOSTNAME="${4:-localhost}"
 
-eval "$(dokku config $APP --export)"
+eval "$(config_export app $APP)"
 DOKKU_DISABLE_DEPLOY="${DOKKU_DISABLE_DEPLOY:-false}"
 
 if [[ "$DOKKU_DISABLE_DEPLOY" = "true" ]]; then
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -67,7 +67,7 @@ case "$1" in
         pluginhook pre-release-buildstep "$APP"
         pluginhook pre-release-buildpack "$APP"
         if [[ -n $(config_export global) ]]; then
-          TMPFILE=$(mktemp /tmp/dokku.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
+          TMPFILE=$(mktemp $DOKKU_ROOT/$APP/.dokku_config.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
           config_export global > $TMPFILE
           id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/00-global-env.sh" < "$TMPFILE")
           test "$(docker wait $id)" -eq 0
@@ -75,7 +75,7 @@ case "$1" in
           rm -rf $TMPFILE
         fi
         if [[ -n $(config_export app $APP) ]]; then
-          TMPFILE=$(mktemp /tmp/dokku.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
+          TMPFILE=$(mktemp $DOKKU_ROOT/$APP/.dokku_config.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
           config_export app $APP > $TMPFILE
           id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/01-app-env.sh" < "$TMPFILE")
           test "$(docker wait $id)" -eq 0
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -7,7 +7,7 @@ case "$1" in
     APP="$2"; IMAGE="dokku/$APP"; IMAGE_SOURCE_TYPE="$3"; TMP_WORK_DIR="$4"
     CACHE_DIR="$DOKKU_ROOT/$APP/cache"
 
-    [[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source "$DOKKU_ROOT/$APP/ENV"
+    eval $(dokku config $APP --export)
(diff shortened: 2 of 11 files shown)


plugins/build-env/pre-build
```diff
@@ -6,8 +6,8 @@ APP="$1"; IMAGE="dokku/$APP"; BUILD_ENV=""
 
 [[ -f "$DOKKU_ROOT/BUILD_ENV" ]] && cat "$DOKKU_ROOT/BUILD_ENV" >> "$DOKKU_ROOT/ENV" && rm "$DOKKU_ROOT/BUILD_ENV"
 
-! (grep -q CURL_CONNECT_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && echo "export CURL_CONNECT_TIMEOUT=5" >> "$DOKKU_ROOT/ENV"
-! (grep -q CURL_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && echo "export CURL_TIMEOUT=30" >> "$DOKKU_ROOT/ENV"
+! (grep -q CURL_CONNECT_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && dokku config:set --global $APP CURL_CONNECT_TIMEOUT=5
+! (grep -q CURL_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && dokku config:set --global $APP CURL_TIMEOUT=30
```
josegonzalez: Could the test fails be because of this^?

josegonzalez: Yep, thats it alright :(

josegonzalez:
@michaelshobbs what is the sed stuff [here](https://github.com/progrium/dokku/blob/master/plugins/config/docker-args-deploy) supposed to be doing?

josegonzalez:
Figured out the issue and now everything is consolidated!

michaelshobbs:
That sed stuff reformats shell `export` statements to the docker env file format.

from https://docs.docker.com/reference/commandline/run/
```
The --env-file flag takes a filename as an argument and expects each line to be in the VAR=VAL format, mimicking the argument passed to --env. Comment lines need only be prefixed with #
```

josegonzalez:
Yeah, I figured it out and everything should be fine now :)

michaelshobbs:
I get why from a code perspective we would want to do this. However, I'm uncomfortable with how often we are calling dokku from within dokku. The more we do this the longer traces get and the hard it is to troubleshoot issues. This is obviously one of the shortcomings with bash. Just my opinion though. 

What if we had an `export_config()`function and `--export` called that? Then we could use that function internally as well?

josegonzalez:
The point of this is so that we can replace the config plugin with something else. If we just shove it into the common functions, that doesn't really help the matter.

michaelshobbs:
Why not put the function in the config plugin and that was other plugins can implement it?

josegonzalez:
As in put it in a `functions` file and source that in other plugins when needed?

michaelshobbs:
Yeah. I don't like the cross plugin sourcing bit, but again, we're doing this in bash....so....compromises?

josegonzalez:
I think this is good to merge, @michaelshobbs ?

docs/development/pluginhooks.md
```diff
@@ -335,10 +335,11 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 # `DOKKU_DISABLE_DEPLOY` env var is set to `true` for an app
 
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_PATH/config/functions"
 
 CONTAINERID="$1"; APP="$2"; PORT="$3" ; HOSTNAME="${4:-localhost}"
 
-[[ -f "$DOKKU_ROOT/$APP/ENV" ]] && source $DOKKU_ROOT/$APP/ENV
+eval "$(dokku config $APP --export)"
```
michaelshobbs: update with `config_export`?

plugins/00_dokku-standard/commands
```diff
@@ -65,15 +66,21 @@ case "$1" in
...
         pluginhook pre-release-buildpack "$APP"
-        if [[ -f "$DOKKU_ROOT/ENV" ]]; then
-          id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/00-global-env.sh" < "$DOKKU_ROOT/ENV")
+        if [[ -n $(config_export global) ]]; then
+          TMPFILE=$(mktemp /tmp/dokku.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
+          config_export global > $TMPFILE
+          id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/00-global-env.sh" < "$TMPFILE")
           test "$(docker wait $id)" -eq 0
           docker commit $id $IMAGE > /dev/null
+          rm -rf $TMPFILE
```
michaelshobbs: we probably should trap and rm. also i wouldn't write config in `/tmp` as it most likely contains passwords and such.

plugins/build-env/pre-build
```diff
@@ -1,22 +1,26 @@
...
 
-[[ -f "$DOKKU_ROOT/BUILD_ENV" ]] && cat "$DOKKU_ROOT/BUILD_ENV" >> "$DOKKU_ROOT/ENV" && rm "$DOKKU_ROOT/BUILD_ENV"
+[[ -f "$DOKKU_ROOT/BUILD_ENV" ]] && cat "$DOKKU_ROOT/BUILD_ENV" >> "$DOKKU_ROOT/ENV" && {
+  dokku_log_info2 "Using a global BUILD_ENV file is deprecated as of 0.3.26"
+  rm "$DOKKU_ROOT/BUILD_ENV"
+}
 
-! (grep -q CURL_CONNECT_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && echo "export CURL_CONNECT_TIMEOUT=5" >> "$DOKKU_ROOT/ENV"
-! (grep -q CURL_TIMEOUT "$DOKKU_ROOT/ENV" > /dev/null 2>&1) && echo "export CURL_TIMEOUT=30" >> "$DOKKU_ROOT/ENV"
+dokku config:get --global CURL_CONNECT_TIMEOUT > /dev/null 2>&1 || dokku config:set --global CURL_CONNECT_TIMEOUT=5
```
michaelshobbs: Can we also allow access to this via a function instead of a full dokku execution?

plugins/00_dokku-standard/commands
josegonzalez: Where should we write it to?

michaelshobbs: The app dir probably makes sense. If you want it looking clean, you could do `$DOKKU_ROOT/$APP/.dokku_config`.

josegonzalez: Fixed the paths. Not sure about the trap stuff - never used it - but I'll take a look!

michaelshobbs: We do similar things in the git plugin and main line even.

https://github.com/progrium/dokku/blob/4e0fed18a6f93eef4801f6c67312e7eae6bf35ac/plugins/git/commands#L11
https://github.com/progrium/dokku/blob/ae2e0bc57c93bbdf07cf676097af5c49957cccdf/dokku#L139

josegonzalez: Awesome love it. So I'll just add that right after I create the temp file :)

plugins/00_dokku-standard/commands
```diff
@@ -65,13 +66,19 @@ case "$1" in
         # *DEPRECATED* in v0.3.22: `pluginhook pre-release-buildstep` will be removed in future releases
         pluginhook pre-release-buildstep "$APP"
         pluginhook pre-release-buildpack "$APP"
-        if [[ -f "$DOKKU_ROOT/ENV" ]]; then
-          id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/00-global-env.sh" < "$DOKKU_ROOT/ENV")
+        if [[ -n $(config_export global) ]]; then
+          TMPFILE=$(mktemp $DOKKU_ROOT/$APP/.dokku_config.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
+          trap 'rm -rf "$TMPFILE" > /dev/null' RETURN
```
michaelshobbs: I think you want this to be 

`trap 'rm -rf "$TMPFILE" > /dev/null' INT TERM EXIT`

plugins/00_dokku-standard/commands
```diff
@@ -65,13 +66,19 @@ case "$1" in
...
+          config_export global > $TMPFILE
+          id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/00-global-env.sh" < "$TMPFILE")
           test "$(docker wait $id)" -eq 0
           docker commit $id $IMAGE > /dev/null
         fi
-        if [[ -f "$DOKKU_ROOT/$APP/ENV" ]]; then
-          id=$(docker run -i -a stdin $IMAGE /bin/bash -c "mkdir -p /app/.profile.d && cat > /app/.profile.d/01-app-env.sh" < "$DOKKU_ROOT/$APP/ENV")
+        if [[ -n $(config_export app $APP) ]]; then
+          TMPFILE=$(mktemp $DOKKU_ROOT/$APP/.dokku_config.XXXXXX) || dokku_log_fail "Cannot create temp file using mktemp in /tmp dir"
```
michaelshobbs: I was also thinking we might want this to be a directory.

i.e. `$DOKKU_ROOT/$APP/.dokku_config/`

plugins/00_dokku-standard/commands
michaelshobbs: actually if this is done in a function you can use return and not runover the other traps. probably better to do that.

plugins/00_dokku-standard/commands
```diff
@@ -65,13 +66,19 @@ case "$1" in
...
+          trap 'rm -rf "$TMPFILE" > /dev/null' RETURN
```
michaelshobbs: trapping twice won't work

PR #1407, opened 2015-08-23
Update variable name in docs

ertrzyiks:
My dockerfile didn't contain EXPOSE line and i tried to add it by variable,  without success. After adding EXPOSE line dokku create a variable with different name, so I believe this is what you wanted to describe in docs.


josegonzalez:
@michaelshobbs I looked through the code and this PR doesn't seem right. Can you verify/merge if necessary?

Issue #854, opened 2014-12-30
Proposal for a common attached-services concept


miraculixx:
I'm not sure proposals like this are warranted, so please let me know if not. This is born out of the observation (and some frustration ;-) that various dokku plugins take different (and somewhat conflicting) approaches in both naming convention, command syntax and functionality. 

### Goal

Facilitate an ecosystem of dokku plugins that provide [attached services](http://12factor.net/backing-services) that follow a common pattern in building such plugins, their use cases and a common command syntax. This is intended as a minimum level.

### Rationale
Currently there seem to be different ways for attached services. Some prefer to attach a service to a single app by default, while the dokku-link plugin allows one service to be linked to multiple apps. Also there are various ways for service plugins to name their environment variables.

This is to propose a standard for attached services, as follows. Ammendments, feedback and proposals welcome.

### Definition

1. A service is a dokku-plugin that, upon plugin installation, builds a base image to run the service in its own container.
2. A service can be linked to any number of dokku apps, but it can also be created without any app specified (deferred linking).
3. When linked, the service is to expose its connection URL to app containers as a single environment variable that follows this dokku-args standard: `-e <SERVICE>_URL=<service>://<user>:<password>@<ip>[/<path>]`, where service is the service name (e.g. MYSQL, RABBITMQ etc.), user/password the access credentials or tokens, ip the linked container's address, and the optional path any further information required.
4. Services can expose further name-spaced env variables via dokku-args or dokku's `config:set` command, e.g. `-e <SERVICE>_SOMEVAR=some-value`. Such variables are to be documented in the plugin's README for easy reference
5. Upon startup of the service container, the plugin creates a credentials file in $DOKKU_ROOT/.$SERVICE/credentials
6. Upon link of a service to an app, the plugin uses dokku-link to create the link
7. Upon removal of a service link from an app, the plugin uses dokku-link to remove the link
8. A service maintains its persistent data (if any) in $DOKKU_ROOT/.$SERVICE/volume, which it maps in its container as /var/$SERVICE. Such volumes are to be recorded in $DOKKU_ROOT/.$SERVICE/volumes
9. A service offers import and export commands to allow users to import/export service state (e.g. backup of persistent data or configuration).
10. A service can be cloned. Cloning means to create a new service that exactly matches the specification of the old service, and copies all data (i.e. does <service>:export <old>, followed by <service>:import <new>>`.

### Common command syntax

[...]

PR #1408, opened 2015-08-23
Add links to official dokku datastorage plugins

josegonzalez:
Something that has been a long time coming, but here are official dokku datastores.

Note that all have been marked as beta (because they are!) and their implementations may change wildly. All plugins appear to work fine under docker 1.7. I plan on formalizing a release process for each, as well as guiding their development. 

All plugins use the official docker images, and I do not intend to change this. Their interface should stay the same, and in fact the diffs between each plugin should be quite minimal.

One thing I plan on noting in the documentation is that these datastores do not provide any sort of backup solution, and it is up to the developer to implement such a thing for their system (if their data is important enough).

All plugins roughly meet the spec proposed by @miraculixx in #854, with a few changes:

- Added a `service:unexpose` command, to remove any exposed ports.
- Added a `service:alias` command, which allows users to change the alias in use by a service.
- Changed `service:delete` to `service:destroy`, to match how the core works/
- Services store their data in `/var/lib/dokku/services/DATASTORE/SERVICE_NAME` instead of the home directory.
- Usage of the `docker-args` pluginhook instead of the `dokku:link` plugin, to reduce dependencies.

### Available datastore plugins

- [Elasticsearch](https://github.com/dokku/dokku-elasticsearch-plugin) [missing clone, connect, export, import]
- [MariaDB](https://github.com/dokku/dokku-mariadb-plugin)
- [Memcached](https://github.com/dokku/dokku-memcached-plugin) [missing clone, export, import]
- [Mongo](https://github.com/dokku/dokku-mongo-plugin)
- [MySQL](https://github.com/dokku/dokku-mysql-plugin)
- [Postgres](https://github.com/dokku/dokku-postgres-plugin)
- [RabbitMQ](https://github.com/dokku/dokku-rabbitmq-plugin) [missing clone, connect, export, import]
- [Redis](https://github.com/dokku/dokku-redis-plugin)
- [RethinkDB](https://github.com/dokku/dokku-rethinkdb-plugin) [missing clone, export, import]

### TODO:

- [x] Add [mongo](https://github.com/docker-library/docs/blob/master/mongo/README.md) plugin.
- [x] Add [rabbitmq](https://github.com/docker-library/docs/blob/master/rabbitmq/README.md) plugin.
- [x] Add [rethinkdb](https://github.com/docker-library/docs/blob/master/rethinkdb/README.md) plugin.
- [x] Finish implementation of `expose`/`unexpose`.
- [x] Ensure services come up on server reboot.
- [x] Document the services command structure and inner workings

Affexted existing datastore maintainers /cc @Kloadut @ohardy @hughfletcher @jlachowski @jeffutter @Flink @luxifer @sekjun9878 @robv @blag

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -29,9 +29,12 @@ The following plugins are available and provided by dokku maintainers. Where not
 | [Elasticsearch (beta)](https://github.com/dokku/dokku-elasticsearch-plugin)                       | [dokku][]             | 0.3.25+               |
 | [MariaDB (beta)](https://github.com/dokku/dokku-mariadb-plugin)                                   | [dokku][]             | 0.3.25+               |
 | [Memcached (beta)](https://github.com/dokku/dokku-memcached-plugin)                               | [dokku][]             | 0.3.25+               |
+| [Mongo (beta)](https://github.com/dokku/dokku-mongo-plugin)                                       | [dokku][]             | 0.3.25+               |
 | [MySQL (beta)](https://github.com/dokku/dokku-mysql-plugin)                                       | [dokku][]             | 0.3.25+               |
 | [Postgres (beta)](https://github.com/dokku/dokku-redis-plugin)                                    | [dokku][]             | 0.3.25+               |
+| [RabbitMQ (beta)](https://github.com/dokku/dokku-rabbitmq-plugin)                                 | [dokku][]             | 0.3.25+               |
 | [Redis (beta)](https://github.com/dokku/dokku-redis-plugin)                                       | [dokku][]             | 0.3.25+               |
+| [RethinkDB (beta)](https://github.com/dokku/dokku-rethinkdb-plugin)                               | [dokku][]             | 0.3.25+               |
 
 ## Community plugins
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -31,7 +31,7 @@ The following plugins are available and provided by dokku maintainers. Where not
 | [Memcached (beta)](https://github.com/dokku/dokku-memcached-plugin)                               | [dokku][]             | 0.3.25+               |
 | [Mongo (beta)](https://github.com/dokku/dokku-mongo-plugin)                                       | [dokku][]             | 0.3.25+               |
 | [MySQL (beta)](https://github.com/dokku/dokku-mysql-plugin)                                       | [dokku][]             | 0.3.25+               |
-| [Postgres (beta)](https://github.com/dokku/dokku-redis-plugin)                                    | [dokku][]             | 0.3.25+               |
+| [Postgres (beta)](https://github.com/dokku/dokku-postgres-plugin)                                 | [dokku][]             | 0.3.25+               |
 | [RabbitMQ (beta)](https://github.com/dokku/dokku-rabbitmq-plugin)                                 | [dokku][]             | 0.3.25+               |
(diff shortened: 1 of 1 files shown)


Flink:
@josegonzalez Great work it seems! :) There is a typo in your comment, the Postgres link points to the redis plugin.

So about community plugins, what does this mean exactly? I suppose they will be deprecated when those plugins won’t be in beta anymore?
Do you need help? Do we need (as current community plugins maintainers) to merge our work with yours?

josegonzalez:
Fixed the typo.

I think that we should probably combine forces. If a community plugin maintainer wants to step up and help out with a particular plugin, then I'm all ears. If a developer wants to maintain their own, thats fine too.

The beta thing is there so that people don't instantly flock to them when they have working installations with the existing plugins.

josegonzalez:
If any existing maintainer would like to join the team for a plugin they currently maintain, let me know and I will add you. I'd like to reduce the amount of work the you all are doing and provide a great experience for our community :)

Flink:
@josegonzalez thanks, I just joined! :grin:
So yeah I think also that we should join forces ;) I’ll still maintain my own plugin for now but I could start to help on the postgres plugin.

How do we proceed?

josegonzalez:
If you'd like to prototype any of the two remaining tasks on one of the plugins, that would be great. I can do the service writeup. Once that's all done, we merge this branch and are off to the races :)

Flink:
Ok great. I’ll have to read/understand your code first so I won’t break anything :grin:
I’ll try to have something this week.

michaelshobbs:
@josegonzalez love it man. I was just having the conversation about "official" datastore plugins on Friday with one of the Ops guys here.

jeffutter:
Great work! I am really excited to have an official, supported plugin. I maintain one of the mongo and postgresql ones and I don't use them much anymore so it will be really great to have officially supported ones that have more eyes on them.

I would be glad to help maintain the official mongo and postgres ones. I can review code and partake in discussions. However, I don't think I will be using them enough to contribute much code.

jeffutter:
Thanks! Also, from skimming the docs and the code. It looks like the official plugins support both ```n apps -> 1 db``` or ```1 app -> 1 db``` or ```n apps -> n dbs```. Correct? If so :+1:

josegonzalez:
That is correct. Usage of the plugins is totally up to the user installing them. I hope that we stick to this for any further plugins, and that the only thing we do is allow users to easily customize their services :)

josegonzalez:
Labeling this as `bc-break` as it will superscede all existing plugins and more or less force our users to use the latest docker, which might break other plugins.

sekjun9878:
@josegonzalez Nice :D I'm guessing you would like us to put a redirect to the official dokku plugins for "combining forces"? I would be happy to contribute and provide feedback to the official repos, but I can't say I use it much anymore.

hhff:
This is so great

josegonzalez:
@sekjun9878 potentially. Once they get merged in, I would, though that's up to you :)

I'll add you to the repos when I'm in my office.

josegonzalez:
I think the plugins are currently in a very good place. 

- All support both 0.3.x and 0.4.x dokku
- Well-tested
- Using latest versions of all datastores
- With few exceptions, implement every action that makes sense.

@Flink up to you when we want to tag a 1.0. We should talk about what that workflow looks like (and document it) but otherwise I'm quite comfortable with these plugins.

PR #1414, opened 2015-08-24
certs plugin

### LINKED ISSUE #1368: Don’t automatically configure certificate’s CN and SANs as domains
Currently, [the certificate’s Common Name and its Subject Alternative Names are automatically added to `VHOST`](https://github.com/progrium/dokku/blob/v0.3.22/plugins/nginx-vhosts/commands#L79-L89). However, your certificate may contain domains which you do not want to use for your particular dokku app.

I would therefore propose to either require configuring additional domains manually ([as documented by Heroku](https://devcenter.heroku.com/articles/ssl-endpoint#dns-and-domain-configuration)) or to only write the CN and SANs to `VHOST` if the file doesn’t already exist.

michaelshobbs:
TODOS:
- [x] remove 'magically' added VHOST entries based on SSL cert information
- [x] implement `certs` plugin 'loosely' based on https://devcenter.heroku.com/articles/ssl-endpoint 
  - only implements `certs:remove`, `certs:info` and `certs:generate` (if someone wants to take a stab at the others, please feel free)
- [x] document `certs` plugin
- [x] document private/shared functions convention for plugins

Note:
- further implements cross-plugin function sharing. as a convention, we've seem to have settled on shareable functions residing in `<plugin_path>/functions` and non-shareable functions residing in `<plugin_path>/commands` (example [here](https://github.com/progrium/dokku/blob/5676dbc4c6bea9837e36aded1897221bcf4cdee6/plugins/nginx-vhosts/commands#L5-L26))

closes #1368 

example `certs:info` output
```
root@dokku:~/dokku# dokku certs:info node-js-app
-----> Fetching SSL Endpoint info for node-js-app...
-----> Certificate details:
=====> Common Name(s):
=====>    test.dokku.me
=====>    www.test.dokku.me
=====>    www.test.app.dokku.me
=====> Expires At: Jan 14 17:57:12 2017 GMT
=====> Issuer: C=US, ST=California, L=San Francisco, OU=Dokku Me, LLC, CN=test.dokku.me
=====> Starts At: Jan 15 17:57:12 2015 GMT
=====> Subject: C=US; ST=California; L=San Francisco; OU=Dokku Me, LLC; CN=test.dokku.me
=====> SSL certificate is self signed.
```


plugins/certs/commands
```diff
@@ -0,0 +1,22 @@
...
+case "$1" in
+  help | certs:help)
+    cat && cat<<EOF
+    certs:add <app> CRT KEY, Add an ssl endpoint to an app.
+    certs:generate <app> DOMAIN, Generate a key and certificate signing request (or self-signed certificate)
+    certs:info <app>, Show certificate information for an ssl endpoint.
+    certs:key <app> CRT KEY [KEY ...], Print the correct key for the given certificate.
+    certs:remove <app>, Remove an SSL Endpoint from an app.
+    certs:update <app> CRT KEY, Update an SSL Endpoint on an app.
+EOF
```
michaelshobbs: working on implementation now

tests/unit/nginx-vhosts.bats
```diff
@@ -109,21 +109,11 @@ assert_error_log() {
...
-  assert_ssl_domain "node-js-app.dokku.me"
-}
-
-@test "(nginx-vhosts) nginx:build-config (with SSL CN mismatch & custom nginx template)" {
-  setup_test_tls
-  custom_ssl_nginx_template
-  deploy_app
-  assert_ssl_domain "node-js-app.dokku.me"
-}
-
```
michaelshobbs: removed magic tests

plugins/certs/commands
josegonzalez: Please update the docs as well to illustrate how this will work.

tests/unit/nginx-vhosts.bats
josegonzalez: So I'm guessing this is a BC break as well? Fine by me, just checking.

plugins/certs/commands
michaelshobbs: For sure. I meant to add a task to redo the entirety of the docs for this; including the "public functions convention.

tests/unit/nginx-vhosts.bats
michaelshobbs: Only in that we won't automagically add domains to your app

josegonzalez: Removing functionality is still a BC break. We're pretty close to a 0.4.0, so I think we can wait till then, yes? May as well rip off all the Band-Aids then :)


(Our next release can probably be 0.4.0 unless herokuish has any bugs.)

michaelshobbs:
Albeit ill-conceived functionality 😜

tests/unit/nginx-vhosts.bats
michaelshobbs: Albeit ill-conceived functionality 😜

josegonzalez:
Output of `certs:info`:

```
jose@mactaku:~/Apps/seatgeek/sghomerunderby git:master $ heroku certs:info
Fetching SSL Endpoint toyama-7189.herokussl.com info for sghomerunderby... done
Certificate details:
Common Name(s): *.domain.com
                domain.com

Expires At:     2019-03-15 05:45 UTC
Issuer:         ISSUER_DATA
Starts At:      2014-12-21 21:24 UTC
Subject:        SUBJECT_DATA
SSL certificate is verified by a root authority.
```

michaelshobbs:
@josegonzalez this is ready to be banged on. review away

josegonzalez:
GTFO tests fail.

Let me see what I can do to test it. I have a wildcard cert somewhere...

docs/application-deployment.md
```diff
@@ -54,7 +54,7 @@ If buildpack detection isn't working well for you or you want to specify a custo
 
 Deployment of `Dockerfile` repos is supported as of v0.3.15. Simply place a Dockerfile in the root of your repo and push to dokku. If you are converting from a heroku/dokku buildpack deployment, ensure you are not setting `$BUILDPACK_URL` or including a `.buildpacks` file in the root of your repo.
 
-By default, we will extract the first `EXPOSE` port and tell nginx to proxy your app to that port. Alternatively, you can set `$DOCKERFILE_PORT` in your app's dokku configuration.
+By default, we will extract the first `EXPOSE` port and tell nginx to proxy your app to that port. Alternatively, you can set `$DOKKU_DOCKERFILE_PORT` in your app's dokku configuration.
```
josegonzalez: @michaelshobbs I think you ran `git cherry-pick` in the wrong branch.

docs/ssl-certificate-management.md
```diff
@@ -0,0 +1,34 @@
...
+=====> Expires At: Aug 24 23:32:59 2016 GMT
+=====> Issuer: C=US, ST=California, L=San Francisco, O=dokku.me, CN=test.dokku.me
+=====> Starts At: Aug 25 23:32:59 2015 GMT
+=====> Subject: C=US; ST=California; L=San Francisco; O=dokku.me; CN=test.dokku.me
+=====> SSL certificate is self signed.
+```
+
+### `certs:remove` command
+
+The `certs:remove` command only works on app-specific certificates. It will `rm` the app-specific tls directory, rebuild the nginx configuration, and reload nginx.
```
josegonzalez: Why only app-specific certs?

josegonzalez:
So is the expectation that people only be able to use self-signed certs with this plugin?

docs/application-deployment.md
michaelshobbs: i rebased master

docs/ssl-certificate-management.md
michaelshobbs: what if i have multiple apps on a dokku server? i didn't want to pull the ssl cert in case another app depended on it. thoughts on best implementation to manage server-wide ssl?

michaelshobbs:
No. See docs for the generate command. We place the SS cert in place and then you can replace it when you get your real cert back if you decide to go that route.

docs/ssl-certificate-management.md
josegonzalez: We could add a `--global` arg. If people *really* want global stuff, they have to deal with it, warts and all. I think that should work?

josegonzalez:
How does someone replace the cert?

michaelshobbs:
RTFM man!! :smile: 

from the new docs
```
If you decide to obtain a CA signed certficate, simply place it in `$DOKKU_ROOT/$APP/tls/server.crt` and run `dokku nginx:build-config $APP`.
```

josegonzalez:
I mean shouldn't we remove the `nginx` version of adding certs then? Seems like that's unnecessary duplication.

michaelshobbs:
so that's a way to import a cert. which doesn't follow the heroku certs ui. hence i didn't use. that's a benefit of using something like ruby (?) for the client. not sure how to pipe in a key and cert file via an ssh connection

josegonzalez:
Lets assume that we don't have to support ssh. Would it then be fine to include that functionality in here and rip it out of the old place?

michaelshobbs:
sure. the current implementation is a bit hokey. would you want to implement the same heroku ui? 

i.e.:
```
certs:add CRT KEY            #  Add an ssl endpoint to an app.
```

docs/ssl-certificate-management.md
michaelshobbs: Besides my loathing of arg parsing in bash, the whole server-wide SSL cert thing doesn't just with the heroku model. I contend that is a special case and should be handled manually.

josegonzalez:
Yeah, I think so?

tests/unit/nginx-vhosts.bats
olalonde: It's a BC that wouldn't affect existing apps though, right? IIRC the domains were added to the domains file so unless you create a new app and add SSL to it, you won't notice that this behaviour has changed.

josegonzalez: It's a bc-break in how we allow users to interface with dokku.

michaelshobbs: Correct on both accounts. :smile:

josegonzalez:
Okay, I moved the existing `nginx:import-ssl` command to `certs:add`. How do you set multiple ssl certs? I don't think thats possible at all, despite what the docs said. Can someone verify that to be the case? I don't see how, given that we hardcode the location of `server.crt` and `server.key`.

michaelshobbs:
Multiple certs were never supported

josegonzalez:
The following heroku commands are left unimplemented:

```
certs:chain CRT [CRT ...], Print the ordered and complete chain for the given certificate.
certs:key CRT KEY [KEY ...], Print the correct key for the given certificate.
certs:rollback, Rollback an SSL Endpoint for an app.
certs:update CRT KEY, Update an SSL Endpoint on an app.
```

The rollback one isn't worth attempting to implement, and I think `certs:update` can be used as an alias for `certs:add` in our case. What do you think of the other two?

josegonzalez:
I've implemented `crt/key` importing via files on disk vs tarball. I think if tests pass, we can safely merge this in, as I have no intention of implementing any of the others (and they can be implemented in the future if need be).

Thoughts (before you head off to the wilderness)?

michaelshobbs:
Works for me. `key` and `chain` didn't make sense to me as i wasn't sure how to implement with openssl. if someone else wants to take a stab, go for it.

plugins/certs/commands
```diff
@@ -0,0 +1,174 @@
...
+is_tar_import() {
+  [[ -t 0 ]] && return 1
+  return 0
+}
+
+is_file_import() {
+  local CRT_FILE="$3"
+  local KEY_FILE="$4"
+
+  if [[ -f $CRT_FILE ]] && [[ -f $KEY_FILE]]; then
```
michaelshobbs: you'll want a space here between `$KEY_FILE` and `]]` :smile:

josegonzalez:
Merging!

PR #1419, opened 2015-08-25
Fix dokku ps <app> over ssh

u2mejc:
Corrects error dialog seen by new users attempting to use the shell's help:

dokku> dokku help
 !    `dokku help` is not a dokku command.
  !    See `dokku help` for a list of available commands.

Resolves the TTY error when connecting over ssh. Updated ps to use has_tty function.

### DIFF
--- a/plugins/ps/commands
+++ b/plugins/ps/commands
@@ -12,7 +12,8 @@ case "$1" in
     ! (is_deployed $APP) && echo "App $APP has not been deployed" && exit 0
 
     for CID in $CONTAINER_IDS; do
-      docker exec -ti "$CID" /bin/sh -c "ps auxwww"
+      has_tty && DOKKU_RUN_OPTS="-i -t"
+      docker exec $DOKKU_RUN_OPTS $CID /bin/sh -c "ps auxwww"
     done
   ;;
 


josegonzalez:
I don't understand the help dialog change. Can you explain that a bit more?

u2mejc:
Happy to. If you read it literally, just like a brand new user may, it asks you (in quotes) to type 'dokku help'.

Here's a sample interaction before PR:
```bash
foo
 !    `foo` is not a dokku command.
 !    See `dokku help` for a list of available commands.
dokku help
 !    `dokku help` is not a dokku command.
 !    See `dokku help` for a list of available commands.
```
:frowning: :trollface: 
After:
```bash
foo
 !    `foo` is not a dokku command.
 !    See `dokku help` for a list of available commands.
help
Usage: dokku [--quiet|--trace|--rm-container|--rm|--force] COMMAND <app> [command-specific-options]

Options:
    apps:cre...
```
:smile:

josegonzalez:
Wait, are you inside of `dokku shell`?

u2mejc:
Correct, aka $ dokku shell (or ssh dokku@dokku.me shell).

u2mejc:
I edited my example above to clarify.

josegonzalez:
I think the expectation is that once you are in a dokku shell, you don't need to use the prefix anymore. Can you revert that change in your PR (and squash the commits)? 

You can open a second PR adding custom MOTD - telling people to not prefix dokku commands - for dokku shell, which will probably be much more useful than this change. There are other places where we call out the `dokku` command in help output, so ^ would cover it.

u2mejc:
np, just pushed up the change.

josegonzalez:
Wait, why did you get a tty error before? Note: I don't seem to have such an error on master. Can you paste the command that results in an error?

u2mejc:
Here you go:
```bash
$ ssh dokku@dokku.me ps node-js-app
time="2015-08-26T21:58:25Z" level=fatal msg="cannot enable tty mode on non tty input" 
```

Do you have ```RequestTTY yes``` in your ssh config? http://progrium.viewdocs.io/dokku/remote-commands/

If so, you may not see this bug. Because that line is buried in the documentation, this brings up a new question: Should we just bubble up this documentation since this *would be bug* is probably lurking in other sub-commands in Dokku or should we fix the TTY bugs (since there is an easy function for it already) when we find them?

josegonzalez:
Here is my `.ssh/config`:

```
Host *
  CheckHostIP yes
  ControlMaster auto
  ControlPath ~/.ssh/master-%r@%h:%p
  SendEnv LANG LC_*
  HashKnownHosts yes
  GSSAPIAuthentication no
  GSSAPIDelegateCredentials no
  RSAAuthentication yes
  PasswordAuthentication yes
  StrictHostKeyChecking no
```

u2mejc:
Hmm somewhere that's being set. Try disabling it with **-T**, my patch allows you to execute without the pseudo-tty:

```bash
$ ssh -T dokku@dokku.me ps node-js-app
````

**Comment by josegonzalez** [comment]
Maybe it's because I'm using `contrib/dokku_client.sh`? Can you try with that?

**Comment by michaelshobbs** [comment]
Given the variance in how different folks might have configured their SSH clients, I like the idea of using `has_tty` where it makes sense.

**Comment by michaelshobbs** [comment]
Also `contrib/dokku_client.sh` forces `-t`

https://github.com/progrium/dokku/blob/ebfcece6200a19dcef5dda79f030f704813d289e/contrib/dokku_client.sh#L97
https://github.com/progrium/dokku/blob/ebfcece6200a19dcef5dda79f030f704813d289e/contrib/dokku_client.sh#L124

**Comment by josegonzalez** [comment]
@michaelshobbs I haven't read up about the flag just yet, but if you feel comfortable merging, ship it :)

**Comment by michaelshobbs** [comment]
Full disclosure, @u2mejc sits right next to me at work. @u2mejc did you test this change with the dokku_client? If not, would you please? It's probably fine but just want to make sure we don't have some weirdness there.

**Comment by u2mejc** [comment]
@michaelshobbs is correct. This patch removes the requirement.

```
NAME
     ssh -- OpenSSH SSH client (remote login program)
...
     -T      Disable pseudo-tty allocation.

     -t      Force pseudo-tty allocation.  This can be used to execute arbitrary screen-based programs on a remote machine, which can be very useful, e.g. when implementing menu services.  Multiple -t options force tty allocation, even if ssh has no local tty.
```

Yep, I tested dokku_client.sh and it does work with or without the patch.

josegonzalez:
@u2mejc good work! Mind making a PR to add MOTD support to `dokku shell`?

Issue #450, opened 2014-02-10
Dokku and cron jobs


vladimirnani:
Is there a possibility to manage a cron jobs using dokku?

Zren:
# NOTICE

If you use `dokku run <app> <command>` inside cron, make sure to use the `--rm-container` (or `--rm`) flag to automatically remove the container afterwards (like so `dokku --rm run <app> <command>`. Otherwise you'll find yourself running out of disk space.

henrik:
Thank you, @Zren. Other things that are good to know (I just set this up):

**You [may need to specify the full path](http://askubuntu.com/a/47801/443378) to dokku.** I did `which dokku` to find out mine is at `/usr/local/bin/dokku`. Thus, my crontab entry has `/usr/local/bin/dokku --rm run my-app my-command`.

**If it still doesn't work, [make sure cron mails you](http://askubuntu.com/a/646831/443378) any STDOUT or STDERR output.** It doesn't need to go to a "real" email address. I did `sudo apt-get install postfix`, chose "Local" during setup, then `sudo tail -f /var/mail/root` to read it (replace "root" with another username if applicable).

josegonzalez:
@henrik would be great if you could contribute some docs on getting cronjobs running. I'm sure other users would benefit from doing so :)

henrik:
@josegonzalez That's a good idea! I'll consider it. Would be nice to contribute back :)

felixyz:
@henrik Alternatively (?), one can put this at the top of ones crontab file:

`````
SHELL=/bin/bash
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
`````

felixyz:
Mostly for the sake of documentation for others looking to do this, here is some notes about my setup:
* Like @henrik, I configured cron to email me with output from jobs (in my case, I want actual emails sent so I get warnings if anything fails)
* However, since I only want to hear from cron if something goes wrong, I installed [cronic](http://habilis.net/cronic/) and run my (Ruby rake) jobs like this:
`*/10 * * * * cronic dokku --rm run my-app bundle exec rake schedule_stuff`
I now get emails if the job fails, but not otherwise.
* Another point discussed above is whether you need to invoke dokku with a full path. In my experience, setting up the cron jobs as the dokku user, I needed a fully qualified path, but as root (and extending the path as described in my previous comment) I didn't. Not sure what is recommended?

[...]

PR #1420, opened 2015-08-26
Add `dokku enter` for connecting to an app container

### LINKED ISSUE #1332: RFC - Implement a dokku enter command
I don't know if it was previously discussed but what about and `dokku enter <name> [bash]` command that allow the admin to easly enter a running container using `bash`.

I often have the need to enter a deployed container just to debug some daemon port bind and running docker commands is pretty slow.

This command should be an alias of something like
```
sudo docker ps
sudo docker exec -it <id> bash
```

josegonzalez:
Add `dokku enter` for connecting to an app container

This command can be used to enter a running container. The following variations of the command exist:

```shell
dokku enter api web
dokku enter api web.1
dokku enter api --container-id ID
```

By default, it runs a `/bin/bash`, but can also be used to run a custom command:

```shell
# just echo hi
dokku enter api web echo hi

# run a long-running command, as one might for a cron task
dokku enter api web python script/background-worker.py
```

Closes #1332


josegonzalez:
Note: going to merge this in for 0.4.0 to encourage users to upgrade.

josegonzalez:
Refs #450 :)

henrik:
Nice! :thumbsup:

I used [this](https://github.com/dokku-alt/dokku-alt/issues/66#issuecomment-62886633) recently which could be slightly simplified if there was a "dokku enter" that didn't require an app name, but it's very slightly indeed, so probably not worth adding:

```
sudo docker exec CONTAINER_ID su postgres -c "psql db_name -c 'CREATE EXTENSION hstore;'"

# could become

dokku enter --container-id CONTAINER_ID su postgres -c "psql db_name -c 'CREATE EXTENSION hstore;'"
```

Not sure if sudo is required in the one and not the other.

josegonzalez:
I think ensuring we scope the docker call to an app is a good idea, otherwise there is potential for a user to run a command against a container that isn't part of their app.

Of note is that you don't need to specify a container id, or even a running app. You could very well just do:

 ````
# Your procfile can have the following entry:
#
#     background: while true; do foo; sleep 2; done
#
# and `background` scaled to 1
#
# Assumes you only run one background task for an instance
# Will do weird stuff if you have multiple background tasks running

dokku ps:scale api background=1 && dokku enter api background some-command && dokku ps:scale api background=0

# or use a one-off container
dokku --rm run api some-command
```

Assuming you want to have a cron file for your server, here is a potential layout:

```
# API backup cron jobs
MAILTO="mail@example.com"
PATH=/usr/local/bin:/usr/bin:/bin

# m   h   dom mon dow   command
# *   *   *   *   *     command to be executed
# -   -   -   -   -
# |   |   |   |   |
# |   |   |   |   +----- day of week (0 - 6) (Sunday=0)
# |   |   |   +------- month (1 - 12)
# |   |   +--------- day of month (1 - 31)
# |   +----------- hour (0 - 23)
# +----------- min (0 - 59)

### HIGH TRAFFIC TIME IS B/W 00:00 - 04:00 AND 14:00 - 23:59
### RUN YOUR TASKS FROM 04:00 - 14:00
### KEEP SORTED IN TIME ORDER

### PLACE ALL CRON TASKS BELOW

0 0 * * * deploy dokku --rm run api some-command

0 1 * * * dokku ps:scale api background=1 && dokku enter api background some-other-command && dokku ps:scale api background=0

### PLACE ALL CRON TASKS ABOVE, DO NOT REMOVE THE WHITESPACE AFTER THIS LINE

```

^ That should probably be in the documentation, which I'd take as a PR for this PR :)

felixyz:
@josegonzalez I completely fail to see the point of having a dummy background container just so I can `enter` it. Or maybe I missed the point of your example?

josegonzalez:
Its just a way to run a command in a new container that isn't already working. If you set mem/vpu limits on a container, this might be desirable.

PR #1421, opened 2015-08-26
First pass at tar plugin.

basicer:


plugins/tar/commands
```diff
@@ -0,0 +1,64 @@
...
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+case "$1" in
+  tar-from|tar:from)
+    verify_app_name "$2"
+    APP=$2
+    URL=$3
+    shift 3
+    curl -# --insecure -L "$URL" | dokku tar-in $APP $@
```
josegonzalez: Should just be `tar:in` and we can remove the aliases.

The other thing is that - to reduce the `DOKKU_TRACE` output - the `tar:in` command can be extracted into it's own function and just executed instead of shelling out to dokku.

plugins/tar/commands
```diff
@@ -0,0 +1,64 @@
...
+    ;;
+
+  tar-build)
+    verify_app_name "$2"
+    APP="$2"; APP_BUILD_LOCK="$DOKKU_ROOT/$APP/.build.lock"
+    APP_BUILD_LOCK_MSG="$APP is currently being deployed or locked. Waiting..."
+    [[ $(flock -n "$APP_BUILD_LOCK" true &>/dev/null ; echo $?) -ne 0 ]] && echo "$APP_BUILD_LOCK_MSG"
+
+    shift 1
+    flock -o "$APP_BUILD_LOCK" dokku tar-build-locked "$@"
```
josegonzalez: Same here re: functions.

plugins/tar/receive-app
```diff
@@ -0,0 +1,12 @@
...
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname $0)/../common/functions"
+
+APP="$1"; REV="$2"
+
+# Don't trigger tar build if there is no tarball.
+if [ ! -f "$DOKKU_ROOT/$APP/src.tar" ]; then
+  cat
+else
+  dokku tar-build $APP
```
josegonzalez: Ditto. Just create a `functions` file that you can source in, similar to how we allow sourcing of `common/functions`.

plugins/tar/commands
josegonzalez: The other thing is `$@` might need quotes.

plugins/tar/commands
```diff
@@ -0,0 +1,64 @@
...
+    pushd $TMP_WORK_DIR > /dev/null
+    tar -x -C "$TMP_WORK_DIR" -f "$DOKKU_ROOT/$APP/src.tar"  
+    #find -name .git -prune -exec rm -rf {} \; > /dev/null
+
+    if [[ -f Dockerfile ]] && [[ "$([[ -f .env ]] && grep -q BUILDPACK_URL .env; echo $?)" != "0" ]] && [[ ! -f ".buildpacks" ]]; then
+      dokku receive "$APP" "dockerfile" "$TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
+    else
+      dokku receive "$APP" "herokuish" "$TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
+    fi
+    ;;
```
basicer: This bit is pretty much copied from the git plugin, the mercurial plugin copies it too.  Should this be moved into common?

josegonzalez:
We should have at least a single test of app deployment via tar. @michaelshobbs Assuming the other tests pass with this, would you do the honors?

plugins/tar/commands
josegonzalez: Maybe once we have all three plugins, that can be the case? Up to you tbh.

josegonzalez:
Seems legit. Any last thoughts before I merge, @michaelshobbs ?

michaelshobbs:
:+1:

plugins/tar/commands
```diff
@@ -0,0 +1,72 @@
...
+    APP=$2
+    URL=$3
+    shift 3
+    curl -# --insecure -L "$URL" | dokku tar-in $APP "$@"
+    ;;
+
+  tar-in|tar:in)
+    APP="$2"
+    verify_app_name "$2"
+    tee "$DOKKU_ROOT/$APP/src.tar" | wc -c
```
michaelshobbs: What is this?

basicer: Basically just confirms the total size of the upload tar, it use to be a little progress bar thing, but I couldn't get `awk` to behave.  We could get rid of it and just `>` it into place.

michaelshobbs: Yeah currently that output has no context and doesn't conform to any other output format we have going on.

plugins/tar/commands
```diff
@@ -0,0 +1,72 @@
...
+
+    # extract tar file
+    chmod 755 $TMP_WORK_DIR
+    pushd $TMP_WORK_DIR > /dev/null
+
+    # Detect a common prefix that all files in the tar have, and strip off each directory found in it
+    COMMON_PREFIX=$(tar -tf "$DOKKU_ROOT/$APP/src.tar" | sed -e 'N;s/^\(.*\).*\n\1.*$/\1\n\1/;D')
+    BOGUS_PARTS=$(echo "$COMMON_PREFIX " | awk 'BEGIN{FS="/"} {print NF-1}')
+
+    dokku_log_info1_quiet "Striping $BOGUS_PARTS worth of directories from tarball"
```
michaelshobbs: What is this output trying to tell me and is it supposed to say "stripping"?

PR #1423, opened 2015-08-26
Set `DYNO_TYPE_NUMBER` environment variable for each container

### LINKED ISSUE #1350: Define worker command when using Dockerfile
As I understand from the [documentation](http://progrium.viewdocs.io/dokku/process-management/) the commands for the server process and the worker process shoud be defined in the `Procfile`.

But how do I define the worker command when I am using a `Dockerfile` instead of the buildpacks? The config `DOKKU_DOCKERFILE_START_CMD` is applied to all process types. web as worker.

josegonzalez:
This will allow users to inspect the underlying environment and conditionally run different commands based upon the `DYNO_TYPE_NUMBER`. The values will be set in the following format:

````
web.1
web.2
worker.1
worker.2
clock.1
````

Users of process managing plugins and dockerfiles can check the `DYNO_TYPE_NUMBER` and set a specific command to run.

Closes #1350

### DIFF
--- a/dokku
+++ b/dokku
@@ -92,6 +92,7 @@ case "$1" in
         # start the app
         DOCKER_ARGS=$(: | pluginhook docker-args $APP deploy)
         DOCKER_ARGS+=" -e DYNO=$PROC_TYPE "
+        DOCKER_ARGS+=" -e DYNO_TYPE_NUMBER=$PROC_TYPE.$CONTAINER_INDEX "
         DOCKER_ARGS+=$(: | pluginhook docker-args-deploy $APP)
         BIND_EXTERNAL=$(pluginhook bind-external-ip $APP)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by josegonzalez** [comment]
I have to test what the env vars look like in the container (do they have quotes or not) before merging

**Comment by josegonzalez** [comment]
Seems legit:

```
root@dokku:~/dokku# dokku enter ruby-sample web
u32414@dc396d0e2d47:~$ env
HOSTNAME=dc396d0e2d47
CURL_CONNECT_TIMEOUT=5
DYNO_TYPE_NUMBER=web.1
CACHE_PATH=/cache
HEROKUISH_VERSION=0.3.2
RACK_ENV=production
DYNO=web
PATH=/app/bin:/app/vendor/bundle/bin:/app/vendor/bundle/ruby/2.1.0/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
SELF=/exec
PWD=/app
LANG=en_US.UTF-8
SELF_EXECUTABLE=/bin/herokuish
HOME=/app
SHLVL=3
GEM_PATH=/app/vendor/bundle/ruby/2.1.0:
CURL_TIMEOUT=30
PORT=5000
_=/usr/bin/env
```

Issue #1205, opened 2015-05-22
Every apt command fails with "Could not get lock ... Temporarily unavailable"


mikker:
I'm not sure when it happened but something seems to have gone awry with my dokku setup and now every apt command fails eg:

```
$ apt-get update && apt-get upgrade -y
Ign http://ppa.launchpad.net trusty InRelease
Hit http://ppa.launchpad.net trusty Release.gpg
Hit http://ppa.launchpad.net trusty Release
Hit http://ppa.launchpad.net trusty/main amd64 Packages
Hit http://ppa.launchpad.net trusty/main i386 Packages
Hit http://ppa.launchpad.net trusty/main Translation-en
Ign http://security.ubuntu.com trusty-security InRelease
Ign http://mirrors.digitalocean.com trusty InRelease
Get:1 http://security.ubuntu.com trusty-security Release.gpg [933 B]
Ign http://mirrors.digitalocean.com trusty-updates InRelease
Get:2 http://security.ubuntu.com trusty-security Release [63.5 kB]
Hit http://mirrors.digitalocean.com trusty Release.gpg
Hit http://mirrors.digitalocean.com trusty-updates Release.gpg
Hit http://mirrors.digitalocean.com trusty Release
Hit http://mirrors.digitalocean.com trusty-updates Release
Get:3 http://security.ubuntu.com trusty-security/main Sources [81.4 kB]
Hit http://mirrors.digitalocean.com trusty/main Sources
Hit http://mirrors.digitalocean.com trusty/universe Sources
Hit http://mirrors.digitalocean.com trusty/main amd64 Packages
Get:4 http://security.ubuntu.com trusty-security/universe Sources [25.2 kB]
Hit http://mirrors.digitalocean.com trusty/universe amd64 Packages
Get:5 http://security.ubuntu.com trusty-security/main amd64 Packages [272 kB]
Hit http://mirrors.digitalocean.com trusty/main i386 Packages
Get:6 http://get.docker.io docker InRelease
Hit http://mirrors.digitalocean.com trusty/universe i386 Packages
Ign http://get.docker.io docker InRelease
Hit http://mirrors.digitalocean.com trusty/main Translation-en
Get:7 http://security.ubuntu.com trusty-security/universe amd64 Packages [104 kB]
Hit http://mirrors.digitalocean.com trusty/universe Translation-en
Get:8 http://security.ubuntu.com trusty-security/main i386 Packages [260 kB]
[...]

PR #1425, opened 2015-08-26
Plugins management

Flink:
As discussed in https://github.com/progrium/dokku/issues/1205, I tried a way to manage plugins and be able to do a `dokku plugins-install` only on core ones.

This PR is here to discuss the features and the details of the implementation.

It’s pretty basic for now but it works as intended :)
So we have three subdirectories in `/var/lib/dokku/plugins`, `core`, `custom` and `enabled`. Plugins are now divided in two categories: core ones and other ones.
The `enabled` directory only links to plugins (either in `core` or `enabled` directories). The core plugins are enabled automatically when installing dokku and cannot be disabled.

There are two new commands:
* `dokku plugins:enable <name>` to enable a third party plugin.
* `dokku plugins:disable <name>` to disable a third party plugin. Core plugins cannot be disabled by this command.

`plugins-install` and `plugins-install-dependencies` now take an optional argument `--core` which when set will instruct those commands to only refer to the core plugins. This should resolve the issue #1205.
This happens by setting PLUGIN_PATH to `/var/lib/dokku/plugins/core` instead of the default `/var/lib/dokku/plugins/enabled` when `--core` is set.

/cc @josegonzalez @michaelshobbs @jimeh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,10 @@ SSHCOMMAND_URL ?= https://raw.github.com/progrium/sshcommand/master/sshcommand
 PLUGINHOOK_URL ?= https://s3.amazonaws.com/progrium-pluginhook/pluginhook_0.1.0_amd64.deb
 STACK_URL ?= https://github.com/gliderlabs/herokuish.git
 PREBUILT_STACK_URL ?= gliderlabs/herokuish:latest
-PLUGINS_PATH ?= /var/lib/dokku/plugins
+DOKKU_LIB_ROOT ?= /var/lib/dokku
+PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/plugins
+CORE_PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/core-plugins
+DISABLED_PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/disabled-plugins
 
 # If the first argument is "vagrant-dokku"...
 ifeq (vagrant-dokku,$(firstword $(MAKECMDGOALS)))
@@ -44,13 +47,15 @@ packer:
 
 copyfiles:
 	cp dokku /usr/local/bin/dokku
-	mkdir -p ${PLUGINS_PATH}
-	find ${PLUGINS_PATH} -mindepth 2 -maxdepth 2 -name '.core' -printf '%h\0' | xargs -0 rm -Rf
+	mkdir -p ${CORE_PLUGINS_PATH} ${PLUGINS_PATH} ${DISABLED_PLUGINS_PATH}
+	rm -rf ${CORE_PLUGINS_PATH}/*
 	find plugins/ -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | while read plugin; do \
-		rm -Rf ${PLUGINS_PATH}/$$plugin && \
-		cp -R plugins/$$plugin ${PLUGINS_PATH} && \
-		touch ${PLUGINS_PATH}/$$plugin/.core; \
+		rm -Rf ${CORE_PLUGINS_PATH}/$$plugin && \
+		rm -rf ${PLUGINS_PATH}/$$plugin && \
+		cp -R plugins/$$plugin ${CORE_PLUGINS_PATH} && \
+		ln -s ${CORE_PLUGINS_PATH}/$$plugin ${PLUGINS_PATH}; \
 		done
+	chown dokku:dokku -R ${PLUGINS_PATH} ${CORE_PLUGINS_PATH} ${DISABLED_PLUGINS_PATH}
 	$(MAKE) addman
 
 addman:
@@ -62,10 +67,10 @@ version:
 	git describe --tags > ~dokku/VERSION  2> /dev/null || echo '~${DOKKU_VERSION} ($(shell date -uIminutes))' > ~dokku/VERSION
 
 plugin-dependencies: pluginhook
-	dokku plugins-install-dependencies
+	dokku plugins-install-dependencies --core
 
 plugins: pluginhook docker
-	dokku plugins-install
+	dokku plugins-install --core
 
 dependencies: apt-update sshcommand pluginhook docker help2man man-db
 	$(MAKE) -e stack
--- a/dokku
+++ b/dokku
@@ -5,8 +5,11 @@ shopt -s nullglob
 export DOKKU_DISTRO=${DOKKU_DISTRO:="ubuntu"}
 export DOKKU_IMAGE=${DOKKU_IMAGE:="gliderlabs/herokuish"}
 export DOKKU_ROOT=${DOKKU_ROOT:=~dokku}
+export DOKKU_LIB_ROOT=${DOKKU_LIB_PATH:="/var/lib/dokku"}
 
-export PLUGIN_PATH=${PLUGIN_PATH:="/var/lib/dokku/plugins"}
+export PLUGIN_PATH=${PLUGIN_PATH:="$DOKKU_LIB_ROOT/plugins"}
+export PLUGIN_CORE_PATH=${PLUGIN_CORE_PATH:="$DOKKU_LIB_ROOT/core-plugins"}
+export PLUGIN_DISABLED_PATH=${PLUGIN_DISABLED_PATH:="$DOKKU_LIB_ROOT/disabled-plugins"}
 export DOKKU_NOT_IMPLEMENTED_EXIT=10
 export DOKKU_VALID_EXIT=0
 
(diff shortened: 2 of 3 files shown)


dokku
```diff
@@ -207,14 +210,22 @@ case "$1" in
     ;;
 
   plugins)
-    ls -1 -d $PLUGIN_PATH/*/
+    for plugin in "$PLUGIN_PATH"/*/; do
```
Flink: here I did this since I think outputting only the name of the plugins is more useful than the complete path. I might be wrong :sweat_smile:

dokku
```diff
@@ -207,14 +210,22 @@ case "$1" in
...
+      basename "$plugin"
+    done
     ;;
 
   plugins-install)
```
Flink: I think `plugins-install` (and `plugins-install-dependencies`, `plugins-update`) should be renamed `plugins:install` to follow the convention used everywhere else.

Flink:
I did not look at the tests at all since I don’t know where we’re going exactly :grin:

dokku
jimeh: :+1:

dokku
jimeh: :+1:

jimeh:
Looks like a good start to my dokku-n00b eyes :) 

And not that you've gotten there yet, but just so it's mentioned here, the upgrade process would obviously need to delete existing core plugins in the old plugins path, and probably move any non-core plugins into the new custom path.

josegonzalez:
@michaelshobbs how much of this would be affected by the `plugn` pr?

michaelshobbs:
There is significant overlap. One of you guys should really take a look at plugn and that branch before going too far down this path. I'd hate for you to waste your time.

Flink:
@jimeh :+1: 

@michaelshobbs I looked at your PR and I don’t think there’s a real problem here. I’m pretty sure (but I didn’t check) that we could replicate what I do here with `plugn` instead of `pluginhook`. Plugn seems to take an env variable for its plugins directory so the mechanism is pretty similar to the one used here.

Also the goal of this PR isn’t the same than yours. Here we try to give a solution (that seems pretty clean to me) to the problem in issue #1205. Another important point I think is that plugin compatibility is not affected at all by this PR.

Flink:
Just found a bug: `source "$(dirname "$0")/../common/functions"` won’t work with symlinks :confused: 
Anyone knows why `source "$PLUGIN_PATH/common/functions"` isn’t used instead?

jimeh:
Really? Shouldn't `source` follow symlinks? Or does `dirname  $0` return the real path rather than the symlinked one?

Flink:
well it won’t work with `ls` too. The thing that works is when you move around with `cd` because your shell keeps an history from where you’ve “entered” the symlinked directory. And since here there is no context, `..` refers to the “real” parent directory.

Flink:
So we’ll need to use `$PLUGIN_PATH` instead in all plugins (but compatibility will be broken :confused:) or we’ll have to drop symlinks and copy the plugins between directories. Other ideas?

jimeh:
Ah, yeah you're right. The `..` does indeed screw it up. Sorry, seemed weird while I was guessing off the top of my head.

Seems symlinks for custom plugins isn't going to work then. But I just did some basic testing and seems custom plugins will work if they simply live in the folder relative to the `common` symlink, rather than being symlinked in themselves.

I tested a folder structure like this locally with some mock plugins:

```
/var
 └── lib
     └── dokku
         ├── core-plugins
         |   ├── common
         |   ├── checks
         |   └── domains
         ├── disabled-plugins
         |   └── mongodb
         └── plugins
             ├── common -> /var/lib/dokku/core-plugins/common
             ├── checks -> /var/lib/dokku/core-plugins/checks
             ├── domains -> /var/lib/dokku/core-plugins/domains
             ├── forego
             └── redis
```

It's mostly just the `enable` and `disable` commands that need to simply move the custom plugin folder back and forth, while refusing to operate on symlinks. It might not be the absolutely cleanest solution, but probably the most decent one to avoid breaking backwards compatibility.

And on a side-note, the example I have above would mean even less backwards compatibility issues, as even installation instructions in existing plugins would still be correct when they instruct you to copy/clone/extract the plugin to `/var/lib/dokku/plugins`.

As for why `dirname "$0"` is used, seems silly, but I guess someone started the convention and others followed... lol

josegonzalez:
Just convention. If you guys want to make a separate pr that uses `$PLUGIN_PATH`, that would work.

Flink:
@jimeh yes I think your solution will work and if we want to keep plugin compatibility we don’t really have a choice.

@josegonzalez we could be then we’ll have to update (almost) all community plugins too :grin: I’m not sure it’s worth the effort.

jimeh:
I think it's worth to change core plugins to use `$PLUGIN_PATH` in an effort to start setting a new and better convention :)... I'd be happy to submit such a PR.

Flink:
Ok since you seem to be motivated, I’ll wait for you PR. Meanwhile, I’ll patch my plugins and forks :wink:

jimeh:
PR is in, but I don't think we can rely on or assume all 3rd party plugins will be updated any time soon. So plugin management is still probably best done via moving them back and forth. But at least new plugins will hopefully follow the new convention and make life easier for everyone in the future.

Flink:
Yes but it’s sad.

We can update and submit PR to a lot of them. There’s also a lot of plugins which aren’t active/maintained anymore :confused: 

Since we’ll need an upgrade script with the new plugins management, we could check for `source $(dirname $0)/../common/functions` in installed ones and print a warning (and not enable them).

Flink:
or better yet, replace automatically the line with the new one but still print a warning message to update plugin (or report the error to the author).

jimeh:
That's one solution, although when/how would that check be executed? I'm assuming as part of `plugins:enable` right?

And come to think of it, the enable/disable thing is nice, but if we use the move solution the benefit is that there's ZERO difference in how things work now. Plugins are enabled by default, they don't run as part of dokku installation, and you can disable plugins via a simple command instead of manually deleting or moving the directory on the server.

The symlink solution feels nicer to me from a architectural point of view, but does alter how the end user installs plugins. An extra step isn't much, but it's still different.

Don't know which one you prefer @Flink, nor am I fully sure which one I prefer, just wanted to point things out for clarity's sake :)

P.S. As a nod towards moving plugins around, OSX actually handles user-installed fonts that way. User fonts live in `~/Library/Fonts`, and if you disable one, it's moved to `~/Library/Fonts (Disabled)`.

Flink:
Well I think I’m like you on this one, the symlink solution seems “better” but what you say is also very true. Moving plugins instead of symlinking them and changing the directory structure will be better from a user point of view.

Flink:
Ok I think this PR is done now and should work as expected. I used the approach suggested by @jimeh with directories and `mv` instead of symlinks. So we don’t need an upgrade script or something like that, it will work out of the box with an existing installation. You just have to upgrade dokku normally :smile: 

/cc @josegonzalez @michaelshobbs

michaelshobbs:
If we're good with this I'd like to merge it into the plugn branch and merge them both into master via that PR; unless we want this in before 0.4.0. Though I didn't think we were releasing another 0.3.x.

Flink:
Well it fixes the issue with debian installation so it can be merged in a 0.3.x branch. But I personally don’t have a preference :)

jimeh:
I don't mind either :)... Personally I worked around the installation issue by having the chef [cookbook](https://github.com/rlister/chef-dokku-simple/pull/4) disable/enable plugins itself before installation/upgrade, so I don't mind waiting till 0.4.x for a fix.

josegonzalez:
We are not doing another 0.3.x release.

Flink:
Ok so it settled ;) @michaelshobbs Why not merge this in master and then rebase your PR?

josegonzalez:
We're not rebasing plugn anymore.

michaelshobbs:
@Flink 6 of one have a dozen of the other. Hence me saying I'd *like* to do it this way. Just a preference.

Flink:
Ok I was just asking :sweat_smile: If it’s easier for you like that then there’s no problem (I don’t really mind :grin:)

dokku
josegonzalez: This probably will need to change with plugn, right @michaelshobbs ?

dokku
```diff
@@ -211,6 +222,23 @@ case "$1" in
     pluginhook update
     ;;
 
+  plugins:disable)
+    PLUGIN="$2"
+    [[ -e $PLUGIN_CORE_PATH/$PLUGIN ]] && echo "Cannot disable a core plugin" && exit 1
+    [[ -e $PLUGIN_DISABLED_PATH/$PLUGIN ]] && echo "Plugin already disabled" && exit 1
+    [[ ! -e $PLUGIN_PATH/$PLUGIN ]] && echo "Plugin does not exist" && exit 1
+    mv "$PLUGIN_PATH/$PLUGIN" "$PLUGIN_DISABLED_PATH"
+    echo "Plugin $PLUGIN disabled"
```
josegonzalez: Pretty sure this should all be done via plugn, right?

michaelshobbs:
Yeah no worries at all. It's probably all the same but my thinking was just that if we're going to overhaul plugins then let's just do it in one go.

Makefile
```diff
@@ -4,7 +4,10 @@ SSHCOMMAND_URL ?= https://raw.github.com/progrium/sshcommand/master/sshcommand
 PLUGINHOOK_URL ?= https://s3.amazonaws.com/progrium-pluginhook/pluginhook_0.1.0_amd64.deb
 STACK_URL ?= https://github.com/gliderlabs/herokuish.git
 PREBUILT_STACK_URL ?= gliderlabs/herokuish:latest
-PLUGINS_PATH ?= /var/lib/dokku/plugins
+DOKKU_LIB_ROOT ?= /var/lib/dokku
+PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/plugins
+CORE_PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/core-plugins
```
josegonzalez: This is never changed in the debian packaging, so we can't merge this...

dokku
michaelshobbs: Yeah but if we're going to have two enabled places then we'll need to manage the plugin path ourselves somehow.

michaelshobbs: There are still some unknowns for me about how this plays along with plugn. Another reason I'd like to pull it into the plugn branch and play around with it. Anyone opposed to this?

Flink: nope

Makefile
Flink: How so? Why couldn’t you update the Makefile?

josegonzalez: Well the debian packaging doesn't use a makefile...

josegonzalez:
If you want to merge this into the plugn branch first, be my guest :) . Just close and rebase it onto that branch.

josegonzalez:
Okay I'm going to close this PR. Please redo the PR and target the `mh-switch-to-plugn` branch. We can merge it into there and then see how we like/dislike the experience :)

PR #1428, opened 2015-08-28
Use `$PLUGIN_PATH` instead of `$(dirname $0)/..`

jimeh:
As briefly discussed in #1425, source the `common/functions` file via an absolute path, rather than determining the relative path by using `dirname`. 3rd-party plugins should follow suit and use the new `$PLUGIN_PATH` convention too.

I only had a few minutes of spare time to put this together, so I haven't run tests locally, and instead will be relying on the PR CI stuff to do it for me :)


Flink:
That was quick! :smile: I think you should also update the documentation to reflect these changes.

jimeh:
Done, apparently my `ack` setup doesn't search through .md files :stuck_out_tongue_closed_eyes:

Flink:
:+1:

PR #1437, opened 2015-09-01
Update compatibility version for several plugins

Flink:
As requested in https://github.com/progrium/dokku/issues/1295#issuecomment-136756883


Flink:
I also cleaned up a little :)

josegonzalez:
Lets use `0.3.26+` instead of just `0.3.26`.

Also, would be ever so lovely if you could test more than just these plugins ;)

Flink:
Ok for `0.3.26+`. For the other plugins, yes but it’ll take some time :grin:. Also we should maybe deprecate some of the plugins listed.

josegonzalez:
For this pr, just change it to `0.3.26+`. You can do deprecations in a different PR, and any other plugins in a separate one as well :)

josegonzalez:
Also note, in a documentation PR update, you can skip tests by adding `[ci skip]` to the commit message.

Flink:
Ah yes you’re right. Ok then!

Flink:
done :)

PR #1444, opened 2015-09-06
Docker cleanup

josegonzalez:


plugins/common/functions
```diff
@@ -320,3 +320,16 @@ release_and_deploy() {
...
 }
+
+docker_cleanup() {
+  # delete all non-running containers
+  # if a container has the label `dokku`,
+  # then that container is not deleted.
+  # This can be used to mark data containers
+  # as non-removable by the cleanup function
+  # shellcheck disable=SC2046
+  docker rm $(comm <(docker ps -aq -f 'status=exited' | sort) <(docker ps -aq -f 'status=exited' -f 'label=dokku' | sort) -3) &> /dev/null || true
```
michaelshobbs: this still looks to be cleaning up all exited containers, no? i thought the point was to only cleanup labeled containers?

josegonzalez: No, it removes lines that exist in [both](http://ss64.com/bash/comm.html):

>    -3   suppress lines that appear in both files 

I also changed it to `dokku.service` in #1448 to give us the ability to use the `dokku` label elsewhere.

PR #1445, opened 2015-09-07
Create CONTRIBUTING.md

### LINKED ISSUE #1435: Create a CONTRIBUTING.md
@michaelshobbs I think it's time we codify how we merge stuff, decide what goes into a release, how people should contribute etc. Do you have thoughts on what is useful data to provide for bugs? What about the structure of how and why people should ask for features vs things that should go into plugins?

We'll also want to extend this to the official plugins, though some of the rules may be different.

/cc @Flink since you're starting to get back into contributing and might have some thoughts as an outsider.

josegonzalez:
Closes #1435

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,191 @@
+# How to contribute
+
+Dokku loves to welcome your contributions. There are several ways to help out:
+
+* Create an [issue](https://github.com/progrium/dokku/issues) on GitHub, if you
+  have found a bug
+* Write test cases for open bug issues
+* Write patches for open bug/feature issues, preferably with test cases
+  included
+* Contribute to the [documentation](http://progrium.viewdocs.io/dokku/)
+
+There are a few guidelines that we need contributors to follow so that we have
+a chance of keeping on top of things.
+
+## Topics
+
+* [Reporting Security Issues](#reporting-security-issues)
+* [Reporting Issues](#reporting-other-issues)
+* [Contributing](#contributing)
+  * [Making Changes](#making-changes)
+  * [Which branch to base the work](#which-branch-to-base-the-work)
+  * [Submitting Changes](#submitting-changes)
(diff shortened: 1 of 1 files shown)


josegonzalez:
I think we can start here and iterate as we see what works?

Flink:
From https://github.com/docker/docker/blob/master/CONTRIBUTING.md#quick-contribution-tips-and-guidelines there’s also a little section on git and I think we should incite good practices when writing commits like using rebase, squash commits, etc. (starting with us ;))

CONTRIBUTING.md
```diff
@@ -0,0 +1,183 @@
...
+  does not already exist.
+  * Clearly describe the issue including steps to reproduce when it is a bug.
+  * Make sure you fill in the earliest version that you know has the issue.
+* Fork the repository on GitHub.
+
+### Making Changes
+
+* Create a topic branch from where you want to base your work.
+  * This is usually the master branch.
+  * Only target an existing branche if you are certain your fix must be on that
```
michaelshobbs: branche?

CONTRIBUTING.md
```diff
@@ -0,0 +1,183 @@
...
+    branch.
+  * To quickly create a topic branch based on master; `git branch
+    master/my_contribution master` then checkout the new branch with `git
+    checkout master/my_contribution`. Better avoid working directly on the
+    `master` branch, to avoid conflicts if you pull in updates from origin.
```
michaelshobbs: maybe reference `git checkout -b` as a single step instead of branch and checkout

michaelshobbs:
:+1: on @Flink's comment

CONTRIBUTING.md
Flink: that’s french for branch :grin:

josegonzalez:
Updated the `Making Changes` section to take @Flink's comment into account. Merging :)

Flink:
Nice :+1:

PR #1458, opened 2015-09-10
set TRACE=true when DOKKU_TRACE is set

michaelshobbs:
this will pass the `TRACE` env var when `DOKKU_TRACE` is set; causing `herokuish` to yield more debug output


josegonzalez:
Seems legit! Ship when ready.

PR #1460, opened 2015-09-11
bump herokuish version to 0.3.3

michaelshobbs:
@josegonzalez do we do anything else to dokku in order to move to a new version of herokuish? 

side note: the docker hub build is still running due to circleci being down earlier.

### DIFF
--- a/deb.mk
+++ b/deb.mk
@@ -1,6 +1,6 @@
 HEROKUISH_DESCRIPTION = 'Herokuish uses Docker and Buildpacks to build applications like Heroku'
 HEROKUISH_REPO_NAME ?= gliderlabs/herokuish
-HEROKUISH_VERSION ?= 0.3.2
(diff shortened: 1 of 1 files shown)


michaelshobbs:
@josegonzalez ok the new image is up on the hub. anything else need to be done here?

josegonzalez:
Might be worth updating the `docker build` command on line 81 to respect `HEROKUISH_VERSION`. I think we've been allowing any old upgrade since 5ever otherwise...

Other than that, looks good to me :)

michaelshobbs:
What do you mean?

josegonzalez:
I meant we should build a specific version of the docker image, though it seems the build script for herokuish [doesn't actually push tags](https://hub.docker.com/r/gliderlabs/herokuish/tags/) so at least this will trigger an upgrade I guess.

PR #1461, opened 2015-09-11
idea: not requiring image-extras

mrluc:
This is more of a thought experiment.

But adding `|| true` to the line that attempts to install `linux-image-extra-...` would mean that dokku could run places with custom kernels, like Linode, by falling back to DeviceMapper. People who really want AUFS could use pv-grub and the StackScript (once a documentation tweak is made).

Is that just a crazy idea? I know that DeviceMapper has worse performance for starting/stopping containers, but as AUFS isn't upstream, it seems like it might work

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -46,7 +46,9 @@ dokku_install_package() {
   curl -sSL https://packagecloud.io/gpg.key | apt-key add -
 
   apt-get update > /dev/null
-  apt-get install -qq -y "linux-image-extra-$(uname -r)" apt-transport-https
+  
+  # if linux-image-extra isn't found for AUFS, allow it to continue and use devicemapper.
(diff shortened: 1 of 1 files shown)


josegonzalez:
Here is where docker requires it in their [install.sh](https://github.com/docker/docker/blob/master/hack/install.sh#L254).

Might be worth us just installing their install into our install:

```
curl -sSL https://get.docker.com/ | sh
```

mrluc:
I don't know if that would cause any other problems, but yeah -- the concern is "why should dokku require aufs if docker doesn't any more?" It could still be mentioned in the hosting-provider-specific docs, as of course it's something that would concern power users.

josegonzalez:
Closing in favor of 1467. @mrluc can you check if the bootstrap.sh from that branch can be run directly on linode?

Issue #1226, opened 2015-06-10
Stackscript for Linode strangeness


chasd00:
1. The "Xenify distro" option wasn't in my profile config. I don't know if Linode has renamed it or what. I opened a ticket with them. When I set "Automount devtmpfs" to no I was able to boot.

2. The stackscript had a couple of errors like this:

E: Some index files failed to download. They have been ignored, or old ones used instead.
E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem.
=> 06/10 21:36:12 Installing pre-requisites

When I manually did the dpkg --config -a then re-ran the stackscript it got further but never completed.

chasd00:
1. Got  a response from Linode "Xenify distro" has been replaced with "Distro helper"

michaelshobbs:
@josegonzalez thoughts here?

chasd00:
Per this blogpost the Linode kernels have what they need to run Dokku out of the box.
https://blog.linode.com/2014/01/03/docker-on-linode/

Here's what I did to get Dokku running at Linode on Ubuntu 14.04 LTS after the free KVM upgrade
1. login as root and do $ssh-keygen
2. curl --silent https://get.docker.io/gpg 2> /dev/null | apt-key add - 2>&1 >/dev/null
3. curl --silent https://packagecloud.io/gpg.key 2> /dev/null | apt-key add - 2>&1 >/dev/null
4. echo "deb http://get.docker.io/ubuntu docker main" > /etc/apt/sources.list.d/docker.list
5. echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ trusty main" > /etc/apt/sources.list.d/dokku.list
6. apt-get update > /dev/null
7. apt-get install -qq -y dokku 
- Ignore the modprobe error
8. reboot and finish through the web install (I chose the web install option)

I didn't have to mess with any of the profile options either ( "Xenify distro" or "Distro Helper" )

nicksergeant:
Confirmed that @chasd00's steps worked for me too, though I switched to Linode's KVM implementation instead of Xen (that's what they're pushing new servers on eventually anyway).

mrluc:
@chasd00's comment seems like a fairly important note for people using Linode! Seems like the documentation should be changed to reflect that a stackscript isn't required for those on KVM, which is what Linode calls "the future here at Linode."

[...]

PR #1463, opened 2015-09-11
Update `Xenify Distro` option

josegonzalez:
Refs #1226


josegonzalez:
@mrluc thoughts on this change? I also updated the stackscript.

mrluc:
Oh, cool, I figured that was in Linode's wheelhouse somehow.

PR #1464, opened 2015-09-14
limit amount of lines in "dokku logs -t"

KernelMadness:
Long running applications are producing enormous amount of logs and currently "dokku logs -t" throws all history from proccess start to console.

This patch limits amounts of log lines to 100, as in "dokku logs" without -t


josegonzalez:
Seems reasonable to me.

PR #1469, opened 2015-09-15
Add nginx:access-logs and nginx:error-logs without sudo requirement

michaelshobbs:
includes and closes #1352


michaelshobbs:
that's the assumption if we `sudo su - dokku`

PR #1473, opened 2015-09-17
Handle crashing containers by using restart=on-failure policy

### LINKED ISSUE #216: Auto Restarting crashed apps
Is there a proposed solution for monitoring dokku apps and auto restarting if the crash?  Would just setting up something like http://godrb.com/ work?

### LINKED ISSUE #398: Node.js and Dokku: How to recover from app crashes?
I finally have my Dokku server running on DigitalOcean and can deploy my Node.js app to it. My procfile looks like this:

```
web: node app.js
```

The problem with this is that once my app crashes for some reason, it won't restart. One solution would be to use the npm package called [Forever](https://npmjs.org/package/forever). This however is a package that needs to be installed globally using the -g tag.

I'm not sure how to go about this in Dokku since it installs dependencies through package.json. Would I need to manually install this package on the Docker container or is there another solution?

Or maybe there is a better way to recover from crashes when using Dokku altogether?

### LINKED ISSUE #1327: Auto restarting crashed containers
Is there any way to tell dokku to automatically attempt to restart crashed apps or apps that are not responding to the `CHECK`? I'm deploying a Node.js server without a Dockerfile.

josegonzalez:
Previously a crashed container would stay down, regardless of exit status. In some cases, it may be useful to restart the container. For example, an application may not be correctly implementing their error handling, or the crash may be caused by a transient error.

By setting the restart policy to `on-failure:N` - where N is a number of max restarts - we can help developers guard against crashing applications. Note that this is not a replacement for proper error handling, nor does this include notifications to a developer when a container is restarted. Those patterns should be implemented application side, or via a feature request to docker.

The value is configurable at the app-level by setting DOKKU_RESTART_LIMIT to a number. By default, containers will be restarted a max of 10 times. If a container crashes during the check-deploy plugin trigger, then the deploy will be marked as a failure.

Closes #216
Closes #398
Closes #1327

### DIFF
--- a/dokku
+++ b/dokku
@@ -77,6 +77,9 @@ case "$1" in
     DOKKU_SCALE_FILE="$DOKKU_ROOT/$APP/DOKKU_SCALE"
     oldids=$(get_app_container_ids $APP)
 
+    DOKKU_RESTART_LIMIT=$(dokku config:get $APP DOKKU_RESTART_LIMIT || echo 10)
+    DOKKU_RESTART_LIMIT=${DOKKU_RESTART_LIMIT:=10}
+
     while read line || [[ -n "$line" ]]; do
       TRIM=${line%#*}
       PROC_TYPE=${TRIM%%=*}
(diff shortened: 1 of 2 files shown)


michaelshobbs:
Seems legit

dokku
```diff
@@ -91,6 +94,7 @@ case "$1" in
 
         # start the app
         DOCKER_ARGS=" -e DYNO='$PROC_TYPE.$CONTAINER_INDEX' "
+        DOCKER_ARGS=" --restart=on-failure:$DOKKU_RESTART_LIMIT"
```
tmikoss: Shouldn't this be `+=`?

josegonzalez: You think you're so clever, don't you!

Patch incoming!

PR #1474, opened 2015-09-17
Reduce trace output

josegonzalez:
Cutting down on a few things that make `dokku trace` output annoying to read through.

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+_ipv4_regex() {
+  echo "([0-9]{1,3}[\.]){3}[0-9]{1,3}"
+}
+
+_ipv6_regex() {
+  local RE_IPV4="$(_ipv4_regex)"
+  local RE_IPV6="([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|"              # TEST: 1:2:3:4:5:6:7:8
+  RE_IPV6="${RE_IPV6}([0-9a-fA-F]{1,4}:){1,7}:|"                         # TEST: 1::                              1:2:3:4:5:6:7::
+  RE_IPV6="${RE_IPV6}([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|"         # TEST: 1::8             1:2:3:4:5:6::8  1:2:3:4:5:6::8
+  RE_IPV6="${RE_IPV6}([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|"  # TEST: 1::7:8           1:2:3:4:5::7:8  1:2:3:4:5::8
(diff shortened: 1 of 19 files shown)


josegonzalez:
Of note: removing `TRACE=1` from the docker options during a trace run will shorten the output a further 3.5k lines.... Definitely worth making optional.

josegonzalez:
Would be incredibly nice if we could refactor plugin commands such that they were native plugn commands as opposed to the case/switch setup we have now. I think that would decrease trace output tremendously. Potentially a bit late to do that, but thats up to you, @michaelshobbs.

Note: I'm happy to revert any of the commits in this PR.

michaelshobbs:
Interesting. I'm down for reducing unnecessary trace output. What are you thinking in regards to using native plugn commands?

josegonzalez:
Probably splitting out each command into a subcommand, like the old refactor PR that did exactly that. That would totally reduce the amount of output, though would be yet another refactor of plugins for 0.4.0 (and a much more backwards-incompatible change). Also not sure how we would use colons in filenames...

michaelshobbs:
But how does that relate to "native plugn commands?"

josegonzalez:
At the moment we send any commands through a `foreach plugin` loop, which adds a bunch of extra output we don't need. Instead we could just do `plugn trigger command_name` to trigger commands, right?

michaelshobbs:
Yeah let me see what I can do. Not sure `trigger` is the right method since that goes through all enabled plugins.

josegonzalez:
Note: that might not be the best route and we might still need to support both methods - the git plugin has `git-*`` as one of it's supported commands - though in the normal case should work quite well.

michaelshobbs:
I say let's merge this and refactor plugin commands (again) afterwards

josegonzalez:
Ha sorry :(

I'll rebase this then.

josegonzalez:
Again, I don't know if refactoring how commands are triggered is the best way forward here. It would be the biggest bc-incompatible change we make thus far and potentially a lot of work. Would also basically break git history.

Then again, if it does look like it's worth it to reduce trace output, then we should break plugin compatibility in a single go. :(

michaelshobbs:
This method is rough on maintenance to be sure because we'll need to maintain the list at the top of each file.

michaelshobbs:
We're going to have some major conflicts between this PR and the nginx one....

josegonzalez:
I'm happy to remove that check for now if that makes it easier to deal with.

michaelshobbs:
It's fine. We can remove it when we refactor plugin commands. Both previous PRs we backwards compatible with the old method of command execution. So no need to hold `0.4.0` for it.

plugins/nginx-vhosts/functions
```diff
@@ -0,0 +1,38 @@
...
+  # Fixes using a wildcard dns service such as xip.io which allows for *.<ip address>.xip.io
+  echo "${RE_IPV4}\$"
+}
+
+get_ipv6_regex() {
+  local RE_IPV6="$(_ipv4_regex)"
+  # Ensure the ip address continues to the end of the line
+  # Fixes using a wildcard dns service such as xip.io which allows for *.<ip address>.xip.io
+  echo "${RE_IPV6}\$"
+}
```
michaelshobbs: I did something similar in #1476

michaelshobbs:
@josegonzalez you want me to bring these commits into #1476 or just handle conflict after the fact?

michaelshobbs:
closing in favor of #1476 which includes changes here

PR #1476, opened 2015-09-18
nginx proxy without VHOST

### LINKED ISSUE #1123: Container IP always change when deploying an app on no-domain server
Hello! I've setup a Dokku app on Digital Ocean and deployed a Node.js app on it. Everything seems working fine but every time I push a new version to the server it changes the running port, for example if the app was running on http://123.123.123.123:49156 after the restart it will run on http://123.123.123.123:49157.

Is there a way to make an app always run on a given port? Thank you.

michaelshobbs:
closes #1123 #1474 

example output with no `$DOKKU_ROOT/VHOST` and an IP in `$DOKKU_ROOT/HOSTNAME` (also can be accomplished by setting `NO_VHOST=1`) 
note: we always use the contents of `$DOKKU_ROOT/HOSTNAME` for urls output in the aforementioned cases

```
$ git push dokku@dokku.me:node-js-app master
Pseudo-terminal will not be allocated because stdin is not a terminal.
Counting objects: 454, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (365/365), done.
Writing objects: 100% (454/454), 218.70 KiB | 0 bytes/s, done.
Total 454 (delta 79), reused 391 (delta 51)
-----> Cleaning up...
-----> Building node-js-app from herokuish...
-----> Fetching custom buildpack
-----> Node.js app detected
...
-----> All checks successful!
=====> node-js-app container output:
       Node app is running at localhost:5000
=====> end node-js-app container output
-----> Running post-deploy
=====> unsupported vhost config found. disabling vhost support
-----> Setting config vars
       NO_VHOST: 1
-----> no nginx port set. setting to random open high port
-----> Setting config vars
       DOKKU_NGINX_PORT: 22578
       Reloading nginx
=====> Application deployed:
       http://10.0.0.2:32824 (container)
       http://10.0.0.2:22578 (nginx)

To dokku@dokku.me:node-js-app
 * [new branch]      master -> master
```


plugins/common/functions
```diff
@@ -330,3 +330,16 @@ docker_cleanup() {
...
+  while true; do
+    local port=$(shuf -i 1025-65535 -n 1)
+    if ! nc -z 0.0.0.0 $port; then
+      echo $port
+      return 0
+    else
+      continue
+    fi
+  done
+  dokku_log_fail "no open ports found...wut?!"
```
josegonzalez: We should be a bit more professional here ;)

michaelshobbs: Bahahahaha!

josegonzalez: I don't just click merge as soon as I see a green button :P

michaelshobbs: Lies!!!!!

michaelshobbs:
I'm not particularly happy with how this ended up. Any feedback on the `nginx_build_config` function would be welcomed.

docs/nginx.md
```diff
@@ -172,7 +180,7 @@ If desired, it is possible to disable vhosts by setting the `NO_VHOST` environme
 dokku config:set myapp NO_VHOST=1
 ```
 
-On subsequent deploys, the nginx virtualhost will be discarded. This is useful when deploying internal-facing services that should not be publicly routeable.
+On subsequent deploys, the nginx virtualhost will be discarded. This is useful when deploying internal-facing services that should not be publicly routeable. As of 0.4.0, nginx will still be configured to proxy your app on some random high port. This allows internal services to maintain the same port between deployments. You may change this port by setting `DOKKU_NGINX_PORT` and/or `DOKKU_NGINX_SSL_PORT` (for services configured to use SSL.)
```
josegonzalez: Should we add port as an option to `nginx:enable` so that users can modify the port to their preferred value?

plugins/00_dokku-standard/commands
```diff
@@ -193,8 +193,18 @@ case "$1" in
       echo "$SCHEME://$(< "$DOKKU_ROOT/VHOST")"
     else
       for PORT_FILE in $DOKKU_ROOT/$APP/PORT.*; do
-        echo "$SCHEME://$(< "$DOKKU_ROOT/HOSTNAME"):$(< "$PORT_FILE")"
+        echo "http://$(< "$DOKKU_ROOT/HOSTNAME"):$(< "$PORT_FILE") (container)"
```
josegonzalez: Why the hardcoded scheme here? What about https?

plugins/nginx-vhosts/functions
```diff
@@ -0,0 +1,204 @@
...
+        SSL_VHOSTS=$(egrep "^${SSL_HOSTNAME_REGEX}$" $VHOST_PATH || true)
+      else
+        SSL_VHOSTS=$(< $DOKKU_ROOT/HOSTNAME)
+      fi
+      NONSSL_VHOSTS=$(egrep -v "^${SSL_HOSTNAME_REGEX}$" $VHOST_PATH 2> /dev/null || true)
+
+      if [[ -z "$DOKKU_NGINX_SSL_PORT" ]]; then
+        if [[ -n "$NO_VHOST" ]]; then
+          dokku_log_info1 "no nginx ssl port set. setting to random open high port"
+          local NGINX_SSL_PORT=$(get_open_port)
```
josegonzalez: It's possible that this subsequent call retrieves the same open port as the previous call. Not sure if we should care about that or not.

plugins/nginx-vhosts/functions
```diff
@@ -0,0 +1,204 @@
...
+      sudo /etc/init.d/nginx reload > /dev/null
+      ;;
+
+    opensuse)
+      sudo /sbin/service nginx reload > /dev/null
+      ;;
+  esac
+}
+
+nginx_build_config() {
```
josegonzalez: This whole method is unfortunately pretty icky. Is there any way to refactor it into separate helper methods to get configuration and then actually build the config?

plugins/nginx-vhosts/commands
```diff
@@ -180,18 +28,52 @@ EOF
...
+  ;;
+
+  nginx:disable)
+    [[ -z $2 ]] && echo "Please specify an app to run the command on" && exit 1
+    verify_app_name "$2"
+    APP="$2"
+
+    eval "$(config_export app $APP)"
+    if [[ -z "$NO_VHOST" ]]; then
+      config_set --no-restart $APP NO_NGINX=1
```
josegonzalez: We should probably prefix this with `DOKKU_`. Any config var that is dokku specific will then be less likely to conflict with a user env var...

plugins/nginx-vhosts/commands
```diff
@@ -180,18 +28,52 @@ EOF
...
+    fi
+  ;;
 
   help | nginx:help)
     cat<<EOF
     nginx:build-config <app>, (Re)builds nginx config for given app
     nginx:access-logs <app> [-t], Show the nginx access logs for an application (-t follows)
     nginx:error-logs <app> [-t], Show the nginx error logs for an application (-t follows)
+    nginx:disable <app>, disable nginx for an application (forces container binding to external interface)
+    nginx:enable <app>, enable nginx for an application
```
josegonzalez: Heh, re: refactoring how we trigger plugin commands, would be cool to see these become `declare=` calls in individual functions somehow. Or at least move the declaration of the help output from a command to where the command is declared.

plugins/certs/commands
```diff
@@ -84,7 +85,7 @@ case "$1" in
       mkdir -p "$DOKKU_ROOT/$APP/tls"
       dokku_log_info1 "Installing certificate and key..."
       mv -f $TMP_WORK_DIR/server.key $TMP_WORK_DIR/server.crt $SSL_PATH
-      [[ -n "$DOMAIN" ]] && (dokku domains:add $APP $DOMAIN || dokku nginx:build-config $APP)
+      [[ -n "$DOMAIN" ]] && (dokku domains:add $APP $DOMAIN || nginx_build_config $APP)
```
josegonzalez: Yay for smaller trace output!

docs/nginx.md
michaelshobbs: eh? <shrug> 

just set the config var...?

plugins/00_dokku-standard/commands
michaelshobbs: this is only if vhost is disabled and we print out the ssl version only if we have a port

plugins/nginx-vhosts/functions
michaelshobbs: yeah this is fraught with peril. we're do a point in time search for an open port. that port could be taken by another process in the future.....

docs/nginx.md
josegonzalez: Fine by me :)

plugins/nginx-vhosts/functions
michaelshobbs: thought about that....maybe/probably? this whole thing is a mess....

plugins/nginx-vhosts/commands
josegonzalez: Now that I think about it, once we're done with any changes we are making, I think it's worth looking through all of our "magic" config vars and prefixing them with `DOKKU_`, while deprecating the old usage. At least then we can grep for their usage.

michaelshobbs:
I'm working on some more refactoring

plugins/config/functions
```diff
@@ -214,6 +214,7 @@ config_unset() {
     ENV_TEMP=$(echo -e "${ENV_TEMP}" | sed "/^export $var=/ d")
 
     config_write "$ENV_TEMP"
+    [[ "$var" == "NO_VHOST" ]] && config_set --no-restart $APP NO_VHOST=0 && DOKKU_CONFIG_RESTART=true
```
michaelshobbs: it was impossible to support transitions from `NO_VHOST=1` to nothing and maintain a single function that told the truth about an app's vhost state.

michaelshobbs:
I think this is a step in the right direction. Could definitely use some more time to breakdown `nginx_build_config` but let's not hold up `0.4.0` for this.

plugins/common/functions
```diff
@@ -330,3 +330,15 @@ docker_cleanup() {
   # shellcheck disable=SC2046
   docker rmi $(docker images -f 'dangling=true' -q) &> /dev/null &
 }
+
+get_open_port() {
+  while true; do
+    local port=$(shuf -i 1025-65535 -n 1)
+    if ! nc -z 0.0.0.0 $port; then
```
u2mejc: I think your if statement is missing the brackets to evaluate this and you have a continue in an iff statement?

michaelshobbs: it twerks:

```
$ get_open_port() {
>   while true; do
>     local port=$(shuf -i 80-81 -n 1)
>     if ! nc -z 0.0.0.0 $port; then
>       echo "open port: $port"
>       return 0
>     else
>       echo "failed port: $port"
>       continue
>     fi
>   done
> }
$ get_open_port
failed port: 80
failed port: 80
open port: 81
```

plugins/common/functions
```diff
@@ -330,3 +330,15 @@ docker_cleanup() {
   # shellcheck disable=SC2046
   docker rmi $(docker images -f 'dangling=true' -q) &> /dev/null &
 }
+
+get_available_port() {
+  while true; do
+    local port=$(shuf -i 1025-65535 -n 1)
```
michaelshobbs:
```
$ dpkg -S /usr/bin/shuf
coreutils: /usr/bin/shuf
```

josegonzalez: ?

michaelshobbs: just showing what provides `shuf`

plugins/config/functions
```diff
@@ -214,6 +214,7 @@ config_unset() {
     ENV_TEMP=$(echo "${ENV_TEMP}" | sed "/^export $var=/ d")
 
     config_write "$ENV_TEMP"
+    [[ "$var" == "NO_VHOST" ]] && config_set --no-restart $APP NO_VHOST=0 && DOKKU_CONFIG_RESTART=true
```
josegonzalez: Why `DOKKU_CONFIG_RESTART=true` here?

plugins/nginx-vhosts/commands
```diff
@@ -1,168 +1,17 @@
 #!/usr/bin/env bash
+[[ " nginx:build-config nginx:access-logs nginx:error-logs help nginx:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
```
josegonzalez: This doesn't have the two new functions you added.

tests/unit/core_ports.bats
```diff
@@ -164,6 +154,16 @@ check_urls() {
...
+@test "(core) dockerfile port exposure" {
+  deploy_app dockerfile
+  run bash -c "grep -A1 upstream $DOKKU_ROOT/$TEST_APP/nginx.conf | grep -q 3000"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+
+  check_urls http://${TEST_APP}.dokku.me
+}
+
```
josegonzalez: Why was the test moved down? Just curious.

plugins/config/functions
michaelshobbs: To ensure consistency

tests/unit/core_ports.bats
michaelshobbs: To be next to the other dockerfile tests

josegonzalez:
yey passes

PR #1478, opened 2015-09-18
Fix naming of herokuish package in installation docs

eljojo:
not that kind of kush

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -42,7 +42,7 @@ dokku ps:rebuildall
 If dokku was installed via a debian package, you can upgrade herokuish via the following command:
 
 ```shell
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -42,7 +42,7 @@ dokku ps:rebuildall
 If dokku was installed via a debian package, you can upgrade herokuish via the following command:
 
 ```shell
-sudo apt-get install Herokush
+sudo apt-get install Herokuish
```
michaelshobbs: probably should be lower case as well

michaelshobbs:
Epic kush!!! :smile:

eljojo:
okey dokey, just lowercased it :snoopdoggemoji:

PR #1482, opened 2015-09-21
Strip the `dokku-` part from plugins on install

Flink:
This way we won’t have a list of plugins with a lot of `dokku-*` in their names.

### DIFF
--- a/plugins/plugin/commands
+++ b/plugins/plugin/commands
@@ -17,7 +17,7 @@ case "$1" in
         ;;
       https:*|git:*)
         PLUGIN_GIT_URL="$2"
-        download_and_enable_plugin $PLUGIN_GIT_URL
+        download_and_enable_plugin "$PLUGIN_GIT_URL" "$3"
         plugn trigger install
         ;;
       *)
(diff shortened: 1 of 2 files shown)


michaelshobbs:
I totally get the reasoning here but I'm not sure we want to 'magically' modify names of plugins. I'm curious if anybody else has thoughts on this. 

(To be clear: I don't think I personally care but I'd like to hear what other users think before merging this in)

hhff:
jah i tend to agree - with such a large mass of bash scripting I think we'd want to keep the "magic" to a minimum

josegonzalez:
You could ask people for confirmation and respect the `--force` flag.

Flink:
@josegonzalez ask for confirmation of installation without the `dokku-` part? What the `--force` flag will do ?

josegonzalez:
A few things:

- We should be able to specify the plugin name.
- If we detect a `dokku-` prefix, we should ask the user to confirm that they want to remove it.
- If the user does `dokku --force plugin:install whatever`, then it will remove the prefix without prompting the user.

Flink:
Ok so `plugin:install` should accept a second argument that specify the local name of the plugin to be installed, right?
If user provides this second argument, do we strip the `dokku-` prefix if it has it? I think we shouldn’t in this case.
And then `--force` should do nothing when user provide the second argument in this case.

josegonzalez:
We don't strip if they provide a name.

michaelshobbs:
I'd prefer if we didn't prompt by default. Sounds annoying

josegonzalez:
Okay, fine by me.

Flink:
Ok, so `--force` is not needed anymore then :)

Flink:
Updated :)

josegonzalez:
Seems legit!

Issue #1487, opened 2015-09-23
dokku plugin:install-dependencies --core fails with "hook: command not found" error


vijayviji:
I'm using windows 10 machine. I synced the git repo@v0.4.0 and when did "vagrant up", it ran many things (including creating VM and installing dokku).

When it reaches  the step "dokku plugin:install-dependencies --core", it fails

==> dokku: /var/lib/dokku/core-plugins/enabled/20_events/dependencies: line 1: hook: command not found
==> dokku: make:
==> dokku: *** [plugin-dependencies] Error 127

And so, the entire installation was getting cancelled. Inside "/var/lib/dokku/core-plugins/enabled/20_events/dependencies" file, it has only "hook" word in it.

*Root Cause:*
This is because, the symlink files don't work in Windows. Basically, "/var/lib/dokku/core-plugins/enabled/20_events/dependencies" is a symlink to "/var/lib/dokku/core-plugins/enabled/20_events/hook". But in windows, git creates a file for symlink and makes the content of the file to be the filename of the symlink.

As vagrant exports the directory to the VM and tries to install from it, all the problems were happening.

vijayviji:
Check out http://stackoverflow.com/questions/5917249/git-symlinks-in-windows

josegonzalez:
This one is a weird issue, since dokku is meant to work on linux and we've never tested a vagrant vm from linux. The *easiest* fix is to just copy the hook into each file. 

Note that this *also* affects other official plugins that use the same symlinking pattern. Not sure what we're going to decide to do here...

vijayviji:
I'm not sure how to solve this for contributors.

But for users, there are two solutions to this. When creating the VM through vagrant, instead of sharing the folder from host to guest, we can sync the git repo again from guest OS and start a fresh install. Other is to create a basic Ubuntu VM and run the "installation" script which does the same thing through inline script.

As no contributor is using vagrant (for virtual box) and that too from windows, I think we can follow this.

josegonzalez:
I use vagrant to contribute to dokku exclusively (other than testing deb packages on a live site and other experiments).

Starting fresh seems like a bad idea since then people will be confused as to why their changes aren't synced into the vm.

vijayviji:
Then we need to create two vm environment in Vagrantfile, one for contributing (dev) and another for production (or users).

https://docs.vagrantup.com/v2/multi-machine/index.html

josegonzalez:
Sure, do you want to come up with a pr for this?

[...]

PR #1490, opened 2015-09-23
Fix for issue #1487

vijayviji:
Created a separated VM environments for developers and users. Also, fixed a minor issue in bootstrap.sh.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,7 +18,6 @@ Vagrant::configure("2") do |config|
   config.ssh.forward_agent = true
 
   config.vm.box = BOX_NAME
-  config.vm.synced_folder File.dirname(__FILE__), "/root/dokku"
 
   config.vm.provider :virtualbox do |vb|
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -55,9 +55,9 @@ dokku_install_package() {
   apt-get update -qq > /dev/null
 
   if [[ -n $DOKKU_CHECKOUT ]]; then
-    apt-get install dokku=$DOKKU_CHECKOUT
+    apt-get install -y dokku-$DOKKU_CHECKOUT
   else
-    apt-get install dokku
+    apt-get install -y dokku
```
josegonzalez: Why the changes here?

Vagrantfile
```diff
@@ -30,15 +29,27 @@ Vagrant::configure("2") do |config|
 
   config.vm.define "empty", autostart: false
 
+  # For normal users
```
josegonzalez: Note: If anything this should either:

- be reversed
- just autodetect if you are on windows and don't sync the directory in that case.

bootstrap.sh
vijayviji: 'apt-get install dokku' was prompting a question. So, I changed it. This is not relevant to the issue. But I faced with this while debugging.

Vagrantfile
vijayviji: Sure, I can change it.

bootstrap.sh
josegonzalez: Gotcha. I think this was fixed already in a separate PR, so if you can rebase onto master, that would be :+1:

vijayviji:
Hi Jose. I rebased the master.

josegonzalez:
Can you squash your commits into one?

vijayviji:
I rebased and squashed.

josegonzalez:
Can you update the virtualbox documentation with a note at the top stating how it works on windows?

PR #1494, opened 2015-09-23
Correctly install packages for DOKKU_TAG=v0.4.0

### LINKED ISSUE #1493: bootstrap.sh incorrectly installs from source for DOKKU_TAG "v0.4.0" through "v0.4.12".
Look at at https://github.com/progrium/dokku/blob/v0.4.0/bootstrap.sh, the [condition to determine the install strategy](https://github.com/progrium/dokku/blob/v0.4.0/bootstrap.sh#L64-L91) looks like this (pseudocode):

```python
if $DOKKU_BRANCH:
    dokku_install_source
elif $DOKKU_TAG:
    # Parse version into $major, $minor, $patch
    if $major >= 0 and $minor >= 3 and $patch >= 13: # <-- "v0.4.0" will never pass
        dokku_install_package
        if $major >= 0 and $minor >= 4 and $patch >= 0: # <-- Unreachable until "v0.4.13"
          dokku plugin:install-dependencies --core
        else:
          dokku plugins-install-dependencies
    else:
        dokku_install_source # <-- So we end up here instead
else:
    dokku_install_package
    dokku plugin:install-dependencies --core
```

All of the tagged install_package paths are contingent on passing `$DOKKU_TAG >= 0.3.14`, but since each version component is compared individually, "v0.4.0" fails the `$patch` test, which looks at `0 >= 13`.

Thus, all attempted installations with `"DOKKU_TAG=v0.4.0"` fall back to `dokku_install_source`.

callahad:
Fixes #1493

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -71,16 +71,16 @@ elif [[ -n $DOKKU_TAG ]]; then
   patch=$(echo $DOKKU_SEMVER | awk '{split($0,a,"."); print a[3]}')
 
   # 0.3.13 was the first version with a debian package
-  if [[ "$major" -eq "0" ]] && [[ "$minor" -ge "3" ]] && [[ "$patch" -ge "13" ]]; then
+  if [[ "$major" -eq "0" ]] && [[ "$minor" -eq "3" ]] && [[ "$patch" -ge "13" ]]; then
     export DOKKU_CHECKOUT="$DOKKU_SEMVER"
     dokku_install_package
     echo "--> Running post-install dependency installation"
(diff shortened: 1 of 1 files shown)


callahad:
Oh lord, this would also start failing for `$DOKKU_TAG >= v0.4.13`, since it would take the first branch into the `>= 0.3.13` stanza. Fixing...

callahad:
Alright, the current revision, 80a2018, should be good to merge.

michaelshobbs:
Good catch.

PR #1500, opened 2015-09-24
Log user name and fingerprint in events

jazzzz:
This allows to see who is deploying.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -98,7 +98,7 @@ dokku_log_plugn_trigger_call() {
   local l_hook
 
   l_hook="$1" ; shift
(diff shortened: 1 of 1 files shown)


josegonzalez:
Where does `NAME` come from?

jazzzz:
> Where does NAME come from?

From `sshcommand`.

josegonzalez:
I'm guessing it sets it as an environment variable, correct?

jazzzz:
> I'm guessing it sets it as an environment variable, correct?

Yes, you can see this in `~dokku/.ssh/authorized_keys`.

PR #1504, opened 2015-09-25
change plugin install doc to show one-step method

### LINKED ISSUE #1503: dokku plugin:install <git url> is not documented
The doc indicates this:

```shell
cd /var/lib/dokku/plugins/available
git clone <git url>
dokku plugin:install
```
whereas 
```shell
dokku plugin:install <git url>
```
should work, but I get this error:

`/var/lib/dokku/plugins/enabled/00_dokku-standard/install: line 16: /etc/init/dokku-redeploy.conf: Permission denied`

It works adding `sudo`, but I'm not sure it should be needed.

michaelshobbs:
closes #1503

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -13,10 +13,17 @@ Let's take a quick look at the current dokku nginx plugin that's shipped with do
 ## Installing a plugin
 
 ```shell
-cd /var/lib/dokku/plugins/available
+# This command requires `root` permissions as the `install` and `install-dependencies`
+# plugin triggers may utilize commands such as `apt-get`. For non-core plugins, please
+# inspect those plugins before running the following command as `root` user.
+sudo plugin:install <git_url>
+
+# previous versions (0.3.x and below) of dokku require a manual process to install plugins
+cd /var/lib/dokku/plugins
 git clone <git url>
-dokku plugin:install
+dokku plugins-install
 ```
+
 ## Creating your own plugin
 
 [See the full documentation](http://progrium.viewdocs.io/dokku/development/plugin-creation).


docs/plugins.md
```diff
@@ -13,9 +13,7 @@ Let's take a quick look at the current dokku nginx plugin that's shipped with do
 ## Installing a plugin
 
 ```shell
-cd /var/lib/dokku/plugins/available
-git clone <git url>
-dokku plugin:install
+dokku plugin:install <git url>
```
josegonzalez: Should we mention that this requires root access?

michaelshobbs: For sure. Any idea where folks got the idea they didn't need root though?

josegonzalez: I mean why would you need root for a directory that you'd expect to be owned by root?

michaelshobbs:
I added a note underneath. Let me know what you think.

docs/plugins.md
```diff
@@ -13,10 +13,10 @@ Let's take a quick look at the current dokku nginx plugin that's shipped with do
...
 ```
+Note: `plugin:install` might trigger actions like `apt-get`, please make sure you inspect any potential plugin's `install` or `install-dependencies` trigger, and (of course) execute this command as the `root` user
+
```
josegonzalez: I think you *still* need root to clone the plugin in the first place.

michaelshobbs: Not necessarily. You could clone it if you run `plugn install` as the `dokku` user. However `plugn trigger install` will fail with the reported error.

```
root@dokku:/var/lib/dokku# ls -lad ./*/{available,enabled}
drwxr-xr-x 21 dokku dokku 4096 Sep 19 00:27 ./core-plugins/available
drwxr-xr-x  2 dokku dokku 4096 Sep 19 00:27 ./core-plugins/enabled
drwxr-xr-x  2 dokku dokku 4096 Sep 22 16:01 ./plugins/available
drwxr-xr-x  2 dokku dokku 4096 Sep 22 16:01 ./plugins/enabled
```

josegonzalez: Interesting, my dokku dir is owned by root...

michaelshobbs: `/var/lib/dokku` is owned by `root`. However we don't install plugins in that directory anymore.

josegonzalez: Basically I think this comment outside of the codeblock should be inside the codeblock and we should have ` sudo` prefix on the command. Something like this:

```shell
# This command requires `root` permissions as the `install` and `install-dependencies`
# plugin triggers may utilize commands such as `apt-get`. For non-core plugins, please
# inspect those plugins before running the following command as `root` user.
sudo plugin:install <git_url>

# previous versions (0.3.x and below) of dokku require a manual process to install plugins
cd /var/lib/dokku/plugins/available
sudo git clone <git url>
sudo dokku plugin:install
```

Does that make sense? Keeping the root notes outside of the codeblock is an easy way for users to get confused and complain that the docs are bad because they can't read ;)

PR #1511, opened 2015-09-28
Add `named-containers` as a core plugin

Flink:
As discussed in https://github.com/progrium/dokku/issues/1483


Flink:
@josegonzalez tests seem to work, the error was: `Failed to destroy btrfs snapshot: operation not permitted`. So not related to my plugin :grin:

michaelshobbs:
@Flink that's a side effect of circleci's infrastructure and is not actually the cause of any test failure. Anyhow, looks like it's green now.

Flink:
@michaelshobbs cool :)

PR #1514, opened 2015-09-29
do not use exit 0 in config functions

michaelshobbs:
using `exit 0` in the config functions broke the build-env plugin by exiting before populating `$BUILD_ENV`
also revealed that our build-env test coverage sucked

this fixes both AND adds a test

`CURL_TIMEOUT` defaults were broken as well. added tests.

### DIFF
--- a/plugins/config/functions
+++ b/plugins/config/functions
@@ -102,7 +102,7 @@ config_all() {
   [[ "$1" == "config" ]] || set -- "config" "$@"
   config_parse_args "$@"
   config_create "$ENV_FILE"
-  is_config_export "$@" && config_export "$DOKKU_CONFIG_TYPE" "$APP" && exit 0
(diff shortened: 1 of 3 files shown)


michaelshobbs:
@josegonzalez can we please get this into `0.4.1`? this is blocking my dokku upgrade project :smile:

michaelshobbs:
@josegonzalez curl defaults were broken too. fix + tests coming

plugins/config/functions
```diff
@@ -132,7 +132,7 @@ config_get() {
 
   config_create "$ENV_FILE"
   if [[ ! -s $ENV_FILE ]]; then
-    exit 0
+    return 0
```
josegonzalez: Lol this might account for bugs in the official storage plugins. Ooops?

Issue #1505, opened 2015-09-25
New containers created every reboot


bobvanderlinden:
Description of problem:

A new docker container is created every time I reboot the machine for Dokku apps.

Output of the following commands

- `dokku version`:
```
v0.4.0
```
- `dokku plugins`:
```
 !     `plugins` is not a dokku command.
 !     See `dokku help` for a list of available commands.
```
- `docker version`:
```
Client:
 Version:      1.8.2
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   0a8c2e3
 Built:        Thu Sep 10 19:19:00 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.8.2
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   0a8c2e3
 Built:        Thu Sep 10 19:19:00 UTC 2015
 OS/Arch:      linux/amd64
```
- `docker info`:
```
Containers: 5
Images: 9
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 19
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 3.19.0-25-generic
Operating System: Ubuntu 14.04.3 LTS
CPUs: 1
Total Memory: 993.2 MiB
Name: dokku
ID: F5XJ:RPWM:PSSQ:OY32:XEHY:TQ6C:YYJQ:K4WG:O4SP:QUA7:63LM:JNHQ
WARNING: No swap limit support
```
- `uname -a`:
```
Linux dokku 3.19.0-25-generic #26~14.04.1-Ubuntu SMP Fri Jul 24 21:16:20 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

Environment details (AWS, VirtualBox, physical, etc.):

Running Dokku on a clean VirtualBox machine with a clean Ubuntu Server 14.04.3 LTS install. 
How was dokku installed?:
```
$ history
    1  sudo apt-get update
    2  sudo apt-get upgrade
    3  wget https://raw.github.com/progrium/dokku/v0.4.0/bootstrap.sh
    4  sudo DOKKU_TAG=v0.4.0 bash bootstrap.sh
    5  sudo sshcommand acl-add dokku bob
        ...
```
How reproducible:
I just did this a second time to make sure it wasn't some mistake:
* Create a new virtual machine in VirtualBox.
[...]

PR #1518, opened 2015-09-30
make sure applications stop (and start) when using ps:restart

bobvanderlinden:
As noted in #1505, `dokku ps:restart` does not actually stop existing docker instances. This can be seen by running `docker ps` before and `docker ps` after `dokku ps:restart`. This same behavior happens upon restarting the system as `dokku ps:restartall` is run upon boot. This can lead to numerous 'zombie' docker processes.

This change solves the problem for me. It now runs `dokku ps:stop` and `dokku ps:start` when running `docker ps:restart`. Before this change `release_and_deploy` was run, which is what `dokku ps:start` does as well, but whichout checking whether the app was already running.

### DIFF
--- a/plugins/ps/commands
+++ b/plugins/ps/commands
@@ -70,7 +70,10 @@ case "$1" in
 
     ! (is_deployed $APP) && echo "App $APP has not been deployed" && exit 0
 
-    release_and_deploy "$APP" "$IMAGE_TAG"
+    if (is_app_running $APP); then
+        dokku ps:stop "$APP"
+    fi
+    dokku ps:start "$APP"
     ;;
 
   ps:restartall)


plugins/ps/commands
```diff
@@ -70,7 +70,10 @@ case "$1" in
 
     ! (is_deployed $APP) && echo "App $APP has not been deployed" && exit 0
 
-    release_and_deploy "$APP" "$IMAGE_TAG"
+    if (is_app_running $APP); then
+        dokku ps:stop "$APP"
+    fi
+    dokku ps:start "$APP"
```
josegonzalez: @michaelshobbs worth turning these into functions and then calling them?

bobvanderlinden: Yes, that might be a good idea. The same goes for these instances:
https://github.com/bobvanderlinden/dokku/blob/fix-psrestart/plugins/ps/commands#L84
https://github.com/bobvanderlinden/dokku/blob/fix-psrestart/plugins/ps/commands#L61

Should these also be turned into functions?

josegonzalez: Probably

michaelshobbs: This breaks zero-downtime....

michaelshobbs:
As noted in #1505, this breaks dokku's current benefit of zero-downtime "actions" (deployments, restart, etc)

bobvanderlinden:
Sorry, I indeed missed the zero-downtime.

PR #1523, opened 2015-10-02
Make room for the longer URL from #1522

callahad:
Since #1522, the URL is now too long to fit inside our `max-width: 650px` box. This bumps the box to `750px` which fits the current URL.

Before:

before

After:

after


josegonzalez:
Note: You can add [ci skip] to changes like this to avoid triggering CI testing.

PR #1524, opened 2015-10-02
Document idiosyncracies with Linode - Fixes #1226

### LINKED ISSUE #1226: Stackscript for Linode strangeness

callahad:


callahad:
Doh! I *almost* remembered to tag this with [ci skip] :)

PR #1530, opened 2015-10-06
Fix nginx configuration for SSL

Flink:
When generating config for SSL vhost, configuration for non ssl vhost
was always added (not the redirect one). This should fix the issue.


michaelshobbs:
i'm pretty sure this breaks at least one of our tests but if not, then cool. :smile:

josegonzalez:
It doesn't! Merging.

PR #1560, opened 2015-10-13
Fixes #1516 Bug when you cannot use both SSL and non-SSL templates at the same time.

### LINKED ISSUE #1516: Certs plugin - nginx.ssl.conf.template is not used for HTTPS urls
Dokku version 0.4.1 / Ubuntu 14.04
I added certificate using new certs plugin.
Now `VHOST` has following entries:
```
mydomain.com
app.mydomain.com
```

URLS file:
```
https://app.mydomain.com
http://mydomain.com
```

Rebuilding:
```
 ssh dokku@jedi nginx:build-config mydomain
-----> Configuring SSL for app.mydomain.com...
-----> Configuring mydomain.com...
       Reloading nginx
```

But the final `nginx.conf` contains both parts from `nginx.conf.template`:
```
server {
  listen      [::]:80;
  listen      80;
  server_name app.mydomain.com;
...
}
server {
  listen      [::]:80;
  listen      80;
  server_name mydomain.com;
...
}
```

So looks like there is a bug and script doesn't use `nginx.ssl.conf.template` for SSL subdomain part.

darklow:


josegonzalez:
You are a lovely human being for doing this :)

darklow:
You're welcome. I'm no expert in bash, so if tests will succeed, I still suggest someone with better expertise checks it out :)
But this fix works for me.

PR #1563, opened 2015-10-15
cleanup dead containers

kdomanski:
I've been experiencing intermittent problems with app containers removal. Old containers would end up "dead" after cleanup. This should take care of it.


josegonzalez:
@michaelshobbs ?

michaelshobbs:
Seems safe enough. @kdomanski can you confirm this works for you when you have dead containers?

kdomanski:
@michaelshobbs Affirmative.

michaelshobbs:
:+1:

PR #1567, opened 2015-10-16
README: Switch to CURL from wget on install

omeid:
dokku already requires curl for installation so it is better to use curl to download the bootstrap script as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,7 @@ Docker powered mini-Heroku. The smallest PaaS implementation you've ever seen. S
 
 To install the latest stable release, you can run the following commands as a user that has access to `sudo`:
 
-    wget https://raw.githubusercontent.com/progrium/dokku/v0.4.2/bootstrap.sh
-    sudo DOKKU_TAG=v0.4.2 bash bootstrap.sh
+    curl https://raw.githubusercontent.com/progrium/dokku/v0.4.2/bootstrap.sh | DOKKU_TAG=v0.4.2 bash
 
 ### Upgrading
 


josegonzalez:
`curl` isn't always available - in fact the bootstrap script installs it if it isn't there - so I don't think this change will work.

omeid:
That is also true for wget, so why require two different tools for the same task?

josegonzalez:
There are quite a few other places where we use `wget` - same for curl.

If you want to switch *all* invocations to one or the other, I don't mind. `wget` appears to be installed with ubuntu by default - at least on my test machines - so that's probably the safer bet.

callahad:
I suggest closing as wontfix. Curl is very nice, but wget is much more commonly available.

In both Ubuntu and Debian, wget packages are set as `Priority: Important`, while curl is `Priority: Optional`. Important packages are those which "should be found on any Unix-like system" and are thus much more likely to be present in default installations of the operating system. For instance, Digital Ocean's images have `wget` but not `curl` by default.

callahad:
More info at https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html#s-priority

>  If you do a default Debian installation all the packages of priority __Standard__ or higher will be installed in your system. If you select pre-defined tasks you will get lower priority packages too.

So all default installations of Debian will have wget but not curl. Ubuntu inherits this policy from Debian.

josegonzalez:
@callahad mind making a pr to switch us away from curl to wget then? That would be nice, since then we don't need to install curl...

PR #1569, opened 2015-10-16
remove docker container after stopping it

bobvanderlinden:
This resolves issues where docker would start old containers upon
reboot. Related to issue #1505.

### DIFF
--- a/dokku
+++ b/dokku
@@ -132,7 +132,11 @@ case "$1" in
 
         # if we can't post-deploy successfully, kill new container
         kill_new() {
-          docker inspect $id &> /dev/null && docker stop $id > /dev/null && docker kill $id > /dev/null
+          docker inspect $id &> /dev/null \
+          && docker stop $id > /dev/null \
+          && docker kill $id > /dev/null \
+          && docker rm $id > /dev/null
+
           trap - INT TERM EXIT
           kill -9 $$
         }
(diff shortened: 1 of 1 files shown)


dokku
```diff
@@ -132,7 +132,11 @@ case "$1" in
 
         # if we can't post-deploy successfully, kill new container
         kill_new() {
-          docker inspect $id &> /dev/null && docker stop $id > /dev/null && docker kill $id > /dev/null
+          docker inspect $id &> /dev/null \
+          && docker stop $id > /dev/null \
+          && docker kill $id > /dev/null \
+          && docker rm $id > /dev/null
```
bobvanderlinden: Can someone comment on whether this is the right approach? `;` instead of `&&` might be better in this case. That way stop, kill and rm are called regardless of their previous failing. (Does `docker stop` fail if the container was already stopped?)

bobvanderlinden:
`ps:restartall` at boot seems still needed due to https://github.com/progrium/dokku/blob/92e9ab84ace17b37bdd4c0dc5df5b216daf3c877/plugins/00_dokku-standard/install#L12-L13
I have not tried whether this is still the case, but it doesn't really seem to hurt to keep it.

dokku
michaelshobbs: Unfortunately this prevents a user from seeing the details of why their deployment failed.

dokku
```diff
@@ -199,6 +203,10 @@ case "$1" in
           docker stop $oldid \
           || docker kill $oldid \
           || plugn trigger retire-container-failed $APP # plugin trigger for event logging
+
+          # Attempt to remove the old container regardless of whether stopping/killing succeeded
+          docker rm $oldid \
```
michaelshobbs: This also breaks the troubleshooting chain. See my [comments](https://github.com/progrium/dokku/issues/1505#issuecomment-148729833)

michaelshobbs:
i think we've resolved to remove this functionality for now. (ref: #1591)

@josegonzalez feel free to re-open if i missed something.

PR #1577, opened 2015-10-18
Add official redirect plugin

### LINKED ISSUE #924: Redirects as a built-in feature
Hey,

Thanks for this great tool! I just saw that domain management is now a built-in feature of dokku, no need to use dokku-domains-plugin anymore. That's great to hear. But one important feature I think pretty much websites implement is redirect: non-www <-> www is required these days. We can do it using dokku-redirects-plugin, but as you implemented domains I thought you could also implement redirects.

Cheers.

Flink:
Closes #924


josegonzalez:
`[ci skip]` is your documentation friend :)

michaelshobbs:
👍🏼👊🏼

Flink:
@josegonzalez yep sorry, I’m always forgetting this one :grin:

josegonzalez:
@Flink might be a good idea to move the non-official one to "deprecated" or something :P

josegonzalez:
Same for any others that you've covered so far.

Flink:
Yeah but I didn’t make a redirect plugin before. So I’m not sure about deprecating plugins I don’t own :grin:

josegonzalez:
Fair enough. We'll keep it for now.

PR #1583, opened 2015-10-20
Fix URLs for docu/ links

frdmn:
I noticed that the links point to a non existent "dokku/" folder.

PR should update said links to the current/proper ones :)

### DIFF
--- a/docs/application-deployment.md
+++ b/docs/application-deployment.md
@@ -18,7 +18,7 @@ dokku apps:create ruby-rails-sample
 
 ### Create the backing services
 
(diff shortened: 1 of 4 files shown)


josegonzalez:
Thanks for the pr!

josegonzalez:
Oh, this is meant to be browsed from the docs site, not from inside the repo. Thats why the links seemed broken. Going to revert this.

frdmn:
Ah crud, my bad! I just checked via GitHub and noticed that they all ended up in 404's.

josegonzalez:
Yeah. If I can figure out how to get viewdocs into circleci, I might be able to get @progrium to turn on auto-deploys for it, at which point I can take a stab at rewriting urls like that...

PR #1585, opened 2015-10-20
make sure dokku can read nginx logs and don't remove other perms

### LINKED ISSUE #1572: nginx:access-logs permission denied
clean install on a new virtual machine and things are running fine but the first time I ran `ssh -t dokku@dokku nginx:access-logs myapp -t` I got permission denied.

Further investigation shows the `/var/log/nginx` folder with `drwxr-x--- www-data adm` permissions, preventing the `dokku` user from accessing the folder.

I'm not sure what the "best" solution is here. If I just allow `+x` on the folder then inotify doesn't work with tail (it needs to be able to read the directory). So the folder probably needs to be changed to the dokku group or something else.

This is using ubuntu 14.04 LTS, dokku v0.4.2.

michaelshobbs:
- add current members of `adm` group to dokku group
- change group of `/var/log/nginx` (recursively) to `dokku`
- add group readable perm to everything in `/var/log/nginx`
- removes the check to bypass running `nginx:*-logs` commands as dokku 

closes #1572

### DIFF
--- a/dokku
+++ b/dokku
@@ -41,7 +41,7 @@ if [[ "${args[0]}" =~ ^--.* ]]; then
 fi
 ! has_tty && DOKKU_QUIET_OUTPUT=1
 
-if [[ $(id -un) != "dokku" && $1 != plugin:*install* && $1 != "plugin:update" && $1 != nginx:*-logs ]]; then
+if [[ $(id -un) != "dokku" && $1 != plugin:*install* && $1 != "plugin:update" ]]; then
   sudo -u dokku -E -H $0 "$@"
   exit
 fi
--- a/plugins/nginx-vhosts/install
+++ b/plugins/nginx-vhosts/install
@@ -32,6 +32,11 @@ fi
 
 echo 'server_names_hash_bucket_size 512;' >| /etc/nginx/conf.d/server_names_hash_bucket_size.conf
 
+# make sure dokku can read the default log dir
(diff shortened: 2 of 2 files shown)


josegonzalez:
We need to document this so that users are aware of the adm change. We should probably add a new `logging` doc anyhow.

michaelshobbs:
When you say "we" need to, are you going to update the docs?

josegonzalez:
Yes.

michaelshobbs:
Cool. Just wanted to clarify.

michaelshobbs:
Need to make sure this doesn't break when running multiple times. I'll do that when I get back in front of a computer

michaelshobbs:
:+1:

mmerickel:
AFAICT this is not actually a fix because nginx's logrotate script switches new log files back to `www-data:adm` so unless that script is switched to `www-data:dokku` or dokku is added to the adm group this is still a problem.

Came back to work on Monday after logrotate and my access-log tail was busted again with permission denied.

josegonzalez:
@mmerickel mind filing a bug?

mmerickel:
Versus reopening #1572 I guess?

josegonzalez:
Sure, I can do that. Commenting on a closed pr definitely wasn't going to help ;)

mmerickel:
Fair enough!

PR #1590, opened 2015-10-22
trigger docker-args-build for dockerfile deployments

### LINKED ISSUE #1588: Host volumes / environment variables not available during container build stage
I have the following set up

```
Build options:
    --link dokku.postgres.xsacdb-database:dokku-postgres-xsacdb-database
    -v /storage/xsacdb/conf:/app/conf
    -v /storage/xsacdb/media:/app/media
Deploy options:
    --link dokku.postgres.xsacdb-database:dokku-postgres-xsacdb-database
    -v /storage/xsacdb/conf:/app/conf
    -v /storage/xsacdb/media:/app/media
Run options:
    --link dokku.postgres.xsacdb-database:dokku-postgres-xsacdb-database
    -v /storage/xsacdb/conf:/app/conf
    -v /storage/xsacdb/media:/app/media
```

However when pushing and triggering a build these directories are not accessible.

```
ls: cannot access /app/conf: No such file or directory
ls: cannot access /app/media: No such file or directory
```

They are available once the container is built

```
dokku run xsacdb bash
$ ls /app/conf
files here!
```

My Dockerfile: https://gist.github.com/wjdp/c19959534f6876214d33

In addition the DATABASE_URL environment variable set my the postgres plugin is also not available at build time.

michaelshobbs:
closes #1588

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -40,7 +40,8 @@ case "$1" in
         plugn trigger pre-build-dockerfile "$APP"
 
         [[ "$DOKKU_DOCKERFILE_CACHE_BUILD" == "false" ]] && DOKKU_DOCKER_BUILD_OPTS="$DOKKU_DOCKER_BUILD_OPTS --no-cache"
-        docker build $DOKKU_DOCKER_BUILD_OPTS -t "$IMAGE" .
+        DOCKER_ARGS=$(: | plugn trigger docker-args-build $APP)
+        docker build $DOCKER_ARGS $DOKKU_DOCKER_BUILD_OPTS -t "$IMAGE" .
 
         plugn trigger post-build-dockerfile "$APP"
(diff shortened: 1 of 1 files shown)


andrey013:
Hello

I have noticed that eg. official mongo plugin creates a `link` option in `DOCKER_OPTIONS_BUILD`. 

https://github.com/dokku/dokku-mongo/blob/master/functions#L92

So when I try to push changes I see this: 
```
remote: flag provided but not defined: --link
remote: See 'docker build --help'.
```

`docker build` does not support some flags like `link` or `volume`.

So I think after this PR plugins should reconsider their build flags since they does matter now.

Or am I doing something wrong?

michaelshobbs:
related: #1609

andrey013:
@michaelshobbs thanks. I've missed that issue. That seems to be it, indeed.

PR #1596, opened 2015-10-23
docker-compose support

josegonzalez:
This PR implements preliminary docker-compose support. It's pretty untested, but building containers and deploying them the dokku way appears to work. A few notes:

- `links` do not work. This is because containers have random names when we start them, and the links depend upon those names being applied to running containers. There isn't a straightforward way to do this since you cannot re-link running containers. I'm sure other compose features are similarly broken.
- We require using a `Procfile.compose` to track the required image for a given process type. It's pretty hacky and I'm not proud of it.
- It goes without saying, but you need both a `Dockerfile` and a `docker-compose.yml` file. Unfortunately there is currently no way to disable `docker-compose` support.
- Tests don't pass, mostly because github is broken atm and I didn't run them locally. We probably want a `docker-compose` test as well. I have a small app that should work.
- You still need to run the following (as root) to get this to work:

 ```shell
curl -L https://raw.githubusercontent.com/0k/shyaml/master/shyaml > /usr/local/bin/shyaml; chmod +x /usr/local/bin/shyaml
curl -L https://github.com/docker/compose/releases/download/1.5.0rc1/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
apt-get install python-yaml
```

Todo:

- [ ] Add a test
- [ ] Display warning if links are specified in `docker-compose.yml`.
- [ ] Create a package for shyaml
- [ ] Create a package for docker-compose (and a PR upstream would be nice as well)
- [ ] Do another pass on documentation

Refs #1564
Refs #1594

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -83,7 +83,7 @@ case "$1" in
         done
 
         # add a compose procfile
-        id=$(docker run -i -a stdin $IMAGE /bin/bash -c "cat > /opt/$process_file" < "$process_file")
+        id=$(docker run -i -a stdin $IMAGE /bin/bash -c "cat > /tmp/$process_file" < "$process_file")
         test "$(docker wait $id)" -eq 0
         docker commit $id $IMAGE > /dev/null
         docker stop $id > /dev/null 2>&1 || true
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -62,12 +62,12 @@ case "$1" in
         dokku_log_verbose_quiet "Pulling service images, it may take some time ..."
         cd $TMP_WORK_DIR && docker-compose $COMPOSE_ARGS pull > /dev/null
 
-        app_stanza_type=$(shyaml get-type app < docker-compose.yml 2> /dev/null || true)
         web_stanza_type=$(shyaml get-type web < docker-compose.yml 2> /dev/null || true)
-        if [[ $app_stanza_type == "struct" ]]; then
-          docker tag -f "$APP"_app "$IMAGE"
-        elif [[ $web_stanza_type == "struct" ]]; then
+        app_stanza_type=$(shyaml get-type app < docker-compose.yml 2> /dev/null || true)
+        if [[ $web_stanza_type == "struct" ]]; then
           docker tag -f "$APP"_web "$IMAGE"
+        elif [[ $app_stanza_type == "struct" ]]; then
+          docker tag -f "$APP"_app "$IMAGE"
         else
           dokku_log_fail "Could not detect app or web stanza in docker-compose.yml"
         fi
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -8,6 +8,7 @@ source "$PLUGIN_AVAILABLE_PATH/nginx-vhosts/functions"
 case "$1" in
   build)
     APP="$2"; IMAGE_SOURCE_TYPE="$3"; TMP_WORK_DIR="$4"; IMAGE=$(get_app_image_name $APP)
+    APP_DIR="$DOKKU_ROOT/$APP/app"
     verify_app_name "$APP"
 
     CACHE_DIR="$DOKKU_ROOT/$APP/cache"
@@ -46,6 +47,34 @@ case "$1" in
         plugn trigger post-build-dockerfile "$APP"
         ;;
 
+      compose)
+        # extract first port from docker-compose.yml
+        COMPOSE_PORT=$(cat docker-compose.yml | shyaml get-value app.ports.0 2> /dev/null || true)
+        [[ ! -n "$COMPOSE_PORT" ]] && COMPOSE_PORT=$(cat docker-compose.yml | shyaml get-value web.ports.0 2> /dev/null || true)
+        [[ -n "$COMPOSE_PORT" ]] && config_set --no-restart $APP DOKKU_COMPOSE_PORT=$COMPOSE_PORT
+        plugn trigger pre-build-compose "$APP"
+
+        FIG_ARGS=$(: | plugn trigger compose-args-build $APP)
+        dokku_log_verbose_quiet "Building docker images, it may take some time ..."
+        cd $TMP_WORK_DIR && docker-compose $FIG_ARGS build > /dev/null
+
+        FIG_ARGS=$(: | plugn trigger compose-args-pull $APP)
+        dokku_log_verbose_quiet "Pulling service images, it may take some time ..."
+        cd $TMP_WORK_DIR && docker-compose $FIG_ARGS pull > /dev/null
+
+        app_stanza_type=$(cat docker-compose.yml | shyaml get-type app 2> /dev/null || true)
+        web_stanza_type=$(cat docker-compose.yml | shyaml get-type web 2> /dev/null || true)
+        if [[ $app_stanza_type == "struct" ]]; then
+          docker tag -f "$APP"_app "$IMAGE"
+        elif [[ $web_stanza_type == "struct" ]]; then
+          docker tag -f "$APP"_web "$IMAGE"
+        else
+          dokku_log_fail "Could not detect app or web stanza in docker-compose.yml"
+        fi
+
+        plugn trigger post-build-compose "$APP"
+        ;;
+
       *)
         dokku_log_fail "Building image source type $IMAGE_SOURCE_TYPE not supported!"
         ;;
@@ -76,6 +105,12 @@ case "$1" in
         # buildstep plugins don't necessarily make sense for dockerfiles. call the new breed!!!
         plugn trigger pre-release-dockerfile "$APP" "$IMAGE_TAG"
         plugn trigger post-release-dockerfile "$APP" "$IMAGE_TAG"
+        ;;
+
+      compose)
+        # buildstep plugins don't necessarily make sense for docker-compose. call the new breed!!!
+        plugn trigger pre-release-compose "$APP" "$IMAGE_TAG"
+        plugn trigger post-release-compose "$APP" "$IMAGE_TAG"
         ;;
 
       *)
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -72,6 +72,22 @@ case "$1" in
           dokku_log_fail "Could not detect app or web stanza in docker-compose.yml"
         fi
 
+        # keep track of all the images
+        process_file="Procfile.compose"
+        touch "$process_file"
+        for key in $(cat docker-compose.yml | shyaml keys); do
+          image_name=$(cat docker-compose.yml | shyaml get-value $key.image 2> /dev/null || echo "$APP"_"$key");
+          image_tag="$IMAGE"-"$key"
+          docker tag -f "$image_name" "$image_tag"
+          echo "$key: $image_tag" >> "$process_file"
+        done
+
+        # add a compose procfile
+        id=$(cat $process_file | docker run -i -a stdin $IMAGE /bin/bash -c "cat > /opt/$process_file")
+        test $(docker wait $id) -eq 0
+        docker commit $id $IMAGE > /dev/null
+        docker stop $id && docker kill $id
+
         plugn trigger post-build-compose "$APP"
         ;;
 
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: plugins/00_dokku-standard/commands [hunk 1]
```diff
@@ -46,6 +47,51 @@ case "$1" in
         plugn trigger post-build-dockerfile "$APP"
         ;;
 
+      compose)
```
michaelshobbs: can we make this docker-compose? or at least something less generic.

plugins/git/commands
```diff
@@ -20,11 +20,17 @@ git_build_app_repo() {
   git submodule update --init --recursive &> /dev/null
   find -name .git -prune -exec rm -rf {} \; > /dev/null
 
-  if [[ -f Dockerfile ]] && [[ "$([[ -f .env ]] && grep -q BUILDPACK_URL .env; echo $?)" != "0" ]] && [[ ! -f ".buildpacks" ]]; then
-    dokku receive "$APP" "dockerfile" "$TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
-  else
-    dokku receive "$APP" "herokuish" "$TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
+  if [[ "$([[ -f .env ]] && grep -q BUILDPACK_URL .env; echo $?)" != "0" ]] && [[ ! -f ".buildpacks" ]]; then
+    if [[ -f Dockerfile ]] && [[ -f docker-compose.yml ]]; then
```
michaelshobbs: i think this is a regression to when we first released this feature where some folks had a dockerfile in their repo but wanted a buildpack deploy.

plugins/00_dokku-standard/commands
```diff
@@ -46,6 +47,51 @@ case "$1" in
...
+        # extract first port from docker-compose.yml
+        COMPOSE_PORT=$(shyaml get-value app.ports.0 < docker-compose.yml 2> /dev/null || true)
+        [[ ! -n "$COMPOSE_PORT" ]] && COMPOSE_PORT=$(shyaml get-value web.ports.0 < docker-compose.yml 2> /dev/null || true)
```
michaelshobbs: if shyaml comes from the python-yaml package then you'll want to add installation of this package to the Makefile.

plugins/ps/functions
```diff
@@ -21,17 +21,21 @@ generate_scale_file() {
     trap 'rm -rf "$TMP_WORK_DIR" > /dev/null' RETURN
 
     copy_from_image "$IMAGE" "/app/Procfile" "$TMP_WORK_DIR" 2>/dev/null || true
+    copy_from_image "$IMAGE" "/opt/Procfile.compose" "$TMP_WORK_DIR" 2>/dev/null || true
```
michaelshobbs: where did `/opt/` come from?

josegonzalez: I needed to store this file somewhere during the `dokku build` step, and `/app/` doesn't exist there yet...

michaelshobbs: do we generate `Procfile.compose` or does the user ship that in their repo?

michaelshobbs:
need docs and tests

michaelshobbs:
Can you describe this deployment scenario from repo to deployment? 

Now that the implementation is out of the way :stuck_out_tongue_winking_eye: what is the justification for adding this new method?

josegonzalez:
Read the attached docs for more reasoning :)

josegonzalez:
Basically this "solves" the issue where Dockerfile deployments currently have no way of running multiple processes. We treat the `docker-compose.yml` as a method of defining separate process types.

plugins/ps/functions
josegonzalez: We generate that file. It's a mapping of `docker-compose.yml` processes to the correct docker image.

It's a really shitty hack and I'm not proud of it, but can't think of any alternatives at this time.

plugins/git/commands
josegonzalez: This still allows that. If a user has a `BUILDPACK_URL` set *or* `.buildpacks` file, it will use the buildpacks for deployment.

Maybe we should introduce a method to force a specific type of deployment instead of interpolating this?

plugins/ps/functions
michaelshobbs: If we generate it then how did it get in the app image under `/opt/`?

plugins/git/commands
michaelshobbs: Ah I see that now. derp.

michaelshobbs:
Can you include some example repo artifacts for this method? Even better a test app would be great.

plugins/ps/functions
josegonzalez: I copy it into the original container image

plugins/00_dokku-standard/commands
```diff
@@ -46,6 +47,51 @@ case "$1" in
...
+        touch "$process_file"
+        for key in $(shyaml keys < docker-compose.yml); do
+          image_name=$(shyaml get-value $key.image < docker-compose.yml 2> /dev/null || echo "${APP}_${key}");
+          image_tag="$IMAGE-$key"
+          docker tag -f "$image_name" "$image_tag"
+          echo "$key: $image_tag" >> "$process_file"
+        done
+
+        # add a compose procfile
+        id=$(docker run -i -a stdin $IMAGE /bin/bash -c "cat > /opt/$process_file" < "$process_file")
```
josegonzalez: From line 75-90, the file is generated and copied.

plugins/ps/functions
michaelshobbs: dokku does this or you did it manually?

plugins/00_dokku-standard/commands
michaelshobbs: oh i see that now. the use of a generic directory like `/opt/` seems odd to me. can we use `/tmp` or `/tmp/build` in the container instead?

josegonzalez:
Heh I've managed to stay this long without having tests!

https://github.com/josegonzalez/dokku-compose-app

It's more or less the sample `docker-compose` application, but modified to understand environment variables for the redis url.

plugins/ps/functions
josegonzalez: I did this.

plugins/00_dokku-standard/commands
josegonzalez: Yeah I can take a stab at that.

mahnunchik:
+1

josegonzalez:
Please do not +1 feature pull requests. We'll merge the features when they are ready :)

josegonzalez:
@michaelshobbs If I add my test app, would you mind adding actual tests for it?

michaelshobbs:
please add a test app in this PR and i can add tests to the documented spec. (side note: make sure those docs clarify the expected behavior)

josegonzalez:
I'm going to see if there is a way to support links. Ideally we can, though I'll need to see how that might work...

josegonzalez:
So for the record, I think we can support docker-compose links by rewriting the yaml file, but that will *also* mean that we'd drop support for zero-downtime deploys. The question then becomes whether or not that is worth doing. 

I'm of the opinion that no, we shouldn't do that. If it is decided that both features would be required in order for this to be merged, then I think we should just not include docker-compose.

Benjamin-Dobell:
I don't know anything about docker-compose, but I will say zero downtime deploys are a hugely important feature - definitely not something I'd be willing to sacrifice in production.

josegonzalez:
The question is do we state that we don't add support for it when using docker-compose? I think no, and as such we'd need to drop docker-compose's linking features since we'd be treating each container as just an image as opposed to using compose to orchestrate the containers. Something to think about.

zenzire:
Do you plan to add supporting docker-compose to dokku?

josegonzalez:
Why?

zenzire:
@josegonzalez I would like to run multiple instances (web, workers) using Dockerfile deployment method (no Procfile). 

Currently if I have Dockerfile, dokku creates only web process and there is no way to run other type of processes (for example workers). Using docker-compose standard, dokku can automatically create web and workers processes which are defined in docker-compose.yml file.

josegonzalez:
How do you imagine linking works? Really what we want is a way to define multiple images or the same image with different start commands. Compose more or less replicates a lot of what dokku does and supporting it properly isn't feasible. 

The above *could* work but would leave everyone upset with how things are implemented/not supported.

zenzire:
I understand your concerns about django-compose which overlaps dokku functionality. And I agree with you.

josegonzalez:
So what would you like us to do? Ideally we do *not* introduce a dokku.yml :)

jvanbaarsen:
@josegonzalez For me the removal of Zero-Downtime deploys would be a dealbreaker. What would be biggest benefit be for the Docker-compose?

josegonzalez:
Being able to run multiple commands for a Dockerfile-based deploy.

zenzire:
@josegonzalez I suggest to add a new option to specify path to Dockerfile (as environmental variable). It gives possibility to create separate Dockerfiles for web and workers.

For example, I can have the following repository structure:
```
/Dockerfile
/worker/Dockerfile
/src/
...
```

And then I can create a worker by defining path to /worker/, e.g. dokku config:set my_app DOCKERFILE_PATH=/worker/

mmerickel:
> @josegonzalez I suggest to add a new option to specify path to Dockerfile (as environmental variable). It gives possibility to create separate Dockerfiles for web and workers.

This is already possible via the DOCKER_BUILD_OPTIONS part of the docker-options plugin. You can specify `dokku docker-options:add myapp build -b worker/Dockerfile` or some such incantation.

josegonzalez:
You can't do that for a specific process though.

mmerickel:
I've had success using a Procfile and the `DYNO` environment variable and a `docker-entrypoint.sh` script to dispatch/start the appropriate worker. I suspect this is the right approach currently. You use the Procfile to declare your worker names and then the `DYNO` var contains "worker:1", etc. so you know which one to start in your entrypoint script. The Dockerfile itself defines the dokku app. Separate Dockerfiles would not fit particularly well into dokku's model I don't think.

josegonzalez:
@mmerickel would you be willing to write an documentation for this?

josegonzalez:
Note: even in such a deploy, you don't get the ability to run different dockerfiles. I suppose that is okay.

josegonzalez:
Closing in favor of using the Procfile behavior that @pmclanahan is going to work on :)

raphaklaus:
@josegonzalez is there any feedback about the feature @pmclanahan is working on?

josegonzalez:
That ended up being our Procfile support as documented [here](http://dokku.viewdocs.io/dokku/deployment/methods/dockerfiles/#procfiles-and-multiple-processes).

blocka:
There is another use-case for this which, I believe, is much more compelling -- dev-prod parity. Currently you can use dokku plugins to link to other services, which is great for production but if you app depends on those services you can no longer use it locally. Docker-compose allows you to have the same exact environment, complete with orchestrated services locally and and in production.

Currently, I get around this by using vagrant to orchestrate local docker containers.

It would be great to be able to deploy a docker-compose file as well. It seems to be supported by heroku as well, so I know i'm not crazy: https://devcenter.heroku.com/articles/local-development-with-docker-compose

josegonzalez:
This just doesn't work with zero-downtime deploys. Docker-compose doesn't handle that properly *at all*, and it's not clear to me how we'd handle conditionally starting/stopping certain services once they are running.

If you'd like to contribute to this, please see me on chat - my nick is `savant` - and we can discuss any potential issues before you start working on this.

michaelshobbs:
@josegonzalez I thought I saw the new v3 compose config supports docker services. Perhaps dokku automatically turns it's zero downtime off in the case of compose and we leave it up to the docker-compose.yml author to do what they want. **shrug**

josegonzalez:
@michaelshobbs you may as well just shove sshfront + docker-compose if you're going to do that. I don't see the value of using docker-compose for infra if you are tossing out the best parts of Dokku.

michaelshobbs:
I get your point. To be fair zero-downtime is just **part** of the build -> deploy -> proxy pipeline. The point about using docker-compose to represent prod environments in development is also valid. Just throwing it out there that this might be possible with the latest version of docker-compose.

blocka:
dokku does offer some other goodies like automatically setting up nginx proxies for your containers, plugins (especially letsencrypt...)

josegonzalez:
I mean if someone wants to do the work of refactoring our pipeline to make it easy to make this an engine, then they may want to speak to me directly on chat.

josegonzalez:
I'm locking this thread as I don't think it's productive - the code here isn't mergeable anymore anyhow - and would rather we go back to the drawing board on how this might fit in.

I mentioned engines above, and have had this convo with @michaelshobbs via video, but if you are interested, please chat with me directly.

Issue #976, opened 2015-02-14
Using with Debian rather than Ubuntu


sammcj:
Is it possible to use Dokku with Debian rather than Ubuntu?

* Debian packages are generally more thoroughly tested
* Debian is lighter-weight than Ubuntu
* Debian has more mature, sensible defaults
* Debian (backports) generally has newer packages than those available in Ubuntu
* Debian Jessie (8) isn't far off being released and uses systemd

josegonzalez:
We don't test Debian, but in theory it would work. We have a trusty repo available at packagecloud.com/dokku/dokku, can you let me know of it works on the latest stable Debian?

josegonzalez:
@sammcj bump

sammcj:
Hey, sorry crazy week at work - yup I'll give it a go!

sammcj:
Ok first problem, the dokku deb is looking for 'lxc-docker' which I'm assuming is what Ubuntu calls Docker - which is interesting since Docker hasn't used lxc for quite some time.

```
root@ext-samm:~# apt-get install lxc-docker
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package lxc-docker is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'lxc-docker' has no installation candidate
```

The official Docker packages are called `docker.io`

```
root@ext-samm:~# dpkg -s docker.io
Package: docker.io
Status: install ok installed
Priority: optional
Section: admin
Installed-Size: 19199
Maintainer: Paul Tagliamonte <paultag@debian.org>
Architecture: amd64
Version: 1.3.3~dfsg1-2
Replaces: docker (<< 1.5~)
Depends: adduser, iptables, init-system-helpers (>= 1.18~), perl, libapparmor1 (>= 2.6~devel), libc6 (>= 2.14), libdevmapper1.02.1 (>= 2:1.02.90), libsqlite3-0 (>= 3.5.9)
Recommends: aufs-tools, ca-certificates, cgroupfs-mount | cgroup-lite, git, xz-utils
Suggests: btrfs-tools, debootstrap, lxc, rinse
Breaks: docker (<< 1.5~)
Conffiles:
 /etc/bash_completion.d/docker 7fcdf73ff720606870336942b7efcd74
 /etc/default/docker ce88dab1dcba6f92903120b9beba2521
 /etc/init.d/docker 870b6e52007eca22390f36e3b2547954
[...]

PR #1607, opened 2015-10-25
Dokku support for Debian Jessie installation

josegonzalez:
First pass at making everything work. Not yet tested of course!

Refs #976

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -115,7 +115,7 @@ validate_nginx() {
 restart_nginx() {
   case "$DOKKU_DISTRO" in
     debian)
-      /usr/sbin/invoke-rc.d nginx reload > /dev/null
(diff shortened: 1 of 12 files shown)


plugins/nginx-vhosts/functions
```diff
@@ -114,6 +114,10 @@ validate_nginx() {
 
 restart_nginx() {
   case "$DOKKU_DISTRO" in
+    debian)
+      /usr/sbin/invoke-rc.d nginx reload > /dev/null
+      ;;
+
```
vincentfretin: Why not simply "systemctl reload nginx" ?

josegonzalez: I asked in #debian on freenode and they said this was the way to do it, and that `invoke-rc.d` would do the right thing, regardless of the init system in use.

josegonzalez:
Tested and working on a local system. Going to test it on a DO instance now. Ideally we can just use the same packages on both installations.

PR #1610, opened 2015-10-26
Add post-stop plugn trigger

kdomanski:
This will enable things like marking applications as manually stopped and preventing them from being restarted after reboot.

### DIFF
--- a/plugins/ps/commands
+++ b/plugins/ps/commands
@@ -41,6 +41,7 @@ case "$1" in
     if [[ -n "$DOKKU_APP_RUNNING_CONTAINER_IDS" ]]; then
       echo "Stopping $APP ..."
       docker stop $DOKKU_APP_RUNNING_CONTAINER_IDS > /dev/null || true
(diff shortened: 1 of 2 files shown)


plugins/ps/commands
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-[[ " ps ps:start ps:stop ps:rebuild ps:rebuildall ps:restart ps:restartall ps:scale help ps:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
+[[ " ps ps:start ps:stop ps:rebuild ps:rebuildall ps:restart ps:restartall ps:restore ps:scale help ps:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
```
josegonzalez: Why this change?

kdomanski: Ooops, mixed up from some other work. Removing.

PR #1612, opened 2015-10-26
Add multiple options to the logs plugin

### LINKED ISSUE #1534: New log prefix breaks JSON format for bunyan
Hi,

while the new log line prefix improves general monitoring capabilities it breaks the log format. Some of our services support logging as JSON for *bunyan*. So that we could use it as follows:

```
ssh dokku@<host> logs <app> -t | bunyan
```
Is there a way to get the raw logs instead? Would it be possible to disable/omit the prefix with an explicit call?

Thanks,
Mirko

Flink:
New version of the logs plugin broke compatibility (as reported in #1534).
So options have been added to handle display:
* -t/--tail to tail logs (already existed)
* -n/--num NUM to display only this number of lines
* -p/--ps PS to display logs only from provided process
* -q/--quiet to display raw logs without color, time and names

This should close #1534

### DIFF
--- a/plugins/logs/commands
+++ b/plugins/logs/commands
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$(dirname "$0")/functions"
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/config/functions"
 source "$PLUGIN_AVAILABLE_PATH/nginx-vhosts/functions"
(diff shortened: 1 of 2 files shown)


plugins/logs/commands
```diff
@@ -11,34 +12,60 @@ case "$1" in
...
+    eval set -- "$TEMP"
+
+    LINE_NUMBERS="100"
+    while true; do
+        case "$1" in
+            -h|--help) usage; exit 0 ;;
+            -t|--tail) DOKKU_LOGS_ARGS+="--follow "; shift ;;
+            -n|--num) LINE_NUMBERS="$2"; shift 2 ;;
+            -p|--ps) ONLY_PROCESS="$2"; shift 2 ;;
+            -q|--quiet) NO_PRETTY_PRINT="true"; shift ;;
```
michaelshobbs: I'm concerned with polluting the global variable space here. We (and I mean that to include me) have not been very good at this and have found ourselves causing unintended bugs in the past because of it.

Perhaps we can prepend these variable with `DOKKU_LOGS_*` or, per the [style guide](https://github.com/progrium/bashstyle), turn this into a function and use `local` variable declarations?

plugins/logs/commands
```diff
@@ -11,34 +12,60 @@ case "$1" in
...
+
+    shift 2;
+    TEMP=$(getopt -o htqn:p: --long help,tail,quiet,num:,ps: -n 'dokku logs' -- "$@")
+    if [[ $? != 0 ]]; then usage >&2 ; exit 1 ; fi
+    eval set -- "$TEMP"
+
+    LINE_NUMBERS="100"
+    while true; do
+        case "$1" in
+            -h|--help) usage; exit 0 ;;
```
michaelshobbs: `-h` diverges from our current UI of `<plugin>:help`. I'd personally like to see us converge on some UI standards.

michaelshobbs:
Just a couple comments inline but overall looks good. Thanks @Flink!

plugins/logs/commands
josegonzalez: Flags are nice but basically impossible in an app like dokku due to our usage of shell :(

josegonzalez: I would remove this one, btw.

josegonzalez:
Might be worth adding a `Logging` doc for this change (and in general). That way we don't need to worry about the `-h` flag.

michaelshobbs:
:+1: on logging docs

plugins/logs/commands
josegonzalez: +1 function.

josegonzalez:
i can add logging docs in a separate PR.

plugins/logs/commands
Flink: I’m not sure that all the `getopt` thing will work in a function. Will have to try

plugins/logs/commands
Flink: So I drop the `-h/--help` flag and instead use `logs:help`?

Flink:
@josegonzalez docs are great but I think usage on the CLI is nice to have. And this is how the `heroku` command works :grin:

plugins/logs/commands
josegonzalez: Yeah. Thats what all the official plugins do, right?

josegonzalez:
I'd say we avoid that for now. Option parsing is hard on bash (check out the `enter` plugin to see how we do it there) and if we rewrite dokku, that's something that would be great to tackle (and add better usage to the cli).

plugins/logs/commands
Flink: Yeah but on official plugins, the `:help` command list all available commands without a detailed usage. Here we only have one command which is explained with all of its options.

Flink:
@josegonzalez I still don’t understand your comment about option parsing. Using `getopt` seems to be the standard way to do this and is quite simple.

plugins/logs/commands
josegonzalez: I would just drop the special `-h` flag and document the help the same way we do elsewhere for now. Changing help output in a one-off place isn't great, and we should instead take a wholistic approach at a later point in time.

plugins/logs/commands
josegonzalez: If the function doesn't work, we need to prepend with `DOKKU_LOGS_*`.

josegonzalez:
@Flink I commented inline with my thoughts as to what needs to be changed.

We should create a separate issue discussing how we want the cli to be used (options, help output, etc.) and then converge on that spec for `0.5.0` I think.

plugins/logs/commands
Flink: Ok then, I’ll drop the flag :)

Flink:
@josegonzalez Ok then ;)

Flink:
Updated :)

plugins/logs/commands
```diff
@@ -11,34 +12,63 @@ case "$1" in
...
+    ;;
+
+  logs:help)
+    usage
     ;;
 
   help)
     cat<<EOF
-    logs <app> [-t], Show the last logs for an application (-t follows)
+    logs <app> [-h] [-t] [-n num] [-q] [-p process], Show the last logs for an application
```
josegonzalez: Still has the -h but I'll fix post-merge.

MarcDiethelm:
The `-h` option is still in the help text as of v0.4.4. FYI

PR #1613, opened 2015-10-26
add ps:restore to start applications which weren't manually stopped

kdomanski:
Allows e.g. restarting previously running apps on reboot.

depends on #1610


michaelshobbs:
Hmm, this is interesting. We could potentially use this in `/etc/init/dokku-redeploy.conf` instead of `ps:restartall`. @josegonzalez thoughts?

josegonzalez:
@kdomanski we merged your `post-stop` plugin trigger, so can you remove that from this pr?

@michaelshobbs how would that work?

michaelshobbs:
@josegonzalez  `sed -i 's/ps:restartall/ps:restore' /etc/init/dokku-redeploy.conf`
because `ps:restore` is a filtered list of apps to restart instead of all apps i think this might make more sense to use to start apps on boot. maybe i'm missing something though.

kdomanski:
@michaelshobbs 
`dokku ps:restore` as a oneshot supervisord task inside a container.
Before we used a systemd unit like [this](https://github.com/experimental-platform/platform-configure/blob/1e1338421f3e517026633bb7f1a431baff177df7/services/dokku-rebuild-apps-protonet.service).

josegonzalez:
@michaelshobbs that might work. Any objections against this PR as is? We can test the reboot functionality separately from this.

josegonzalez:
@michaelshobbs Can we merge as is and have a separate pr that redoes our restart strategy?

michaelshobbs:
yes

PR #1622, opened 2015-10-27
Various small doc updates

kimausloos:


docs/configuration-management.md
```diff
@@ -21,13 +21,13 @@ You can set multiple environment variables at once:
...
 
-When setting variables with whitespace, you need to escape the whitespace:
+When setting variables with whitespaces, you need to escape them:
 
 ```shell
 dokku config:set node-js-app KEY=\"VAL\ WITH\ SPACES\"
 ```
 
-When setting or unsetting environment variables, you may wish sometimes to avoid an application restart. This is useful when developing plugins or when setting multiple environment variables in a scripted manner. To do so, use the `--no-restart` flag:
+When setting or unsetting environment variables, you sometimes may wish to avoid an application restart. This is useful when developing plugins or when setting multiple environment variables in a scripted manner. To do so, use the `--no-restart` flag:
```
josegonzalez: Just remove `sometimes` here.

docs/deployment/one-off-processes.md
```diff
@@ -4,8 +4,9 @@ Sometimes you need to either inspect running containers or run a one-off command
 
 ## Run a command in an app environment
 
-```
-run <app> <cmd>                                  Run a command in the environment of an application
+```shell
+# Run a command in the environment of an application
+run <app> <cmd>
```
josegonzalez: Do not reformat these codeblocks please. Will mean updating the help message requires more work.

josegonzalez:
Protip: Avoid test runs using `[ci skip]` in your commit messages :)

callahad:
I've updated the patch to address the comments above. Suggest closing in favor of #1657.

PR #1634, opened 2015-10-31
bats is available in homebrew

callahad:
Leaving ci on to make sure this doesn't break anything :)

### DIFF
--- a/tests.mk
+++ b/tests.mk
@@ -1,10 +1,12 @@
+SYSTEM := $(shell sh -c 'uname -s 2>/dev/null')
+
 shellcheck:
-ifeq ($(shell shellcheck > /dev/null 2>&1 ; echo $$?),127)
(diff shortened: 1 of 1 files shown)


josegonzalez:
Good call. I do this for the official plugins: https://github.com/dokku/dokku-postgres/blob/master/Makefile

callahad:
Makefiles are terrible at conditionals, especially when you start wanting to have three branches: Darwin, Ubuntu, and Debian... there *has* to be a better way. :)

Maybe calling out to an `install_deps.sh` script or something? Leaving that as an exercise for another time.

josegonzalez:
@michaelshobbs might have some thoughts on that. Personally I prefer the makefile approach in this case.

michaelshobbs:
@callahad agreed. However I figured this didn't need an external script since it's used in development and CI.

PR #1635, opened 2015-10-31
Remove deprecated Linode stackscript

callahad:
This is nolonger relevant since we rewrote the Linode docs in commit
81523b3bd01910d485b20dba255e298d7cff0a33

[ci skip]


josegonzalez:
Seems legit.

PR #1636, opened 2015-10-31
Fix dokku compatibility with systemd and sysv-init systems

jindrichskupa:
I added backward compatibility for dokku-installer with SysV init. I am going to prepare dokku works on debian (wheezy, jessie) and ubuntus versions (not only on LTS).

### DIFF
--- a/contrib/dokku-installer.rb
+++ b/contrib/dokku-installer.rb
@@ -3,21 +3,26 @@ require "sinatra"
 require "open3"
 
 if ARGV[0] == "onboot"
-  File.open("/etc/init/dokku-installer.conf", "w") do |f|
-    f.puts "start on runlevel [2345]"
-    f.puts "exec #{File.absolute_path(__FILE__)} selfdestruct"
-  end if File.directory?("/etc/init")
-  File.open("/etc/systemd/system/dokku-installer.service", "w") do |f|
-    f.puts "[Unit]"
-    f.puts "Description=Dokku web-installer"
-    f.puts ""
-    f.puts "[Service]"
-    f.puts "ExecStart=#{File.absolute_path(__FILE__)} selfdestruct"
-    f.puts ""
-    f.puts "[Install]"
-    f.puts "WantedBy=multi-user.target"
-    f.puts "WantedBy=graphical.target"
-  end if File.directory?("/etc/systemd/system")
+  if File.realpath("/sbin/init") != "/sbin/init" then
+    File.open("/etc/init/dokku-installer.conf", "w") do |f|
+      f.puts "start on runlevel [2345]"
+      f.puts "exec #{File.absolute_path(__FILE__)} selfdestruct"
+    end if File.directory?("/etc/init")
+    File.open("/etc/systemd/system/dokku-installer.service", "w") do |f|
+      f.puts "[Unit]"
+      f.puts "Description=Dokku web-installer"
+      f.puts ""
+      f.puts "[Service]"
+      f.puts "ExecStart=#{File.absolute_path(__FILE__)} selfdestruct"
+      f.puts ""
+      f.puts "[Install]"
+      f.puts "WantedBy=multi-user.target"
+      f.puts "WantedBy=graphical.target"
+    end if File.directory?("/etc/systemd/system")
+  else
+    FileUtils.cp(File.absolute_path(__FILE__)+"init", "/etc/init.d/dokku-installer", options)
+    `update-rc.d -f dokku-installer defaults`
+  end
   File.open("/etc/nginx/conf.d/dokku-installer.conf", "w") do |f|
     f.puts "upstream dokku-installer { server 127.0.0.1:2000; }"
     f.puts "server {"
@@ -28,7 +33,7 @@ if ARGV[0] == "onboot"
     f.puts "}"
   end
   `rm -f /etc/nginx/sites-enabled/*`
-  puts "Installed Upstart service and default Nginx virtualhost for installer to run on boot."
+  puts "Installed Upstart/Systemd/SysV-init service and default Nginx virtualhost for installer to run on boot."
   exit
 end
 
--- a/debian/postinst
+++ b/debian/postinst
@@ -45,8 +45,8 @@ case "$1" in
 
     dokku plugin:install --core
 
-    rm -f /home/dokku/VERSION
-    cp /var/lib/dokku/STABLE_VERSION /home/dokku/VERSION
+    rm -f ${DOKKU_ROOT}/VERSION
+    cp /var/lib/dokku/STABLE_VERSION ${DOKKU_ROOT}/VERSION
 
     if [ -f /etc/init/dokku-installer.conf ] && service dokku-installer status 2> /dev/null | grep waiting; then
         sudo service dokku-installer start
--- a/debian/preinst
+++ b/debian/preinst
@@ -3,17 +3,25 @@ set -e
 
 . /usr/share/debconf/confmodule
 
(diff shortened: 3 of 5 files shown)


contrib/dokku-installer.rb
```diff
@@ -3,21 +3,26 @@
...
-    f.puts "Description=Dokku web-installer"
-    f.puts ""
-    f.puts "[Service]"
-    f.puts "ExecStart=#{File.absolute_path(__FILE__)} selfdestruct"
-    f.puts ""
-    f.puts "[Install]"
-    f.puts "WantedBy=multi-user.target"
-    f.puts "WantedBy=graphical.target"
-  end if File.directory?("/etc/systemd/system")
+  if File.realpath("/sbin/init") != "/sbin/init" then
```
josegonzalez: Not sure I understand this change, can you explain?

debian/preinst
```diff
@@ -3,17 +3,25 @@ set -e
...
     if [ "$RET" = "true" ]; then
       INIT_CONF="/etc/init/dokku-installer.conf"
+      INIT_SCRIPT="/etc/init.d/dokku-installer"
       NGINX_CONF="/etc/nginx/conf.d/dokku-installer.conf"
 
-      rm -f $INIT_CONF
-      touch $INIT_CONF
-      echo 'start on runlevel [2345]' >> $INIT_CONF
-      echo 'exec /usr/local/share/dokku/contrib/dokku-installer.rb selfdestruct' >> $INIT_CONF
+      if [[ ${INIT} == "systemd" ]]; then
```
josegonzalez: Isn't this reversed?

debian/postinst
```diff
@@ -45,8 +45,8 @@ case "$1" in
 
     dokku plugin:install --core
 
-    rm -f /home/dokku/VERSION
-    cp /var/lib/dokku/STABLE_VERSION /home/dokku/VERSION
+    rm -f ${DOKKU_ROOT}/VERSION
+    cp /var/lib/dokku/STABLE_VERSION ${DOKKU_ROOT}/VERSION
```
josegonzalez: `$DOKKU_ROOT` is always `/home/dokku`, no?

contrib/dokku-installer.rb
jindrichskupa: Yes, this code doesnt work on systems with SysV-init, because there is no systemd directory, where to place dokku service. In case SysV-init initscript is installed in other cases systemd is used.

debian/preinst
jindrichskupa: I dont think so or please explain. This is same change as in dokku-installer.rb for service installation.

josegonzalez: What is the value of `${INIT}` on ubuntu 14.04 systems?

jindrichskupa: I expect nothing, its defined by init system detection:
+INIT=${INIT:-$([[ -h /sbin/init  ]] && echo -n systemd || echo -n sysvinit)}

jindrichskupa:
Jose thanks for review, but I will send new pull request with fixed used initscript for CI.

debian/preinst
josegonzalez: Don't we want to keep the init file as is then?

josegonzalez:
Can you just update this PR instead of sending a new pr?

jindrichskupa:
Done. I hope it will be better.

jindrichskupa:
Finally fixed - its passing bash codestyle.

contrib/dokku-installer.rb
josegonzalez: I don't want to merge this because the check on the path here changes how dokku works for ubuntu (you are switching to this new rbinit file) and it's unclear to me how that is better than the status quo.

jindrichskupa: Ok, I can explain - On ubuntu it works same, but on older debian or older ubuntu releases it switching from upstart/systemd to sysv-init. This code: File.realpath("/sbin/init") != "/sbin/init" detects sysvinit or upstart/systemd by checking if /sbin/init is symlink to systemd/upstart or it is regular file. Is it more clear? I dont push on you, I am ok with PR refuse. This PR solves problem on Debian Wheezy, where its impossible to install dokku without manual modifications.

josegonzalez: It's not a symlink on Ubuntu 14.04 which uses upstart, so the file that is templates out is your new init file. Is my reading of this logic correct?

debian/postinst
michaelshobbs: it's overridable in `dokku`

contrib/dokku-installer.rb
josegonzalez: Oh I see what is going on. You're trying to add support for older debian systems!

josegonzalez:
This PR isn't mergeable anymore - we removed the old ruby installer in favor of a python one - so do you mind re-implementing the changes against master? I think we should keep the current start/stop behavior on debian, but we can definitely use that new init file for older debian versions.

contrib/dokku-installer.rb
jindrichskupa: Right ! Ok, I will close this.

josegonzalez: Are you no longer interested in making this change?

jindrichskupa: Yes, for this time. I will correct it and re-work in future. I am planning
create a RHEL port, so I will make more changes in this way. Thanks.

2015-12-23 8:25 GMT+01:00 Jose Diaz-Gonzalez <notifications@github.com>:

> In contrib/dokku-installer.rb
> <https://github.com/dokku/dokku/pull/1636#discussion_r48325969>:
>
> > -  File.open("/etc/init/dokku-installer.conf", "w") do |f|
> > -    f.puts "start on runlevel [2345]"
> > -    f.puts "exec #{File.absolute_path(__FILE__)} selfdestruct"
> > -  end if File.directory?("/etc/init")
> > -  File.open("/etc/systemd/system/dokku-installer.service", "w") do |f|
> > -    f.puts "[Unit]"
> > -    f.puts "Description=Dokku web-installer"
> > -    f.puts ""
> > -    f.puts "[Service]"
> > -    f.puts "ExecStart=#{File.absolute_path(__FILE__)} selfdestruct"
> > -    f.puts ""
> > -    f.puts "[Install]"
> > -    f.puts "WantedBy=multi-user.target"
> > -    f.puts "WantedBy=graphical.target"
> > -  end if File.directory?("/etc/systemd/system")
> > +  if File.realpath("/sbin/init") != "/sbin/init" then
>
> Are you no longer interested in making this change?
>
> —
> <https://github.com/dokku/dokku/pull/1636/files#r48325969>.
>



-- 
Jindrich Skupa
Tel: +420602807773
E-mail: jindrich.skupa@gmail.com

josegonzalez: I'll pull in your other changes in a different pr so that adding sysv-init support is easy.

debian/postinst
josegonzalez: I added this change in #1800.

josegonzalez:
Feel free to open a new PR with your changes :)

jindrichskupa:
Ok, I will for sure. Thanks.

2015-12-23 10:00 GMT+01:00 Jose Diaz-Gonzalez <notifications@github.com>:

> Feel free to open a new PR with your changes :)
>
> —
> <https://github.com/dokku/dokku/pull/1636#issuecomment-166837037>.
>



-- 
Jindrich Skupa
Tel: +420602807773
E-mail: jindrich.skupa@gmail.com

PR #1637, opened 2015-11-01
Fix all the shellcheck

callahad:
Runs a single `find` command, rather than `find | xargs | egrep | egrep | awk | sed | grep | xargs`

Uses a slightly different heuristic for figuring out what to lint, but produces equivalent results. Here's a diff of the two methods, piped into `grep '^In '`:

```diff
diff --git a/result.txt b/result.txt
index 72e496d..f3f022d 100644
--- a/result.txt
+++ b/result.txt
@@ -10,42 +10,9 @@ In ./dokku line 98:
 In ./plugins/00_dokku-standard/commands line 4:
 In ./plugins/00_dokku-standard/commands line 5:
 In ./plugins/00_dokku-standard/commands line 6:
-In ./plugins/20_events/backup-check line 3:
-In ./plugins/20_events/backup-export line 3:
-In ./plugins/20_events/backup-import line 3:
-In ./plugins/20_events/bind-external-ip line 3:
-In ./plugins/20_events/check-deploy line 3:
 In ./plugins/20_events/commands line 4:
-In ./plugins/20_events/dependencies line 3:
-In ./plugins/20_events/docker-args-build line 3:
-In ./plugins/20_events/docker-args-deploy line 3:
-In ./plugins/20_events/docker-args-run line 3:
-In ./plugins/20_events/git-post-pull line 3:
-In ./plugins/20_events/git-pre-pull line 3:
 In ./plugins/20_events/hook line 3:
 In ./plugins/20_events/install line 3:
-In ./plugins/20_events/nginx-hostname line 3:
-In ./plugins/20_events/nginx-pre-reload line 3:
-In ./plugins/20_events/post-build-buildpack line 3:
-In ./plugins/20_events/post-build-dockerfile line 3:
-In ./plugins/20_events/post-delete line 3:
-In ./plugins/20_events/post-deploy line 3:
-In ./plugins/20_events/post-domains-update line 3:
-In ./plugins/20_events/post-release-buildpack line 3:
-In ./plugins/20_events/post-release-dockerfile line 3:
-In ./plugins/20_events/pre-build-buildpack line 3:
-In ./plugins/20_events/pre-build-dockerfile line 3:
-In ./plugins/20_events/pre-delete line 3:
-In ./plugins/20_events/pre-deploy line 3:
-In ./plugins/20_events/pre-receive-app line 3:
-In ./plugins/20_events/pre-release-buildpack line 3:
-In ./plugins/20_events/pre-release-dockerfile line 3:
-In ./plugins/20_events/receive-app line 3:
-In ./plugins/20_events/receive-branch line 3:
-In ./plugins/20_events/retire-container-failed line 3:
-In ./plugins/20_events/tags-create line 3:
-In ./plugins/20_events/tags-destroy line 3:
-In ./plugins/20_events/update line 3:
 In ./plugins/apps/commands line 34:
 In ./plugins/apps/commands line 4:
 In ./plugins/apps/post-delete line 3:
@@ -87,14 +54,10 @@ In ./plugins/config/commands line 4:
 In ./plugins/config/commands line 5:
 In ./plugins/config/docker-args-deploy line 3:
 In ./plugins/config/docker-args-deploy line 4:
-In ./plugins/config/docker-args-run line 3:
-In ./plugins/config/docker-args-run line 4:
 In ./plugins/config/functions line 3:
 In ./plugins/docker-options/commands line 31:
 In ./plugins/docker-options/commands line 4:
-In ./plugins/docker-options/docker-args-build line 27:
 In ./plugins/docker-options/docker-args-deploy line 27:
-In ./plugins/docker-options/docker-args-run line 27:
 In ./plugins/domains/commands line 4:
 In ./plugins/domains/commands line 5:
 In ./plugins/domains/commands line 6:
@@ -147,12 +110,12 @@ In ./plugins/ps/post-stop line 4:
 In ./plugins/ps/pre-deploy line 3:
 In ./plugins/ps/pre-deploy line 4:
 In ./plugins/shell/commands line 21:
-In ./plugins/shell/inputrc line 1:
 In ./plugins/tags/commands line 4:
 In ./plugins/tags/commands line 5:
 In ./plugins/tags/functions line 3:
 In ./plugins/tar/commands line 4:
 In ./plugins/tar/receive-app line 3:
+In ./tests/apps/clojure/check_deploy line 1:
 In ./tests/unit/test_helper.bash line 115:
 In ./tests/unit/test_helper.bash line 131:
 In ./tests/unit/test_helper.bash line 23:
```

Aside from `plugins/shell/inputrc`, which isn't actually a shell script, all of the removed files are symlinks to other scripts in the same directory, so no sense in linting them twice. Or thirty times, in the case of `plugsin/20_events/*`.


callahad:
Hm, so Linux doesn't like OS X's `-E`, and OS X doesn't have Linux's `-regextype`... one sec.

michaelshobbs:
Does this work on both Linux and Mac now?

callahad:
Yep! Works on Debian and OS X

michaelshobbs:
Can we assume all shellcode will exist in files marked as executable, though?

callahad:
I think that's a pretty reasonable policy. Only two scripts, `plugins/config/functions` and ` tests/unit/test_helper.bash` weren't already marked as such, but happy to defer to your judgment.

michaelshobbs:
I'm saying that we don't have that policy and I'm not sure we want it since we actually source a lot of stuff already. 

For example, there's actually no reason for the `functions` files to be executable.

callahad:
I think we *do* have that policy, though. All [hook invocations are guarded](https://github.com/progrium/plugn/blob/b5372ff92aefb74c7da7857f8ddaf1d35694cf03/bashenv/plugn.bash#L36) by `[[ -x "$plugin/$hook" ]] && $plugin/$hook "$@"`, for instance.

In fact, of all the shell scripts in the repo, the only ones that *do not* necessarily need to be executable are:

- The 8 `./plugins/*/functions` files
- The `./tests/unit/test_helper.bash` file

This approach is more robust than what's in master since it relies on intrinsic file properties rather than a complex pipeline to parse human-readable output from `file`, ensuring that it will work, even on machines that don't have their locale set to `"en_*"`. It also forks way fewer processes, so it runs negligibly more quickly. ;)

What I actually find compelling is that the single `find` command make the work to be done much more explicit, understandable, and maintainable.

In contrast to the current rule, this pull request correctly:

- Avoids duplicate work by not linting symlinks to the same file
- Does not skip linting `./tests/apps/.websocket.disabled/check_deploy` or `./tests/apps/clojure/check_deploy`

callahad:
If you'd like to drop `+x` from the functions files, we could instead pass `--external-sources` to ShellCheck which instructs it to follow `source` directives and lint the included files. We'd have to set `PLUGIN_AVAILABLE_PATH` and `PLUGIN_CORE_AVAILABLE_PATH` in ShellCheck's environment so it could properly track lines like `source "$PLUGIN_AVAILABLE_PATH/tags/functions"`, but that's reasonably easy to do.

michaelshobbs:
Oh yeah. To be clear, I prefer the single `find` command to the 'pipeline' I created. 😄

Just don't want to miss linting something in the future by changing this make target to assume something that's not necessarily explicit.

callahad:
Yep! The only tricky things are scripts that are only meant to be sourced. Fortunately, ShellCheck *will* attempt to follow `source` calls and lint the included files, which should make it impossible for things to fall through the cracks.

The only limitation is that ShellCheck requires a constant path, and all of ours use a variable like `$PLUGIN_AVAILABLE_PATH`. ShellCheck won't follow those by default, but it *will* warn about them ([SC1090](https://github.com/koalaman/shellcheck/wiki/SC1090)), and we can get it to follow them by providing a comment with the resolved path, like `"# shellcheck source=src/lib.sh`.

That *would* mean adding ~90 comment lines to the repo, but it also means that so long as we don't see `SC1090`, we know we're not missing anything.

callahad:
Ok, from testing, it looks like I may be slightly misunderstanding shellcheck's behavior, but it *will* warn for basically every conceivable issue arising from calling `source`, so we can make sure that those files are also passed to shellcheck.

callahad:
I just exploded the scope of this PR. I apologize. Each individual commit should be reasonably sized and scoped.

It looks like the version of shellcheck in the Ubuntu repo is a bit old (3.3). Version 4.1 adds quite a few additional lints. This PR fixes everything that 4.1 yells at and drops the executable bit from all of the `functions` files, which don't actually need to be executable.

If we source a file that isn't itself passed to shellcheck, shellcheck will complain ([SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091)). This ensures that we're not missing anything when linting.

Ironically, the `# shellcheck source=...` annotations required to fix [SC1090](https://github.com/koalaman/shellcheck/wiki/SC1090) in ShellCheck 4.x cause parse errors in ShellCheck 3.x. Joy. Dropping those for now.

callahad:
Well, doing this all at once was a bad idea. Closing this PR and replacing with smaller ones. :)

PR #1643, opened 2015-11-02
Fix generated nginx config when NO_VHOST=1

### LINKED ISSUE #1625: Missing server block in nginx conf when NO_VHOST=1
This is the only content in nginx conf generated by dokku:
```
upstream app_name {
  server 127.0.0.1:32773;
}
```

As per doc, what I understand is when NO_VHOST=1, dokku will still use nginx to proxy requests to dokker container.

Flink:
Fixes #1625


josegonzalez:
Should this include tests?

PR #1645, opened 2015-11-03
a git repo can be accessed via ssh

rvalyi:
say you want to do:
```
dokku plugin:install ssh://git@gitserver.example.com:port/path
```

currently this fails silently. But it works with my patch.


michaelshobbs:
👍🏼

u2mejc:
:+1: :shipit:

PR #1651, opened 2015-11-04
Resolve SC2029: 'this expands on client side'

callahad:
https://github.com/koalaman/shellcheck/wiki/SC2029

This patch addes inline directives where this is the desired behavior,
instead of globaly excluding the check.


callahad:
This CI failure looks like some intermittent host problem. There's literally no way this patch could have caused it. :(

PR #1652, opened 2015-11-04
Resolve SC2143: 'Instead of [ -n $(foo | grep bar) ], use foo | grep -q bar'

callahad:
...we actually don't violate this

https://github.com/koalaman/shellcheck/wiki/SC2143

### DIFF
--- a/tests.mk
+++ b/tests.mk
@@ -61,11 +61,10 @@ lint:
 	# these are disabled due to their expansive existence in the codebase. we should clean it up though
 	# SC2034: VAR appears unused - https://github.com/koalaman/shellcheck/wiki/SC2034
 	# SC2086: Double quote to prevent globbing and word splitting - https://github.com/koalaman/shellcheck/wiki/SC2086
-	# SC2143: Instead of [ -n $(foo | grep bar) ], use foo | grep -q bar - https://github.com/koalaman/shellcheck/wiki/SC2143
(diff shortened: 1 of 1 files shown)


callahad:
Bogus CI failure; the lint stage passed, and the only change in this PR is to the parameters passed to shellcheck. Looks like an issue with btrfs on the CI box.

callahad:
Actually, I'm seeing the btrfs failures on successful CI runs, so that's not what's causing things to blow up...

callahad:
So this is where it failed happened:

```
remote: -----> Building test-go-4254 from herokuish...
remote: WARNING: Your kernel does not support memory swappiness capabilities, memory swappiness discarded.
remote: -----> Adding BUILD_ENV to build environment...
remote: WARNING: Your kernel does not support memory swappiness capabilities, memory swappiness discarded.
remote: WARNING: Your kernel does not support memory swappiness capabilities, memory swappiness discarded.
remote: WARNING: Your kernel does not support memory swappiness capabilities, memory swappiness discarded.
remote: You cannot attach to a stopped container, start it first
To dokku@dokku.me:test-go-4254
error: failed to push some refs to 'dokku@dokku.me:test-go-4254'
+ failed git-push
+ echo '************ git-push failed ************'
************ git-push failed ************
```

If only we knew why the pre-receive hook failed... either way, it couldn't have been this patch :)

PR #1656, opened 2015-11-05
Resolve SC2001: 'See if you can use ${var//search/replace} instead'

callahad:
https://github.com/koalaman/shellcheck/wiki/SC2001


callahad:
```bash
$ cat sc2001.sh
#!/usr/bin/env bash

FOO="hello world world world"

A="$(echo "$FOO" | sed -e "s:world:dokku:g")"
B="${FOO//world/dokku}"

echo "$A"
echo "$B"

$ ./sc2001.sh
hello dokku dokku dokku
hello dokku dokku dokku
```

Issue #1654, opened 2015-11-05
Rely on distro-provided nginx


callahad:
## Resolution

- [ ] Pin the .deb requirements to `nginx (>= 1.4.6)`
- [ ] Remove the calls to `apt-add-repository` that set up the nginx PPA
- [ ] Ensure all included features still work
- [ ] Update the docs

As a bonus, once we stop using `apt-add-repository`, we can also

- [ ] drop `software-properties-common` and `python-software-properties` from package dependencies

## Rationale

(See discussion below for more flavor)

- The version in the LTS repo works fully for dokku's core needs.
- The LTS package is less volatile (it *only* receives security fixes), so the underlying platform is less likely to break during routine, unattended upgrades. E.g., there's no risk of things breaking overnight when a new release of nginx gets uploaded on the PPA.
- By policy, the LTS team will backport a security fix from 1.10 to 1.4 if they have to. The PPA is weill within its rights to just tell you to upgrade to 1.10.

## Discussion

(ported from #1629)

@callahad:

> I'd also request: "Works with distro-provided nginx." Right now, the nginx-vhosts/dependencies file tries to add a PPA and install nginx from it, which seems a bit extreme. I want the host system to be as bland and stable as possible. Newer software is easy enough to run inside my deployed containers. :)
> 
> As it stands, that would mean supporting: nginx 1.4.6 on Ubuntu 14.04 and 1.6.2 on Debian 8. From reading the changelogs, I don't think we depend on anything introduced in 1.6 or later, and there are no open CVEs in the distro-provided packages.

@michaelshobbs:

> @callahad the (current) primary install base is definitely Ubuntu and thus, we would need to be comfortable with nginx 1.4.6. IIRC, the move to PPA installation was to jump to 1.6.x from 1.4.x.

@callahad:

 > Are we actually depending on anything in 1.6, though? The PPA line was added [back in 2013](https://github.com/progrium/dokku/commit/de906a22aaf651d0cd78812ef823601e10403f00) with the intention of upgrading to 1.4 for websocket support. Now that 1.4 is in 14.04 LTS's repos, maybe we don't need the PPA?

@mmerickel: 

> The PPA is 1.8. LTS is pretty out of date.

@michaelshobbs:

> I think we don't have enough information to determine the impacts of downgrading nginx due to the ability to customize with a config template.
> 
> We've already been shipping with a PPA (as was noted) since 2013. What do we as a project gain by removing this?

@callahad: 

[...]

PR #1660, opened 2015-11-05
Fixup debian/control for Debian

callahad:
* Bump `Version` field.
* Set minimum nginx version to 1.4.6, per #1654
* Depend on both `rubygem-foo` (Ubuntu) and `ruby-foo` (Debian).
* Drop defunct `lxc-docker` package name
* Add distro-default `docker.io (>= 1.6.2)` as an alternative.

  Version 1.6.2 is in both `trusty-updates` and `jessie-backports`,
  while `trusty` itself has `0.9.1`. We seriously want to avoid that.

This also updates the docs so people don't forget to bump the version
field in the future.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,9 +1,12 @@
 Package: dokku
-Version: 0.3.12
-Section: base
+Version: 0.4.3
+Section: web
 Priority: optional
 Architecture: amd64
-Depends: locales, git, make, curl, software-properties-common, docker-engine | docker-engine-cs | lxc-docker, gcc, python-software-properties, man-db, herokuish, sshcommand
-Pre-Depends: nginx, dnsutils, ruby, ruby-dev, rubygem-rack, rubygem-rack-protection, rubygem-sinatra, rubygem-tilt, apparmor, cgroupfs-mount | cgroup-lite, plugn, sudo
+Depends: locales, git, make, curl, gcc, man-db, herokuish, sshcommand,
+ docker-engine-cs | docker-engine | lxc-docker (>= 1.6.2) | docker.io (>= 1.6.2),
+ software-properties-common, python-software-properties
+Pre-Depends: nginx (>= 1.4.6), dnsutils, apparmor, cgroupfs-mount | cgroup-lite, plugn, sudo,
+ ruby, ruby-dev, rubygem-rack , rubygem-rack-protection, rubygem-sinatra, rubygem-tilt
 Maintainer: Jose Diaz-Gonzalez <dokku@josediazgonzalez.com>
 Description: Docker powered mini-Heroku in around 100 lines of Bash
--- a/docs/development/release-process.md
+++ b/docs/development/release-process.md
@@ -10,6 +10,7 @@ To propose a release, the following tasks need to be performed:
 - The installable version must be changed in the `README.md` file.
 - The installable version must be changed in the `contrib/dokku-installer.rb` file.
 - The installable version must be changed in the `docs/template.html` file.
+- The installable version must be changed in the `debian/control` file.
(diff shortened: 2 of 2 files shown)


debian/control
```diff
@@ -1,9 +1,9 @@
 Package: dokku
-Version: 0.3.12
+Version: 0.4.3
```
josegonzalez: Note: this actually never *needs* to be bumped, as the deb.mk script uses sed to fix it in the package. But no need to change here :)

debian/control
```diff
@@ -1,9 +1,9 @@
...
 Section: base
 Priority: optional
 Architecture: amd64
-Depends: locales, git, make, curl, software-properties-common, docker-engine | docker-engine-cs | lxc-docker, gcc, python-software-properties, man-db, herokuish, sshcommand
-Pre-Depends: nginx, dnsutils, ruby, ruby-dev, rubygem-rack, rubygem-rack-protection, rubygem-sinatra, rubygem-tilt, apparmor, cgroupfs-mount | cgroup-lite, plugn, sudo
+Depends: locales, git, make, curl, software-properties-common, docker-engine-cs | docker-engine | docker.io (>= 1.6.2), gcc, python-software-properties, man-db, herokuish, sshcommand
+Pre-Depends: nginx (>= 1.4.6), dnsutils, ruby, ruby-dev, rubygem-rack | ruby-rack, rubygem-rack-protection | ruby-rack-protection, rubygem-sinatra | ruby-sinatra, rubygem-tilt | ruby-tilt, apparmor, cgroupfs-mount | cgroup-lite, plugn, sudo
```
josegonzalez: Good catch.

debian/control
```diff
@@ -1,9 +1,9 @@
 Package: dokku
-Version: 0.3.12
+Version: 0.4.3
 Section: base
 Priority: optional
 Architecture: amd64
-Depends: locales, git, make, curl, software-properties-common, docker-engine | docker-engine-cs | lxc-docker, gcc, python-software-properties, man-db, herokuish, sshcommand
```
josegonzalez: Can you re-add `lxc-docker`? It's just 1.6.2, so if we make it at least that version, then we should be fine, right? I'd rather not break systems that haven't switched their remote just yet.

debian/control
callahad: Huh, so it does. There's a part of me that would like to move a bit more to standard .deb build tools rather than the ad-hoc stuff in deb.mk, but I'm still learning those bits myself :)

callahad:
Actually, hang on... it looks like Ubuntu also uses, say, `ruby-sinatra` instead of `rubygem-sinatra`, and that the `rubygem-*.deb` files are coming from `deb.mk` which uses `fpm`.

Any reason we can't just use the ruby / rack / sinatra that ship with the distros? They may be old, but they're only used on initial configuration, so that should be fine.

debian/control
josegonzalez: `dpkg-deb` is pretty standard, no? :D

josegonzalez:
TBH I would rather we use the same package for now on both systems, and in the future switch to using a python virtualenv app (or even a golang binary) for the initial setup.

debian/control
callahad: I have no idea what I'm doing

callahad:
Be still, my beating heart. I'd love to do this in Python. Will update PR to leave the Ruby parts alone for now.

josegonzalez:
*bump* where is my update :)

callahad:
There, you happy now? ;-p

callahad:
(I did rearrange things so that, e.g., all of the Ruby packages are on the same line. This should make for better diffs in the future, at the cost of a slightly messier diff now)

Issue #1617, opened 2015-10-27
Error Upgrading to 0.4.3


mlavin:
I installed dokku through the apt repo and trying to run `apt-get upgrade` to the latest version is running into an error:
```
...
User 'dokku' already exists
docker:x:999:dokku
gpasswd: user 'syslog
mark' does not exist
dpkg: error processing package dokku (--configure):
 subprocess installed post-installation script returned error exit status 3
Errors were encountered while processing:
 dokku
E: Sub-process /usr/bin/dpkg returned an error code (1)
```
From what I can tell this comes from the `nginx-vhosts/install`. In particular the change from f9e636ddb1c7b2442b1a7ae3cea7ce7aa49ed121 added the call to `gpasswd`. As it would happen my server has a group named `admin` which my user belongs to so `egrep ^adm /etc/group | awk -F ":" '{ print $4 }'` returns two lines 

```
syslog
mark
```
and triggers the error. It would seem that changing this regex to `^adm:` would match only the `adm` group rather than all groups which begin with `adm` but my Debian packaging skills are admittedly sub-par.

michaelshobbs:
Whoops. I'll patch that.

callahad:
This should be fixed, since the above PR has been merged. Should we close it?

starding:
Get into the same trouble when I installed dokku 0.4.7.

ubuntu version:
VERSION="14.04.3 LTS, Trusty Tahr"

dokku version:
0.4.7

error details:

...
User 'dokku' already exists
docker:x:999:yangchen,starding,yudong,dokku
/var/lib/dokku/plugins/enabled/00_dokku-standard/commands
/var/lib/dokku/plugins/enabled/20_events/commands
/var/lib/dokku/plugins/enabled/apps/commands
/var/lib/dokku/plugins/enabled/backup/commands
/var/lib/dokku/plugins/enabled/certs/commands
/var/lib/dokku/plugins/enabled/config/commands
/var/lib/dokku/plugins/enabled/docker-options/commands
/var/lib/dokku/plugins/enabled/domains/commands
/var/lib/dokku/plugins/enabled/enter/commands
/var/lib/dokku/plugins/enabled/git/commands
/var/lib/dokku/plugins/enabled/logs/commands
/var/lib/dokku/plugins/enabled/nginx-vhosts/commands
/var/lib/dokku/plugins/enabled/plugin/commands
gpasswd: user 'syslog
developer' does not exist
dpkg: error processing package dokku (--configure):
[...]

PR #1662, opened 2015-11-05
Only return users of group 'adm'

xadh00m:
This fixes the issue where all groups starting with 'adm' are returned which broke the installation.


xadh00m:
Fixes #1617

mlavin:
:sparkling_heart: thank you! :sparkling_heart:

PR #1664, opened 2015-11-06
add $REV to pre-receive-app call

michaelshobbs:
I have a modified dokku-slack fork that inspects the incoming rev to extract the commit author and post it to slack in the `post-deploy` phase. This change facilities an easy way to get at the incoming revision. Seems like it would be useful for other applications as well.


josegonzalez:
You need to also update the docs for this hook.

michaelshobbs:
Oh yea. Good call. Hold please.

PR #1665, opened 2015-11-06
Replace curl with wget; see #1567

callahad:
Let's see if this passes CI...

(it should, but there are some subtleties in curl vs wget that I'm going to double-check while CI runs)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,11 +99,10 @@ plugn:
 	tar xzf /tmp/plugn_latest.tgz -C /usr/local/bin
 
 docker: aufs
-	apt-get install -qq -y curl
 	egrep -i "^docker" /etc/group || groupadd docker
 	usermod -aG docker dokku
 ifndef CI
-	curl -sSL https://get.docker.com/ | sh
+	wget -nv -O - https://get.docker.com/ | sh
 ifdef DOCKER_VERSION
 	apt-get install -qq -y docker-engine=${DOCKER_VERSION} || (apt-cache madison docker-engine ; exit 1)
 endif
@@ -123,7 +122,7 @@ ifdef BUILD_STACK
 else
 ifeq ($(shell echo ${PREBUILT_STACK_URL} | egrep -q 'http.*://|file://' && echo $$?),0)
 	@echo "Start importing herokuish from ${PREBUILT_STACK_URL}"
-	docker images | grep gliderlabs/herokuish || curl --silent -L ${PREBUILT_STACK_URL} | gunzip -cd | docker import - gliderlabs/herokuish
+	docker images | grep gliderlabs/herokuish || wget -nv -O - ${PREBUILT_STACK_URL} | gunzip -cd | docker import - gliderlabs/herokuish
 else
 	@echo "Start pulling herokuish from ${PREBUILT_STACK_URL}"
 	docker images | grep gliderlabs/herokuish || docker pull ${PREBUILT_STACK_URL}
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -7,7 +7,7 @@
 
 # We wrap this whole script in a function, so that we won't execute
 # until the entire script is downloaded.
-# That's good because it prevents our output overlapping with curl's.
+# That's good because it prevents our output overlapping with wget's.
 # It also means that we can't run a partially downloaded script.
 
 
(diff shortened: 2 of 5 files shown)


josegonzalez:
Nice! I'm in the middle of making a release, but this will go into 0.4.5 :)

callahad:
Heh. You're presuming it won't Break Everything. Waiting for 0.4.5 is a good idea :)

docs/troubleshooting.md
```diff
@@ -138,12 +138,12 @@ If you see output similar this when deploying:
  !     tar: Error is not recoverable: exiting now
 ```
 
-it might that the curl command that is supposed to fetch the buildpack (anything in the low megabyte file size range) takes too long to finish, due to slowish connection.  To overwrite the default values (connection timeout: 3 seconds, total maximum time for operation: 30 seconds), edit `/home/dokku/ENV` like the following:
+it might that the command that is supposed to fetch the buildpack (anything in the low megabyte file size range) takes too long to finish, due to slowish connection.  To overwrite the default values (connection timeout: 3 seconds, total maximum time for operation: 30 seconds), edit `/home/dokku/ENV` like the following:
```
josegonzalez: The english here was always solid :P

plugins/build-env/pre-build-buildpack
```diff
@@ -6,8 +6,8 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
 APP="$1"; IMAGE=$(get_app_image_name $APP); BUILD_ENV=""
 verify_app_name "$APP"
 
-[[ -z $(config_get --global CURL_CONNECT_TIMEOUT) ]] && config_set --global CURL_CONNECT_TIMEOUT=5
-[[ -z $(config_get --global CURL_TIMEOUT) ]] && config_set --global CURL_TIMEOUT=30
+[[ -z $(config_get --global WGET_CONNECT_TIMEOUT) ]] && config_set --global WGET_CONNECT_TIMEOUT=5
+[[ -z $(config_get --global WGET_TIMEOUT) ]] && config_set --global WGET_TIMEOUT=30
```
josegonzalez: These are internal environment variables for the buildpacks, so they should be reverted.

callahad:
So, one potential issue with this is that both `wget --quiet` and `curl --silent` are both completely silent, and we still want to see log entries when something fails.

Curl has `--show-error`, to still print when something blows up, but wget doesn't have anything like it. The closest wget has is `--no-verbose`, which prints a single status line for all requests, in addition to any errors:

```
# curl -sS is silent, but still prints errors
$ curl -sSL 'http://httpbin.org/status/200'
$ curl -sSL 'http://httpbin.org/status/404'
$ curl -sSL 'http://httpbin.org/status/500'
$ curl -sSL 'http://example.invalid'
curl: (6) Could not resolve host: example.invalid

# wget -q is completely silent
$ wget -qO- 'http://httpbin.org/status/200'
$ wget -qO- 'http://httpbin.org/status/404'
$ wget -qO- 'http://httpbin.org/status/500'
$ wget -qO- 'http://example.invalid'
$ wget -qO- 'http://example.invalid'

# wget -nv is almost silent, but prints on every request
$ wget -nv -O- 'http://httpbin.org/status/200'
2015-11-06 08:17:57 URL:http://httpbin.org/status/200 [0/0] -> "-" [1]
$ wget -nv -O- 'http://httpbin.org/status/400'
http://httpbin.org/status/404:
2015-11-06 08:18:00 ERROR 404: NOT FOUND.
$ wget -nv -O- 'http://httpbin.org/status/500'
http://httpbin.org/status/500:
2015-11-06 08:18:05 ERROR 500: INTERNAL SERVER ERROR.
$ wget -nv -O- 'http://example.invalid'
wget: unable to resolve host address 'example.invalid'
```

We can either:

- lose printing on errors with `wget -q`
- get a print on every request with `wget -nv`
- try to emulate curl's behavior with `wget -nv ... &2>1 | grep '^wget: '`

callahad:
And, if we need to be able to pass the output along, we can send stderr into a subshell, and stdout wherever

```bash
$ wget -nv -O - 'http://example.invalid' 2> >(grep '^wget: ') > /dev/null
wget: unable to resolve host address 'example.invalid'
```

josegonzalez:
I do a trick where I redirect all output to a temporary file, and if there is an error, I dump the contents of that file. See [here](https://github.com/seatgeek/bash-aptfile/blob/master/bin/aptfile#L130) and [here](https://github.com/seatgeek/bash-aptfile/blob/master/bin/aptfile#L115).

callahad:
Yeah, I need to revisit things, especially around what output goes where in the bats tests.

callahad:
Dropping this PR. We *could* probably make wget to work everywhere, but curl is *way* more developer friendly for us, especially its `--show-error` flag. Instead of excising it completely, let's just avoid using curl during bootstrapping, and keep the .deb package dependency on curl, so that we can use it once Dokku is installed.

Issue #1658, opened 2015-11-05
dokku certs:add only supports tarball


hhff:
`root@server:~# dokku certs:add production server.crt server.key`

outputs:

`Tar archive containing server.crt and server.key expected on stdin`

It should take those files separately, right?

josegonzalez:
Interesting. Are the `server.crt` and `server.key` in the directory where you ran that command?

hhff:
Yessir - but in my case they're actually named `example_com.crt` `example_com.key`

u2mejc:
ok to close?

hhff:
thanks dudes!

PR #1667, opened 2015-11-07
Fix dokku certs:add file input bug

u2mejc:
- fixes issue #1658
- add better output for argument files missing
- don't delete (move) the certs passed in

### DIFF
--- a/plugins/certs/commands
+++ b/plugins/certs/commands
@@ -11,11 +11,17 @@ is_tar_import() {
 }
 
 is_file_import() {
-  local CRT_FILE="$3"
-  local KEY_FILE="$4"
+  local CRT_FILE="$1"
+  local KEY_FILE="$2"
 
-  if [[ -f $CRT_FILE ]] && [[ -f $KEY_FILE ]]; then
-    return 0
+  if [[ $CRT_FILE ]] && [[ $KEY_FILE ]]; then
+    if [[ ! -f $CRT_FILE ]]; then
+      dokku_log_fail "CRT file specified not found, please check file paths"
+    elif [[ ! -f $KEY_FILE ]]; then
+      dokku_log_fail "KEY file specified not found, please check file paths"
+    else
+      return 0
+    fi
   fi
 
   return 1
@@ -55,8 +61,8 @@ certs_set() {
   fi
 
   mkdir -p "$DOKKU_ROOT/$APP/tls"
-  mv "$CRT_FILE" "$DOKKU_ROOT/$APP/tls/server.crt"
-  mv "$KEY_FILE" "$DOKKU_ROOT/$APP/tls/server.key"
+  cp "$CRT_FILE" "$DOKKU_ROOT/$APP/tls/server.crt"
+  cp "$KEY_FILE" "$DOKKU_ROOT/$APP/tls/server.key"
   cd $DOKKU_ROOT
   rm -rf $TEMP_DIR
   nginx_build_config $APP


plugins/certs/commands
```diff
@@ -55,8 +61,8 @@ certs_set() {
   fi
 
   mkdir -p "$DOKKU_ROOT/$APP/tls"
-  mv "$CRT_FILE" "$DOKKU_ROOT/$APP/tls/server.crt"
-  mv "$KEY_FILE" "$DOKKU_ROOT/$APP/tls/server.key"
+  cp "$CRT_FILE" "$DOKKU_ROOT/$APP/tls/server.crt"
+  cp "$KEY_FILE" "$DOKKU_ROOT/$APP/tls/server.key"
```
michaelshobbs: i suppose it doesn't matter as `$TEMP_DIR` get removed but why change to `cp`?

michaelshobbs:
I'm not certain this fixes #1658. This plugin could definitely use some tests anyway, though. Care to take a stab at that and include the case from #1658? You can use the test certs in `tests/server_ssl*.tar`.

plugins/certs/commands
u2mejc: @michaelshobbs no you tell me, why you delete my certs bro? That was my only copy!!! :wink:

michaelshobbs:
The way I read it, we are working in a temp folder and not on the input. Maybe I misread it?

u2mejc:
ah sorry I understand your confusion now, github isn't showing you the whole file. The temp dir is for the tar file and unrelated to passing in file arguments from the command line. 

Here's an example:
dokku certs:add ruby-sample /home/myuser/server.crt /home/myuser/server.key
CRT_FILE=$3 

If you *mv* $CRT_FILE, your literally deleting the user's /home/myuser/server.crt. The verbiage add doesn't imply delete to the user and I can't think of a single program I would expect to delete my key and force me to retrieve a new copy from the CA. Does heroku delete the keys from your laptop when it copies to the server? I can see where that might be confusing.

michaelshobbs:
No I think the Heroku client just copies them up to the paas. I figured I just misread it. Still would be great to get at least one test written for each input type. i.e. file and tarball. 

Care to take a stab at that?

u2mejc:
It's 5:59 on friday, happy to! :+1:

michaelshobbs:
I definitely did not mean right this moment sir. 😄️

hhff:
thanks duuuuuudes

josegonzalez:
@u2mejc *bump* I also don't see how it solves the bug.

u2mejc:
Bummer, @michaelshobbs @josegonzalez when you cloned and tested it, what was the results of your tests?

josegonzalez:
I didn't try it, I'm just saying it doesn't *look* like it does.

plugins/certs/commands
```diff
@@ -11,11 +11,17 @@ is_tar_import() {
 }
 
 is_file_import() {
-  local CRT_FILE="$3"
-  local KEY_FILE="$4"
-
-  if [[ -f $CRT_FILE ]] && [[ -f $KEY_FILE ]]; then
-    return 0
+  local CRT_FILE="$1"
+  local KEY_FILE="$2"
```
michaelshobbs: i see the fix now!!!

michaelshobbs:
@josegonzalez no tests!! why u merge bro!? :smile:

u2mejc:
Sorry I didn't have a chance to respond sooner! I'll work on tests and submit a separate PR.

josegonzalez:
@michaelshobbs I'm too trigger happy, sorry :(

josegonzalez:
I saw you say you understood it so I was like LG2ME

michaelshobbs:
I was just so excited I finally saw the feex 😜️

PR #1670, opened 2015-11-08
Add apps rename

zachfeldman:
This adds the

`dokku apps:rename <old-app> <new-app>`

function to dokku.

I honestly wasn't sure how to run the tests but I'm reasonable confident mine works although I'm sure we could also test that the new name is there :)

### DIFF
--- a/docs/application-deployment.md
+++ b/docs/application-deployment.md
@@ -100,6 +100,16 @@ This will prompt you to verify the application's name before destroying it. You 
 dokku --force apps:destroy APP
 ```
 
+### Renaming a deployed app
+
+You can rename a deployed app using the `apps:rename` CLI tool:
+
+```shell
+dokku apps:rename OLD_NAME NEW_NAME
+```
+
+This will copy all of your app's contents into a new app directory with the name of your choice, delete your old app, then rebuild the new version of the app and deploy it. All of your config variables, including database urls, will be preserved.
+
 ### Adding deploy users
 
 While it is possible to use password-based authorization to push to Dokku, it is preferable to use key-based authentication for security. You can add your public key to the dokku user's `authorized_keys` file with the following command:
--- a/tests/unit/apps.bats
+++ b/tests/unit/apps.bats
@@ -26,3 +26,23 @@ load test_helper
   echo "status: "$status
   assert_success
 }
+
+@test "(apps) apps:rename" {
+  deploy_app
+  run bash -c "dokku apps:rename $TEST_APP great-test-name"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run bash -c "dokku apps | grep $TEST_APP"
+  echo "output: "$output
+  echo "status: "$status
+  assert_output ""
+  run bash -c "curl --silent --write-out '%{http_code}\n' `dokku url great-test-name` | grep 404"
+  echo "output: "$output
+  echo "status: "$status
+  assert_output ""
+  run bash -c "dokku --force apps:destroy great-test-name"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+}
--- a/plugins/apps/commands
+++ b/plugins/apps/commands
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-[[ " apps apps:create apps:destroy help apps:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
+[[ " apps apps:create apps:rename apps:destroy help apps:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/apps/functions"
@@ -13,7 +13,25 @@ case "$1" in
   apps:create)
     apps_create $2
     ;;
+  
+  apps:rename)
+    [[ -z $2 ]] && dokku_log_fail "Please specify an app to run the command on"
+    [[ -d "$DOKKU_ROOT/$3" ]] && dokku_log_fail "Name is already taken"
+    OLD_APP="$2"
+    NEW_APP="$3"
 
+    mkdir -p "$DOKKU_ROOT/$NEW_APP"
+    docker run --rm -v "$DOKKU_ROOT/$OLD_APP/cache:/cache" "dokku/$OLD_APP" chmod 777 -R /cache
+    rm -rf "$DOKKU_ROOT/$OLD_APP/cache"
+    cp -a "$DOKKU_ROOT/$OLD_APP/." "$DOKKU_ROOT/$NEW_APP"
+    dokku apps:destroy $OLD_APP --force
+    sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/URLS"
+    sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/VHOST"
+    dokku ps:rebuild $NEW_APP
+    plugn trigger post-app-rename $OLD_APP $NEW_APP
+    echo "Renaming $OLD_APP to $NEW_APP... done"
+    ;;
+ 
   apps:destroy)
     [[ -z $2 ]] && dokku_log_fail "Please specify an app to run the command on"
     [[ "$2" == "tls" ]] && dokku_log_fail "Unable to destroy tls directory"
@@ -52,6 +70,7 @@ case "$1" in
     apps, List your apps
     apps:create <app>, Create a new app
     apps:destroy <app>, Permanently destroy an app
+    apps:rename <old-app> <new-app>, Rename an app
 EOF
     ;;
 


docs/application-deployment.md
```diff
@@ -100,6 +100,16 @@ This will prompt you to verify the application's name before destroying it. You
...
+### Renaming a deployed app
+
+You can rename a deployed app using the `apps:rename` CLI tool:
+
+```shell
+dokku apps:rename OLD_NAME NEW_NAME
+```
+
+This will copy all of your app's contents into a new app directory with the name of your choice, delete your old app, then rebuild the new version of the app and deploy it. All of your config variables, including database urls, will be preserved.
+
```
josegonzalez: Please note that links may not persist.

zachfeldman: Which links?

josegonzalez: From other containers. For instance, linking one of the official datastores to the old application. I don't think the link will persist, and this should be documented.

zachfeldman: It actually does persist because as noted, the config variables are the same. It just has the old name.

Should I just note that, for instance, if you link a database through a config variable and used the postgresql plugin, your link will still work but your database instance will have the old app name?

zachfeldman:
image

First issue when trying to rename an app this morning, involving permissions. Any ideas @josegonzalez or anyone else?

zachfeldman:
(Btw, I tried running it with sudo when it didn't work before...I know that's not a good way to test and it had no effect)

josegonzalez:
We should probably *not* copy app cache over.

docs/application-deployment.md
mmerickel: Probably an event should be emitted for plugins to handle updating their metadata otherwise this will always be half-broken with regards to plugin config. For example dokku-postgres can tell you which database is linked with which containers and will need a way to update that.

zachfeldman:
The ways to "limit" a `cp` command are pretty annoying and complicated.

would there be major issues if we just `rm -rf ./cache` before copy? I'd assume this directory gets rebuilt when an app gets built anyway, right?

josegonzalez:
You could try using rsync.

I think you'd need to delete the cache from inside the container running that app.

zachfeldman:
@josegonzalez I didn't want to have another dependency though - are we positive the `rsync` command is installed on all possible Dokku machines? 

Sorry if that's the case and I'm just being a n00b!

josegonzalez:
Yeah you are right, they are probably not installed.

[Here](https://github.com/dokku/dokku-postgres/blob/master/commands#L88) is what the plugins do to delete data, we can probably do the same for the cache directory.

zachfeldman:
Seems like I don't have permission to delete the cache directory, I tried with a simple:

`sudo rm -rf "$DOKKU_ROOT/$OLD_APP/cache"`

zachfeldman:
But obviously you can't login as sudo with `dokku` user. Ideas?

josegonzalez:
You can start up a container and delete the cache within that container, no?

zachfeldman:
Ok, cache directory problem is fixed! Took me some time to understand how Docker handles mounting directories, permissions, etc.

I think this is basically ready to merge if all the tests pass.

josegonzalez:
The tests are the last thing that should be taken care of before merging. Since I personally stay away from testing as much as possible - my code is great! - I think this will need either more tests from you or a maintainer as a "sponsor" for the testing.

josegonzalez:
Looks good otherwise, great work!

zachfeldman:
Sure, let me know what tests you think are missing? I thought the test I added covered success of all commands and that would be good

josegonzalez:
I'd say that we want to ensure the new app is running and the old app does not exist.

zachfeldman:
Ok sure, I can definitely write some code for that but I noticed that the `apps:destroy` test doesn't ensure the old app is deleted, nor does the `apps:create` test ensure that an app was created, just that the command ran successfully, unless I'm reading the tests wrong.

I'm aware my latest push broke the same type of test for my `apps:rename`. But I guess my point is, if that level of testing is good enough for those other commands, why is it not good enough for mine?

zachfeldman:
Ok, it looks like my test failed because the "my-cool-guy-test-app" that gets created with the test helpers isn't a, "real" app, so there's not images to choose from in `docker images` when we try to delete the cache directory.

I'm fine with maybe adding more depth to this command by creating a more, "real" sample app, but are there any guidelines to what kind of app I should put in this test app container?

michaelshobbs:
Yeah `create_app` creates a blank app but `deploy_app` will actually give you a running app.

zachfeldman:
So I tried adding deploy_app but it's taking forever now to get through that test, sometimes it doesn't even get through at all...I'll keep on plugging though!

zachfeldman:
Does anyone have a better way of running the whole test suite than on an active Dokku instance ? I can't seem to get past deploying a "real" app to test this function...looking for some guidance here. Thanks!

michaelshobbs:
I'm not sure I understand what you mean by "an active Dokku instance." We have [this documentation](http://progrium.viewdocs.io/dokku/development/testing/) about running tests locally.

tests/unit/apps.bats
```diff
@@ -26,3 +26,11 @@ load test_helper
   echo "status: "$status
   assert_success
 }
```
u2mejc: ^ Here, need to destroy_app before you create it again below.

michaelshobbs: @u2mejc that test does an `apps:destroy` though

zachfeldman:
Let's see if this works on circleci...

meanwhile, it seems that we'll also need to change the URLS and VHOST file, see:
https://github.com/progrium/dokku/issues/1716

Is there any issue with simply doing a string replace on these files? Just want to make sure that doesn't break something important or there aren't any functions that already take care of these files.

zachfeldman:
Seems as if `deploy_app` times out each time...any ideas?

tests/unit/apps.bats
```diff
@@ -26,3 +26,12 @@ load test_helper
...
+
+@test "(apps) apps:rename" {
+  deploy_app
+  run bash -c "dokku apps:rename $TEST_APP great-test-name"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run bash -c "dokku apps:destroy great-test-name"
```
michaelshobbs: you'll want to pass in `--force`. it's hanging on confirmation of `apps:destroy`

michaelshobbs: I think I'd like to see this block like this if that's cool with you @zachfeldman 
```
@test "(apps) apps:rename" {
  deploy_app
  run bash -c "dokku apps:rename $TEST_APP great-test-name"
  echo "output: "$output
  echo "status: "$status
  assert_success
  run bash -c "dokku --force apps:destroy great-test-name"
  echo "output: "$output
  echo "status: "$status
  assert_success
}
```

zachfeldman:
Ok, so now I'm changing the app URL and other tests are failing - due to 404s. Seems related! Any ideas here?

zachfeldman:
Also thanks @michaelshobbs for the suggestion to force destroy!

zachfeldman:
Some relevant error messages:

````
WARNING: Your kernel does not support memory swappiness capabilities, memory swappiness discarded.
Error deleting container: Error response from daemon: Cannot destroy container a1f5731b7f552da3f7c6da94978083658f07685cd913082b69a87dcb658988d4: Driver btrfs failed to remove root filesystem a1f5731b7f552da3f7c6da94978083658f07685cd913082b69a87dcb658988d4: Failed to destroy btrfs snapshot: operation not permitted
Error response from daemon: Cannot destroy container 0e4ba7cf84176557370d51a5264208d83de3a451c98aa64180a02b616b1cd0be: Driver btrfs failed to remove root filesystem 0e4ba7cf84176557370d51a5264208d83de3a451c98aa64180a02b616b1cd0be: Failed to destroy btrfs snapshot: operation not permitted
Error: failed to remove containers: [0e4ba7cf84176557370d51a5264208d83de3a451c98aa64180a02b616b1cd0be]
````

**Comment by michaelshobbs** [comment]
The real error is this:
```
remote: /bin/sh: 1: cd: can't cd to ../pcre-8.36
remote: make[1]: *** [../pcre-8.36/Makefile] Error 2
remote:        make[1]: Leaving directory `/cache/nginx-1.6.2'
remote: make: *** [build] Error 2
remote: cp: cannot stat '/tmp/build/nginx/*': No such file or directory
```

#1715 fixes this
rebase from master and you should be good

**Comment by zachfeldman** [comment]
You da bomb hobbs @michaelshobbs . Trying it now!

**Comment by josegonzalez** [comment]
Mind squashing before I merge?

**Comment by michaelshobbs** [comment]
I have some thoughts/questions on testing and datasource plugins

- can we explicitly test that the old app name and url are no longer in use?
- can we explicitly test that the new url responds?
- @josegonzalez should we trigger `post-app-renamed` or something for (datastore or other) plugins to do their thing? not sure what implications this has on those types of plugins

**Comment by josegonzalez** [comment]
All valid points, and yeah, a trigger would be good. May make sense to do `post-app-` as a prefix for these types of triggers from now on (same for `create`/`delete`). We can deprecate any old ones in 0.5.x.

**Comment by zachfeldman** [comment]
Yay passing tests!

Let me look into these other issues, will report back soon

**Comment by zachfeldman** [comment]
...if anyone has any code samples/tips for triggers, that would be helpful

**Comment by josegonzalez** [comment]
```shell
plugn trigger post-app-rename $OLD_NAME $NEW_NAME
```

zachfeldman:
Added additional tests. It's a bit hard to test that a URL is not in use, but I have tested that the old app name is not in the results of `dokku apps` and the new url doesn't return a 404. I also added the plugin trigger.

Tests passed but I rebased and now they're running again!

michaelshobbs:
@zachfeldman thanks for sticking with this man. Much appreciated. For reference on handling the URL stuff check out https://github.com/progrium/dokku/blob/v0.4.5/tests/unit/nginx-vhosts.bats

michaelshobbs:
EDIT 2: Forgot to highlight the line numbers: https://github.com/progrium/dokku/blob/v0.4.5/tests/unit/nginx-vhosts.bats#L48-L54

zachfeldman:
Hey @michaelshobbs and @josegonzalez sorry for the delay on finishing this up. I took another look and I'm a bit confused why my pull request hasn't been merged yet, I have more tests in it than any of the other apps: commands. Regarding testing to make sure an app url is not in use, I'm not sure if this is possible because it seems that if an app url is not present, Dokku will render the "default" app anyway. So you'll never receive a 404 for a URL that is no longer in use. @michaelshobbs thanks for the link but this is testing that a 200 is returned so I'm not sure if it's helpful. Let me know where you'd like to go from here!

josegonzalez:
Looks good to me then, I'll merge.

zachfeldman:
Thanks @josegonzalez and @michaelshobbs for all of your help on this! Very exciting!

PR #1671, opened 2015-11-08
User Auth plugin trigger

josegonzalez:
I came up with a proof of concept for limiting commands to a given user. I'm not convinced this is the absolute best way, but I do think that it's the most flexible approach given how dokku is structured.

### DIFF
--- a/dokku
+++ b/dokku
@@ -51,7 +51,7 @@ fi
 
 if [[ $(id -un) != "dokku" && $1 != plugin:*install* && $1 != "plugin:update" ]]; then
   sudo -u dokku -E -H $0 "$@"
-  exit
+  exit $?
 fi
 
 if [[ -n "$SSH_ORIGINAL_COMMAND" ]]; then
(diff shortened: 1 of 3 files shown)


josegonzalez:
@michaelshobbs thoughts on this?

dokku
```diff
@@ -50,8 +50,9 @@ fi
 ! has_tty && DOKKU_QUIET_OUTPUT=1
 
 if [[ $(id -un) != "dokku" && $1 != plugin:*install* && $1 != "plugin:update" ]]; then
+  export SSH_USER=$(id -un)
   sudo -u dokku -E -H $0 "$@"
-  exit
+  exit $?
```
michaelshobbs: is this change to `exit` pertinent to this change? (just curious)

michaelshobbs:
seems fine for a first pass

dokku
josegonzalez: Nope, just wanted to change it while i was there.

michaelshobbs: :+1:

PR #1682, opened 2015-11-11
set nullglob when looking for PORT files

### LINKED ISSUE #1678: 00_dokku-standard/commands: line 171: /home/dokku/<app>/PORT.*: No such file or directory
I'm getting this after upgrading from 0.3.x to 0.4.4 and after deploy:

```
/var/lib/dokku/plugins/enabled/00_dokku-standard/commands: line 171: /home/dokku/<app>/PORT.*: No such file or directory
```

I see that I have `NO_VOST=1` set so the command is doing the right thing. Question is how do I get this PORT file that it is looking for? This app has no external port that needs to be exposed.

michaelshobbs:
When deploying an app without a web proc **and** `$NO_VHOST` set, dokku will not generate a file that matches [this glob](https://github.com/progrium/dokku/blob/v0.4.4/plugins/00_dokku-standard/commands#L170) and thus will yield error output as indicated in #1678. 

It's also probably good code hygiene to guard nullglobs like this anyway. We do it almost everywhere else except for `backup-export/import` triggers but we know we are going to sort those out soon anyway.

NOTE: This is not a fatal error and does not cause dokku to fail deployment. 

closes #1678

### DIFF
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -167,9 +167,11 @@ case "$1" in
     fi
 
     if [[ "$(is_app_vhost_enabled $APP)" == "false" ]]; then
+      shopt -s nullglob
       for PORT_FILE in $DOKKU_ROOT/$APP/PORT.*; do
         echo "http://$(< "$DOKKU_ROOT/HOSTNAME"):$(< "$PORT_FILE") (container)"
       done
+      shopt -u nullglob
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Whats the original bug?

michaelshobbs:
@josegonzalez updated the PR description

josegonzalez:
Got it, seems legit :)

PR #1683, opened 2015-11-12
Add bats test for certs plugin

u2mejc:
Add tests for cert plugin, would catch this: #1658

Related PR discussion: https://github.com/progrium/dokku/pull/1667


hhff:
:+1:

PR #1684, opened 2015-11-12
Cause certs:remove to return non zero on error

u2mejc:
This was caught when writing the certs.bats. This patch will return
non zero on error instead of zero which can casue false negatives
in testing.

### DIFF
--- a/plugins/certs/commands
+++ b/plugins/certs/commands
@@ -154,7 +154,7 @@ case "$1" in
       rm -rf $APP_SSL_PATH
       plugn trigger post-domains-update $APP
     else
-      dokku_log_info1 "An app-specific SSL endpoint is not defined"
+      dokku_log_fail "An app-specific SSL endpoint is not defined"
     fi
     ;;
 


plugins/certs/commands
```diff
@@ -154,7 +154,7 @@ case "$1" in
       rm -rf $APP_SSL_PATH
       plugn trigger post-domains-update $APP
     else
-      dokku_log_info1 "An app-specific SSL endpoint is not defined"
+      dokku_log_fail "An app-specific SSL endpoint is not defined"
```
michaelshobbs: `dokku_log_fail` returns 1

see [here](https://github.com/progrium/dokku/blob/master/plugins/common/functions#L88-L91)

michaelshobbs:
Is it fair to assume you tested this change with #1683?

josegonzalez:
Can you rebase this and add a test?

u2mejc:
@josegonzalez The test already exists in #1683. The certs.bats is testing for non zero exit code, but currently `certs:remove` returns zero on error without this patch.

Issue #1627, opened 2015-10-29
backup:import is broken


dustMason:
I just tried to use dokku's backup:export and backup:import commands to carry an app config through a server teardown and was unable to import the tarball created by the initial backup:export command.

To export, I did `dokku backup:export filename`. I then booted a fresh server and installed dokku and scp'd up the tarball. When I ran `dokku backup:import filename`, I first got

`/var/lib/dokku/plugins/enabled/backup/commands: line 53: dokku_log_fail: command not found`

Using the -f flag resulted in the same output. I had a look at the source and decided to check out the contents of the tarball. I noticed that the `.dokku_backup_apps` file list included non-app directories like `.ssh` and `.basher`. I removed those and made a new tarball, which I was able to import.

Did I use this feature incorrectly? It seems like those hidden directories weren't supposed to be in the `.dokku_backup_apps` list.

michaelshobbs:
Unfortunately, this feature has been neglected over the past few months and I don't think all plugins are supporting it either. I know from conversations with @josegonzalez, this will be part of our development focus in the short term. 

In the meantime, a full backup of `/home/dokku` and `/var/lib/dokku` should accomplish the same thing.

dustMason:
Good to know, thanks for the reply. In the meantime hopefully this issue will help future backer-uppers.

Seralto:
I'm having problems with backup:import too, where I can generate the file and restore it, but it only creates empty folders.

josegonzalez:
Yeah, the backup feature is hella broken. There is an issue open to consolidate how we do configuration in dokku (probably for 0.5.0) which will make backups a bit more trivial. At the moment, the only method we have is tarring the two directories - as @michaelshobbs mentioned - though you'll want to stop any official services you have in the /var/lib/dokku/services directory first (or pause all writes while backing it up).

Seralto:
Thanks for the reply josegonzalez.

u2mejc:
I was going to take a crack at this bug, but I got to thinking, is this a feature we want to keep? Considering that any app is rebuildable from source? Correct me if I'm wrong, but heroku only supports pg:backup for the postgres databases. Is anyone storing session data on the docker file system? If someone mounted a volume, I would say they could just rsync that to a new instance or be keeping snapshots for disaster recovery.

Documentation may be more maintainable for migrating to a new server:
```
$ git remote rm dokku
$ git remote add dokku dokku@[dokku.me:dokku.me]
$ git push dokku [master]
```
[...]

PR #1687, opened 2015-11-13
Deprecate and remove dokku backup plugin, replace with documentation.

u2mejc:
This PR deprecates the currently broken backup plugin. See discussion here: #1627


josegonzalez:
Do you mind removing all the backup triggers from core plugins as well?

u2mejc:
Will do, maybe I should rebase too :yum:

docs/backup-recovery.md
```diff
@@ -0,0 +1,38 @@
...
+```
+dokku postgres:export [db_name] > [db_name].dump
+dokku postgres:import [db_name] < [db_name].dump
+```
+
+## Volumes and Static Assets
+Dokku doesn't enforce a [300mb] limit on apps, but it's best practice to keep binary assets outside of git. Since containers are considered volatile in Dokku, external stores like s3 or volumes should be used for non-volatile items like user uploads. The [dokku-volume] plugin also supports dumping and restoring data:
+
+> volume:dump&nbsp;&nbsp;&nbsp;<app> <container_path> > filename.tar.gz&nbsp;&nbsp;&nbsp;&nbsp;Dump volume data to filename.tar.gz
+> volume:restore&nbsp;&nbsp;&nbsp;&nbsp;<app> <container_path> < filename.tar.gz&nbsp;&nbsp;&nbsp;&nbsp;Restore volume data from filename.tar.gz
```
josegonzalez: We dont have a volume plugin? (should we)

docs/backup-recovery.md
```diff
@@ -0,0 +1,38 @@
+# Backup and Recovery
+
+## Backup and Migration Tutorial
```
josegonzalez: newlines after headers please.

docs/backup-recovery.md
jvanbaarsen: I think a persisted storage plugin would be a great addition. Since not everyone can host their assets on S3 because of laws

docs/backup-recovery.md
```diff
@@ -0,0 +1,38 @@
+# Backup and Recovery
+
```
josegonzalez: Add a note stating the backup plugin was removed in 0.5.x as it was completely broken or something.

docs/backup-recovery.md
jvanbaarsen: This might be a little bit related: https://github.com/dokku/dokku/issues/1695

josegonzalez: Open an issue. Personally I think it can be handled by docker-options, but that might be too hard for some users. Porcelain over plumbing is always nice.

jvanbaarsen: See: https://github.com/dokku/dokku/issues/1820

josegonzalez: @u2mejc I would remove this section for now since we don't have an official volumes plugin.

josegonzalez:
_bump_

u2mejc:
Blocked pending new core storage plugin: https://github.com/dokku/dokku/pull/1830

josegonzalez:
 Show all checks

u2mejc:
Updated backup-recovery.md to reflect new core storage plugin.

josegonzalez:
@u2mejc can you rebase this?

u2mejc:
> @u2mejc can you rebase this?
No it's going to be totally broken now! If only we had a branch that was continuously being rebased! ;-P

j/k, I'll see how bad it is later today.

michaelshobbs:
Can you squash this into one commit por favor?

PR #1691, opened 2015-11-14
Fix "Must specify DOMAIN" error over ssh

### LINKED ISSUE #1614: Getting "Must specify DOMAIN" with wildcard domains on ssh (eg. *.myapp.dokku.me)
I tried various quotes/escaping combinations but seems that [this check](https://github.com/progrium/dokku/blob/39ba53607aba743b4728c0ff6cc5c4eecac06e3c/plugins/domains/commands#L62) always triggers with a leading * when the command is invoked via ssh:

     ssh dokku@dokku.me dokku domains:add '*.myapp.dokku.me'

but works after entering the server

     ssh root@dokku.me
     dokku domains:add '*.myapp.dokku.me'

I realize it might be some bash/ssh quoting issue, still maybe something can be done to mitigate it

u2mejc:
Resolves #1614

### DIFF
--- a/dokku
+++ b/dokku
@@ -60,7 +60,9 @@ if [[ -n "$SSH_ORIGINAL_COMMAND" ]]; then
     xargs $0 <<<$SSH_ORIGINAL_COMMAND
     exit $?
   else
+    set -f
     $0 $SSH_ORIGINAL_COMMAND
+    set +f
     exit $?
   fi
 fi
--- a/plugins/domains/commands
+++ b/plugins/domains/commands
@@ -65,7 +65,7 @@ case "$1" in
       exit 1
     fi
 
-    if [[ $(egrep ^"$3"$ "$DOKKU_ROOT/$APP/VHOST" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
+    if [[ $(grep --fixed-strings "$3" "$DOKKU_ROOT/$APP/VHOST" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
       echo "$3 is already defined for $APP"
       exit 1
     fi


dokku
```diff
@@ -60,7 +60,9 @@ if [[ -n "$SSH_ORIGINAL_COMMAND" ]]; then
     xargs $0 <<<$SSH_ORIGINAL_COMMAND
     exit $?
   else
+    set -f
     $0 $SSH_ORIGINAL_COMMAND
+    set +f
```
u2mejc: This disables globbing which was causing the *.balh to disappear from the end.

plugins/domains/commands
```diff
@@ -65,7 +65,7 @@ case "$1" in
       exit 1
     fi
 
-    if [[ $(egrep ^"$3"$ "$DOKKU_ROOT/$APP/VHOST" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
+    if [[ $(grep --fixed-strings "$3" "$DOKKU_ROOT/$APP/VHOST" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
```
u2mejc: This is a related bug that I caught, \*.example.com matches www.example.com. Since it's one domain per line, using fixed strings allows us to match the entire line, including \*'s. grep because it's better supported and threw less errors in testing.

michaelshobbs:
the previous `egrep` command is a regex match that looks for the fixed string. `^` is beginning of the line and `$` is end of line. So it translates to matching a fixed string. Maybe I'm missing the point here?

michaelshobbs:
Oh I'm dumb. In the wildcard case, `egrep` will treat the string as a regexp.

michaelshobbs:
However this does break adding domains that match substrings....
i.e.

```
root@dokku:~/dokku# dokku apps:create testapp
Creating testapp... done
root@dokku:~/dokku# dokku domains:add testapp www.test.dokku.me
-----> Creating new /home/dokku/testapp/VHOST...
-----> Setting config vars
       DOKKU_NGINX_PORT: 80
-----> Configuring testapp.dokku.me...
-----> Configuring www.test.dokku.me...
-----> Configuring testapp.dokku.me...
-----> Configuring www.test.dokku.me...
-----> Added www.test.dokku.me to testapp
root@dokku:~/dokku# dokku domains:add testapp test.dokku.me
test.dokku.me is already defined for testapp
root@dokku:~/dokku# dokku domains testapp
=====> testapp Domain Names
testapp.dokku.me
www.test.dokku.me
```

josegonzalez:
Closing in favor of #1709. Thanks for the initial work @u2mejc!

PR #1712, opened 2015-11-24
filter incompatible docker option when building dockerfile vs herokuish apps

### LINKED ISSUE #1609: Cannot run first build using docker and postgres
I'm new to dokku so sorry if I've missed anything obvious.

I created a new app and linked a postgres container to it. I then tried to push my docker-based app into dokku for the first time. There was an error about the `--link` flag being 'provided but not defined'.

I poked around a bit and I think it's because the docker-options plugin uses the same args for build and run. If you've linked a postgres container to the app before you've built it, then you'll get a `--link` provided in the build stage.

I think I've skirted the problem by unlinking postgres, pushing into dokku, then relinking. Not sure if this is the right thing to do though.

michaelshobbs:
i'm not sure this is the best solution but it works in leiu of making the plugins handle it

closes #1609

### DIFF
--- a/docs/development/plugin-triggers.md
+++ b/docs/development/plugin-triggers.md
@@ -459,7 +459,7 @@ dokku config:set --no-restart $APP MANUALLY_STOPPED=1
 
 - Description:
 - Invoked by: `dokku build`
-- Arguments: `$APP`
+- Arguments: `$APP $IMAGE_SOURCE_TYPE`
 - Example:
 
 ```shell
--- a/plugins/00_dokku-standard/commands
+++ b/plugins/00_dokku-standard/commands
@@ -23,7 +23,7 @@ case "$1" in
         [[ -d $CACHE_DIR ]] || mkdir $CACHE_DIR
         plugn trigger pre-build-buildpack "$APP"
 
-        DOCKER_ARGS=$(: | plugn trigger docker-args-build $APP)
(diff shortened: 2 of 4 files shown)


mmerickel:
Is `--link` the only option expected to be used by plugins at this stage not used by herokuish/buildstep? My first impression is that this fix is short-sighted but I'm also suffering from the issue so I'm happy if this works as a temporary fix.

What about dockerfile plugins that don't work for herokuish builds like `--build-arg` or `--file`?

michaelshobbs:
Probably not and it is short-sighted indeed. However, leaving this broken is probably not good either.

mmerickel:
Yeah I was just bringing up the inverse problem where a plugin is expecting a dockerfile deployment.

It does seem beneficial to tell the plugin which deployment type is happening to really solve this issue.

michaelshobbs:
Updated to take into account the inverse problem. This is by no means exhaustive but gets us a pattern to filter args for now.

josegonzalez:
@michaelshobbs you want to add that static buildpack and a test app to herokuish?

michaelshobbs:
I can yeah. My fork or the official one?

josegonzalez:
Your fork. Or just move your fork to the dokku org and be done with it. Official fork?

michaelshobbs:
Yeah I'll revert this change here and open a new PR

PR #1718, opened 2015-11-26
Add post-create hook

josegonzalez:
Note that this is invoked *before* the application is deployed, so it's usage is quite different than a potential post-successful-create hook.

### DIFF
--- a/plugins/apps/commands
+++ b/plugins/apps/commands
@@ -2,6 +2,7 @@
 [[ " apps apps:create apps:destroy help apps:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/apps/functions"
 
 case "$1" in
   apps)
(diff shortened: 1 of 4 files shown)


plugins/apps/commands
```diff
@@ -16,6 +16,7 @@ case "$1" in
 
     mkdir -p "$DOKKU_ROOT/$APP"
     echo "Creating $APP... done"
+    plugn trigger post-create $APP
```
michaelshobbs: `post-app-create` perhaps?

plugins/git/commands
```diff
@@ -14,6 +14,7 @@ git_build_app_repo() {
   chmod 755 $TMP_WORK_DIR
   unset GIT_DIR GIT_WORK_TREE
   pushd $TMP_WORK_DIR > /dev/null
+  [[ ! -d "$DOKKU_ROOT/$APP" ]] && dokku apps:create $APP
```
michaelshobbs: Why is this necessary?

josegonzalez: Otherwise we won't actually call the hook on every app creation. If I push a repo up without first creating the app, it skips the above...

josegonzalez: We could also just manually call the hook here instead.

plugins/apps/commands
josegonzalez: The analogue for app:destroy is `post-delete`.

michaelshobbs: :+1:

plugins/git/commands
michaelshobbs: I'd personally prefer this to be a function call rather than an invocation of a new `dokku` process.

PR #1725, opened 2015-11-27
Copy csr from temporary directory into app ssl directory

### LINKED ISSUE #1724: Generating self signing ssl cert gives chmod error
Hi there.

When I try to run "dokku certs:generate APP DOMAIN" I get the openSSL questions and all seems fine. Yet it finished with this statement:

-----> Installing certificate and key...
chmod: cannot access ‘/home/dokku/everseller/tls/server.csr’: No such file or directory

When I run "dokku ps:restartall" to restart my app I see these lines:

-----> No matching configured domains for everseller found in SSL certificate. Your app will show as insecure in a browser if accessed via SSL -> this is ok

-----> Please add appropriate domains via the dokku domains command -> my app has the appropriate domain installed.

-----> Creating https nginx.conf -> So I guess some SSL stuff is getting configured?

The app restarts successfully and serves my app fine without ssl. But when requesting the app over https the connections get refused.

What would be a solution? :-)

I did not have this issue under 0.4.4.

Thanks!

josegonzalez:
Closes #1724

### DIFF
--- a/plugins/certs/commands
+++ b/plugins/certs/commands
@@ -93,7 +93,7 @@ case "$1" in
 
       mkdir -p "$APP_SSL_PATH"
       dokku_log_info1 "Installing certificate and key..."
-      mv -f $TMP_WORK_DIR/server.key $TMP_WORK_DIR/server.crt $APP_SSL_PATH
+      mv -f $TMP_WORK_DIR/server.key $TMP_WORK_DIR/server.crt $TMP_WORK_DIR/server.csr $APP_SSL_PATH
       chmod 750 $APP_SSL_PATH
       chmod 640 $APP_SSL_PATH/server.key $APP_SSL_PATH/server.crt $APP_SSL_PATH/server.csr
       [[ -n "$DOMAIN" ]] && (dokku domains:add $APP $DOMAIN || nginx_build_config $APP)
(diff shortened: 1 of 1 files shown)


plugins/certs/commands
```diff
@@ -93,14 +93,14 @@ case "$1" in
...
-      mv -f $TMP_WORK_DIR/server.key $TMP_WORK_DIR/server.crt $APP_SSL_PATH
+      mv -f $TMP_WORK_DIR/server.key $TMP_WORK_DIR/server.crt $TMP_WORK_DIR/server.csr $APP_SSL_PATH
       chmod 750 $APP_SSL_PATH
       chmod 640 $APP_SSL_PATH/server.key $APP_SSL_PATH/server.crt $APP_SSL_PATH/server.csr
       [[ -n "$DOMAIN" ]] && (dokku domains:add $APP $DOMAIN || nginx_build_config $APP)
       dokku_log_info1 "The following is a certificate signing request that can be used"
       dokku_log_info1 "to generate an 'officially' signed SSL certificate for $APP at $DOMAIN"
       dokku_log_info1 "by a CA of your choosing."
-      cat server.csr
+      cat "$APP_SSL_PATH/server.csr"
```
michaelshobbs: Derp on me here I think. #1732 changes this behavior and fixes this bug. There's no real reason to keep the csr around IMO.

josegonzalez:
So lets close this.

PR #1745, opened 2015-12-03
Change herokuish to recommended package

### LINKED ISSUE #1738: Can Deb package Recommend instead of Depend on herokuish
We have a dokku installation environment that is completely behind a firewall, and the only internet access is to proxy apt (and other) package repositories.  We currently have a working proxy to be able to GET and install the dokku deb package.  However, installing in this manner fails while installing the herokuish package, because it runs a curl to github.  

We are using Dokku with Dockerfile based building application building, so the herokuish package is unneeded in our situation.

Changing the herokuish package to Recommends instead of Depends would allow me to skip the installation of that package by specifying ```--no-install-recommends```, but herokuish would still be installed by other users by default.

beverku:
Fixes #1738 

Changing the herokuish package to Recommends instead of Depends allows users to skip the installation of that package by specifying --no-install-recommends.  Herokuish will still be installed by default using apt-get

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,8 @@ Version: 0.4.5
 Section: web
 Priority: optional
 Architecture: amd64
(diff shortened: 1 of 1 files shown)


josegonzalez:
Nice :)

PR #1750, opened 2015-12-04
New graphite & statsd plugin with grafana frontend added

jlachowski:
I've just created a new graphite/statsd plugin with grafana frontend (based on Jose Diaz-Gonzalez postgres plugin). This PR is adding the plugin link and info into dokku doc's.


josegonzalez:
@jlachowski What do you say about making this an official plugin? You'd still have full admin control over the repo, we just give it a better standing in the rankings :)

jlachowski:
No problem, it souncd great :-).
It's still WIP in some parts but, it's already functional and I use it in production.
Fell free to make it an official one - do you need me to do something?

josegonzalez:
Just added you to the org. Please initiate a transfer and we'll be off to the races :) @byrnedo is doing similar for the dokku-nats plugin, so maybe you two could share notes and let me know what we can do to make these sorts of moves snappier :)

Issue #1641, opened 2015-11-02
Fix lintian issues


josegonzalez:
I just randomly ran the latest deb package through [lintian](https://lintian.debian.org/manual/section-2.2.html) and came up with the following violations:

```
root@dokku:~# lintian dokku_0.4.3_amd64.deb
warning: the authors of lintian do not recommend running it with root privileges!
E: dokku: changelog-file-missing-in-native-package
W: dokku: control-file-is-empty conffiles
E: dokku: no-copyright-file
W: dokku: malformed-prompt-in-templates dokku/hostname
W: dokku: malformed-prompt-in-templates dokku/key_file
W: dokku: postrm-does-not-purge-debconf
W: dokku: missing-debconf-dependency-for-preinst
E: dokku: extended-description-is-empty
W: dokku: unknown-section base
E: dokku: dir-in-usr-local usr/local/bin/
E: dokku: file-in-usr-local usr/local/bin/dokku
W: dokku: file-in-unusual-dir usr/local/bin/dokku
E: dokku: dir-in-usr-local usr/local/share/
E: dokku: dir-in-usr-local usr/local/share/dokku/
E: dokku: dir-in-usr-local usr/local/share/dokku/contrib/
E: dokku: file-in-usr-local usr/local/share/dokku/contrib/dokku-installer.rb
W: dokku: file-in-unusual-dir usr/local/share/dokku/contrib/dokku-installer.rb
E: dokku: dir-in-usr-local usr/local/share/man/
E: dokku: dir-in-usr-local usr/local/share/man/man1/
E: dokku: file-in-usr-local usr/local/share/man/man1/dokku.1
W: dokku: file-in-unusual-dir usr/local/share/man/man1/dokku.1
W: dokku: script-not-executable var/lib/dokku/core-plugins/available/config/functions
```

Would be great to fix as many of these as possible so that we can be good packaging citizens :)

michaelshobbs:
Why does `config/functions`need to be executable?

josegonzalez:
Probably because of the shebang.

jvanbaarsen:
@josegonzalez / @michaelshobbs I think it would be possible to remove the shebang right? Since it only holds functions, and the file itself will never be executed. Or am I overlooking something?

josegonzalez:
Yeah that's possible.

jvanbaarsen:
@josegonzalez I added a PR to remove that shebang

josegonzalez:
Closing as there is a pr open.

PR #1767, opened 2015-12-10
Remove shebang from config/functions

jvanbaarsen:
In order to be one step closer to lintian compliance, lets remove the shebang
from this file so it doesn;t have to be executable

See #1641

### DIFF
--- a/plugins/config/functions
+++ b/plugins/config/functions
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 


josegonzalez:
Note: we still don't have tests running on CircleCI. They are aware of the issue and investigating. Merging as I've tested this change locally.

jvanbaarsen:
@josegonzalez What are the issues with Circle?

josegonzalez:
They apparently can't seem to enable it because I'm somehow not an admin of the dokku org in their system...

jvanbaarsen:
@josegonzalez Ah ok, did you consider an alternative? (Semaphore maybe?)

josegonzalez:
No, I haven't. I think @michaelshobbs is looking into it.

If anything, we'll setup our own CI on Digital Ocean. They sponsor us so we can probably setup an internal jenkins thing or similar to run tests.

jvanbaarsen:
ah ok :-) I hope the semaphore / circle solutions works though :) Otherwise its another thing you have to maintain ;)

PR #1771, opened 2015-12-12
Make plugin hooks send out more information

jvanbaarsen:
At this point you only get information that "something" happend on the given
hook. By adding more information to the hook, plugin creators can do more
interesting stuff.

* The first params remains the `$APP` name (to maintain backwards compatability)
* The second params sends in the action that happend, for domain this can be:
  `add`, `clear` or `remove`
* The third param sends in the argument that was passed to the original
  function, for domain that would be a list of all the domains that was given.

One reason this addition would be great for plugin owners, is that when you have
a web tool over Dokku, and someone creates a domain over the CLI, we a plugin
can send back information about the event to the webapp over an API. That way
the webapp is always up-to-date with the Dokku installation.

### DIFF
--- a/plugins/domains/commands
+++ b/plugins/domains/commands
@@ -75,7 +75,7 @@ case "$1" in
     for DOMAIN in "$@"; do
       echo "$DOMAIN" >> "$DOKKU_ROOT/$APP/VHOST"
     done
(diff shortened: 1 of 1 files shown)


michaelshobbs:
lgtm

PR #1776, opened 2015-12-15
fix version constraints

t-8ch:
The version of docker.io in jessie-backports is 1.6.2~dfsg1-1~bpo8+1.
This is regarded as lower than 1.6.2 by dpkg because `~` has a special meaning
in regards to version sort in dpkg.

This leads to:
```
The following packages have unmet dependencies:
 dokku : Depends: docker-engine-cs but it is not installable or
                  docker-engine but it is not installable or
                  lxc-docker (>= 1.6.2) but it is not installable or
                  docker.io (>= 1.6.2) but 1.6.2~dfsg1-1~bpo8+1 is to be installed
         Recommends: herokuish but it is not going to be installed
```

By changing the contraints the version in backports will be accepted by dpkg.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Version: 0.4.6
 Section: web
 Priority: optional
 Architecture: amd64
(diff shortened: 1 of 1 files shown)


josegonzalez:
Good catch.

PR #1779, opened 2015-12-16
Add link to new zero-downtime Let's Encrypt Plugin to docs

sseemayer:
I've created a new plugin for dokku that allows securing dokku apps using the now public-beta free certificate authority Let's Encrypt by one simple `dokku letsencrypt <myapp>`.

As opposed to the existing [sgulseth/dokku-letsencrypt](https://github.com/sgulseth/dokku-letsencrypt) plugin, no stopping of the dokku nginx server is required during the host verification process and it won't mess up permissions of a pre-existing systemwide let's encrypt installation's data in `/etc/letsencrypt` because it only works in its own isolated Let's Encrypt working directory.


josegonzalez:
You and @sgulseth should collaborate :)

PR #1784, opened 2015-12-17
Tiny fix in SSL documentation

duboff:
From what I can tell, this was meant for per app config and does not work without mentioning the app name.


michaelshobbs:
Good catch!

Issue #1783, opened 2015-12-16
Install is broken for Debian


antoine-lizee:
I get a 'bad Gateway' when trying to get to the web-admin after a fresh install depicted in /docs/getting-started/install/debian.md ([here](http://dokku.viewdocs.io/dokku/getting-started/install/debian/) online) on a system with debian (stable) Jessie.

I tracked down the bug to be sinatra not loading because `apt-get` pre-installs `rubygem-sinatra` from the ppa we just added (https://packagecloud.io/dokku/dokku/ubuntu/ trusty main). This package installs the gem for an incorrect version of ruby (1.9 instead of the default 2.1 on Debian Jessie).

3 remarks:

1. Am I right to try to install Dokku following these instructions on debian? If these are Ubuntu specific, it needs to be mentionned... If it's supposed to be debian compatible, then we gotta fix it :-)
1. Was there a proper way to debug that? I tracked down the bug by actually restarting myself the dokku-installer init script (that failed with sinatra not found).
1. For those who got the same problem, installing the correct packages and restart the init script did the trick:
```
sudo apt-get install ruby-rack ruby-rack-protection ruby-sinatra ruby-tilt
sudo sh /etc/init/dokku-installer.conf &
```

u2mejc:
Hi @antoine-lizee,

I was able to reproduce your issue in an EC2 instance. You might be right, reading quickly you would probably just noticed a list of steps, but the doc actually describes those as a way to automate the install instead of using the bootstrap script (like ubuntu does). 

I belive the install script has the same issue, the only thing I did after the install was:
```bash
gem install sinatra
sh /etc/init/dokku-installer.conf 
```

Debian isn't our primary target, but I don't think it'll be a hard fix. I'll poke this again in the morning when some of the guys are online.

antoine-lizee:
Thanks @u2mejc 
We could also package for debian by just changing the pre-depends to the right debian alternatives (see above), and store the package in `debian/jessie` of the package repo if we wanted to keep the apt-get philosophy. I would agree that just installing the gems afterwards is an easier path to deployment for these platforms though. Just mentioning it on the `debian.md` would be nice then! I've put in a quick PR, with suggested changes.

antoine-lizee:
Just adding a line maybe in the doc to be even clearer. See #1786.

michaelshobbs:
@u2mejc @josegonzalez I'm fine with updating docs but how hard would it be to make this installation target work?

[...]

PR #1786, opened 2015-12-17
make it clear that the debian install notes are not bug-free for debian.

antoine-lizee:
See #1783

### DIFF
--- a/docs/getting-started/install/debian.md
+++ b/docs/getting-started/install/debian.md
@@ -18,6 +18,8 @@ sudo apt-get install dokku
 sudo dokku plugin:install-dependencies --core
 ```
 
+These steps are not guaranteed to work on a non-Ubuntu platform. In particular, you'll need to install by hand the sinatra gem on Debian if you want to use the web admin for configuration.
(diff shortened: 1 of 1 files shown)


docs/getting-started/install/debian.md
```diff
@@ -18,6 +18,8 @@ sudo apt-get install dokku
 sudo dokku plugin:install-dependencies --core
 ```
 
+These steps are not guaranteed to work on a non-Ubuntu platform. In particular, you'll need to install by hand the sinatra gem on Debian if you want to use the web admin for configuration.
+
```
michaelshobbs: Might be better to lay out what you did in the code block?

antoine-lizee: Yeah, but these are not relevant for the *Ubuntu* unattended release, so I'll just adde them in a different code block. I think @josegonzalez suggestion of putting the gem install out of the package management is best the best long term solution?

michaelshobbs:
@antoine-lizee mind squashing this?

PR #1789, opened 2015-12-17
Make it possible to skip a deploy

lvillani:
The rationale here is to make it possible to `git push dokku master` a base image that is used by other applications that are effectively being deployed.

Used in this manner, Dokku becomes a lightweight alternative to a private Docker registry.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -316,12 +316,23 @@ release_and_deploy() {
       IMAGE_SOURCE_TYPE="dockerfile"
     fi
 
+    DOKKU_APP_SKIP_DEPLOY="$(dokku config:get $APP DOKKU_SKIP_DEPLOY || true)"
+    DOKKU_GLOBAL_SKIP_DEPLOY="$(dokku config:get --global DOKKU_SKIP_DEPLOY || true)"
+
+    DOKKU_SKIP_DEPLOY=${DOKKU_APP_SKIP_DEPLOY:="$DOKKU_GLOBAL_SKIP_DEPLOY"}
+
     dokku_log_info1 "Releasing $APP ($IMAGE)..."
     dokku release "$APP" "$IMAGE_SOURCE_TYPE" "$IMAGE_TAG"
-    dokku_log_info1 "Deploying $APP ($IMAGE)..."
-    dokku deploy "$APP" "$IMAGE_TAG"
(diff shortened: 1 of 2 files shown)


michaelshobbs:
You'll need to rebase this in order to get tests to pass.

plugins/common/functions
```diff
@@ -316,12 +316,20 @@ release_and_deploy() {
       IMAGE_SOURCE_TYPE="dockerfile"
     fi
 
+    DOKKU_APP_SKIP_DEPLOY="$(dokku config:get $APP DOKKU_APP_SKIP_DEPLOY || true)"
```
michaelshobbs: We'll probably want to support global with an override. See this as an example: https://github.com/dokku/dokku/blob/v0.4.6/dokku#L96-L102

lvillani:
Rebased and updated to add support for global setting. For some reason the test is still failing though.

lvillani:
Ah, it seems it was a temporary problem. Looks green now.

plugins/common/functions
```diff
@@ -316,12 +316,23 @@ release_and_deploy() {
...
+    DOKKU_SKIP_DEPLOY=${DOKKU_APP_SKIP_DEPLOY:="$DOKKU_GLOBAL_SKIP_DEPLOY"}
+
     dokku_log_info1 "Releasing $APP ($IMAGE)..."
     dokku release "$APP" "$IMAGE_SOURCE_TYPE" "$IMAGE_TAG"
-    dokku_log_info1 "Deploying $APP ($IMAGE)..."
-    dokku deploy "$APP" "$IMAGE_TAG"
-    dokku_log_info2 "Application deployed:"
-    dokku urls $APP | sed "s/^/       /"
+
+    if [ "$DOKKU_SKIP_DEPLOY" != "true" ]; then
```
michaelshobbs: Sorry I missed this in the first pass, but following our [coding standards](https://github.com/progrium/bashstyle), can you swap out the deprecated `[` for the preferred `[[`? Thanks!

lvillani: Right! I'll try to fix it ASAP (I'm on a train and have bad connectivity, atm)

lvillani: Should be OK now.

michaelshobbs: Excellent! Thank you

PR #1791, opened 2015-12-17
don't run app deploy tests and spread out unit tests to 4 containers

### LINKED ISSUE #1787: pare down test suite
In thinking about the test suite and how to make it run faster, I'd like to propose dropping the deploy tests and only run bats tests. Given that `herokuish` now runs through all of the buildpacks in its test suite, it seems redundant that we do it as well.

Dropping those two test runs from dokku, would allow us to split up the bats test across all four of the free containers circleci provides us.

Thoughts?

michaelshobbs:
closes #1787


josegonzalez:
Why this change?

michaelshobbs:
See #1787

michaelshobbs:
This isn't quite working for some reason yet. I'll be able to work on this more over the weekend.

PR #1794, opened 2015-12-22
Replace dokku-installer.rb with dokku-installer.py

### LINKED ISSUE #1783: Install is broken for Debian

josegonzalez:
The python version should allow us to remove any ruby dependencies and should also slightly speed up dokku installation. All LSB-compliant distributions should have python, so we should be okay depending upon it.

Closes #1783
Closes #1786


josegonzalez:
Tested against ubuntu and works fine :)

michaelshobbs:
I'm +1 for anything that `--` ruby code :stuck_out_tongue_closed_eyes:

josegonzalez:
I'm going to merge this. If there are any issues in usage, we can sort them out after the release.

antoine-lizee:
LGTM! Thanks for this. Unfortunately I don't really have the skills/time to quickly test it on a fresh Debian install right now, but I don't see why it wouldn't work. Will make sure to try it next time I install dokku on Debian.

PR #1797, opened 2015-12-22
ensure we run plugin commands as root

### LINKED ISSUE #1768: thought: don't try to install plugins, unless root.
**running `v0.4.5`**
I came across the same problem as #1503 
`/var/lib/dokku/plugins/enabled/00_dokku-standard/install: line 16: /etc/init/dokku-redeploy.conf: Permission denied`

I was using the `sshcommand` interface where there's no sudo.

The confusing part was the cloning and everything worked fine, so I wasn't sure at which step I was, and it isn't obvious how to recover from this state. 

(what if there *where* tasks after the one that failed? impossible to tell without reading the source.)

Isn't it better to check the privileges before proceeding? and provide a helpful error, etc?

michaelshobbs:
closes #1768

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,10 +72,10 @@ version:
 	git describe --tags > ~dokku/VERSION  2> /dev/null || echo '~${DOKKU_VERSION} ($(shell date -uIminutes))' > ~dokku/VERSION
 
 plugin-dependencies: plugn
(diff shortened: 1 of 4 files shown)


josegonzalez:
Love it.

PR #1807, opened 2015-12-30
Allow updating submodules at any revision

josegonzalez:
Using `--depth=1` breaks any submodules that are not at HEAD.

/cc @mmerickel


josegonzalez:
I'm going to merge this without comment. It breaks existing applications that don't use HEAD of their submodules, which I think is pretty bad.

We can't do a `--depth=1 || --depth=all` as that will break (git will see the existing submodules and bork on the alternative). The only other method would be to retry the temp repository with a new base directory, but I don't feel like doing that (PRs welcome though!).

@mmerickel feel free to revisit this since I know it's a painful topic for you :(

mmerickel:
No problem, I shouldn't have thrown the submodule fix in there. It's not a feature I use myself. I naively assumed that --depth=1 would clone the commit the submodule was pinned to... not just head.

PR #1810, opened 2015-12-31
Dockerfile installation requirements

josegonzalez:
WIP

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -29,6 +29,7 @@ hostname -f > /dev/null 2>&1 || {
   exit 1
 }
 
+echo "--> Ensuring we have the proper dependencies"
 apt-get update -qq > /dev/null
 [[ $(lsb_release -sr) == "12.04" ]] && apt-get install -qq -y python-software-properties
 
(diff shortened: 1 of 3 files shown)


josegonzalez:
Currently the install fails as soon as the `postinst` debian package hook is hit.

My first pass at getting a dockerfile setup working is getting it running under ubuntu, which hopefully means using a package based-solution - who wants to maintain multiple installation methods, right? - but maybe I should just take the "install from source" tact...

josegonzalez:
From some [random debconf docs](http://www.fifi.org/doc/debconf-doc/tutorial.html#AEN198) I found on the interwebs:


> A few things can commonly go wrong when you convert something over to debconf:
> - Your postinst uses debconf and starts a daemon that doesn't close all inherited file descriptors (all such daemons are buggy, really). This makes debconf hang, because the debconf frontend waits for the daemon to close the fd's before continuing. Note that if you use confmodule, the program probably needs to close fd's 0, 1, 2, and 3.
>
>   To fix:
>   - Redirect fd's to /dev/null before running the daemon
>   - Or fix the daemon.
>   - Or call the "stop" command at the end of your postinst, to let the frontend know you're done.

It's actually `db_stop`, but adding it seems to fix the install process at least.

Issue #1076, opened 2015-03-30
Docker hosted dokku


Downchuck:
This puts Dokku into a docker container -- for ease of use, I'm just using docker-flatip and socat for glue.

Dockerfile

``` Dockerfile
FROM ubuntu:latest

ENV DOKKU_TAG=v0.3.16

RUN apt-get update
RUN apt-get -y install libpcre3 libssl-dev
RUN apt-get -y install libpcre3-dev
RUN apt-get -y install wget zip gcc

## Dokku
RUN apt-get -y install curl git make ruby
RUN apt-get -y install dnsutils dropbear
RUN wget -qO- https://raw.githubusercontent.com/progrium/dokku/v0.3.16/bootstrap.sh | bash \
 && cd /root/dokku && make -i dokku-installer \
 && rm /etc/nginx/conf.d/dokku-installer.conf \
 && rm /etc/init/dokku-installer.conf \
 && rm -rf /home/dokku \
 && rm -rf /var/lib/apt/lists/*;
# RUN apt-get -y install nginx

EXPOSE 22
EXPOSE 80
VOLUME ["/home/dokku"]

ADD start /start
RUN chmod 755 /start

CMD ["/start"]
```

/start

``` sh
#!/bin/bash
set -e

DOKKU_VERISON="v0.3.16"
DOKKU_HOME="/home/dokku"
VHOST=$(hostname)

mkdir -p -m 0700 ${DOKKU_HOME}
chown -R dokku:dokku ${DOKKU_HOME}

echo "${VHOST}" > ${DOKKU_HOME}/HOSTNAME
echo "${VHOST}" > ${DOKKU_HOME}/VHOST
echo "${DOKKU_VERSION}" > ${DOKKU_HOME}/VERSION
cat <<E > ${DOKKU_HOME}/ENV
export CURL_CONNECT_TIMEOUT=5
export CURL_TIMEOUT=30
E

[ -f "/var/run/docker.sock" ] || \
 ln -s ${DOKKU_HOME}/docker.sock /var/run/docker.sock

echo "Starting dropbear...";
dropbear
echo "Starting nginx...";
nginx -g 'daemon off;'
```

On the host, you will need to link the unix domain socket for docker:

``` sh
socat unix-listen:/opt/dokku/data/docker.sock,fork unix-connect:/var/run/docker.sock
```

[...]

PR #1812, opened 2016-01-02
Debian packaging updates

### LINKED ISSUE #1641: Fix lintian issues

### LINKED ISSUE #1654: Rely on distro-provided nginx

josegonzalez:
> Note: Even with these change, we will still only officially support whatever the current LTS of Ubuntu is at any point in time. This isn't due to favoritism or anything, we just don't have the time, resources, or desire to support anything else. 
>
> If someone wants to jump in and make the latest CentOS *just work*, I'd be happy to give them support/commit bit. FWIW, we've already allowed some (now untested) SUSE support from #751.
>
> As for the core, our next steps are to officially support Dockerfile-based deployments. You can follow along in #1076 for more details.

- Fix as many lintian issues as possible. Closes #1641.
- Add actually tested support for Ubuntu 15.04, 15.10, and Debian Jessie. Closes #1654.
- Fix issue where we made unnecessary init files for dokku-installer.
- Add python2.7 and debconf as predependencies

Todo:

- [x] Fix dokku-redeploy for systemd
- [x] Test on Ubuntu 14.04
- [x] Test on Ubuntu 15.04
- [x] Test on Ubuntu 15.10
- [x] Test on Debian Jessie 8.2

### DIFF
--- a/contrib/dokku-installer.py
+++ b/contrib/dokku-installer.py
@@ -38,6 +38,10 @@ def check_boot():
     systemd_dir = os.getenv('SYSTEMD_DIR', '/etc/systemd/system')
     nginx_dir = os.getenv('NGINX_DIR', '/etc/nginx/conf.d')
 
+    if os.path.exists(init_dir):
+        with open('{0}/dokku-installer.conf'.format(init_dir), 'w') as f:
+            f.write("start on runlevel [2345]\n")
+            f.write("exec {0} selfdestruct\n".format(os.path.abspath(__file__)))
     if os.path.exists(systemd_dir):
         with open('{0}/dokku-installer.service'.format(systemd_dir), 'w') as f:
             f.write("[Unit]\n")
@@ -49,10 +53,6 @@ def check_boot():
             f.write("[Install]\n")
             f.write("WantedBy=multi-user.target\n")
             f.write("WantedBy=graphical.target\n")
-    elif os.path.exists(init_dir):
-        with open('{0}/dokku-installer.conf'.format(init_dir), 'w') as f:
-            f.write("start on runlevel [2345]\n")
-            f.write("exec {0} selfdestruct\n".format(os.path.abspath(__file__)))
     if os.path.exists(nginx_dir):
         with open('{0}/dokku-installer.conf'.format(nginx_dir), 'w') as f:
             f.write("upstream dokku-installer { server 127.0.0.1:2000; }\n")
--- a/contrib/dokku-installer.py
+++ b/contrib/dokku-installer.py
@@ -38,10 +38,6 @@ def check_boot():
     systemd_dir = os.getenv('SYSTEMD_DIR', '/etc/systemd/system')
     nginx_dir = os.getenv('NGINX_DIR', '/etc/nginx/conf.d')
 
-    if os.path.exists(init_dir):
-        with open('{0}/dokku-installer.conf'.format(init_dir), 'w') as f:
-            f.write("start on runlevel [2345]\n")
-            f.write("exec {0} selfdestruct\n".format(os.path.abspath(__file__)))
     if os.path.exists(systemd_dir):
         with open('{0}/dokku-installer.service'.format(systemd_dir), 'w') as f:
             f.write("[Unit]\n")
@@ -53,6 +49,10 @@ def check_boot():
             f.write("[Install]\n")
             f.write("WantedBy=multi-user.target\n")
             f.write("WantedBy=graphical.target\n")
+    elif os.path.exists(init_dir):
+        with open('{0}/dokku-installer.conf'.format(init_dir), 'w') as f:
+            f.write("start on runlevel [2345]\n")
+            f.write("exec {0} selfdestruct\n".format(os.path.abspath(__file__)))
     if os.path.exists(nginx_dir):
         with open('{0}/dokku-installer.conf'.format(nginx_dir), 'w') as f:
             f.write("upstream dokku-installer { server 127.0.0.1:2000; }\n")
--- a/deb.mk
+++ b/deb.mk
@@ -83,13 +83,13 @@ deb-dokku: deb-setup
 	mkdir -p /tmp/tmp /tmp/build
 
 	cp -r debian /tmp/build/DEBIAN
-	mkdir -p /tmp/build/usr/bin
+	mkdir -p /tmp/build/usr/local/bin
 	mkdir -p /tmp/build/var/lib/dokku/core-plugins/available
 	mkdir -p /tmp/build/usr/share/man/man1
 	mkdir -p /tmp/build/usr/share/dokku/contrib
 	mkdir -p /tmp/build/usr/share/doc/dokku
 
-	cp dokku /tmp/build/usr/bin
+	cp dokku /tmp/build/usr/local/bin
 	cp LICENSE /tmp/build/usr/share/doc/dokku/copyright
 	cp -r plugins/* /tmp/build/var/lib/dokku/core-plugins/available
 	find plugins/ -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | while read plugin; do touch /tmp/build/var/lib/dokku/core-plugins/available/$$plugin/.core; done
--- a/deb.mk
+++ b/deb.mk
@@ -87,8 +87,10 @@ deb-dokku: deb-setup
 	mkdir -p /tmp/build/var/lib/dokku/core-plugins/available
 	mkdir -p /tmp/build/usr/share/man/man1
 	mkdir -p /tmp/build/usr/share/dokku/contrib
+	mkdir -p /tmp/build/usr/share/doc/dokku
 
 	cp dokku /tmp/build/usr/bin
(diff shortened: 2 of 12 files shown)


micheal:
Sweet. Let me know if I can help with testing anything.

josegonzalez:
Note: one thing I am changing is being a bit more strict about the data we keep around after a user performs a `purge`. I'm going to destroy all applications, as well as any dokku-related configuration (tls, dokkurc, etc.).

I may introduce a `pre-dokku-purge` trigger as well, so plugins can take advantage of this instead of implementing it in the packaged distribution.

MikeSchroll:
Just tying things together, by leaving a note here regarding a challenge with Ubuntu 15.10 (and maybe soon 16.04): https://github.com/dokku/dokku/issues/2034

michaelshobbs:
@MikeSchroll what's the issue?

josegonzalez:
Looks like wily has this [package available](http://packages.ubuntu.com/wily/sigil) as the thing that broke #2034. *sigh*

I'll take a look at renaming the package, though if someone installs both, they are hosed.

Issue #1795, opened 2015-12-22
Support docker-swarm


aisensiy:
I read the document in docker it said that it is possible to use dokku with swarm and I install dokku and a 3-node swarm cluster with vagrant and is there some method to let dokku's default docker host is the cluster?

michaelshobbs:
Currently dokku is a single host solution. I think deis is the proposed multi host alternative, currently. I know some folks have been tinkering a bit with swarm and dokku but not sure where all that stands. I would definitely be interested in helping to spec out a swarm implementation, though.

Would you mind linking to that document from docker?

aisensiy:
@michaelshobbs  https://docs.docker.com/swarm/

aisensiy:
@michaelshobbs We are trying to use deis to run a PaaS but it has no concept of addon or plugin in deis, which means I can just run a web application without cache, database.

josegonzalez:
@aisensiy I don't think there is currently a way to script set the `-H` option on all docker commands. You could try the docker-options plugin, but that will break for the `dokku logs` command I think, among others.

I'll try and see if I can set this up this weekend, would be interesting.

aisensiy:
@josegonzalez Thanks for your reply, I tried with `-H` option but still no work...

blackxored:
I've played with this quite a bit, being able to have dokku running in a "swarmy", however most of the time gives me errors on not being able to find the image in the host, and I stopped there.

josegonzalez:
@blackxored what did you do to modify dokku? I'm definitely into any changes necessary.

One thing I think that might be a PITA is that images are built on a single box, and I don't know if swarm knows how to locate "local" images. Might that be the reason why you can't easily scale up/down instances?

Seems related: https://github.com/docker/swarm/issues/806

Not sure of a good way to get around ^ if that is a real issue, other than creating a registry, pushing the image there, *then* scaling...

blackxored:
@josegonzalez In 360 view I had to pass/obey/verify DOCKER_HOST, DOCKER_TLS_VERIFY, plus DOCKER_CERT_PART, had to have it respect custom-made overlay net $DOCKER_NETWORK_NAME, and due to the nature of multi-host, disable the host's nginx and run in in a container instead. But between the volatile image errors on deployment and the fact that I was half-way into having it send the right IP to nginx's configuration I kinda gave up at the moment.

blackxored:
The ghost images issues might have to do with the cleanup dokku does everytime, can someone confirm?
[...]

PR #1814, opened 2016-01-02
fix: use new Go template for extracting IP

blackxored:
I was running into issues with the IP addresses, and it seems that it's using an incorrect/outdated syntax for retrieving the container's IP Address. In multi-host networking `NetworkSettings.IPAddress` is empty.

The docker documentation at https://docs.docker.com/engine/reference/commandline/inspect/
uses a different syntax which I've tested to work both on multi-host and standard bridge networking.

If you're interested in the other issues and the work/hacks I've been doing related to #1795, let me know. In the meantime, I think this fix is needed.

### DIFF
--- a/dokku
+++ b/dokku
@@ -141,7 +141,7 @@ case "$1" in
             ipaddr=127.0.0.1
           else
             id=$(docker run -d -e PORT=$port $DOCKER_ARGS $IMAGE $START_CMD)
-            ipaddr=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' $id)
+            ipaddr=$(docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $id)
           fi
         else
           id=$(docker run -d $DOCKER_ARGS $IMAGE $START_CMD)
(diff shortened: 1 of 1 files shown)


michaelshobbs:
What versions of docker is this compatible? It doesn't seem to jive with the docker version on CircleCI.

blackxored:
This is my setup:
```
Client:
 Version:      1.9.1
 API version:  1.21
 Go version:   go1.4.3
 Git commit:   a34a1d5
 Built:        Fri Nov 20 17:56:04 UTC 2015
 OS/Arch:      darwin/amd64

Server:
 Version:      swarm/1.0.1
 API version:  1.21
 Go version:   go1.5.2
 Git commit:   744e3a3
 Built:
 OS/Arch:      linux/amd64
```

michaelshobbs:
I tested it locally and it definitely works on docker `1.9.1` but circle currently runs on `1.8.2`. In this PR's current form, I think we'll need to wait to merge until circle updates their docker version.

blackxored:
Let me see if I can come up with some backward-compatible fix.

michaelshobbs:
:+1: 

I also messaged the Circle folks to find out when 1.9.x support will be available.

josegonzalez:
Good cach @blackxored. We'd be happy to have any other patches you think are useful/relevant, even if tests don't pass. We can figure out testing at a later date :)

PR #1825, opened 2016-01-05
Frog staging

darinherrmann:
Merge tag 'v0.4.9' into frog-staging


josegonzalez:
I don't think you meant to do this?

PR #1827, opened 2016-01-05
use docker 1.9.0 on circleci

michaelshobbs:
reference: https://discuss.circleci.com/t/docker-1-9-0-is-available/366


josegonzalez:
I'm going to start collecting stuff for an impending 0.5.x release. Should we drop support for old versions of docker?

michaelshobbs:
I think it's reasonable to support `docker >=1.9.0` in `dokku 0.5.x`

Issue #1220, opened 2015-06-07
dokku cleanup removed all "exited" containers


palamccc:
dokku should only remove containers created by dokku. It should not remove containers created by others.

I'm running host, with dokku and other docker images. I use 'data-only' container pattern to store data in a 'non-running' container. But dokku removes my non-running data container as well.
http://container42.com/2013/12/16/persistent-volumes-with-docker-container-as-volume-pattern/

https://github.com/progrium/dokku/blob/4aedb6507a3bdec1ff29be2ff60366075186a72e/dokku#L170

josegonzalez:
How are we to detect that you're running a non-dokku provided container?

3onyc:
Possible solution is to label ([--label](https://docs.docker.com/reference/commandline/cli/#create)) containers created with Dokku, then modify clean-up to only remove those. Note that any plug-ins need to be modified to also label their containers.

You'd need to modify dokku itself to accomplish this though.

palamccc:
Is it possible to run all docker build commands with "--rm" option, it will automatically remove container after doing the work. so there won't be any need for cleanup.

3onyc:
That would potentially leave images and containers, for example the Docker daemon crashing would prevent containers (with `--rm`) from being deleted.

michaelshobbs:
it's also very useful to have the intermediate images/containers around debug deployments. Dokku was never (from what I can tell) specifically written to "play nice" with non-dokku managed images and containers.

Having said that, we can definitely label dokku core containers but as @3onyc mentioned, plugins will also need to be modified. 

Anybody else have thoughts on this?

zimbatm:
I have just been hit by that issue. Maybe the doc should say more clearly that dokku is not meant to work with other docker containers.

josegonzalez:
@michaelshobbs Not sure I understand your first comment:

> it's also very useful to have the intermediate images/containers around debug deployments. Dokku was never (from what I can tell) specifically written to "play nice" with non-dokku managed images and containers.

In any case, I would rather we just state in the docs that dokku doesn't play nice with externally-managed containers. We should definitely not delete data-only containers, though I'm not sure what we can do here. Maybe support a special, dokku-specific label that we'll use to ensure we don't kill certain containers?

alessio:
:+1: for containers labelling

josegonzalez:
[...]

PR #1828, opened 2016-01-05
label all dokku containers

michaelshobbs:
refs: #1220


docs/development/plugin-creation.md
```diff
@@ -90,3 +90,4 @@ A few notes:
   ```
 - From time to time you may want to allow other plugins access to (some of) your plugin's functionality. You can expose this by including a `functions` file in your plugin for others to source. Consider all functions in that file to be publicly accessible by other plugins. Any functions not wished to be made "public" should reside within your plugin trigger or commands files.
 - As of 0.4.0, we allow image tagging and deployment of said tagged images. Therefore, hard-coding of `$IMAGE` as `dokku/$APP` is no longer sufficient. Instead, for non `pre/post-build-*` plugins, use `get_running_image_tag()` & `get_app_image_name()` as sourced from common/functions. See the [plugin triggers](http://dokku.viewdocs.io/dokku/development/plugin-triggers) doc for examples.
+- As of 0.5.0, we will be using container labels to help cleanup intermediate containers with `dokku cleanup`. If calling `docker run`, include `$DOKKU_GLOBAL_RUN_ARGS`. This will ensure you intermediate containers labelled correctly.
```
josegonzalez: - `we will be using` => `we use`
- `If calling` => `If manually calling` (should we provide a wrapper for this?).

dokku
```diff
@@ -29,6 +29,9 @@ export DOKKU_VALID_EXIT=0
 export DOKKU_LOGS_DIR=${DOKKU_LOGS_DIR:="/var/log/dokku"}
 export DOKKU_EVENTS_LOGFILE=${DOKKU_EVENTS_LOGFILE:="$DOKKU_LOGS_DIR/events.log"}
 
+export DOKKU_CONTAINER_LABEL=dokku
+export DOKKU_GLOBAL_RUN_ARGS="--label=$DOKKU_CONTAINER_LABEL"
```
josegonzalez: Might also be useful for partial swarm support.

michaelshobbs: :+1:

docs/development/plugin-creation.md
michaelshobbs: I'm on the fence on the wrapper topic. I thought about it but meh

Issue #1558, opened 2015-10-13
Consolidated dokku app and plugin configuration


josegonzalez:
One thing we've been sorely lacking is a single way to refer and store configuration in the dokku world. 

- Application-related plugins usually store their data in the app's git directory in `/home/dokku`
- The official datastores store their data (and mount information) in `/var/lib/dokku/services/SERVICE/`
- Community plugins use stuff like `/home/dokku/.o_mysql`
- Dokku-wide config is usually stored in `/home/dokku`

Because of all the random ways in which data can be stored, there isn't exactly a simple way to save the state of an existing dokku installation. You can very easily push the container images to a registry and have backups occur from datastores, but actual config? Fffft.

Our `backup` plugin has been more or less ignored in the `0.3.x` line - totally my fault - and i think we need to start getting serious about this.

At one point, there was a note in a roadmap to store everything in a git directory. I think it's time we head down that road for 0.5.x.

---

### generic spec

- Data should be stored in a directory such as `/var/lib/dokku/config`.
  - `/var/lib/dokku/config/apps` - stores application data. Each application should get it's own directory, corresponding to the application name.
  - `/var/lib/dokku/config/plugins` - stores plugin data. Each plugin should get it's own directory, corresponding to the plugin name.
- We should provide a new `dokku-config` binary that is aware of the above paths and can interact with them.
  - Apps will interface with it using the `--app APP` flag, and can perform general crud actions against their own config.
  - Plugins will interface with it using the `--plugin PLUGIN` flag. Ditto on crud actions
  - The binary is a generic key-value store, and will be capable of triggering hooks on actions. For instance, we might want to log access, or ensure all actions get saved to an external datastore (git!).
- The backup system will be retrofitted to only work with the above system. If you want a backup, you basically get a tar of that directory. Users that want a backup of datastores should interface with those datastores either directly or through their plugin interfaces.
  - The official datastores all have ways of backing up/restoring data, so this is not a problem. We also cover all major datastores.

### open questions

- Do we handle encryption? I don't think it's worth it.
- How do we handle stuff like nginx? Does the nginx include path just change here?
- Should we serialize the data somehow, to json maybe?
- How do we ensure that data doesn't get accidentally overwritten by a shitty plugin/app/user?
- What language should it be written in? Golang is the obvious one to me, though handling plugins in it is annoying.

@Flink @michaelshobbs thoughts?

josegonzalez:
#1571

michaelshobbs:
[...]

PR #1830, opened 2016-01-06
Add core storage plugin to manage docker bind mounts

### LINKED ISSUE #1820: Create an official Dokku-Persisted Storage plugin
It would be great to have a plugin that can handle persisted storage in a nice way. The reason for this, is that not everyone can store data on S3 (because of local law they are prohibited to store anything on servers owned by US companies.)

I think the following API with be awesome: `dokku storage:create app_name`. This will then create a folder in `/home/dokku/storage/`. And the volume will be mounted into the container on `/shared/`.

Downside of the above API is that we only support one shared folder per app, but I think for starters that would be good enough.

Relates to: https://github.com/dokku/dokku/issues/1695

u2mejc:
First rev of storage core plugin to add non-volatile storage, resolves #1820.

### DIFF
--- a/docs/dokku-storage.md
+++ b/docs/dokku-storage.md
@@ -1,0 +1,45 @@
+# Dokku Core Storage Plugin
+
+> Spanking New as of 0.5.0
+
+As of Dokku 0.5.0, the preferred method to mount external containers to a dokku managed container, is to use the dokku storage plugin.
+
+
+```shell
+$ dokku storage:help
+    storage:list <app>, List bind mounts for app's container(s) (host:container)
+    storage:mount <app> <host-dir:container-dir>, Create a new bind mount
+    storage:unmount <app> <host-dir:container-dir>, Remove an existing bind mount
+```
+
+## Ideology and Background
+The storage plugin requires explicit paths on the host side. This is intentional to ensure that new users avoid running into unexpected results with implicit paths that may not exist (a feature deprecate in [Docker 1.9.0](https://github.com/docker/docker/releases/tag/v1.9.0])). The container directory is created for the mount point in the container. Any existing directory contents are not accessible after a mount is added to the container. Dokku creates a new directory `/var/lib/dokku/data/storage` during installation, it's the general consensus that new users should use this directory. Mounts are only available at run and deploy times, you must redeploy (restart) an app to mount or unmount to an existing app's container.
+
+## Usage
+This example demonstrates how to mount the recommended directory to `/storage` inside the container:
+```
+$ dokku storage:mount app-name /var/lib/dokku/data/storage:/storage
+```
+Dokku will then mount the shared contents of`/var/lib/dokku/storage` to `/storage` inside the container.
+
+## Use Cases
+
+##### Persistent storage
+
+Dokku is powered by Docker containers, which recommends in their [best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#containers-should-be-ephemeral) that containers be treated as ephemeral. In order to manage persistent storage for web applications, like user uploads or large binary assets like images, a directory outside the container should be mounted.
+
+##### Shared storage between containers
+
+When scaling your app, you may require a common location to access shared assets between containers, a storage mount can be used in this situation.
+
+##### Shared storage across environments
+
+Your app may be used in a cluster that requires containers or resources not running on the same host access your data. Mounting a shared file service (like S3FS or EFS) inside your container will give you great flexibility.
+
+##### Backing up
+
+Your app may have services that are running in memory and need to be backed up locally (like a key store). Mount a non ephemeral storage mount will allow backups that are not lost when the app is shut down.
+
+## Docker-Options Note
+
+The storage plugins is compatible with storage mounts created with the docker-options. The storage plugin will only list mounts from the deploy phase.
--- a/plugins/storage/commands
+++ b/plugins/storage/commands
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+[[ " storage storage:list storage:mount storage:unmount storage:help help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/storage/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/docker-options/functions"
+
+case "$1" in
+  storage:list)
+    # List all bound mounts
+    passed_phases="deploy"
+    [[ ! $2 ]] && dokku_log_fail "storage:list requires an app to list."
+    verify_app_name "$2" && APP="$2"
+    echo "$APP volume bind-mounts:"
+    get_bind_mounts "$(get_phase_file_path "$passed_phases")"
+    ;;
+
+  storage:mount)
+    # Add bind-mount, redeploy app if running
+    passed_phases=(deploy run)
+    [[ -z $3 || $4 ]] && dokku_log_fail "storage:mount requires an app and a two paths devided by colon."
+    verify_app_name "$2" && APP="$2"
+    verify_paths "$3" && MOUNT_PATH="$3"
+    check_if_path_exists "$MOUNT_PATH" "$(get_phase_file_path "${passed_phases[@]}")" && dokku_log_fail "Mount path already exists."
+    add_passed_docker_option passed_phases[@] "-v $MOUNT_PATH"
+    ;;
+
+  storage:unmount)
+    # Remove bind-mount, restart app if running
+    passed_phases=(deploy run)
+    [[ -z $3 || $4 ]] && dokku_log_fail "storage:unmount requires an app and a two paths devided by colon."
+    verify_app_name "$2" && APP="$2"
+    verify_paths "$3" && MOUNT_PATH="$3"
+    check_if_path_exists "$MOUNT_PATH" "$(get_phase_file_path "${passed_phases[@]}")" || dokku_log_fail "Mount path does not exist."
+    remove_passed_docker_option passed_phases[@] "-v $MOUNT_PATH"
+    ;;
+
+  help | storage:help)
+    cat<<EOF
+    storage:list <app>, List bind mounts for app's container(s) (host:container)
+    storage:mount <app> <host-dir:container-dir>, Create a new bind mount
+    storage:unmount <app> <host-dir:container-dir>, Remove an existing bind mount
+EOF
+    ;;
+
+  *)
+    exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
+    ;;
+
+esac
--- a/tests/unit/20_storage.bats
+++ b/tests/unit/20_storage.bats
@@ -1,0 +1,40 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+setup() {
+  create_app
+}
+
+teardown() {
+  destroy_app
+}
+
+@test "(storage) storage:mount, storage:list, storage:umount" {
+  run /bin/bash -c "dokku storage:mount $TEST_APP /tmp/mount:/mount"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run /bin/bash -c "dokku storage:list $TEST_APP | grep -qe '^\s*/tmp/mount:/mount'"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run /bin/bash -c "dokku storage:mount $TEST_APP /tmp/mount:/mount"
+  echo "output: "$output
+  echo "status: "$status
+  assert_output "Mount path already exists."
+  assert_failure
+  run /bin/bash -c "dokku storage:unmount $TEST_APP /tmp/mount:/mount"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run /bin/bash -c "dokku storage:list $TEST_APP | grep -vqe '^\s*/tmp/mount:/mount'"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run /bin/bash -c "dokku storage:unmount $TEST_APP /tmp/mount:/mount"
+  echo "output: "$output
+  echo "status: "$status
+  assert_output "Mount path does not exist."
+  assert_failure
+}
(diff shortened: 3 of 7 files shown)


jvanbaarsen:
:heart: Thanks for working on this!

michaelshobbs:
@u2mejc will you also be adding tests? 😍️😄️

plugins/storage/commands
```diff
@@ -0,0 +1,53 @@
...
+[[ " storage storage:list storage:mount storage:unmount storage:help help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/storage/functions"
+
+case "$1" in
+  storage:list)
+    # List all bound mounts
+    [[ ! $2 ]] && dokku_log_fail "storage:list requires an app to list."
+    unset APP
```
michaelshobbs: Why are we unsetting `APP` here?

michaelshobbs:
I'm concerned with the potential danger of this plugin interfacing with the same config files as the `docker-options` plugin. Additionally it seems there is some (at least logical) code duplication between the this and the `docker-options` plugin. 

Thoughts on re-using the `docker-options` plugin to implement this feature?

josegonzalez:
I'd only accept a merge if it reused docker-options.

u2mejc:
Totally can add tests. 

Since dokku is a wrapper for docker, and docker-options can easily do exactly what this plugin is doing, I especially tested it to make sure that they don't collide. I mean it's a txt file with one liners, the new plugin specifically -v lines. 

Not positive what you mean reuse? Like functions or you mean call dokku docker-options? That would saved me worlds of time. :stuck_out_tongue_closed_eyes: 

So strip all of the cases and do this:
```dokku docker-options:remove $APP deploy -v $MOUNT_PATH```

**Comment by michaelshobbs** [comment]
This plugin attempts to create a `/home/dokku/${APP}/DOCKER_OPTIONS_DEPLOY` file similar to [this code](https://github.com/dokku/dokku/blob/v0.4.9/plugins/docker-options/commands#L37). Thus the two plugins will interact with the same file in similar ways. More to the point one could use this plugin as well as the `docker-options` plugin to add `-v` args.

So instead of doing that with two different plugins, let's reuse [these functions](https://github.com/dokku/dokku/blob/v0.4.9/plugins/docker-options/commands#L37) and the like. That way if we want to change anything about that file, we can do so in one place. 
NOTE: you'll need to do a bit of reorganizing and move them to `plugins/docker-options/functions` and source that file in both plugins.

Then you can implement `storage:mount` and `storage:unmount` similar to `docker-options:add` and `docker-options:remove`
https://github.com/dokku/dokku/blob/v0.4.9/plugins/docker-options/commands#L124-L139

```
    get_passed_docker_option "$@"
    add_passed_docker_option passed_phases[@] "${passed_docker_option[@]}"
```
and
```
    get_passed_docker_option "$@"
    remove_passed_docker_option passed_phases[@] "${passed_docker_option[@]}"
```

Let me know if that makes sense or not.

**Comment by u2mejc** [comment]
> NOTE: you'll need to do a bit of reorganizing and move them to plugins/docker-options/functions and source that file in both plugins.

It's ok to cross reference one plugin's functions from other plugins outside common? Isn't that like taboo or something?

**Comment by michaelshobbs** [comment]
No we do that all over the place. The taboo bit is one plugin manipulating another's persistence layer. So we introduced `functions` to expose a plugin's functions to other plugins in case they want to get or set things another plugin is responsible for.

**Comment by josegonzalez** [comment]
@michaelshobbs we need to cleanup places where I used functions for private/public interfaces :P

**Comment by michaelshobbs** [comment]
@josegonzalez open a new issue with examples please?

**Comment by u2mejc** [comment]
Now reuses docker-option functions (separated into plugins/docker-options/functions) and adds bats tests for storage plugin.

File: plugins/storage/commands [hunk 2]
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+[[ " storage storage:list storage:mount storage:unmount storage:help help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/storage/functions"
+source "$PLUGIN_AVAILABLE_PATH/docker-options/functions"
```
michaelshobbs: Since a user cannot disable a core plugin, replace `PLUGIN_AVAILABLE_PATH` with `PLUGIN_CORE_AVAILABLE_PATH` when sourcing core plugin functions.

tests/unit/20_storage.bats
```diff
@@ -0,0 +1,50 @@
...
+
+@test "(storage) storage:list" {
+  run /bin/bash -c "dokku storage:mount $TEST_APP /tmp/mount:/mount"
+  run /bin/bash -c "dokku storage:list $TEST_APP | grep -qe '^\s*/tmp/mount:/mount'"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+}
+
+@test "(storage) storage:mount" {
```
michaelshobbs: `@test "(storage) storage:mount"` and `@test "(storage) storage:list"` seem to be doing the same thing. I would drop the first and relabel the second one to indicate it's doing both items.

u2mejc: The list is empty if you don't populate it first. The other tests rely on list returning a result, but also as you said, it doesn't add much time.

plugins/storage/commands
u2mejc: Can do.

tests/unit/20_storage.bats
michaelshobbs: Yeah I was literally thinking you could delete lines 13-20 and change line 21 to say `@test "(storage) storage:mount, storage:list"`

michaelshobbs: Or even just have a single test block that does all the things because you can't list or unmount without mount"

u2mejc:
Single test block and PLUGIN_AVAILABLE_PATH done. :+1:

plugins/docker-options/functions
```diff
@@ -0,0 +1,109 @@
...
+    fi
+  done
+  dokku_log_fail "Phase(s) must be one of [${phases[*]}]"
+}
+
+get_phase_file_path() {
+  local phase=$1
+  [[ $phase ]] || dokku_log_fail "Phase(s) must be one of [${phases[*]}]"
+  local prefix="DOCKER_OPTIONS_"
+  phase_file_path="${DOKKU_ROOT}/${APP}/${prefix}${phase^^}"
```
josegonzalez: We should turn this into not a global, and echo it if necessary.

michaelshobbs:
Excellent! and thanks! :+1:

u2mejc:
Updated PR pending https://github.com/dokku/dokku/pull/1870 merging and rebase! :+1:

josegonzalez:

ebeigarts:
It would be great if it could use some default host folder `dokku storage:mount <app> /app/uploads`. The default could be `/var/lib/dokku/storage/$APP/app/uploads`.

josegonzalez:
@u2mejc thoughts on that?

u2mejc:
I internally didn't add that as a feature, because I like to be explicit in my infrastructure.

But I don't really mind adding if you ( @josegonzalez ) disagree with me. Related: Docker has depricated host side arbitrary dirs that do not exist: [Auto-creating missing host paths for bind mounts](https://docs.docker.com/engine/deprecated/#auto-creating-missing-host-paths-for-bind-mounts)

josegonzalez:
Yeah you know what, explicit is better than implicit. We should instead provide guidance as to where people should place their files (to make backups easier). Thoughts on doing *that*? Our deb package/Makefile can `mkdir -p` that path :)

u2mejc:
Sweet! :tada: 
I've seen /var segregated to it's own partition to control /var/log's growth before, but otherwise I don't have a super strong opinion where to steer people.

josegonzalez:
`/var/lib` is usually where I see data placed...

michaelshobbs:
For example `/var/lib/dokku/storage/<APP>/uploads` perhaps?

josegonzalez:
Maybe `/var/lib/dokku/plugins/storage`?

michaelshobbs:
That directory gets modified via the `make copyfiles` quite a bit. I'd keep it out of the `plugins` sub dir. Not sure it makes sense there anyway.

josegonzalez:
oh hmm maybe not then. `storage` works then :P

ebeigarts:
For other official plugins we have

```
/var/lib/dokku/services/elasticsearch
/var/lib/dokku/services/mariadb
/var/lib/dokku/services/memcached
/var/lib/dokku/services/mongo
/var/lib/dokku/services/postgres
/var/lib/dokku/services/redis
```

As the plugin is called `storage`, I think it should go in `/var/lib/dokku/services/storage`.

Also I think there should be a default, so I don't have to give `root` access to the machine for users to create their folders. Also this will simplify backups (all apps will have storage in the same place).

```
ssh root@dokku.example.com "mkdir -p /var/lib/dokku/services/storage/<app>/app/uploads"
ssh dokku@dokku.example.com -- storage:mount /var/lib/dokku/services/storage/<app>/app/uploads :/app/uploads
```

vs

```
ssh dokku@dokku.example.com -- storage:mount /app/uploads
```

josegonzalez:
How about:

```
/var/lib/dokku/data/storage
```

Reason being so that we can store configuration for these in:

```
/var/lib/dokku/config/storage/APP_NAME
```

Similar to the spec here: https://github.com/dokku/dokku/issues/1558

michaelshobbs:
What's the actionable item here? Documentation? Code? Both?

josegonzalez:
@u2mejc Mind updating the docs to match the path I mentioned in my last comment? Specifically:

```
/var/lib/dokku/data/storage
```

As the prefix.

We should also have the plugin create the path in it's `install` pluginhook like so:

```shell
#!/usr/bin/env bash

main() {
  local storage_dir="${DOKKU_LIB_ROOT}/data/storage"

  if [[ ! -d "$storage_dir" ]]; then
    mkdir -p "$storage_dir"
    chown dokku:dokku "$storage_dir"
  fi
}

main "$@"
```

That chown is so that it's not owned by root (which can happen if run via the deb package). I don't know if it should be a recursive chown.

ebeigarts:
I saw that only "deploy" phase is used. Is there a reason why "run" is not included? I would expect that by running `dokku run <app> bash` I would able to read the mounted dir.

u2mejc:
I'll work on this and rebase at the same time.

@ebeigarts I'll look at adding run support.

u2mejc:
@ebeigarts I added run support too. :beer:

josegonzalez:
Looks good! Some comments:

- Version number should be bumped
- Still needs that `install` plugin trigger
- Needs docs

u2mejc:
@josegonzalez Working on docs, it doesn't require an install step?

u2mejc:
bumped version :+1:

josegonzalez:
@u2mejc to setup the "preferred" directory.

Also, please move those docs into this pr.

u2mejc:
> to setup the "preferred" directory.

Can do.
> Also, please move those docs into this pr.

That breaks the other PR's *feature* of documentation. I can squash both PRs into one if you like, but they're otherwise dependent on each other: Removing backup relies on documenting what to do instead. Adding storage plugin relies on docs from backups.

See here: https://github.com/u2mejc/dokku/blob/u2mejc-nobackup/docs/backup-recovery.md

josegonzalez:
I think we need separate documentation from backup recovery on how to use this plugin. If your other pr supplements that, thats cool, but this needs to be documented separately.

u2mejc:
Added doc. :+1:

docs/dokku-storage.md
```diff
@@ -0,0 +1,36 @@
+# Dokku Event Logs
```
josegonzalez: Wrong heading :P

plugins/storage/commands
```diff
@@ -0,0 +1,50 @@
...
+    # Remove bind-mount, restart app if running
+    passed_phases=(deploy run)
+    [[ -z $3 || $4 ]] && dokku_log_fail "storage:unmount requires an app and a two paths devided by colon."
+    verify_app_name "$2" && APP="$2"
+    verify_paths "$3" && MOUNT_PATH="$3"
+    check_if_path_exists "$MOUNT_PATH" "$(get_phase_file_path "${passed_phases[@]}")" || dokku_log_fail "Mount path does not exist."
+    remove_passed_docker_option passed_phases[@] "-v $MOUNT_PATH"
+    ;;
+
+  help | storage:help | storage)
```
josegonzalez: @michaelshobbs what do you think about making the default method here map to help? We don't follow that pattern anywhere else (though I'm not strictly opposed).

docs/dokku-storage.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Usage
+This example demonstrates how to mount the recommended directory to `/storage` inside the container:
+```
+$ dokku storage:mount app-name /var/lib/dokku/data/storage:/storage
+```
+Dokku will then mount the shared contents of`/var/lib/dokku/storage` to `/storage` inside the container.
+
+## Use Cases
+##### Persistent storage
```
josegonzalez: Can you add newlines before/after each heading?

plugins/storage/commands
michaelshobbs: I'd rather have uniformity here

u2mejc: +1 I'd like to see all plugin to this, instead of just error.

michaelshobbs: What standard are you proposing we implement here?

michaelshobbs: @u2mejc that question was for you

u2mejc: @michaelshobbs I would +1 more plugins displaying help when ran with no arguments

michaelshobbs:
Can you spec out the full standard? It's not that simple since some plugin commands do something functional when no args are given.

u2mejc:
Word, not sure I have time right now though.

michaelshobbs:
In the meantime then, let's not hold this up. Let's remove that default for now and we can revisit when we've thought it through more.

josegonzalez:
@u2mejc what @michaelshobbs said. We can merge once thats done :)

Issue #1831, opened 2016-01-06
Wrong .key files calculation


mbajur:
Hello,
i'm trying to install SSL cert for my app hosted with dokku. I'm following this tutorial: http://dokku.viewdocs.io/dokku/deployment/ssl-configuration/

I have two files, `server.crt` and `server.key` and i'm taring them with `tar cvf cert.tar server.crt server.key` which gives me:

![image](https://cloud.githubusercontent.com/assets/581569/12142447/680b25a8-b478-11e5-8b6c-a28d1f0a1ac7.png)

Next step is to upload that tar to my dokku app. So, i'm doing it with `dokku certs:add <app> < cert.tar` and dokku fails with a following output:

![image](https://cloud.githubusercontent.com/assets/581569/12142461/8aec6db6-b478-11e5-9803-22a1b551275b.png)

I'm not sure what's that `./._server.key` file doing there but it's not present after manually untaring cert.tar locally and running `ls -la`. I'm also not sure if that's the reason why dokku thinks there are two key files in the tar but no matter what, something is really not right in here.

Have you any ideas what's going on? Am i doing something wrong or i've just found some dokku bug?
Thanks in advance for any clues!

ps. i'm on **dokku 0.4.9**

jvanbaarsen:
Just tested this on my Dokku version (0.4.6) and the problem is there as well.

jvanbaarsen:
@mbajur On what machine are you creating the tar? OSX?

mbajur:
I'm in a horrible situation that my clients production app is hosted using dokku and i can not add SSL to it while it's a goal for today - is there any temporary workaround for this? Or am i totally done for? :D 

Yup, i'm creating my tar on OSX

jvanbaarsen:
Can you try taring with: `COPYFILE_DISABLE=1 tar -cvf cert-key.tar server.crt server.key`

jvanbaarsen:
OSX has the weird habit of adding `./._*` files to tar archives. I'll work on a workaround for this issue. For now, the above should help you out!

mbajur:
Great! Works like a charm, thanks :)

jvanbaarsen:
Glad I could help :-)

PR #1834, opened 2016-01-06
Make sure we ignore hidden files in the SSL cert check

### LINKED ISSUE #1832: Ignore osx's hidden files when untarring a CERT zip
See #1831 and http://unix.stackexchange.com/questions/9665/create-tar-archive-of-a-directory-except-for-hidden-files

jvanbaarsen:
OSX creates hidden files: `./._server.key` and `./._server.crt`. The Dokku cert
plugin is choking on those files. We should probably just ignore them, and go on
with our lives

Fixes #1832 
Ref #1831


michaelshobbs:
:+1: Thanks for adding tests too!! :heart_eyes:

PR #1835, opened 2016-01-06
Add support for more versions of docker-engine

josegonzalez:


jvanbaarsen:
Just curious, what would be the benefit/reasoning for this?

josegonzalez:
This fixes herokuish compatibility with lxc-docker installs, as well as tutum etc.

PR #1836, opened 2016-01-07
support scripts.dokku. in app.json

### LINKED ISSUE #1824: Create a plugin that runs certain commands after deployment
When you deploy  a Ruby on Rails application, and you have database changes, you need to migrate those, I can imagine that for certain other tech stacks "After deploy commands" are also common.

What do you think about a plugin that looks for a `after_deploy.sh` file in the `/app` directory, and if that file is present, run it. So you can have all kind of after_deploy code in there.

I see it working kinda the same as the `HEALTH_CHECK` file.

Looking forward to hear what you think.

michaelshobbs:
- supports scripts.dokku.predeploy and scripts.dokku.postdeploy in (/app/app.json in the app image)
- adds a couple json parsing utility functions
- adds `IMAGE_TAG` `post-deploy` plugn trigger call
- and oh yeah, adds a test and docs :P

closes #1824

### DIFF
--- a/docs/deployment/deployment-tasks.md
+++ b/docs/deployment/deployment-tasks.md
@@ -1,0 +1,55 @@
+# Deployment Tasks
+
+Sometimes you need to run a command on at deployment time, but before an app is completely deployed.
+Common use cases include:
+
+* Checking a database is initialised
+* Running database migrations
+* Any commands required to set up the server (e.g. something like a Django `collectstatic`)
+
+## `app.json` and `scripts.dokku`
+
+Dokku accomplishes this by using an `app.json` file. We (mostly) use the same format as Heroku's [app.json](https://devcenter.heroku.com/articles/app-json-schema).
+However, dokku currently only supports the nodes `scripts.dokku.predeploy` and `scripts.dokku.postdeploy`.
+Simply place an `app.json` file in the root of your repository.
+NOTE: `app.json` is only supported in buildpack-deployed apps and postdeploy changes are not committed to the app image.
+
+### Example app.json
+
+```
+{
+  "name": "barebones nodejs",
+  "description": "A barebones Node.js app using Express 4.",
+  "keywords": [
+    "nodejs",
+    "express"
+  ],
+  "repository": "https://github.com/michaelshobbs/node-js-sample",
+  "scripts": {
+    "dokku": {
+      "predeploy": "touch /app/predeploy.test",
+      "postdeploy": "touch /app/postdeploy.test"
+    }
+  },
+  "env": {
+    "SECRET_TOKEN": {
+      "description": "A secret key for verifying the integrity of signed cookies.",
+      "value": "secret"
+    },
+    "WEB_CONCURRENCY": {
+      "description": "The number of processes to run.",
+      "generator": "echo 5"
+    }
+  },
+  "image": "gliderlabs/herokuish",
+  "addons": [
+    "dokku-postgres",
+    "dokku-redis"
+  ],
+  "buildpacks": [
+    {
+      "url": "https://github.com/heroku/heroku-buildpack-nodejs"
+    }
+  ]
+}
+```
--- a/dokku
+++ b/dokku
@@ -207,7 +207,7 @@ case "$1" in
     done < "$DOKKU_SCALE_FILE"
 
     dokku_log_info1 "Running post-deploy"
-    plugn trigger post-deploy $APP $port $ipaddr
+    plugn trigger post-deploy $APP $port $ipaddr $IMAGE_TAG
 
     # kill the old container
     if [[ -n "$oldids" ]]; then
--- a/docs/development/plugin-triggers.md
+++ b/docs/development/plugin-triggers.md
@@ -405,7 +405,7 @@ dokku postgres:link $POSTGRES $APP
 
 - Description: Allows running of commands after a deploy has completed. Dokku core currently uses this to switch traffic on nginx.
 - Invoked by: `dokku deploy`
-- Arguments: `$APP $INTERNAL_PORT $INTERNAL_IP_ADDRESS`
+- Arguments: `$APP $INTERNAL_PORT $INTERNAL_IP_ADDRESS $IMAGE_TAG`
 - Example:
 
 ```shell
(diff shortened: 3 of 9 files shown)


dokku
```diff
@@ -195,7 +195,7 @@ case "$1" in
     done < "$DOKKU_SCALE_FILE"
 
     dokku_log_info1 "Running post-deploy"
-    plugn trigger post-deploy $APP $port $ipaddr
+    plugn trigger post-deploy $APP $port $ipaddr $IMAGE_TAG
```
josegonzalez: This should really be cleaned up imo. `port` and then `ip`?

docs/deployment/deployment-tasks.md
```diff
@@ -0,0 +1,54 @@
...
+Common use cases include:
+
+* Checking a database is initialised
+* Running database migrations
+* Any commands required to set up the server (e.g. something like a Django `collectstatic`)
+
+## `app.json` and `scripts.dokku`
+
+Dokku accomplishes this by using an `app.json` file. We (mostly) use the same format as Heroku's [app.json](https://devcenter.heroku.com/articles/app-json-schema).
+However, dokku currently only supports the nodes `scripts.dokku.predeploy` and `scripts.dokku.postdeploy`.
```
josegonzalez: Should we support `scripts.postdeploy` as well?

josegonzalez:
What happens for `Dockerfile` users that don't have bash in their containers?

dokku
michaelshobbs: wut?

docs/deployment/deployment-tasks.md
michaelshobbs: No

michaelshobbs:
Yeah the current implementation only supports buildpack apps anyway since we don't `nullglob` on sourcing files from `/app/.profile.d/*`. Not sure dockerfile support makes sense since `app.json` is a Herokuism anyway. Thoughts?

michaelshobbs:
Tests are failing because CircleCi did not include their `docker cp` patch in the `1.9.0-circleci` daemon binary we're using now. We can downgrade back to `1.8.2-circleci` but I'd like to see if they will patch the newer version first. I've opened a support ticket that asks for this.

docs/deployment/deployment-tasks.md
josegonzalez: Right, what I'm saying is should we _also_ implement their `scripts.postdeploy` hook at app creation?

josegonzalez:
Yeah lets drop the dockerfile reference.

dokku
josegonzalez: The `pre-deploy` hook is:

```shell
plugn trigger pre-deploy $APP $IMAGE_TAG
```

Maybe we should follow that and instead do:

```shell
plugn trigger pre-deploy $APP $IMAGE_TAG $ipaddr $port
```

Or similar? What uses the port/ipaddr from here that cannot retrieve it from some other helper?

michaelshobbs: Oh I see. This would break any `pre-deploy` plugins that used those arguments. While I agree with you that this should probably happen at some point, I don't think it should go in this PR.

josegonzalez: I meant `post-deploy` in my second example.

docs/deployment/deployment-tasks.md
michaelshobbs: Maybe. It feels a little confusing from a user perspective though. As that wasn't part of the original issue (#1824), I propose we punt on this one and see if we get any user feedback asking for it.

michaelshobbs:
Noooooo! :stuck_out_tongue_winking_eye:  This really doesn't need to wait for `0.5.0`.

dokku
michaelshobbs: Yeah I meant that too :)

josegonzalez:
I'd like to leave features up for 0.5.0 to push people to upgrade :)

michaelshobbs:
Bah! Ok :crying_cat_face:

josegonzalez:
Is there a good reason why we shouldn't incentivize updates?

michaelshobbs:
Meh?

josegonzalez:
Note, I don't mind doing it in 0.4.x if that's your wish.

michaelshobbs:
no. `0.5.0` works for me

michaelshobbs:
I've split out the commits that don't belong here into other PRs. Once those tests pass, I'll update this one.

michaelshobbs:
I'm wondering if `postdeploy` makes any sense. The way dokku deploys is illustrated below.
```
# copy app to container
docker run
docker wait
docker commit dokku/$APP

# trigger pre-build
# build
docker run
docker wait
docker commit dokku/$APP
# tigger post-build

# trigger pre-release
# release
docker run ...
docker wait
docker commit dokku/$APP
# trigger post-release

# trigger pre-deploy (run app.json: scripts.dokku.predeploy)
# deploy
docker run /start web (this is the deployed app container)
docker wait
docker commit dokku/$APP
# trigger check-deploy
# trigger post-deploy (run app.json: scripts.dokku.postdeploy)
```

If an app.json postdeploy script makes changes to the app image, the app will never see those changes because our container is already running with the previous image layer.

As far as I can tell, we have 2 options.
* drop scripts.dokku.postdeploy support
* don't commit the scripts.dokku.postdeploy container
  * committing this step ends up creating an image layer that is never used by the app and drops the image name from `docker ps` output
```
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
2fcdcdb6d725        f84f300e0aed        "/start web"        25 minutes ago      Up 25 minutes                           node-js-express.web.1
```

I'm leaning towards option 2. Thoughts?

josegonzalez:
Option 2

michaelshobbs:
Assuming tests pass, I think this is good to go.

josegonzalez:

jvanbaarsen:
:tada: Awesome! @michaelshobbs Thanks for working on this :heart:

josegonzalez:
Don't know why help generation is failing very weird, as master passes...

michaelshobbs:
Must have been something weird with your rebase. I rebased and forced-push. Tests pass now

josegonzalez:
Merge when ready :)

PR #1837, opened 2016-01-07
add back some deploy tests that test dokku functionality

michaelshobbs:
This adds back a few app deploy tests that exercise dokku-specific functionality. For example, when we moved to docker-1.9.0 and CircleCI assured me that `docker cp` was supported in that binary, we missed that it actually didn't work because I had removed the only test (deploy-test-checks-root) that used it. 

Unfortunately this is also true of the new feature in #1836. 

reference: https://discuss.circleci.com/t/unable-to-use-docker-cp-but-it-worked-2-days-ago/1137/8


michaelshobbs:
They got back to me. Testing the new version now. Then I'll rebase #1836.

PR #1843, opened 2016-01-07
Move let's encrypt plugin to official plugins

sseemayer:
Since dokku-let'sencrypt is a official plugin now (thanks @josegonzalez !), update the plugin list accordingly.


josegonzalez:
I'm always happy to make extremely useful, high-quality plugins official :)

PR #1844, opened 2016-01-08
named-containers: handle multiple old containers and don't attempt to rename a dead container

michaelshobbs:
- handles multiple "previous" containers given the name filter
- skips attempting to rename "dead" containers
- adds some output to indicate what the crap it's trying to do

this should solve the majority of the variance we've seen with this plugin


josegonzalez:
Good catch.

PR #1848, opened 2016-01-09
Move docker-options functions to functions file, rework phase_file_path

u2mejc:
@josegonzalez clean up per request in PR https://github.com/dokku/dokku/pull/1830

I pulled the docker-options functions out here too, so it won't be part of the first storage plugin commit when I rebase.

### DIFF
--- a/plugins/docker-options/functions
+++ b/plugins/docker-options/functions
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+AVAILABLE_PHASES=(build deploy run)
+
+get_phase_file_path() {
+  local phase_file_prefix="DOCKER_OPTIONS_"
+  echo "${DOKKU_ROOT}/${APP}/${phase_file_prefix}${phase^^}"
+}
+
+get_phases() {
+  local passed_phases_list="$1"
+  local -r phases_allowed=$(sed -e 's/ /\|/g' <<< "${AVAILABLE_PHASES[@]}")
+  local phase
+  local passed_phases
+  if [[ -n "$passed_phases_list" ]]; then
+    IFS=',' read -ra passed_phases <<< "$passed_phases_list"
+    for phase in "${passed_phases[@]}"; do
+      [[ "$phase" = @($phases_allowed) ]] || dokku_log_fail "Phase(s) must be one of [${AVAILABLE_PHASES[*]}]"
+    done
+    echo "${passed_phases[@]}"
+  fi
+}
+
+create_phase_file_if_required() {
+  local phase_file_path="$1"
+  [[ -f "$phase_file_path" ]] || {
+    touch "$phase_file_path"
+  }
+}
+
+display_phase_options() {
+  local phase="$1"
+  local phase_file_path="$2"
+  echo "${phase^} options:"
+  sed -e 's/^/    /' "$phase_file_path"
+}
+
+display_all_phases_options() {
+  local phases=("${AVAILABLE_PHASES[@]}")
+  for phase in "${phases[@]}"; do
+    local phase_file_path
+    phase_file_path="$(get_phase_file_path)"
+    if [[ -s "$phase_file_path" ]]; then
+       display_phase_options "$phase" "$phase_file_path"
+    fi
+  done
+}
+
+display_passed_phases_options() {
+  local passed_phases=("${!1}")
+  local phase
+  for phase in "${passed_phases[@]}"; do
+    local phase_file_path
+    phase_file_path="$(get_phase_file_path)"
+    if [[ ! -s "$phase_file_path" ]]; then
+      echo "${phase^} options: none"
+    else
+      display_phase_options "$phase" "$phase_file_path"
+    fi
+  done
+}
+
+add_passed_docker_option() {
+  local passed_phases=("${!1}")
+  shift
+  local passed_option_string="$*"
+  local phase
+  for phase in "${passed_phases[@]}"; do
+    local phase_file_path
+    phase_file_path="$(get_phase_file_path)"
+    create_phase_file_if_required "$phase_file_path"
+    echo "${passed_option_string}" >> "$phase_file_path"
+    all_phase_options="$(< "$phase_file_path")"
+    echo -e "${all_phase_options}" | sed '/^$/d' | sort -u > "$phase_file_path"
+  done
+}
+
+remove_passed_docker_option() {
+  local passed_phases=("${!1}")
+  shift
+  local passed_option_string="$*"
+  local phase
+  for phase in "${passed_phases[@]}"; do
+    local phase_file_path
+    phase_file_path="$(get_phase_file_path)"
+    [[ ! -s "$phase_file_path" ]] || {
+      all_phase_options="$(< "$phase_file_path")"
+      all_phase_options=$(echo -e "${all_phase_options}" | sed "s#^${passed_option_string}\$##")
+      echo -e "${all_phase_options}" | sed '/^$/d' | sort -u > "$phase_file_path"
+    }
+  done
+}
--- a/plugins/docker-options/commands
+++ b/plugins/docker-options/commands
@@ -1,118 +1,14 @@
 #!/usr/bin/env bash
-[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
+[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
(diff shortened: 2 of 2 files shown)


michaelshobbs:
I didn't follow in the other PR what was trying to be accomplished with regards to `phase_file_path`. Can you clarify here?

u2mejc:
docker-options was using a global var $phase_file_path, and it was being set inside nested functions. Both myself and Jose had to dig to find where it was being set. 

Basicly this was the change:
```bash
-    get_phase_file_path $phase
+    local phase_file_path=$(get_phase_file_path $phase)
```

michaelshobbs:
ah! thanks for the clarification

plugins/docker-options/functions
```diff
@@ -0,0 +1,109 @@
...
+    fi
+  done
+  dokku_log_fail "Phase(s) must be one of [${phases[*]}]"
+}
+
+get_phase_file_path() {
+  local phase=$1
+  [[ $phase ]] || dokku_log_fail "Phase(s) must be one of [${phases[*]}]"
+  local prefix="DOCKER_OPTIONS_"
+  phase_file_path="${DOKKU_ROOT}/${APP}/${prefix}${phase^^}"
```
josegonzalez: I think you're still setting this as a global var?

michaelshobbs:
@u2mejc yeah I don't see the diff you referenced either.
```
-    get_phase_file_path $phase
+    local phase_file_path=$(get_phase_file_path $phase)
```

u2mejc:
$ git my bad... Corrected and rebased. :sweat_smile:

plugins/docker-options/functions
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
+[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit $DOKKU_NOT_IMPLEMENTED_EXIT
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+get_app() {
+  [[ -z $1 ]] && dokku_log_fail "Please specify an app to run the command on"
+  verify_app_name "$1"
+  APP="$1"
```
josegonzalez: This is still global...

plugins/docker-options/functions
```diff
@@ -0,0 +1,109 @@
...
+      return 0
+    fi
+  done
+  dokku_log_fail "Phase(s) must be one of [${phases[*]}]"
+}
+
+get_phase_file_path() {
+  local phase=$1
+  local prefix="DOCKER_OPTIONS_"
+  echo "${DOKKU_ROOT}/${APP}/${prefix}${phase^^}"
```
u2mejc: This? Returns string from echo.

plugins/docker-options/functions
```diff
@@ -0,0 +1,109 @@
...
+  local phase=$1
+  local phase_file_path=$2
+  echo "${phase^} options:"
+  sed -e 's/^/    /' $phase_file_path
+}
+
+display_all_phases_options() {
+  local phases=(build deploy run)
+  for phase in "${phases[@]}"; do
+    local phase_file_path=$(get_phase_file_path $phase)
```
u2mejc: To set these.

plugins/docker-options/functions
josegonzalez: Set `APP` as a local var.

u2mejc:
@josegonzalez @michaelshobbs Looks like circle's lint'ing might be a little too confident of itself here, I linted the file before pushing (hence double quoting all of the naked vars). Want to take a look to see if we can merge this?

michaelshobbs:
```
root@dokku:~/dokku# make lint
# these are disabled due to their expansive existence in the codebase. we should clean it up though
# SC2034: VAR appears unused - https://github.com/koalaman/shellcheck/wiki/SC2034
# SC2086: Double quote to prevent globbing and word splitting - https://github.com/koalaman/shellcheck/wiki/SC2086
linting...

In ./plugins/docker-options/commands line 9:
PHASE_FILE_PATH='${DOKKU_ROOT}/${APP}/${PHASE_FILE_PREFIX}${phase^^}'
                ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

make: *** [lint] Error 123
```

u2mejc:
Thank you @michaelshobbs, did you read my message?

michaelshobbs:
Yes. Fix the lint issue.

u2mejc:
Ah sorry, I wasn't familiar with shellcheck disables, updating...

josegonzalez:
Uh, we don't want to disable it, we want to use double-quotes, no?

u2mejc:
I moved that path to the top of the file since instead of bearing it inside a function so it will be easy to update when we deprecate ~dokku. The script uses eval below.

michaelshobbs:
Given the choice to use `eval`, disabling that particular lint check makes sense. However, we want the `functions` file to be self-sufficient by default. So moving lines 7-9 in the `commands` file to `functions` would make that true.

Additionally, I think that means you can remove the use of eval and subsequently change the single-quotes to double-quotes. Feel free to correct me if I'm wrong there.

u2mejc:
One concern of @josegonzalez, that $APP was being set in some buried functioned, was rectified by setting it in the routine calling the function. 

Related, the path of the file(s) is now editable by simply updating one line at the top instead of multiple locations inside functions and will help us with the deprecation of ~dokku. The eval updates the path based on the APP name, the prefix, and the phase. I can remove the top var and move it back down to use double quotes but it adds more code / complexity to maintain.

u2mejc:
I'm working on a update to move the var to functions and correct the lint errors now.

plugins/docker-options/functions
```diff
@@ -0,0 +1,84 @@
...
+}
+
+display_phase_options() {
+  local phase="$1"
+  local phase_file_path="$2"
+  echo "${phase^} options:"
+  sed -e 's/^/    /' "$phase_file_path"
+}
+
+display_all_phases_options() {
```
josegonzalez: This should take in an `APP` as `$1`, and then use that to construct `$PHASE_FILE_PATH`.

plugins/docker-options/functions
```diff
@@ -0,0 +1,84 @@
...
+  local phases=("${AVAILABLE_PHASES[@]}")
+  for phase in "${phases[@]}"; do
+    local phase_file_path="$(eval echo $PHASE_FILE_PATH)"
+    if [[ -s "$phase_file_path" ]]; then
+       display_phase_options "$phase" "$phase_file_path"
+    fi
+  done
+}
+
+display_passed_phases_options() {
```
josegonzalez: Same basically everywhere...

plugins/docker-options/functions
michaelshobbs: Problem is `PHASE_FILE_PATH` is used in a for loop that defines `phase` which is a component of `PHASE_FILE_PATH`. I think moving towards a more readable version of this plugin would be best.

josegonzalez: Fair enough, but in this case `$APP` might not be set before the functions are sourced in, right?

michaelshobbs: I think we should bring back `get_phase_file_path()` and call it thusly:

```
...
  for phase in "${passed_phases[@]}"; do
    local phase_file_path=$(get_phase_file_path $phase)
...
```

josegonzalez: fine :(

michaelshobbs: @josegonzalez Can you enumerate your issue(s) with this suggestion?

josegonzalez: I just think that having a global variable makes things more difficult to read, and we should avoid them where possible.

michaelshobbs: In the proposed solution, where is there a global variable?

michaelshobbs: Oh! I should be more clear here.

```
get_phase_file_path() {
  local phase=$1
  local prefix="DOCKER_OPTIONS_"
  echo "${DOKKU_ROOT}/${APP}/${prefix}${phase^^}"
}
...
  for phase in "${passed_phases[@]}"; do
    local phase_file_path=$(get_phase_file_path $phase)
...
```

michaelshobbs: There's no additional global (`DOKKU_ROOT` and `APP` are global all over) in what I'm proposing.

josegonzalez: `APP` totally should not be global all over, which is what I'm trying to get away from here as well.

michaelshobbs: In the context of a given plugin, `APP` being set globally by a given case in commands is fine. The proposal to add `APP` as an arg to every function seems much too complex for the provided 'safety` you're looking to add.

josegonzalez: Okay, I guess as long as tests pass and you are happy, its fine by me.

u2mejc:
Updated to appease the shellcheck gawds. Added get_phase_file_path function per @michaelshobbs.

josegonzalez:
http://www.ingawdswetrust.com/

u2mejc:
Rebased to pick up circleci bash upgrade.

josegonzalez:
I'm pretty sure this broke tests for all the plugins, see [here](https://travis-ci.org/dokku/dokku-redis/builds/103014451). @michaelshobbs and I are investigating, and will probably put out a 0.4.13 release to fix the issue, since 0.4.12 didn't seem to :(

michaelshobbs:
The failed tests are due to the version of bash not matching what is on ubuntu 14.04. See #1862

PR #1854, opened 2016-01-10
Update year in footer (2013-2016)

fedosov:
Otherwise project seems abandoned.


josegonzalez:
Question: Did you actually think the project was abandoned because the copyright year date didn't include this year?

fedosov:
No, but only because I am visiting `dokku` repository often and can see date of the last commit. :)

PR #1855, opened 2016-01-11
Add skip_keyfile option for deb package

### LINKED ISSUE #1851: Unattended upgrade is not working
When running an Unattended upgrade, I still get the following question:

```
Configuring dokku
-----------------
Enable the web-based config page?
```

I use the following command: `$ sudo apt-get install -qq -y dokku`. Is this something I'm doing wrong?

jvanbaarsen:
There are cases where we don't want the check for keyfile, this will add a
config option, that allows you to skip this particular check.

Fixes #1851

### DIFF
--- a/debian/preinst
+++ b/debian/preinst
@@ -27,7 +27,8 @@ case "$1" in
       exit 0
     fi
 
-    if [ -z "${DEBCONF_RECONFIGURE}" ]; then
+    db_get "dokku/skip_key_file"
+    if [ -z "${DEBCONF_RECONFIGURE}" ] && [ "$RET" != "true" ]; then
       db_get "dokku/key_file"
       if [ ! -f "$RET" ]; then
           echo "Error: keyfile '$RET' not found."
--- a/debian/templates
+++ b/debian/templates
@@ -13,6 +13,11 @@ Description: Hostname or IP for server:
 Type: string
 Default: dokku.me
 
+Template: dokku/skip_key_file
+Description: Skip key file requirement?
+Type: boolean
+Default: false
+
 Template: dokku/key_file
 Description: Keyfile for initial user:
 Type: string
--- a/docs/getting-started/install/debian.md
+++ b/docs/getting-started/install/debian.md
@@ -37,4 +37,5 @@ After setting the desired options, proceed with the installation as described ab
 | dokku/web_config   | boolean | true                  | Use web-based config for below options                                   |
 | dokku/vhost_enable | boolean | false                 | Use vhost-based deployments (e.g. <app>.dokku.me)                        |
 | dokku/hostname     | string  | dokku.me              | Hostname, used as vhost domain and for showing app URL after deploy      |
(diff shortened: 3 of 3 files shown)


josegonzalez:
- What happens if this value isn't set?
- Doesn't this need to be handled somehow [here](https://github.com/dokku/dokku/blob/master/debian/config)?
- Do we need a debconf template for it [here](https://github.com/dokku/dokku/blob/master/debian/templates)?

jvanbaarsen:
@josegonzalez Let me check every one of those cases!

For the template, I only feel that is needed if we also document this feature, what do you think?

josegonzalez:
Oh yeah, this also needs to be documented.

jvanbaarsen:
:+1: On it!

jvanbaarsen:
@josegonzalez When I add the following to debian/config, the installation will break:

```
db_input "high" "dokku/skip_key_file" || false
```

Can I only put `true` on the right side?

josegonzalez:
Idk @3onyc implemented this. You're gonna have to go debian packaging spelunking :D

jvanbaarsen:
@josegonzalez Ok, what about I just make sure I get it to work for the old scenario, and the new scenario, If that works without touching the config, would that be ok for you?

josegonzalez:
Yeah. The following should work:

- Fresh install with web config
- Fresh install/Upgrade with debconf
- Fresh install/Upgrade with prompts

jvanbaarsen:
@josegonzalez Also, when I add it to the `template` file, it will ask in the GUI prompts if you want to skip the key_file. I feel that should not be there, right? Since its only for fully unattended installations?

josegonzalez:
If thats the case, I guess you can leave it off.

3onyc:
@jvanbaarsen If I recall correctly only items listed in `debian/config` get asked to the user, so `dokku/skip_key_file` shouldn't be in there. However it should be added to `debian/templates`

Also, looking at your use case, you're already asking for IP/port to install Dokku, can't you also prompt for a public key there?

jvanbaarsen:
@3onyc I'm skipping those questions as well with the config setters. Why should it be in debian/templates though?

jvanbaarsen:
@josegonzalez I feel like this is ready to be merged

3onyc:
@jvanbaarsen It's the standard file to define all debconf values in, if we omit it Debian packaging related utilities won't know it exists.

jvanbaarsen:
@3onyc I don't think that is a problem, since it mostly there for people that know what it is for. And not really for general use. @josegonzalez Opinion?

josegonzalez:
I would go with what @3onyc says.

jvanbaarsen:
:+1:

jvanbaarsen:
@josegonzalez Updated PR

josegonzalez:
@3onyc I'll wait on your decision before merging :)

jvanbaarsen:
@3onyc Bump :-)

3onyc:
Already merged, but :+1:.

jvanbaarsen:
@3onyc / @josegonzalez Thanks!

PR #1862, opened 2016-01-13
install bash 4.3.x on circleci

michaelshobbs:
needed by #1848 

```
$ bash --version
GNU bash, version 4.2.25(1)-release (x86_64-pc-linux-gnu)
$ bats tests/unit/20_docker-options.bats
...
11 tests, 6 failures
$ sudo apt-get install -y -q bash=$(apt-cache show bash | egrep "^Version: 4.3" | head -1 | awk -F: '{ print $2 }' | xargs)
...
Setting up bash (4.3-7ubuntu1.5) ...
...
$ bash --version
GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)
$ bats tests/unit/20_docker-options.bats
...
11 tests, 0 failures
```

### DIFF
--- a/tests/ci/parallel_runner.sh
+++ b/tests/ci/parallel_runner.sh
@@ -19,6 +19,9 @@ setup_circle() {
   sudo -E CI=true make -e install
   sudo -E make -e setup-deploy-tests
   make -e ci-dependencies
+  # circleci runs Ubuntu 12.04 and thus a previous version of bash (4.2) than 14.04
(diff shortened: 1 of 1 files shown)


u2mejc:
Thank you @michaelshobbs!

Here is the super old thread on the bash bug in question: http://lists.gnu.org/archive/html/bug-bash/2014-03/msg00065.html

josegonzalez:
Merge when tests pass :)

PR #1863, opened 2016-01-14
Fix a typo

znz:


michaelshobbs:
Eek! That's a major typo 😁️

Perhaps there's something wrong with the IPv6 tests?

josegonzalez:
Lmao. Welp merging.

michaelshobbs:
@josegonzalez i think this is the original commit: https://github.com/dokku/dokku/commit/467264d31f929c13134ecd95eed08cc0ef99eded#diff-24a947d0d17068c3d22fbc3fb60fa554

josegonzalez:
Welp that just means your original tests sucked :D

michaelshobbs:
Perhaps

michaelshobbs:
@znz did you hit a bug before fixing this? i'm curious because i'd like to fix the test suite to catch this. if so, can you share with us your configuration? (specifically any dokku config that contained IPv6 addresses. `/home/dokku/HOSTNAME`, `/home/dokku/VHOST`, `/home/dokku/$APP/VHOST`)

znz:
No, I did not hit this bug.
I found it when I watch `git log -p --stat`.

PR #1864, opened 2016-01-14
no longer need to set nullglob

znz:


michaelshobbs:
:+1: thanks for the contribution!

PR #1870, opened 2016-01-15
Remove arg check from docker-options/functions, global var cleanup

### LINKED ISSUE #1869: docker-options/functions has an arg check
docker-options/functions has an arg check which is preventing other plugins from sourcing it as it causes an exit.

```bash
[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
```

I git blame @michaelshobbs for not catching that :yum:

u2mejc:
Removes another global var and resolves #1869

### DIFF
--- a/plugins/docker-options/commands
+++ b/plugins/docker-options/commands
@@ -4,6 +4,8 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_CORE_AVAILABLE_PATH/docker-options/functions"
 
+declare APP # global assigned in case
+
 case "$1" in
   # Display applications docker options
   docker-options)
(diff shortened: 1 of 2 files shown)


plugins/docker-options/functions
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-[[ " docker-options docker-options:add docker-options:remove help docker-options:help " == *" $1 "* ]] || exit "$DOKKU_NOT_IMPLEMENTED_EXIT"
```
michaelshobbs: this will need to move to `commands`

u2mejc: agreed, one sec

plugins/docker-options/commands
```diff
@@ -4,6 +4,8 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
```
u2mejc: Look at line two above here... :smile:

michaelshobbs: ah!! github folding!!!!!

u2mejc:
This fixes a bug I introduced in the 0.4.x branch, and passed circle ci, I'm going to merge.

josegonzalez:
I'll make another release so that 0.4.x doesn't have an unstable final release.

u2mejc:
:+1:

PR #1871, opened 2016-01-18
support bash 4.2 so we don't have to modify all plugin test envs

michaelshobbs:
My thinking here is if we’re going to have a baseline bash version requirement, then we should have a damn good reason for it. Currently, there is no real need to support `bash>=4.3`; especially not for just one line of code.

Given that this PR is so small and still maintains `bash<4.3` support, I think it’s reasonable to maintain that support for now. If we end up finding ourselves having to go through uncomfortable machinations to support `bash>=4.2`, then we punt and force the `bash>=4.3` requirement.

### DIFF
--- a/plugins/docker-options/functions
+++ b/plugins/docker-options/functions
@@ -17,7 +17,9 @@ get_phases() {
   local phase
   local passed_phases
   if [[ -n "$passed_phases_list" ]]; then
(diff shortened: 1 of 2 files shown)


josegonzalez:
Closing, this is no longer necessary since i just updated all the envs...

Issue #1661, opened 2015-11-05
CHECKS run on non-web containers


mcolyer:
The [documentation](http://progrium.viewdocs.io/dokku/process-management/#the-web-proctype) seems to indicate that CHECKS should only happen on `web` procs, however from the output below (on 0.4.3) it seems to be running the CHECKS for every type:

```
-----> DOKKU_SCALE file found
=====> web=1
=====> worker=1
-----> Running pre-flight checks
       For more efficient zero downtime deployments, create a file CHECKS.
       See http://progrium.viewdocs.io/dokku/checks-examples.md for examples
       CHECKS file not found in container: Running simple container check...
-----> Waiting for 10 seconds ...
-----> Default container check successful!
=====> xxx container output:
=====> end xxx container output
-----> Running pre-flight checks
       For more efficient zero downtime deployments, create a file CHECKS.
       See http://progrium.viewdocs.io/dokku/checks-examples.md for examples
       CHECKS file not found in container: Running simple container check...
-----> Waiting for 10 seconds ...
-----> Default container check successful!
=====> xxx container output:
=====> end xxx container output
```

josegonzalez:
We run a basic "are the containers up" check on deploy. I'd have to look at the code to see what exactly we should be doing/documented as doing.

michaelshobbs:
We definitely run a default timeout-style check on all containers. 

https://github.com/progrium/dokku/blob/de832c596a92631f70f6c83eea4bce8a8efe59e5/plugins/checks/check-deploy#L82-L107

It should also be noted the documentation states dokku runs **custom checks** on the web proctype.
```
The web proctype is the only proctype that will invoke custom checks as defined by a CHECKS file.
```

mcolyer:
> We definitely run a default timeout-style check on all containers.

:+1:, thanks for clarifying.

josegonzalez:
We can definitely be more clear in the docs (and in fact just point people to the zero-downtime deploys page).

trevorturk:
I'd like to request an option to skip checks for non web containers. 

In my case, I'm checking that my Rails app is up, and the other process is a Resque worker. It's safe to fire up another worker in any case, so I'd like to skip the check. 

Maybe the CHECKS file should be divided by process type DOKKU_SCALE and we could set SKIP=true or something along those lines? 

(Sorry, I'm not sure where feature requests like this should go...)

josegonzalez:
@trevorturk mind opening up another issue and explaining how this feature might work?

PR #1872, opened 2016-01-19
Update log to better reflect what's going on with CHECKS

kenips:
I ran into the same confusion as #1661. Let's update the log?


kenips:
sorry, pushed something silly. fixed now!

michaelshobbs:
Seems reasonable to me. Thanks!

PR #1876, opened 2016-01-21
Document `cleanup` command

MarcDiethelm:
Noticed that this very useful command is not in `dokku help`. Don't know if this is on purpose, but here's a PR.

### DIFF
--- a/dokku
+++ b/dokku
@@ -251,6 +251,7 @@ case "$1" in
     echo "Options:"
 
     cat<<EOF | plugn trigger commands help | sort | column -c2 -t -s,
+    cleanup, Remove exited containers that are managed by dokku
     help, Print the list of commands
 EOF
     ;;


josegonzalez:
Purposeful. If you run that right after a deploy, the containers will force quit, dropping open connections.

Dokku already runs it at the beginning of every deploy so I don't see too much value in running this manually.

michaelshobbs:
Yeah I've gone back and forth in my head on this one as well. Still on the fence.

josegonzalez:
The problem is that those containers aren't exited during subsequent deploys, so many deploys in rapid succession _could_ cause memory/disk space issues. I think that is the better problem to solve, not exposing the plumbing behind dokku to users.

That said, how do we better communicate/deprecate old containers to users? Is there a way to "detect" that an app has `N` old containers and we expect `M`, and then warn users when it's getting bad, and force a stop when it's really bad?

josegonzalez:
We've decided internally that we don't want to document this.

Issue #1885, opened 2016-01-27
<command>:help hangs for certs, enter, tags, and tar


u2mejc:
Creating issue to reference in PR.

u2mejc:
Closing, PR submitted.

PR #1887, opened 2016-01-27
Fix <command>:help hangs for certs, enter, tags, tar

u2mejc:


josegonzalez:
Good catch, these have been broken for all of 0.4.x :P

u2mejc:
Fixes #1885

PR #1892, opened 2016-01-29
nginx-vhosts rewrite

### LINKED ISSUE #1629: nginx-vhosts expected behaviors
Over the length of time I've been using and contributing to dokku, the nginx functionality and  implementation complexity of said functionality has increased significantly. It doesn't seem (to me at least) that we have documented fully what the plugin is meant to do, what cases are automated, and what cases are left to the reader to implement with a custom template.

My personal position on this (and with software development in general, I think) is we should do a handful of things very very well and document the edge cases to give users the ability to implement them in their own way. (read: batteries included but replaceable)

This issue is an attempt to document and clarify what this plugin does and subsequently clean up the implementation to this end. I will enumerate my current understanding of this and willfully open the conversation to the community in hopes to solidify the expectation of functionality, and subsequently help prevent the historic regression cycle we've seen in this plugin.

### baseline assumptions
* nginx listener on port 80 
  * no-ssl
  * Loadbalancer (LB) with SSL termination. *not sure how this differs from no-ssl given our current templates*
* nginx listener on port 443 (ssl hosted on box with 301 redirection from port 80)
* some list of domains (`server_name` args) for which to answer
* custom nginx template may override any of the aforementioned scenarios using documented template variables
* dokku will validate (`nginx -t`) all nginx config before successfully deploying an app if VHOST support is enabled for said app

### VHOST/domain support
* local ssl configurations (i.e. non-ssl-terminated LBs) require nginx but **not** a domain (i.e. self-signed certs for development or internal services perhaps)
* non-ssl configurations **do not** require a domain or nginx
* `VHOST` support is on by default but can be disabled both explicitly or implicitly
  * (explicit) `NO_VHOST` env variable
    * nginx listens on random high port
    * container listens on external interface
  * (implicit) if either `$DOKKU_ROOT/VHOST` or `$DOKKU_ROOT/HOSTNAME` contain IPv4/IPv6 addresses, `NO_VHOST` is set for the deployed app
* nginx can be disabled completely
  * removes the app-specific nginx config and restart a container that listens on a random high port on the external interface of the host
* app-specific domains/VHOST support (`$DOKKU_ROOT/VHOST`) works even without a global VHOST (`$DOKKU_ROOT/VHOST`)
* all domains added with the domains plugin can be used to access the app
* default domain is `<appname>.$(cat $DOKKU_ROOT/HOSTNAME)`, if...
  * `$DOKKU_ROOT/$APP/VHOST` doesn't exist
  * `$DOKKU_ROOT/VHOST` and `$DOKKU_ROOT/HOSTNAME` do not contain IPv4 or IPv6 addresses
  * `$DOKKU_ROOT/VHOST` exists 
  * `nginx-hostname` plugin trigger returns nothing

### ssl specifics
* server-wide **AND** app-specific SSL certificates are supported
  * app-specifc certs take precedence over server-wide certs
* SSL certs with SANs are supported
* nginx will proxy the same app on all SSL domains (as found in the defined cert) **and** any additional non-SSL domains added with the domains plugin (without redirection)

I may have missed some cases but I think this illustrates how insane we've allowed our nginx support to become and I hope we can come to consensus on a more restrained scope of functionality or a more rational implementation. Now go ahead, talk amongst ya'selves.

refs: #1625 #1581 #1597 #1575 #674
those are just the open issues regarding nginx config.

michaelshobbs:
This is a work in progress, **PLEASE DO NOT MERGE.** I wanted to get this up for initial feedback. The following checklist will show current progress. **Please pick this apart!**
- [x] refactor into functions where possible; reduce global variables. needed access to some other plugin functionality, so I did this in those plugins as well
- [x] genericize enable/disable proxy and port variables
- [x] create proxy plugin
- [x] support proxy with no domains attached
- [x] install sigil via `Makefile`
- [x] remove magic `NO_VHOST` variable. explicitly use `domains:disable/domains:enable`
- [x] install sigil via package
- [x] re-implement nginx config template using [sigil](https://github.com/gliderlabs/sigil)
- [x] drop server-wide ssl certs
- [x] add UI to manage global vhost
- [x] with an ssl-enabled app, redirect all configured domains to 443
- [x] show warning if no configured domains are contained in the SSL certificate details
- [x] add ability to disable zero-downtime per app (added commands to checks plugin)
- [x] support multiple exposed ports for dockerfile apps
- [x] unify on one custom template name (`nginx.conf.sigil`)
- [x] only look for nginx template in app and discard after deploy (allows easy revert to default template)
- [x] look for custom template in `/app` or `WORKDIR`
- [x] usage docs
- [x] migration docs

open questions:
- [x] if we see a dockerfile with udp port exposures, do we automatically disable proxy or do we just leave it to the user? we will leave this to the user. the nginx-vhosts plugin will ignore udp ports. however, a different proxy implementation (haproxy perhaps) could be made to do this.
- [x] user configurable upstreams (this can be done with a custom template or we could add an include dir similar to this ` include {{ .DOKKU_ROOT }}/{{ .APP }}/nginx.conf.d/*.conf;` **this can be handled via a custom template now**

closes #1629


josegonzalez:
Can we list the udp and tcp ports separately? And pass them in somehow to the enabled proxy plugin. That way the plugin can do as it wants.

I think we disable the proxy plugin and treat it like a background app for our nginx implementation. Nothing else we can do anyhow.

michaelshobbs:
@josegonzalez I'm not clear on the context of your two sections there. Can you clarify please?

josegonzalez:
>  if we see a dockerfile with udp port exposures, do we automatically disable proxy or do we just leave it to the user?

michaelshobbs:
Ah ok. Haven't even touched dockerfile stuff yet but I play around with some test deploys and report back.

plugins/00_dokku-standard/functions
```diff
@@ -0,0 +1,43 @@
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/proxy/functions"
+
+get_app_urls() {
```
josegonzalez: Should we consolidate the `common` and `00_dokku-standard` plugins?

michaelshobbs: Probably.

plugins/nginx-vhosts/functions
```diff
@@ -93,147 +35,158 @@ restart_nginx() {
...
-    [[ "$(is_app_vhost_enabled $APP)" == "false" ]] && disable_app_vhost $APP --no-restart
+  tail "$NGINX_LOGS_ARGS" "$NGINX_LOGS_PATH"
+}
 
-    if [[ -z "$DOKKU_NGINX_PORT" ]]; then
-      if [[ "$(is_app_vhost_enabled $APP)" == "false" ]]; then
+configure_nginx_ports() {
+  local desc="configure nginx listening ports"
+  local APP=$1; verify_app_name "$APP"
+  local RAW_TCP_PORTS="$(get_app_raw_tcp_ports $APP)"
```
josegonzalez: Should we prefix this with DOKKU?

michaelshobbs: Not necessary as it's a local var, right?

plugins/nginx-vhosts/functions
```diff
@@ -93,147 +35,158 @@ restart_nginx() {
...
-      if [[ "$(is_ssl_enabled $APP)" == "app" ]]; then
-        SSL_DIRECTIVES=$(cat <<EOF
-ssl_certificate     $APP_SSL_PATH/server.crt;
-ssl_certificate_key $APP_SSL_PATH/server.key;
-EOF
-)
-      elif [[ "$(is_ssl_enabled $APP)" == "global" ]]; then
-        SSL_DIRECTIVES=""
-      fi
+  if (is_ssl_enabled $APP) && [[ -z "$DOKKU_NGINX_SSL_PORT" ]] && [[ -z "$RAW_TCP_PORTS" ]]; then
```
josegonzalez: Maybe assign this conditional to a variable. Not sure wtf is going on here.

plugins/nginx-vhosts/functions
michaelshobbs: My thinking is any var that gets set as a config variable should be prefixed but as a local var in a function, it's not necessary

plugins/nginx-vhosts/functions
```diff
@@ -251,13 +204,18 @@ EOF
...
     fi
 
-    if [[ "$(is_app_vhost_enabled $APP)" == "true" ]]; then
+    if ([[ -n "$NONSSL_VHOSTS" ]] || [[ -n "$SSL_VHOSTS" ]]) && [[ "$IS_APP_VHOST_ENABLED" == "true" ]]; then
       echo "# THIS FILE IS GENERATED BY DOKKU - DO NOT EDIT, YOUR CHANGES WILL BE OVERWRITTEN" > $URLS_PATH
-      xargs -i echo "https://{}" <<< "${SSL_VHOSTS}" >> $URLS_PATH
-      xargs -i echo "http://{}" <<< "${NONSSL_VHOSTS}" >> $URLS_PATH
+      if [[ -n "$RAW_TCP_PORTS" ]]; then
+        for p in $RAW_TCP_PORTS; do
+          xargs -i echo "$SCHEME://{}:${p}" <<< "$(echo ${SSL_VHOSTS} ${NONSSL_VHOSTS} | tr ' ' '\n' | sort -u)" >> $URLS_PATH
```
josegonzalez: Too bad we can't include `:` in the port, then we could just set RAW_TCP_PORTS to an empty string and consolidate this for with the else below.

plugins/checks/functions
```diff
@@ -0,0 +1,59 @@
...
+  if [[ -z "$DOKKU_CHECKS_ENABLED" ]] || [[ "$DOKKU_CHECKS_ENABLED" == "1" ]];then
+    local status=true
+  else
+    local status=false
+  fi
+
+  echo $status
+}
+
+checks_main() {
```
josegonzalez: Why are these functions here? I prefer to have them in the commands themselves unless they are actually sourced by other plugins.

plugins/certs/functions
```diff
@@ -22,22 +17,14 @@ is_ssl_enabled() {
...
-  case "$(is_ssl_enabled $APP)" in
-    app)
-      SSL_PATH="$DOKKU_ROOT/$APP/tls"
-    ;;
-
-    global)
-      SSL_PATH="$DOKKU_ROOT/tls"
-    ;;
-  esac
+  local SSL_PATH="$DOKKU_ROOT/$APP/tls"
```
josegonzalez: Fuck yeah.

plugins/00_dokku-standard/commands
```diff
@@ -199,6 +162,7 @@ case "$1" in
     docker version
     dokku_log_info1 "docker daemon info: "
     docker -D info
+    dokku_log_info1 "sigil version: $(sigil -v)"
```
josegonzalez: Remember to add this to our contributing doc and the wiki template.

plugins/00_dokku-standard/commands
```diff
@@ -149,46 +151,7 @@ case "$1" in
 
   url | urls)
```
josegonzalez: Is there a reason we need to support both?

plugins/nginx-vhosts/functions
michaelshobbs: I could move this to `get_app_urls`

plugins/config/functions
```diff
@@ -168,20 +170,20 @@ config_set() {
...
-    VALUE=$(echo ${var} | cut -d"=" -f2-)
+    local KEY=$(echo ${var} | cut -d"=" -f1)
+    local VALUE=$(echo ${var} | cut -d"=" -f2-)
 
     if [[ $KEY =~ [a-zA-Z_][a-zA-Z0-9_]* ]]; then
-      RESTART_APP=true
-      ENV_TEMP=$(echo "${ENV_TEMP}" | sed "/^export $KEY=/ d")
-      ENV_TEMP="${ENV_TEMP}
+      local RESTART_APP=true
+      local ENV_TEMP=$(echo "${ENV_TEMP}" | sed "/^export $KEY=/ d")
```
josegonzalez: does this redeclare ENV_TEMP (it's already declared above).

plugins/nginx-vhosts/commands
```diff
@@ -66,8 +21,6 @@ case "$1" in
     nginx:build-config <app>, (Re)builds nginx config for given app
     nginx:access-logs <app> [-t], Show the nginx access logs for an application (-t follows)
     nginx:error-logs <app> [-t], Show the nginx error logs for an application (-t follows)
-    nginx:disable <app>, Disable nginx for an application (forces container binding to external interface)
-    nginx:enable <app>, Enable nginx for an application
```
josegonzalez: How do we disable proxies now?

plugins/nginx-vhosts/functions
josegonzalez: gotcha, makes sense. What about the other local vars?

plugins/nginx-vhosts/commands
josegonzalez: nvm found it

plugins/checks/functions
josegonzalez: Ditto basically in every other plugin.

josegonzalez:
Other than my comments, looks good. What are features you would drop given the chance?

plugins/checks/functions
michaelshobbs: They very well could be sourced/used by community plugins

plugins/00_dokku-standard/commands
michaelshobbs: If you just want one url for your app. It's not that big of a deal IMO

plugins/config/functions
michaelshobbs: We're just overwriting like we did before

plugins/nginx-vhosts/functions
michaelshobbs: What do you mean?

plugins/checks/functions
josegonzalez: They aren't yet though.

michaelshobbs:
Dockerfile support :smiley:

josegonzalez:
lol no

plugins/checks/functions
michaelshobbs: Actually this functions file is sourced by `plugins/checks/check-deploy`

plugins/nginx-vhosts/functions
michaelshobbs: moved

plugins/certs/functions
michaelshobbs: :metal:

plugins/00_dokku-standard/commands
michaelshobbs: :+1:

plugins/checks/functions
josegonzalez: Right but this specific function is only used in the command file. In fact, most of the functions in this class aren't used outside the command file.

plugins/config/functions
josegonzalez: Does it rescope the override?

michaelshobbs: I don't believe so

michaelshobbs:
@josegonzalez I added a migration doc but not sure how you want to bubble it up. Feedback welcomed!

michaelshobbs:
I'll squash these commits once we have consensus that this is good to go

plugins/proxy/commands
```diff
@@ -0,0 +1,40 @@
...
+    ;;
+
+  proxy:disable)
+    [[ -z $2 ]] && dokku_log_fail "Please specify an app to run the command on"
+    proxy_disable $2 --no-restart
+    ;;
+
+  help | proxy:help)
+    cat<<EOF
+    ps <app>, Show proxy for app
```
u2mejc: s/ps/proxy

u2mejc:
@michaelshobbs What is the expected resulting when disabling proxy on apps with Dockerfiles that expose ports?

michaelshobbs:
We should restart the app with a container that exposes those ports

u2mejc:
I'm still poking around, but it looks like dockerfile ports are not being exposed as expected on push or proxy change.

michaelshobbs:
@u2mejc this is slightly broken still

```
root@dokku:~# git push dokku@dokku.me:dockertest master
...
-----> Cleaning up...
-----> Building dockertest from dockerfile...
-----> Setting config vars
       DOKKU_DOCKERFILE_PORTS: 3001/udp 3000/tcp 3003
remote: build context to Docker daemon 4.608 kB
Step 1 : FROM node:4
 ---> ce67537dc510
...
=====> Application deployed:
       http://dockertest.dokku.me:3000
       http://dockertest.dokku.me:3003
root@dokku:/tmp/dockerfile# dokku proxy:disable dockertest
-----> Disabling proxy for app (dockertest)
-----> Setting config vars
       DOKKU_DISABLE_PROXY: 1
-----> VHOST support disabled, deleting dockertest/VHOST
-----> VHOST support disabled, deleting dockertest/URLS
...
=====> Application deployed:
       http://dokku.me:32769 (container)
root@dokku:/tmp/dockerfile# docker ps
CONTAINER ID        IMAGE                     COMMAND                  CREATED             STATUS              PORTS                                                                       NAMES
8942bb82e290        dokku/dockertest:latest   "/bin/sh -c 'npm star"   2 minutes ago       Up About a minute   0.0.0.0:32769->3000/tcp, 0.0.0.0:32768->3001/udp, 0.0.0.0:32768->3003/tcp   dockertest.web.1
```

plugins/proxy/functions
```diff
@@ -0,0 +1,74 @@
...
+  local APP_PROXY_ENABLED=true
+
+  local DOKKU_DISABLE_PROXY=$(config_get $APP DOKKU_DISABLE_PROXY)
+  if [[ -n "$DOKKU_DISABLE_PROXY" ]]; then
+    local APP_PROXY_ENABLED=false
+  fi
+  echo $APP_PROXY_ENABLED
+}
+
+get_app_proxy_type() {
```
u2mejc: This function is always static, should we add a check for DOKKU_DISABLE_PROXY?

michaelshobbs: This functionality isn't fully fleshed out yet as we don't have another proxy yet.

u2mejc: bah why doesn't github actually show you context following the line you clicked?!

michaelshobbs: Updated `get_app_proxy_type()`

u2mejc:
>EDIT: Specifically the url. The ports won't be the same because we can't start up a new container with the same external ports but we should print out the ports correctly.

only if the ports are colliding though, correct? I think that's why I was asking. If a user is sacrificing proxy (zero downtime) for direct access, should they be able to assume they get the same port each time?

u2mejc:
Should we purge nginx logs when apps:destroy is used? Presently I can't look at the logs from dokku when the app is destroyed, but if I recreate it and disable proxy, I can still read the old logs using nginx:[ access error ]-logs. Also they eat space.

michaelshobbs:
I think that gets a bit complicated. Think about this flow....

- dokku disable:proxy app
  - removes nginx config
  - restarts app container with exposed ports

Currently subsequent pushes would end up changing the external port binding. In order to avoid that, we would have to stop the running container (obviously incurring some downtime), then start up a new one. This means we need to disable zero downtime checks if we have a dockerfile and no proxy enabled. Then re-enable them if the user turns the proxy back on. That's the least of my concern though. The biggest deal is we're gonna have another conditional for starting the deployed container that says if we're a dockerfile app and we have exposed ports, stop the running container and start this new one. 

I don't know that this edge usage case is worth all this extra logic. I'm open to contrary opinion.

michaelshobbs:
IMO, log management is up to the user.

josegonzalez:
If you are turning off the proxy in order to get a performance boost from direct access, I say you're also probably running something like `registrator` or whatever to do dns for your applications, in which case no, I don't think we need to keep the same ports for you. Your service discovery tool should handle that, and that's an edge case I don't want to consider. If people *really* want that, they can contribute it.

michaelshobbs:
:+1:

u2mejc:
> I don't know that this edge usage case is worth all this extra logic. I'm open to contrary opinion.

My thought is we should do it as we can't assume that a dokku user is using a service discovery process. If a user has an external proxy in front of the dokku host, they ~~may~~ will want a stable port they can always expect to be there. We simply document that ~~nginx~~ the proxy is dooku's way of enabling zero-downtime. 

Also, I'm happy to help if it's cool.

michaelshobbs:
@u2mejc see @josegonzalez's (comment)[https://github.com/dokku/dokku/pull/1892#issuecomment-183552892]
I think it's more a question of code complexity and is that increased complexity worth it for this use case. I think we should kick this can down the road and see how the community responds.

josegonzalez:
Disable the proxy == no zero-downtime. Ports just change. Want a light-weight way of having the same ports without the overhead of nginx or haproxy? Create a proxy plugin that does it via iptables.

I don't particularly like the idea of:

- check if new container passes checks.
- turn off all existing containers.
- start the new ones.

Sounds risky, people will complain that their apps don't work, and will be a mess to keep track of.

u2mejc:
I mean it was a "feature" of dokku years ago if there was port collision, the old container was simply shut down first. But I'll defer to you guys since I don't utilize the use case, so I have not vested interest.

josegonzalez:
Do you have docs for that feature?

plugins/proxy/functions
michaelshobbs: Had to back this out. If we check here then we don't clean up nginx artifacts and restart the app when running `proxy:disable`. Will sort this out later.

michaelshobbs:
Fixed `get_app_urls()` for non-proxied dockerfile apps with exposed ports.

This dockerfile
```
FROM node:4

EXPOSE 3001/udp
EXPOSE  3000/tcp
EXPOSE 3003

COPY . /app
WORKDIR /app
RUN npm install

CMD npm start
```

Yields these URLS:
```
       dokku.me:32779->3000/tcp (container)
       dokku.me:32773->3001/udp (container)
       dokku.me:32778->3003/tcp (container)
```

michaelshobbs:
Please let me know if I missed anything here

u2mejc:
> Do you have docs for that feature?
you have to go way back! like here: https://github.com/dokku/dokku/blob/cf7a2a00d46241feb7049b02b284de80b9d82964/dokku#L31
But I think it's a mute issue.

I'll poke around more today.

michaelshobbs:
@josegonzalez I'd like to drop the custom template path variable and support it similarly to how we handle the `Procfile` in #1915

josegonzalez:
What do you mean?

michaelshobbs:
Meaning we'll search for it in `/app` or `WORKDIR` (only in the case of dockerfile apps of course)

michaelshobbs:
I don't like the additional complexity it adds and we don't support this anywhere else. Additionally there is a community plugin to help store your "dokku-isms" in a subfolder and move them appropriately at deploy time.

https://github.com/mikexstudios/dokku-app-configfiles

josegonzalez:
Fine by me, just document it.

michaelshobbs:
Ah! That plugin is out of date but the point still stands about additional complexity and it being the only place we support this sort of custom pathing. Feels janky.

michaelshobbs:
@josegonzalez I'll need @pmclanahans's changes to `copy_from_image`

josegonzalez:
Time to break any installs of master!

MorrisJobke:
@sseemayer This is maybe interesting for the Let's encrypt plugin. I guess the plugin should be ported to the new templating.

PR #1897, opened 2016-02-04
Add vim swap files to .gitignore

u2mejc:
Yes, because I still use vim!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build
 tmp
 *.deb
 .ruby-version
+.*.sw[a-z]


josegonzalez:
Use your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

jvanbaarsen:
@u2mejc Why don't you turn of the swap creation? I have done so a while ago never missed it :D see: https://github.com/jvanbaarsen/dotfiles/blob/master/vimrc#L266. also YAY vim :tada:

u2mejc:
Agreed, the global is probably better.

@jvanbaarsen Because the swap is created for a reason, data recovery. I don't have vim creating backups (name~) while I'm working with git, but say my Mac :hankey: 's the bed, you can use vim -r to recover your working copy. Now I could move the swaps to another directory like /tmp, but I often simply open vim and paste in a working copy that I'm refactoring, so the working path gives me context. 

Saving is retro on a Mac (some apps don't have a save functions anymore), with background jobs, screen sessions, and buffers, I treat `:w` like committing changes. :+1: to vim!

josegonzalez:
@u2mejc Note, I normally wouldn't care except I'm #TeamNano.

u2mejc:
@josegonzalez Oh noz, this is why families shouldn't talk religion! :fire:

jvanbaarsen:
xD

PR #1904, opened 2016-02-10
Fix issue when Dockerfile present but BUILDPACK_URL is set

### LINKED ISSUE #1903: BUILDPACK_URL environment variable is ignored when Dockerfile is present
Hi all,

I'm experiencing an issues while trying to deploy [zendesk/samson](https://github.com/zendesk/samson) using buildpack instead of the bundled Dockerfile.

What I see is, that despite I have `BUILDPACK_URL: https://github.com/heroku/heroku-buildpack-ruby` set under the application, Dokku still tries to build the application with the bundled Dockerfile. [Gist](https://gist.github.com/cu12/9fbdc4ebe4d998df44a1)

As I can see in the [documentation](http://dokku.viewdocs.io/dokku/deployment/dockerfiles/) - point it out if I missed something - this is contrary to the desired behavior.

cu12:
Fixes  #1903

### DIFF
--- a/tests/unit/10_build-env.bats
+++ b/tests/unit/10_build-env.bats
@@ -44,6 +44,20 @@ teardown() {
   assert_success
 }
 
+@test "(build-env) buildpack if Dockerfile present" {
+  if [ $TEST_APP = dockerfile ]; then
+    run dokku config:set --no-restart $TEST_APP BUILDPACK_URL='https://github.com/heroku/heroku-buildpack-php'
+    echo "output: "$output
+    echo "status: "$status
+    assert_success
+
+    run deploy_app
+    echo "output: "$output
+    echo "status: "$status
+    assert_failure
+  fi
+}
+
 @test "(build-env) buildpack failure" {
   run dokku config:set --no-restart $TEST_APP BUILDPACK_URL='https://github.com/dokku/fake-buildpack'
   echo "output: "$output
@@ -55,3 +69,5 @@ teardown() {
   echo "status: "$status
   assert_failure
 }
+
+
--- a/plugins/git/commands
+++ b/plugins/git/commands
@@ -2,6 +2,7 @@
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/apps/functions"
+source "$PLUGIN_AVAILABLE_PATH/config/functions"
 
 git_build_app_repo() {
   verify_app_name "$1"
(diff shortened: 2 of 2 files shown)


josegonzalez:
I can't merge this as it doesn't use the config plugin to interact with the env file. Can you fix that?

cu12:
Sure, I wasn't checking hard enough to find the proper way I guess ;)

josegonzalez:
Search for `config_get` :)

cu12:
It seems to be that CircleCI died during the test, but the linting was ok.

josegonzalez:
@michaelshobbs should we have a test for this?

michaelshobbs:
Probably

josegonzalez:
@cu12 mind adding a test so we don't break this functionality in the future?

cu12:
Sure, no problem.

josegonzalez:
@cu12 *bump* :)

cu12:
I started to cook up something for this one, but I only have minimal to none experience with bats. Hold on :)

josegonzalez:
I leave the sense-making to @michaelshobbs (you'll notice I've never written a test for dokku ever).

tests/unit/10_build-env.bats
```diff
@@ -44,6 +44,20 @@ teardown() {
   assert_success
 }
 
+@test "(build-env) buildpack if Dockerfile present" {
```
michaelshobbs: I don't think this does what you think it does. Let me know what we're trying to do here and I'll slap something together.

cu12: @michaelshobbs: My way of thinking here was, that we know that the `Dockerfile` application is NodeJS, so if I specify the PHP buildpack it should fail. 

There might be a better solution to this, so if you have anything in your mind, please let me know.

michaelshobbs: The default app that `deploy_app` attempts to deploy does not contain a `Dockerfile` 
My understanding of this PR is to allow an app that contains a `Dockerfile` be deployed via buildpack if `BUILDPACK_URL` is set. In order to test that it seems like we should instead set `BUILDPACK_URL` to the nodejs buildpack, then call `deploy_app dockerfile` and test for success. Example:

```
@test "(build-env) buildpack deploy with Dockerfile" {
  run dokku config:set --no-restart $TEST_APP BUILDPACK_URL='https://github.com/heroku/heroku-buildpack-nodejs'
  echo "output: "$output
  echo "status: "$status
  assert_success

  run deploy_app dockerfile
  echo "output: "$output
  echo "status: "$status
  assert_success
}
```

Side note: Unless I'm missing something (which happens often depending on the day) this new test, does the same thing as the test below. Definitely correct me if I'm wrong there please. :smile: 

Also, apologies but you'll need to rebase this. We just did a big restructure for commands. I'm happy to take care of it if you'd prefer not to. Just let me know.

cu12: If I understand this correctly, the test above will also be a success if Dokku builds `$TEST_APP` using `Dockerfile`, ignoring the fact that it should build using the buildpack set in `BUILDPACK_URL` instead. 

I'd be glad if you could rebase this as I'm pretty much slow to respond here due to a deadline :)

michaelshobbs: Fair point about the test....I'll noodle on a better approach. Happy to rebase too. Now get back to "real" work :smile:

michaelshobbs:
Closing in favor of rebased #1980

PR #1911, opened 2016-02-12
bugfix: git url pattern

2mia:
github git urls format is "git@...". `dokku plugin:install` command will fail due to format not being matched.


michaelshobbs:
good catch

PR #1912, opened 2016-02-12
passing -i -t to docker exec only if tty present

mattberther:
When executing `dokku enter` from a crontab, I get the following error:

`cannot enable tty mode on non tty input`

Following a similar model to the run command in 00_dokku-standard/commands, this pull request should allow for `dokku enter` to work in a crontab.


michaelshobbs:
Thank you sir!

Issue #1564, opened 2015-10-15
Support Procfile commands with a Dockerfile-based deployment


pmclanahan:
I read through #1350 but I'm still a bit confused. I'd expect the following to work (dokku 4.1):

1. Have a Dockerfile deployment that works.
2. Add a Procfile that defines different commands for a `web` and a `worker` process type.
3. Deploy and have the `web` process scale to 1 using the command from the Procfile.
4. Scale the `worker` type to 1 using `ps:scale` and have another instance of the docker image started running the `worker` command from the Procfile.

Is this a thing that's possible? I'm unclear on why It'd be required to do this in the app/image itself since buildpack-based deploys do this exactly and are also docker images. If you do think it's possible w/o breaking some kind of other feature compatibility then I'm happy to work on it.

michaelshobbs:
As I'm sure you've found, that scenario does not currently work. I consider the dockerfile deployment as *beta* and could use more rfc/feature dev for sure. I think the current implementation is still my original "first pass" and to that point all assumptions were that the dockerfile author would dictate what gets run via the `RUN` verb. Please feel free to amend this (or a new issue) with what you're thinking, how it'll play with the rest of the feature set and let's flesh out what the next version of dockerfile support looks like.

Some notes on your current scenario.

1. If we remove the check for a dockerfile, we'll need a better way to determine which deployment method a user wants
2. seems reasonable
3. what if there is no `web` process. i think we currently handle this scenario anyway
4. i definitely like getting the same `ps` functionality with both deployment methods

any other dockerfile users have thoughts or comments on this?

pmclanahan:
Thanks for the thorough reply. That all seems quite reasonable and cool. 

> If we remove the check for a dockerfile, we'll need a better way to determine which deployment method a user wants

Wasn't suggesting that. I was more using that step as a baseline; a kind of "steps to reproduce". I can see how that sounded bad. The Dockerfile deploy works just fine now IMHO. I'm just wishing for the ability to alter the run command via the Procfile. Apologies for the confusing wording.

pmclanahan:
> what if there is no web process. i think we currently handle this scenario anyway

I think so, yeah. I think it'd be a required thing if you wanted a Procfile. Could potentially fall back to the CMD from the Dockerfile as the default for `web` if it's not there, but it seems quite reasonable to require a valid Procfile to contain a `web` type.

callahad:
[...]

PR #1915, opened 2016-02-14
Use Procfile for process types in Dockerfile apps

pmclanahan:
This is my initial attempt at fixing #1564. I'm opening a PR now to see how the tests in CI go and to get some feedback on direction. If I need to be doing something different I figured earlier was better than later to find that out.

Still todo:

- [x] Documentation of the feature
- [x] Tests for the feature

### DIFF
--- a/docs/deployment/dockerfiles.md
+++ b/docs/deployment/dockerfiles.md
@@ -39,3 +39,33 @@ dokku config:set APP DOKKU_DOCKERFILE_START_CMD="--harmony server.js"
 To tell docker what to run.
 
 Setting `$DOKKU_DOCKERFILE_CACHE_BUILD` to `true` or `false` will enable or disable docker's image layer cache. Lastly, for more granular build control, you may also pass any `docker build` option to `docker`, by setting `$DOKKU_DOCKER_BUILD_OPTS`.
+
+### Procfiles and Multiple Processes
+
+You can also customize the run command using a `Procfile`, much like you would on Heroku or
+with a buildpack deployed app. The `Procfile` should contain one or more lines defining [process
+types and associated commands](https://devcenter.heroku.com/articles/procfile#declaring-process-types).
+When you deploy your app a Docker image will be built, the `Procfile` will be extracted from the image
+(it must be in the folder defined in your `Dockerfile` as `WORKDIR` or `/app`) and the commands
+in it will be passed to `docker run` to start your process(es). Here's an example `Procfile`:
+ 
+```
+web: bin/run-prod.sh
+worker: bin/run-worker.sh
+```
+
+And `Dockerfile`:
+
+```
+FROM debian:jessie
+WORKDIR /app
+COPY . ./
+CMD ["bin/run-dev.sh"]
+```
+
+When you deploy this app the `web` process will automatically be scaled to 1 and your Docker container
+will be started basically using the command `docker run bin/run-prod.sh`. If you want to also run
+a worker container for this app, you can run `dokku ps:scale worker=1` and a new container will be
+started by running `docker run bin/run-worker.sh` (the actual `docker run` commands are a bit more
+complex, but this is the basic idea). If you use an `ENTRYPOINT` in your `Dockerfile`, the lines
+in your `Procfile` will be passed as arguments to the `ENTRYPOINT` script instead of being executed.
--- a/dokku
+++ b/dokku
@@ -137,7 +137,9 @@ case "$1" in
 
         if [[ -z "$DOKKU_HEROKUISH" ]]; then
           DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
-          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
+          DOKKU_DOCKERFILE_START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || true)
+          DOKKU_PROCFILE_START_CMD=$(get_cmd_from_procfile $APP $PROC_TYPE)
+          START_CMD=${DOKKU_DOCKERFILE_START_CMD:-$DOKKU_PROCFILE_START_CMD}
         fi
 
         if [[ "$PROC_TYPE" == "web" ]]; then
--- a/plugins/ps/functions
+++ b/plugins/ps/functions
@@ -9,8 +9,31 @@ print_dokku_scale_file() {
   done < "$DOKKU_SCALE_FILE"
 }
 
+extract_procfile() {
+  local APP="$1"
+  local IMAGE_TAG="$(get_running_image_tag $APP)"
+  local IMAGE="$(get_app_image_name $APP $IMAGE_TAG)"
+  local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
+  verify_app_name "$APP"
+
+  copy_from_image "$IMAGE" "Procfile" "$DOKKU_PROCFILE" 2>/dev/null || true
+  if [[ -f "$DOKKU_PROCFILE" ]]; then
+    dokku_log_info1_quiet "App Procfile file found ($DOKKU_PROCFILE)"
+  else
+    dokku_log_info1_quiet "No Procfile found in app image"
+  fi
+}
+
+remove_procfile() {
+  local APP="$1"; local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
+  if [[ -f "$DOKKU_PROCFILE" ]]; then
+    rm -f "$DOKKU_PROCFILE"
+  fi
+}
+
 generate_scale_file() {
   local APP="$1"; local IMAGE_TAG="$2"; local IMAGE=$(get_app_image_name $APP $IMAGE_TAG); local DOKKU_SCALE_FILE="$DOKKU_ROOT/$APP/DOKKU_SCALE"
+  local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
   verify_app_name "$APP"
 
   copy_from_image "$IMAGE" "/app/DOKKU_SCALE" "$DOKKU_ROOT/$APP" 2>/dev/null || true
@@ -20,18 +43,18 @@ generate_scale_file() {
     TMP_WORK_DIR=$(mktemp -d)
     trap 'rm -rf "$TMP_WORK_DIR" > /dev/null' RETURN
 
-    copy_from_image "$IMAGE" "/app/Procfile" "$TMP_WORK_DIR" 2>/dev/null || true
-    local PROCFILE="$TMP_WORK_DIR/Procfile"
-    if [[ ! -e "$PROCFILE" ]]; then
-      echo "web=1" >> $DOKKU_SCALE_FILE
-    else
+    if [[ -f $DOKKU_PROCFILE ]]; then
       while read -r line || [[ -n "$line" ]]; do
-        [[ "$line" =~ ^#.* ]] && continue
+        if [[ -z "$line" ]] || [[ "$line" == "#"* ]]; then
+          continue
+        fi
         NAME=${line%%:*}
         NUM_PROCS=0
         [[ "$NAME" == "web" ]] && NUM_PROCS=1
         [[ -n "$NAME" ]] && echo "$NAME=$NUM_PROCS" >> $DOKKU_SCALE_FILE
-      done < "$PROCFILE"
+      done < $DOKKU_PROCFILE
+    else
+      echo "web=1" >> $DOKKU_SCALE_FILE
     fi
     dokku_log_info1_quiet "New DOKKU_SCALE file generated"
   else
--- a/plugins/ps/post-deploy
+++ b/plugins/ps/post-deploy
@@ -5,4 +5,5 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
 APP="$1"
 
+remove_procfile "$APP"
 dokku config:set --no-restart $APP DOKKU_APP_RESTORE=1
--- a/plugins/ps/pre-deploy
+++ b/plugins/ps/pre-deploy
@@ -5,4 +5,5 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
 APP="$1"
 
-generate_scale_file $APP
+extract_procfile "$APP"
+generate_scale_file "$APP"
--- a/tests/unit/10_ps-dockerfile.bats
+++ b/tests/unit/10_ps-dockerfile.bats
@@ -116,3 +116,128 @@ teardown() {
   echo "status: "$status
   assert_success
 }
+
+@test "(ps) dockerfile with procfile" {
+  deploy_app dockerfile-procfile
+  run bash -c "dokku ps:stop $TEST_APP"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  for CID_FILE in $DOKKU_ROOT/$TEST_APP/CONTAINER.*; do
+    run bash -c "docker ps -q --no-trunc | grep -q $(< $CID_FILE)"
+    echo "output: "$output
+    echo "status: "$status
+    assert_failure
+  done
+
+  run bash -c "dokku ps:start $TEST_APP"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  for CID_FILE in $DOKKU_ROOT/$TEST_APP/CONTAINER.*; do
+    run bash -c "docker ps -q --no-trunc | grep -q $(< $CID_FILE)"
+    echo "output: "$output
+    echo "status: "$status
+    assert_success
+  done
+
+  run bash -c "dokku ps:restart $TEST_APP"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  for CID_FILE in $DOKKU_ROOT/$TEST_APP/CONTAINER.*; do
+    run bash -c "docker ps -q --no-trunc | grep -q $(< $CID_FILE)"
+    echo "output: "$output
+    echo "status: "$status
+    assert_success
+  done
+
+  run bash -c "dokku ps:rebuild $TEST_APP"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  for CID_FILE in $DOKKU_ROOT/$TEST_APP/CONTAINER.*; do
+    run bash -c "docker ps -q --no-trunc | grep -q $(< $CID_FILE)"
+    echo "output: "$output
+    echo "status: "$status
+    assert_success
+  done
+}
+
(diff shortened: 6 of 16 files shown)


josegonzalez:
Tests pass. Can you add docs for this so I can compare it to what we think it should do/our current handling of dockerfile deploys? This is obviously a breaking change, so I'm curious as to what the impact might be on our users.

pmclanahan:
Will do. I'll try to get that done tonight.

pmclanahan:
I could also make this feature off by default and enabled via an environment variable. So something like `dokku config:set DOKKU_USE_PROCFILE=true` would turn this on? Maybe that would allow for a more smooth roll-out?

josegonzalez:
No, I think we should roll it out by default. We break functionality in 0.5.x, so no need to keep BC. Ideally if it works one way on heroku, it works that way on dokku. 

What should happen if we define an entry point?

pmclanahan:
The way it's written is that the entries in the Procfile would be run as arguments to the entrypoint script. I think as long as it's documented that providing a Procfile with a Dockerfile deployment means that the commands in the Procfile are used as the argument to `docker run` then it should make sense to people and they can use that to their advantage. I'm not sure how it would work otherwise. I could run an entrypoint experiment on heroku to check though.

mmerickel:
> We break functionality in 0.5.x, so no need to keep BC.

All I request is that so long as there's no `Procfile` at the root of the repo we keep things working as-is. :-)

pmclanahan:
@mmerickel that is the idea for sure.

@josegonzalez my first pass at documentation is in.

pmclanahan:
@mmerickel and since it actually won't have any effect unless it's in the docker image, you could add `Procfile` to `.dockerignore` and still have a `Procfile` in your repo if you so desired and it wouldn't affect the dokku deployment.

plugins/ps/pre-deploy
```diff
@@ -5,4 +5,5 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
 APP="$1"
 
+get_procfile $APP
```
michaelshobbs: We don't need a herokuish app's Procfile. Can we only extract the `Procfile` in the necessary case?

pmclanahan: I switched the `generate_scale_file` function to use the local `Procfile` extracted by this step, so it is used in either case. I figured it'd be the best thing since it was being extracted by that command anyway but thrown away.

michaelshobbs: Ah I missed that. Personally, I'd like to get away from cluttering the `/home/dokku/APP` directory. However, I believe that concern is being addressed in another issue (#1558)

pmclanahan: That said, if you'd prefer I can revert `generate_scale_file` changes and only do this for dockerfile apps, but it would mean double extraction for dockerfile apps to generate the scale file and save the procfile for use by `dokku deploy`.

pmclanahan:
I tried an experiment on Heroku with a Dockerfile with an `ENTRYPOINT` and I can't get it to successfully deploy at all. They, unlike dokku, impose some very strict rules on Dockerfiles for use with Heroku. They don't use Docker; their CLI tool extracts a slug from a local docker image and sends it to their API. So I guess they don't support `ENTRYPOINT` at all.

plugins/ps/pre-deploy
michaelshobbs: I'll defer to @josegonzalez

michaelshobbs:
Yeah that's the conversation I'd like to have on #1564

pmclanahan:
I personally feel it's a heroku bug rather than how they'd like it to work. They don't (can't) support the full docker feature-set, but dokku can do a better job. If we're going to ignore Procfiles for all but buildpack deployments then we need a better solution than the `DYNO` variable, but it could be as simple as a `PROCESS_TYPE` variable.

michaelshobbs:
@pmclanahan would you mind bringing this discussion to #1564?

pmclanahan:
Just did

michaelshobbs:
Gracias!

pmclanahan:
Just added `DOKKU_PROC_TYPE` variable as well. I can take or leave it, but added it in a separate commit as an easier alternative to using `DYNO`. Could go with just that or document how best to use `DYNO`. Could also just be `DYNO_TYPE` since the history file says that `DYNO_TYPE_NUMBER` should exist, though I can't find a reference to it in the code.

pmclanahan:
Ah. NM. I see that was later removed and `DYNO` became what `DYNO_TYPE_NUMBER` was to be more heroku compatable.

josegonzalez:
@pmclanahan Yeah, remove that please :)

pmclanahan:
gone

dokku
```diff
@@ -137,7 +137,13 @@ case "$1" in
...
         if [[ -z "$DOKKU_HEROKUISH" ]]; then
           DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
-          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
+          if [[ -z "$START_CMD" ]]; then
+            START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
+          fi
+          if [[ -z "$START_CMD" ]]; then
+            # get cmd from Procfile
+            START_CMD=$(get_cmd_from_procfile $APP $PROC_TYPE)
+          fi
```
michaelshobbs: Maybe shorten this block to:
```
DOKKU_DOCKERFILE_START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
DOKKU_PROCFILE_START_CMD=$(get_cmd_from_procfile $APP $PROC_TYPE)
START_CMD=${DOKKU_DOCKERFILE_START_CMD:="$DOKKU_PROCFILE_START_CMD"}
```

josegonzalez: `config_get`, not `dokku config:get`.

michaelshobbs: Yeah that's the new standard. I don't think that code has been refactored yet

```
DOKKU_DOCKERFILE_START_CMD=$(config_get $APP DOKKU_DOCKERFILE_START_CMD)
DOKKU_PROCFILE_START_CMD=$(get_cmd_from_procfile $APP $PROC_TYPE)
START_CMD=${DOKKU_DOCKERFILE_START_CMD:-$DOKKU_PROCFILE_START_CMD}
```

michaelshobbs: You'll also need to add this line under the source command for `common/functions`

```
source "$PLUGIN_ENABLED_PATH/common/functions"
```

pmclanahan: I like that. It's updated.

dokku
```diff
@@ -137,7 +137,9 @@ case "$1" in
 
         if [[ -z "$DOKKU_HEROKUISH" ]]; then
           DOKKU_DOCKERFILE_PORT=$(dokku config:get $APP DOKKU_DOCKERFILE_PORT || true)
-          START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD || $START_CMD)
+          DOKKU_DOCKERFILE_START_CMD=$(dokku config:get $APP DOKKU_DOCKERFILE_START_CMD)
```
michaelshobbs: @josegonzalez mentioned switching this to `config_get`. however, the main `dokku` script hasn't been refactored for this in mind and the `dokku config:get` pattern is all over the place. I vote we don't include that change in this PR.

josegonzalez: Fair enough.

dokku
pmclanahan: > You'll also need to add this line under the source command for common/functions

You mean up around line 35? Does that not include the functions from the common plugin?

dokku
josegonzalez: @pmclanahan ^

plugins/ps/functions
```diff
@@ -20,18 +33,16 @@ generate_scale_file() {
...
 
-    copy_from_image "$IMAGE" "/app/Procfile" "$TMP_WORK_DIR" 2>/dev/null || true
-    local PROCFILE="$TMP_WORK_DIR/Procfile"
-    if [[ ! -e "$PROCFILE" ]]; then
-      echo "web=1" >> $DOKKU_SCALE_FILE
-    else
+    if [[ -f $DOKKU_PROCFILE ]]; then
       while read -r line || [[ -n "$line" ]]; do
-        [[ "$line" =~ ^#.* ]] && continue
+        if [[ -z "$line" ]] || [[ "$line" == "#"* ]]; then continue; fi
```
michaelshobbs: Not sure this change is necessary. Can you clarify?

michaelshobbs: Also one-line if blocks are frowned upon by @progrium :smile:

pmclanahan: Sure, the old way wasn't skipping empty lines. This also is what was in that other bask procfile parser (shoreman). I'm happy to make it a well indented if block though :)

michaelshobbs: Yes please and squash. Then I think we're good to go.

plugins/ps/functions
```diff
@@ -9,8 +9,21 @@ print_dokku_scale_file() {
   done < "$DOKKU_SCALE_FILE"
 }
 
+get_procfile() {
+  local APP="$1"; local IMAGE_TAG="$3"; local IMAGE=$(get_app_image_name $APP $IMAGE_TAG); local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
```
michaelshobbs: Also, why is `IMAGE_TAG` set to `$3` and not `$2`? We never call `get_procfile()` with more than 1 arg. Additionally that brings up the question about supporting image tags.

michaelshobbs: I think what you want is this:
```
local APP="$1"
local IMAGE_TAG="$(get_running_image_tag $APP)" 
local IMAGE=$(get_app_image_name $APP $IMAGE_TAG)
local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
```

pmclanahan: I wasn't sure about that `$3` either. I got it from the function below that, `generate_scale_file` which had similar input requirements. I'll update to what you have above though.

michaelshobbs: Ah yeah that's a typo. My bad.

pmclanahan: Should I also update `generate_scale_file` to use `get_running_image_tag`?

michaelshobbs: Ah now that you mention it, just switch it to be `$2`
i.e.
```
local APP="$1"
local IMAGE_TAG="$2" 
local IMAGE=$(get_app_image_name $APP $IMAGE_TAG)
local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
```

michaelshobbs: ref: #1923

pmclanahan:
I think we're good here as soon as tests pass again.

docs/deployment/dockerfiles.md
```diff
@@ -39,3 +39,34 @@ dokku config:set APP DOKKU_DOCKERFILE_START_CMD="--harmony server.js"
 To tell docker what to run.
 
 Setting `$DOKKU_DOCKERFILE_CACHE_BUILD` to `true` or `false` will enable or disable docker's image layer cache. Lastly, for more granular build control, you may also pass any `docker build` option to `docker`, by setting `$DOKKU_DOCKER_BUILD_OPTS`.
+
+### Procfiles
```
josegonzalez: `Procfiles and Multiple Processes`

docs/deployment/dockerfiles.md
```diff
@@ -39,3 +39,34 @@ dokku config:set APP DOKKU_DOCKERFILE_START_CMD="--harmony server.js"
...
+
+You can also customize the run command using a `Procfile`, much like you would with Heroku or
```
josegonzalez: `on Heroku`

docs/deployment/dockerfiles.md
```diff
@@ -39,3 +39,34 @@ dokku config:set APP DOKKU_DOCKERFILE_START_CMD="--harmony server.js"
...
+WORKDIR /app
+COPY . ./
+CMD ["bin/run-dev.sh"]
+```
+
+When you deploy this app the `web` process will automatically be scaled to 1 and your Docker container
+will be started basically using the command `docker run bin/run-prod.sh`. If you want to also run
+a worker container for this app, you can run `dokku ps:scale worker=1` and a new container will be
+started by running `docker run bin/run-worker.sh` (the actual `docker run` commands are a bit more
+complex, but this is the basic idea). So if you use an `ENTRYPOINT` in your `Dockerfile` the lines
```
josegonzalez: Missing comma after `Dockerfile`

docs/deployment/dockerfiles.md
```diff
@@ -39,3 +39,34 @@ dokku config:set APP DOKKU_DOCKERFILE_START_CMD="--harmony server.js"
...
+in your `Procfile` will be passed as arguments to the `ENTRYPOINT` script instead of being executed since
```
josegonzalez: Drop the `since this is how docker's run command works` bit.

docs/deployment/dockerfiles.md
josegonzalez: Remove the `So` part of this sentence.

plugins/ps/pre-deploy
josegonzalez: Can we delete the file if it's from a dockerfile deploy?

pmclanahan: Not sure what you mean @josegonzalez. We need it in both cases for generating the scale file, and especially in dockerfile deploy for discovering what commands to run.

josegonzalez: I meant in the `/home/dokku/APP` directory, and only extract it when necessary.

pmclanahan:
Rebased to fix merge conflict and updated docs based on feedback.

plugins/ps/pre-deploy
pmclanahan: Could probably do that, or could go back to never storing it and reading it from the image every time. It just seemed like a useful pattern to keep info necessary for starting containers in app dir like the SCALE file is.

michaelshobbs:
We keep the scale file because we can set it via the command line as well as ship it with the app. No need to keep the Procfile around really

pmclanahan:
So maybe keep things as they are but remove the file in `plugins/ps/post-deploy`?

plugins/ps/post-deploy
```diff
@@ -4,5 +4,8 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
 APP="$1"
+DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
+
+[ -f "$DOKKU_PROCFILE" ] && rm -f "$DOKKU_PROCFILE"
```
pmclanahan: @josegonzalez @michaelshobbs like so?

plugins/ps/post-deploy
```diff
@@ -4,5 +4,8 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
 APP="$1"
+DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
+
+[[ -f "$DOKKU_PROCFILE" ]] && rm -f "$DOKKU_PROCFILE"
```
michaelshobbs: Is there a simple way to do this in `pre-deploy`?

pmclanahan: The file is primarily needed in the deploy command.

michaelshobbs: Sorry I meant to say in the deploy phase. IIRC, you call `get_cmd_from_procfile()` in `dokku`'s deploy phase. Thus, perhaps we can do everything in one swing as opposed to extracting the `Procfile` in `pre-deploy`, using it in `deploy` and removing it in `post-deploy`.

What do you think?

pmclanahan: It's extracted in pre-deploy currently because `generate_scale_file` which is in pre-deploy now uses the extracted file. Thus it makes sense to do cleanup in post-deploy. We could of course move all of this into deploy in the `dokku` executable, but I don't see a down side of keeping it as it is.

pmclanahan: And actually `extract_procfile` is probably a better function name than `get_procfile`. I can change it to be more explicit, and add a `cleanup_procfile` or `remove_procfile` function to make `post-deploy` cleaner.

michaelshobbs: Seems reasonable. Thanks for talking it out.

tests/unit/10_ps-dockerfile.bats
```diff
@@ -116,3 +116,126 @@ teardown() {
...
+    shopt -u nullglob
+    run bash -c "[[ -z \"$CIDS\" ]]"
+    echo "output: "$output
+    echo "status: "$status
+    assert_success
+  done
+}
+
+@test "(ps:scale) procfile commands extraction" {
+  deploy_app dockerfile-procfile
```
michaelshobbs: This test seems to be doing double duty given we already test the deployment of this app type in the previous test.

Instead, I recommend running `create_app`, generating a `Procfile` manually at `$DOKKU_ROOT/$TEST_APP/DOKKU_PROCFILE` and then running the `get_cmd_from_procfile()` tests. This way we don't waste the cycles testing something other than the specific thing we want to test. We follow this pattern in the Dockerfile port extraction tests. https://github.com/dokku/dokku/blob/v0.4.13/tests/unit/30_core_ports_2.bats#L100-L124

pmclanahan: good catch. updated and pushed to my branch.

pmclanahan:
For some reason github isn't indicating the passing test run. But [it did pass](https://circleci.com/gh/dokku/dokku/2150). So hopefully we're nearly done here?

michaelshobbs:
lgtm

@josegonzalez

josegonzalez:
Ship it at your leisure, @michaelshobbs.

josegonzalez:
Note: I backed out this change while I attempt to make a release of dokku for 0.4.x. This change will ship with 0.5.x.

josegonzalez:
Alright, it's now in master :)

pmclanahan:
W00t! :cake: Thanks guys! Pleasure working with ya.

michaelshobbs:
No no, thank you @pmclanahan :facepunch:

vladfaust:
If I specify `ENTRYPOINT []` or `ENTRYPOINT [""]` in the Dockerfile, and `web: npm run web` in the Procfile, the build fails with `No command specified`. However, empty entrypoint is indeed legal and works locally (`docker run container npm run web`).

josegonzalez:
@vladfaust pleas file a new issue rather than commenting on an 8 year old closed one. Thank you.

PR #1934, opened 2016-02-22
fix get_running_image_tag() with docker 1.10.x

michaelshobbs:
docker 1.10.x changed the format of `docker ps -a`; which we should rely on anyway. This change uses `docker inspect` output instead. All other instances of `docker ps ...| egrep ...` are simplistic enough for now.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -167,7 +167,7 @@ get_running_image_tag() {
   verify_app_name "$APP"
 
   CIDS=( $(get_app_container_ids $APP) )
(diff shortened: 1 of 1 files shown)


michaelshobbs:
This is also backwards compatible with docker `1.9.x`

josegonzalez:
Why not just always use docker inspect? Hopefully they break that interface less often...

michaelshobbs:
Sometimes we just want running containers. `docker ps` gives us that.

PR #1935, opened 2016-02-22
remove unnecessary nginx test

michaelshobbs:
We removed `NGINX_CUSTOM_TEMPLATE_PATH` from the nginx rewrite but I forgot to remove the test.


josegonzalez:
Looks like the test passed anyhow?

michaelshobbs:
Indeed it did but it wasn't actually testing anything anyway

Issue #1638, opened 2015-11-01
Client "contrib/dokku_client.sh" populates automatically parameters wrongly.


davivcgarcia:
**Description of problem:**

Client *contrib/dokku_client.sh* populates automatically parameters wrongly for a few Dokku commands, such as *mongo:create* and *postgres:create*. It uses the app name as the name parameter, ignoring the parameter given.

**Outputs:**

- `uname -a`:

```
Linux dokku 3.13.0-66-generic #108-Ubuntu SMP Wed Oct 7 15:20:27 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

- `docker version`:

```
Client:
 Version:      1.8.3
  API version:  1.20
  Go version:   go1.4.2
  Git commit:   f4bf5c7
  Built:        Mon Oct 12 05:37:18 UTC 2015
  OS/Arch:      linux/amd64
 
 Server:
  Version:      1.8.3
  API version:  1.20
  Go version:   go1.4.2
  Git commit:   f4bf5c7
  Built:        Mon Oct 12 05:37:18 UTC 2015
  OS/Arch:      linux/amd64
```

- `docker -D info`:

```
 Containers: 2
 Images: 57
 Storage Driver: aufs
  Root Dir: /var/lib/docker/aufs
  Backing Filesystem: extfs
  Dirs: 61
  Dirperm1 Supported: false
 Execution Driver: native-0.2
 Logging Driver: json-file
 Kernel Version: 3.13.0-66-generic
 Operating System: Ubuntu 14.04.3 LTS
 CPUs: 4
 Total Memory: 7.798 GiB
 Name: dokku
 ID: NDUS:6JUF:ZNSI:INLE:4CLG:MCX7:5GUW:PP6W:OAKW:OQIH:LM4N:U2SM
 WARNING: No swap limit support
```

- `dokku version`:

```
0.4.3
```

- `dokku plugin`:

```
  00_dokku-standard    0.4.0 enabled    dokku core standard plugin
  20_events            0.4.0 enabled    dokku core events logging plugin
  apps                 0.4.0 enabled    dokku core apps plugin
  backup               0.4.0 enabled    dokku core backup plugin
  build-env            0.4.0 enabled    dokku core build-env plugin
  certs                0.4.0 enabled    dokku core certificate management plugin
  checks               0.4.0 enabled    dokku core checks plugin
  common               0.4.0 enabled    dokku core common plugin
  config               0.4.0 enabled    dokku core config plugin
[...]

PR #1953, opened 2016-03-01
subcommands/bashstyle refactor

### LINKED ISSUE #1579: Plugin command structure
Currently, plugin command structure (i.e. anything that is not a trigger. think `dokku ps:restart`) lives in a `commands` file that must be "hunted for." 

execution today:
```
implemented=0
for script in $PLUGIN_ENABLED_PATH/*/commands; do
  set +e; $script "$@" ; exit_code=$? ; set -e
  if [[ "$exit_code" -eq "$DOKKU_NOT_IMPLEMENTED_EXIT" ]]; then
    continue
  fi

  implemented=1
  if [[ "$exit_code" -ne "$DOKKU_VALID_EXIT" ]]; then
    exit $exit_code
  fi
done
if [[ "$implemented" -eq 0 ]]; then
  dokku_log_warn "\`$*\` is not a dokku command."
  dokku_log_warn "See \`dokku help\` for a list of available commands."
  exit 1
fi    
```

I propose this be simplified in a similar manner as #1039 and #592. 

### benefits
- more efficient execution
- less TRACE output leads to easier debugging
- explicit plugin namespace prevents plugins from polluting other plugin namespaces. this also prevents an error in one plugin from blocking the execution of another

proposed execution example:
```
PLUGIN_NAME="$1"
if [[ -x $PLUGIN_ENABLED_PATH/$PLUGIN_NAME/command ]]; then
  "$PLUGIN_ENABLED_PATH/$PLUGIN_NAME/command" "$@"
elif [[ -x $PLUGIN_ENABLED_PATH/${PLUGIN_NAME%%:*}/subcommands/${1#*:} ]]; then
  "$PLUGIN_ENABLED_PATH/${PLUGIN_NAME%%:*}/subcommands/${1#*:}" "$@"
fi
```
refs: #592 #1039

---

### general spec
- each command lives in `plugins/<plugin_name>/subcommands/<name_of_command>`
- current UI will be maintained (i.e. `<plugin_name>:<subcommand>`)
- top-level command (i.e. `dokku ps`) will be implemented in `plugins/<plugin_name>/command`
- old method of "hunting" may be supported as an interim but will be deprecated
- execution of plugin triggers are not changed by this

### open questions
- Can we actually do this with all of the core plugins? probably not as `git` comes to mind
- If we still support the old method, do we need to wait for a major release to merge this?

thoughts?

michaelshobbs:
- [x] moves commands to there own files
- [x] implements `--app` refs #1638 
- [x] everything that can be a function is a function (i.e. `dokku` is not called and we have limited global vars defined
- [x] docs
- [x] tests (with `--app` support)
- [x] fixes client issue #1638

refs #592 #1039
closes #1579 #1638


docs/development/plugin-creation.md
```diff
@@ -8,7 +8,89 @@ If you create your own plugin:
...
+
+  *)
+    exit $DOKKU_NOT_IMPLEMENTED_EXIT
+    ;;
+
+esac
+```
+
+
+### Sample plugin - old structure (still supported but not advised)
```
josegonzalez: Just remove the old method. One true way.

michaelshobbs: Sounds good to me

plugins/20_events/subcommands/default
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+events_main_cmd() {
+  local desc="shows contents of dokku events log"
```
josegonzalez: Worth moving these to declare? *shrug*

michaelshobbs: Yeah I can. Just `declare desc=`, right?

plugins/20_events/subcommands/default
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+events_main_cmd() {
```
josegonzalez: Was hoping we'd just call all the functions `main`, but this is certainly useful if we ever wanted to start sourcing the subcommands.

plugins/20_events/subcommands/default
josegonzalez: Yep

plugins/20_events/subcommands/default
michaelshobbs: Yeah the thinking there was clarity in trace output

josegonzalez:
:+1:

PR #1958, opened 2016-03-03
Enable debug output for dokku global exports in trace

### LINKED ISSUE #1957: RFC: enable debug output for dokku global exports in trace
- Description of problem:

Dokku exports are not constant, `dokku trace on` does not yield custom set variables in stdout.

- Proposal:

Move `[[ $DOKKU_TRACE ]] && set -x` toward the top of the `dokku` file to capture global env vars in trace.

Thoughts? Be a better mind reader?

u2mejc:
Resolves #1957

### DIFF
--- a/dokku
+++ b/dokku
@@ -2,6 +2,11 @@
 set -eo pipefail
 shopt -s nullglob
 
+export DOKKU_ROOT=${DOKKU_ROOT:=~dokku}
+[[ -f $DOKKU_ROOT/dokkurc ]] && source "$DOKKU_ROOT/dokkurc"
+[[ -d $DOKKU_ROOT/.dokkurc ]] && for f in $DOKKU_ROOT/.dokkurc/*; do source "$f"; done
+[[ $DOKKU_TRACE ]] && set -x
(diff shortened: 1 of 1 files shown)


dokku
```diff
@@ -2,6 +2,8 @@
 set -eo pipefail
 shopt -s nullglob
 
+[[ $DOKKU_TRACE ]] && set -x
+
```
josegonzalez: How would `dokku trace on` be affected by this?

u2mejc: amending, but the answer was because non dokku user's call the dokku command again with sudo :wink:

u2mejc:
hmm that's a chicken and egg issue looking at it now...

u2mejc:
There we go. The path I stumbled on was for dokkurc. The DOKKU_ROOT path could be set in the dokkurc file, but it's not possible to change the path before calling the file. That's the chicken and egg, but this change doesn't affect that.

josegonzalez:
Looks fine to me, will let @michaelshobbs merge :)

PR #1959, opened 2016-03-03
Allow customizing ssh port for the default client

### LINKED ISSUE #1942: dokku_client.sh can't connect to SSH with custom port
Description of problem:
My remote SSH is not on port 22, so the `dokku` command does nothing.

Proposed solution:
As there's already a DOKKU_HOST env var, there could be a DOKKU_PORT optional env var.
If set with a port number, use it, otherwise fallback to default 22.

josegonzalez:
Closes #1942

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -41,6 +41,8 @@ if [[ -z $DOKKU_HOST ]]; then
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


u2mejc:
ah I can't read either of your minds. If you want me to just pump out code, you should say so. :wink:

josegonzalez:
lol *shrug* I'm sure you'll figure it out eventually :D

PR #1975, opened 2016-03-08
Fixed Nginx websocket configuration issue

Benjamin-Dobell:
The quotes are dropped during template generation, so they should be escaped. Also switched to using uppercase leading character as there seems to be better compatibility:
https://github.com/websocket-rails/websocket-rails/issues/55

As an example this change was required in order to run http://web.sockethook.io/ - which is a nice and simple test case because zero configuration is required.

### DIFF
--- a/plugins/nginx-vhosts/templates/location.config
+++ b/plugins/nginx-vhosts/templates/location.config
@@ -10,7 +10,7 @@
     proxy_pass  http://{{ .APP }};
     proxy_http_version 1.1;
     proxy_set_header Upgrade $http_upgrade;
-    proxy_set_header Connection "upgrade";
+    proxy_set_header Connection \"Upgrade\";
     proxy_set_header Host $http_host;
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_set_header X-Forwarded-For $remote_addr;


michaelshobbs:
I don't see this in master. Can you share how I might recreate the issue?

```
root@dokku:~/dokku# cat /home/dokku/node-js-app/nginx.conf
server {
  listen      [::]:80;
  listen      80;
  server_name node-js-app.dokku.me;
  access_log  /var/log/nginx/node-js-app-access.log;
  error_log   /var/log/nginx/node-js-app-error.log;

  location    / {
    gzip on;
    gzip_min_length  1100;
    gzip_buffers  4 32k;
    gzip_types    text/css text/javascript text/xml text/plain text/x-component application/javascript application/x-javascript application/json application/xml  application/rss+xml font/truetype application/x-font-ttf font/opentype application/vnd.ms-fontobject image/svg+xml;
    gzip_vary on;
    gzip_comp_level  6;

    proxy_pass  http://node-js-app;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $http_host;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_set_header X-Forwarded-For $remote_addr;
    proxy_set_header X-Forwarded-Port $server_port;
    proxy_set_header X-Request-Start $msec;
  }
  include /home/dokku/node-js-app/nginx.conf.d/*.conf;
}

upstream node-js-app {
  server 172.17.0.2:5000;
}
```

Benjamin-Dobell:
D'oh! It may well already be fixed.

My hosts are all running the latest build from apt-get, not master. I made the change to the template manually on my host and only afterwards thought I better share the change. It does look as though templates have definitely changed (more modular).

So my `\"` vs. `"` change may well be moot. However, "upgrade" vs "Upgrade" may still be a thing. For testing I recommend you just clone https://github.com/factor-io/websockethook and push it to straight to Dokku (no configuration required). Then just visit the root URL for the app in your browser - it's a webpage that will specifically output whether or not websockets are working.

Sorry, I should really test that myself but don't have any hosts I can upgrade to master and my Ubuntu VMs are well neglected!

michaelshobbs:
Ah yes. This is indeed an issue in 0.4.14. We completely rewrote the nginx plugin for 0.5.x and that's what is in master at the moment. The correct workaround until 0.5.0 will be a custom template. Heads up though, the templating engine going forward will be sigil. Check out the [docs][migration-guide] in master to see the changes.

[migration-guide](https://github.com/dokku/dokku/blob/master/docs/appendices/0.5.0-migration-guide.md)

PR #1981, opened 2016-03-09
only remove containers with dokku label

### LINKED ISSUE #1220: dokku cleanup removed all "exited" containers

michaelshobbs:
closes #1220

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -414,11 +414,11 @@ docker_cleanup() {
   declare desc="cleans up all exited/dead containers and removes all dangling images"
   # delete all non-running containers
   # shellcheck disable=SC2046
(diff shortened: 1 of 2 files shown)


josegonzalez:
Should we have a test for this?

michaelshobbs:
Yes, and **you** should write it :stuck_out_tongue_winking_eye:

josegonzalez:
Closing, #wontfix, #notabug.

PR #1987, opened 2016-03-12
Do not restart stopped processes on config:set/unset

### LINKED ISSUE #1889: Request: Make config_set and config_unset respect DOKKU_APP_RESTORE
I've been upgrading the containers used by a few services (through the mysql plugin), here's my current workflow:

    docker pull mysql:5.7.10
    dokku ps:stop app
    dokku mysql:clone app-database app-database-new
    dokku mysql:unlink app-database app
    dokku mysql:link app-database-new app
    dokku ps:start app

Unfortunately, as soon as I unlink the old database it triggers a `dokku ps:restart app` (from here https://github.com/dokku/dokku/blob/master/plugins/config/functions#L226) even though DOKKU_APP_RESTORE has been set to 0.

I think it'd be a good idea if `config_set` and `config_unset` didn't restart the app if DOKKU_APP_RESTORE is set as 0.

josegonzalez:
Closes #1889

### DIFF
--- a/plugins/config/functions
+++ b/plugins/config/functions
@@ -202,7 +202,8 @@ ${var}"
     plugn trigger post-config-update "$APP" "set" "$@"
   fi
 
-  if [[ "$DOKKU_CONFIG_RESTART" == "true" ]]; then
+  local DOKKU_APP_RESTORE=$(config_get "$APP" DOKKU_APP_RESTORE || true)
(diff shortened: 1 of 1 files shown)


josegonzalez:
@michaelshobbs looks good?

michaelshobbs:
+1

PR #1990, opened 2016-03-13
Highlight features that are yet to come

### LINKED ISSUE #1989: Default docs showing `master` is misleading and confusing
I just spent hours the other day trying to configure [Persistent Storage](http://dokku.viewdocs.io/dokku/dokku-storage/) and [Deployment Tasks](http://dokku.viewdocs.io/dokku/deployment/deployment-tasks/) on my brand new dokku instance. Yet I couldn't get it to work, which was very frustrating. Until I realised, though I followed the recommended procedure of said document to install the latest version, I wasn't viewing the docs _of the version I had installed_ – I was looking at master.

Sure, there is a box on the bottom right stating I look at "master" and there is a big warning on the top, telling me this is only available from 0.5, but after just having installed dokku exactly the way this particular doc told me to, I was assuming to be on a version of that or higher – but not lower. If that was docs for master, I'd assume its recommended way of installation would have been to install from source (and I would probably have done that). 

That is very frustrating. Not only because I wasted hours trying to configure a feature that isn't shipped yet, but also because prior I was browsing the feature list of the docs to let me decide _whether_ dokku is suitable for my needs. And it appeared to be – having all these features linked and described how to configure them. But they are unstable or at least not yet released. Unstable shouldn't really be the front page of the docs – that is very misleading. Or at least have the box on the top state that this feature _isn't released yet_ rather than saying which (future) version number it will be in.

ligthyear:
- Tone down "new as of version" boxes
 - Replace all "New as of 0.5" with "yet to be released" text
 - Use strong red only for "yet to be released" boxes

Fixes #1989 

It now looks like this:

screen shot 2016-03-13 at 11 31 16
screen shot 2016-03-13 at 11 25 59

### DIFF
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -238,8 +238,8 @@ blockquote {
   margin-bottom: 30px;
 }
 blockquote.new-as-of {
-  border-left: 4px solid #f0ad4e;
-  background-color: #f0dc6e;
+  border-left: 4px solid #d1f2a5;
+  background-color: #effab4;
 }
 blockquote.not-yet-released {
   border-left: 4px solid #ef5b58;
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -238,6 +238,10 @@ blockquote {
   margin-bottom: 30px;
 }
 blockquote.new-as-of {
+  border-left: 4px solid #f0ad4e;
+  background-color: #f0dc6e;
+}
+blockquote.not-yet-released {
   border-left: 4px solid #ef5b58;
   background-color: #f9ad76;
 }
--- a/docs/deployment/deployment-tasks.md
+++ b/docs/deployment/deployment-tasks.md
@@ -1,6 +1,6 @@
 # Deployment Tasks
 
-> not yet released and only available in master
+> Not yet released and only available in master
 
 Sometimes you need to run a command on at deployment time, but before an app is completely deployed.
 Common use cases include:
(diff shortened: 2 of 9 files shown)


docs/assets/style.css
```diff
@@ -238,6 +238,10 @@ blockquote {
   margin-bottom: 30px;
 }
 blockquote.new-as-of {
+  border-left: 4px solid #f0ad4e;
+  background-color: #f0dc6e;
```
josegonzalez: How about:

```
#7ad375
#a6e22d
```

for colors

josegonzalez: That would be green, btw.

docs/deployment/deployment-tasks.md
```diff
@@ -1,6 +1,6 @@
 # Deployment Tasks
 
-> New as of 0.5.0
+> not yet released and only available in master
```
josegonzalez: Should be uppercase

docs/template.html
```diff
@@ -252,6 +252,8 @@ <h1 class="heading navbar-brand">
       Array.prototype.forEach.call(blockquotes, function (el, i) {
         if (el.innerHTML.indexOf('New as of') !== -1) {
           addClass(el, 'new-as-of');
+        } else if (el.innerHTML.indexOf('not yet released') !== -1) {
```
josegonzalez: probably can call `lower()` on `innerHtml`

josegonzalez:
I made a pr to your pr.

docs/assets/style.css
josegonzalez: Ah, how about:

```
#D1F2A5
#EFFAB4
```

PR #1991, opened 2016-03-13
Only chown of existing files

MorrisJobke:
* this caused issues during installation if the file is not existent

@michaelshobbs @josegonzalez This fixed an issue with the ArchLinux installation where the installation of the plugins aborted because the VHOST file was not available. Is this file needed or just optional? Because in Line 9 of this file the VHOST is created only if the host name could be resolved by `dig`.

### DIFF
--- a/plugins/00_dokku-standard/install
+++ b/plugins/00_dokku-standard/install
@@ -60,4 +60,8 @@ fi
 EOF
 chmod +x /etc/update-motd.d/99-dokku
 
-chown dokku:dokku "$DOKKU_ROOT/HOSTNAME" "$DOKKU_ROOT/VHOST"
+chown dokku:dokku "$DOKKU_ROOT/HOSTNAME"
+
+if [[ -f  "$DOKKU_ROOT/VHOST" ]]; then
+  chown dokku:dokku "$DOKKU_ROOT/VHOST"
+fi


MorrisJobke:
I added this fix to my 0.4.14 version for ArchLinux to be able to provide a working package.

plugins/00_dokku-standard/install
```diff
@@ -60,4 +60,8 @@ fi
 EOF
 chmod +x /etc/update-motd.d/99-dokku
 
-chown dokku:dokku "$DOKKU_ROOT/HOSTNAME" "$DOKKU_ROOT/VHOST"
+chown dokku:dokku "$DOKKU_ROOT/HOSTNAME"
+
+if [[ ! -f  "$DOKKU_ROOT/VHOST" ]]; then
```
michaelshobbs: Am I reading this right? `chmod` this if it's not a file? Can you clarify here?

MorrisJobke: -.- Typo. Should be chown

MorrisJobke: And of course not negated.

MorrisJobke: It just worked, because I added the VHOST before installation - I retested without the VHOST file and then it works now.

PR #2012, opened 2016-03-22
minor bug fixes around app.json and docker-options

michaelshobbs:
- remove newlines in docker args
- make sure we set HOME and exit with status from app.json script
- actually test docker-options are ending up in container config


josegonzalez:
:+1:

Issue #1952, opened 2016-03-01
Dokku run commands defined in Procfile?


pascalw:
Heroku allows you to run commands defined in the Procfile using `heroku run`. For example `heroku run console` will run the command defined in your Procfile as the `console` process type.

I noticed that Dokku doesn't seem to support this. Would you accept a PR that would add this functionality? I'm thinking it could work as follows:

* On `dokku run` check if the provided command name corresponds with a proc type defined in the Procfile.
  * If match -> execute the command specified for that proc type
  * Else -> execute the command as is

This should be mostly backwards compatible. Theoretically if someone would have a Procfile with a proc type that corresponds to a name of something in `$PATH` this could break. Perhaps we could even check if the name is in `$PATH` and skip Procfile execution if it is.

josegonzalez:
Sure, go for it. Remember to write tests and add documentation :)

josegonzalez:
@pascalw *bump*

pascalw:
Hey sorry have been very busy. I've looked into how to implement it and it's pretty trivial (herokuish supports executing procttypes directly) but haven't gotten around to writing the tests and docs.

Will try to work on it this week.

pascalw:
Thinking about this some more I think Dockerfile deployments now also support Procfiles right? In that case it's better not to rely on Herokuish to execute the Procfiles command but use Dokku's Procfile parsing since that would work in both cases.

josegonzalez:
Works for me.

michaelshobbs:
What happens if a `Procfile` key matches a command? (ex. `bash: echo 'you probably should not do this`)

Would the `Procfile` key take precedence?

josegonzalez:
Yes, that seems reasonable to me.

We could probably output "Detected command from Procfile, running that command" as a warning...

pascalw:
@michaelshobbs I described this in the original issue. 
The most robust approach would be to not run the Procfile command if the key corresponds with a name in $PATH in the container, however that's not particularly easy I think and also might yield some unexpected results.

Outputting a message when this behavior kicks in seems reasonable to me.

josegonzalez:
Closing as this is taken care of by #1952.

PR #2018, opened 2016-03-23
Support running Procfile commands using `dokku run`

pascalw:
As discussed in #1952.

This is just a first stab, I've got a couple of things I would like some feedback on.

### DIFF
--- a/plugins/00_dokku-standard/subcommands/run
+++ b/plugins/00_dokku-standard/subcommands/run
@@ -25,11 +25,11 @@ dokku_run_cmd() {
   has_tty && local DOKKU_RUN_OPTS+=" -i -t"
   is_image_herokuish_based "$IMAGE" && local EXEC_CMD="/exec"
 
-  # TODO: should we cleanup the procfile after run?
   DOKKU_QUIET_OUTPUT=1 extract_procfile "$APP"
 
   POTENTIAL_PROCFILE_KEY="$1"
   PROC_CMD=$(get_cmd_from_procfile "$APP" "$POTENTIAL_PROCFILE_KEY" || echo '')
+  remove_procfile "$APP"
 
   if [ ! -z "$PROC_CMD" ]; then
     dokku_log_info1 "Found '$POTENTIAL_PROCFILE_KEY' in Procfile, running that command"
--- a/plugins/00_dokku-standard/subcommands/run
+++ b/plugins/00_dokku-standard/subcommands/run
@@ -26,7 +26,7 @@ dokku_run_cmd() {
   is_image_herokuish_based "$IMAGE" && local EXEC_CMD="/exec"
 
   # TODO: should we cleanup the procfile after run?
-  extract_procfile "$APP"
+  DOKKU_QUIET_OUTPUT=1 extract_procfile "$APP"
 
   POTENTIAL_PROCFILE_KEY="$1"
   PROC_CMD=$(get_cmd_from_procfile "$APP" "$POTENTIAL_PROCFILE_KEY" || echo '')
--- a/plugins/00_dokku-standard/subcommands/run
+++ b/plugins/00_dokku-standard/subcommands/run
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/ps/functions"
 source "$PLUGIN_AVAILABLE_PATH/config/functions"
 
 dokku_run_cmd() {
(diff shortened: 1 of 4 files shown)


plugins/00_dokku-standard/subcommands/run
```diff
@@ -23,6 +24,16 @@ dokku_run_cmd() {
   [[ "$DOKKU_RM_CONTAINER" ]] && local DOKKU_RUN_OPTS="--rm"
   has_tty && local DOKKU_RUN_OPTS+=" -i -t"
   is_image_herokuish_based "$IMAGE" && local EXEC_CMD="/exec"
+
+  # TODO: should we cleanup the procfile after run?
+  extract_procfile "$APP"
```
pascalw: @josegonzalez @michaelshobbs I noticed that Procfiles are cleaned up after deployment. To be able to read the Procfile on `dokku run` I had to extract it first. Is this desirable? Unfortunately it yields a log message which is a bit distracting: `-----> App Procfile file found (/home/dokku/nodejs/DOKKU_PROCFILE)`.

plugins/00_dokku-standard/subcommands/run
```diff
@@ -23,6 +24,16 @@ dokku_run_cmd() {
...
+  PROC_CMD=$(get_cmd_from_procfile "$APP" "$1" || echo '')
+
+  if [ ! -z $PROC_CMD ]; then
+    dokku_log_info1 "Found '$1' in Procfile, running that command"
```
pascalw: Is `dokku_log_info1` the correct log function to use for these kinds of messages?

josegonzalez: Yes, I think so.

We should move `$1` to a variable...

tests/unit/30_core_1.bats
```diff
@@ -128,6 +128,16 @@ build_nginx_config() {
...
 }
 
+@test "(core) run command from Procfile" {
+  deploy_app
+  run /bin/bash -c "dokku run $TEST_APP console -c 'echo hi' | tail -n 1"
+  echo "output: "$output
+  echo "status: "$status
+
+  assert_success
+  assert_output "hi"
```
pascalw: This tests is currently failing because of an issue I ran into with how herokuish handles arguments.

Observe this:

```sh
root@dokku:/home/vagrant# docker run --label=dokku -i -t dokku/nodejs:latest /exec bash -c 'echo hi'

root@dokku:/home/vagrant#
```

```sh
root@dokku:/home/vagrant# docker run --label=dokku -i -t dokku/nodejs:latest bash -c 'echo hi'
hi
root@dokku:/home/vagrant#
```

Could this be a bug in herokuish?

plugins/00_dokku-standard/subcommands/run
michaelshobbs: I'd use `dokku_log_info1_quiet` I think. Gives the option to silence that ouput.

plugins/00_dokku-standard/subcommands/run
michaelshobbs: You could set `$DOKKU_QUIET_OUTPUT` before the call and unset it afterwards. **shrug**

tests/unit/30_core_1.bats
michaelshobbs: This works though:
```
root@dokku:~/dokku# dokku run node-js-app echo hi
hi
```

plugins/00_dokku-standard/subcommands/run
josegonzalez: Or just set it for that command:

```
DOKKU_QUIET_OUTPUT=1 extract_procfile "$APP"
```

pascalw: Ah cool, added that.

tests/unit/30_core_1.bats
pascalw: Yes, it seems to be an issue specifically with quoted named arguments. I've updated the test case to not rely on that and I will open a separate issue on herokuish.

plugins/00_dokku-standard/subcommands/run
```diff
@@ -23,6 +24,18 @@ dokku_run_cmd() {
   [[ "$DOKKU_RM_CONTAINER" ]] && local DOKKU_RUN_OPTS="--rm"
   has_tty && local DOKKU_RUN_OPTS+=" -i -t"
   is_image_herokuish_based "$IMAGE" && local EXEC_CMD="/exec"
+
+  DOKKU_QUIET_OUTPUT=1 extract_procfile "$APP"
+
+  POTENTIAL_PROCFILE_KEY="$1"
+  PROC_CMD=$(get_cmd_from_procfile "$APP" "$POTENTIAL_PROCFILE_KEY" || echo '')
+  remove_procfile "$APP"
```
pascalw: I'm now cleaning up the Procfile after using it, just like `deploy` does.

It would be nice if `get_cmd_from_procfile` would take care of extracting the Procfile and cleaning it up. Potential downside of that is that if you would call it multiple times it would extract the Procfile every time, but the overhead of that is pretty small I think and also seems unlikely you would call this multiple times. What do you think @josegonzalez @michaelshobbs?

josegonzalez: Eh, I think this is fine as is.

josegonzalez: Eh, this seems fine to me.

josegonzalez:
@michaelshobbs this seems good to go. Thoughts?

plugins/00_dokku-standard/subcommands/run
pascalw: Alright fine be me :-)

PR #2024, opened 2016-03-26
Delete bad symlinks without confirmation

josegonzalez:
This fixes issues where the deleted backup plugin causes warnings during `plugn` runs.


michaelshobbs:
:+1:

PR #2035, opened 2016-03-28
Fix ps_restart to not exit early

stesie:
ps_restartall_cmd (aka `dokku ps:restartall`) calls the `ps_restart` function from a loop, so if `ps_restart` does an `exit 0` if it finds an app that hasn't been deployed (in my case due to `dokku ps:scale web=0`), then it exits the `ps:restartall` loop early and hence just doesn't restart all the apps.

### DIFF
--- a/plugins/ps/functions
+++ b/plugins/ps/functions
@@ -128,9 +128,11 @@ ps_restart() {
   local APP="$1"; verify_app_name "$APP"
   local IMAGE_TAG=$(get_running_image_tag "$APP")
 
-  ! (is_deployed "$APP") && echo "App $APP has not been deployed" && exit 0
-
-  release_and_deploy "$APP" "$IMAGE_TAG"
+  if is_deployed "$APP"; then
+    release_and_deploy "$APP" "$IMAGE_TAG"
+  else
+    echo "App $APP has not been deployed"
+  fi
 }
 
 ps_scale() {


plugins/ps/functions
```diff
@@ -128,9 +128,11 @@ ps_restart() {
   local APP="$1"; verify_app_name "$APP"
   local IMAGE_TAG=$(get_running_image_tag "$APP")
 
-  ! (is_deployed "$APP") && echo "App $APP has not been deployed" && exit 0
-
-  release_and_deploy "$APP" "$IMAGE_TAG"
+  if is_deployed "$APP"; then
+    release_and_deploy "$APP" "$IMAGE_TAG"
+  else
+    echo "App $APP has not been deployed"
```
josegonzalez: Should we return 1 here at least?

stesie: I don't think so. After all IMHO that even isn't an error, there's just nothing to be done and then echo just shows a log message. All current callers (ps/subcommands/restartall, nginx-vhosts/proxy-disable and nginx-vhosts/proxy-enable) don't need the information.  So YAGNI.

Even worse, if we `return 1` here we would have to change all callers (since they `set -e` they then would `exit 1`)

josegonzalez:
@michaelshobbs beep beep you're a jeep.

Issue #1943, opened 2016-02-25
dokku ps:restore fails due to CHECKS file


z0mt3c:
Description of problem:

For me it seems a ```CHECKS``` prevents ```dokku ps:restore``` from restoring the applications/instances. I noticed it because suddenly the my services didn't come back after system reboots Whenever i published an app containing a ```CHECKS``` file none of my applications were started after reboot (ps:restore). (Even there were some without a ```CHECKS``` file as well.) May they were started and terminated after a few seconds. May it's just my fault? Did i miss something? Any idea?

My ```CHECKS``` file looked like this:
```properties
WAIT=2
ATTEMPTS=5
/
```

My ```upstart/dokku-redeploy.log``` looked like this when it went wrong (with ```CHECKS``` file) for the particular app:
```logs
Restoring app PROJECT_NAME ...
-----> Releasing PROJECT_NAME (dokku/PROJECT_NAME:latest)...
-----> Deploying PROJECT_NAME (dokku/PROJECT_NAME:latest)...
-----> DOKKU_SCALE file found (/home/dokku/PROJECT_NAME/DOKKU_SCALE)
=====> web=1
-----> Running pre-flight checks
-----> Attempt 1/5 Waiting for 2 seconds ...
       CHECKS expected result:
       http://localhost/ => ""
 !
curl: (7) Failed to connect to 172.17.0.3 port 5000: Connection refused
 !     Check attempt 1/5 failed.
-----> Attempt 2/5 Waiting for 2 seconds ...
       CHECKS expected result:
       http://localhost/ => ""
 !
curl: (7) Failed to connect to 172.17.0.3 port 5000: Connection refused
 !     Check attempt 2/5 failed.
-----> Attempt 3/5 Waiting for 2 seconds ...
       CHECKS expected result:
       http://localhost/ => ""
 !
curl: (7) Failed to connect to 172.17.0.3 port 5000: Connection refused
 !     Check attempt 3/5 failed.
-----> Attempt 4/5 Waiting for 2 seconds ...
       CHECKS expected result:
       http://localhost/ => ""
 !
curl: (7) Failed to connect to 172.17.0.3 port 5000: Connection refused
 !     Check attempt 4/5 failed.
-----> Attempt 5/5 Waiting for 2 seconds ...
       CHECKS expected result:
       http://localhost/ => ""
 !
curl: (7) Failed to connect to 172.17.0.3 port 5000: Connection refused
 !     Check attempt 5/5 failed.
[...]

PR #2039, opened 2016-03-29
Update upgrading.md

christiangenco:
Updated upgrading instructions from [the trouble spots I had in upgrading](https://github.com/dokku/dokku/issues/1943#issuecomment-202658888) with @michaelshobbs's advice.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -12,6 +12,10 @@ Migration guides contain information regarding the new features introduced in ea
 
 ## General Information
 
+Upgrading is not advised on running apps. Stop them before with `dokku ps:stop <app>`.
+
+Before using `apt-get install`, run `apt-get update` to download the latest packages.
+
 If dokku was installed via a debian package, you can upgrade dokku via the following command:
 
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -12,6 +12,10 @@ Migration guides contain information regarding the new features introduced in ea
 
 ## General Information
 
+Upgrading is not advised on running apps. Stop them before with `dokku ps:stop <app>`.
```
josegonzalez: Is this true, @michaelshobbs ?

michaelshobbs: I would think so. I wouldn't upgrade on a running "production" host. Specifically when we require a particular version of docker that could potentially cause that package to be upgraded would cause the docker daemon to restart. Thus, resulting in downtime.

michaelshobbs: It should be noted that docker >=1.11 promises your containers don't need to be restarted if the docker daemon is restarted. However live upgrades in a production level environment (without fail overs) are not recommended.

PR #2050, opened 2016-03-30
Implement repo:gc and repo:purge-cache

### LINKED ISSUE #1570: Add support for purging the npm cache
My package.json references an npm module at the tip of a private repo.  Works great, but since dokku caches the npm folder, it never checks if there's a new commit on that package.

It would be nice if something similar to the [Heroku repo plugin](https://github.com/heroku/heroku-repo) existed so I could dokku repo:purge_cache.

josegonzalez:
These are helper plugins that mimic the repo plugin: https://github.com/heroku/heroku-repo

The following commands will remain unimplemented:

- `clone`: not useful for the host client to clone
- `download`: currently not possible for us to "host" a tarball of the repo
- `reset`: we still store files in the repo (though we shouldn't) so we cannot easily reinitialize the repository.

Also fixes a potential race condition if two separate processes create an application's `$CACHE_DIR`.

Closes #1570

### DIFF
--- a/plugins/repo/subcommands/purge-cache
+++ b/plugins/repo/subcommands/purge-cache
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+repo_purge_cache() {
+  declare desc="deletes the contents of the build cache stored in the repository"
+  local cmd="repo:purge-cache"
+  [[ -z $2 ]] && dokku_log_fail "Please specify an app to run the command on"
+  local APP="$2";
+  verify_app_name "$APP"
+
+  local IMAGE=$(get_app_image_name "$APP"); local CACHE_DIR="$DOKKU_ROOT/$APP/cache"
+
+  if [[ -d $CACHE_DIR ]]; then
+    docker run "$DOKKU_GLOBAL_RUN_ARGS" --rm -v "$CACHE_DIR:/cache" "$IMAGE" find /cache -depth -mindepth 1 -maxdepth 1 -exec rm -Rf {} \; || true
+    mkdir -p "$CACHE_DIR" || true
+  fi
+}
+
+repo_purge_cache "$@"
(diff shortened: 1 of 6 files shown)


michaelshobbs:
Do we need this given that we only clone a depth of 1? Also, tests.

plugins/repo/subcommands/purge-cache
```diff
@@ -0,0 +1,16 @@
...
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+
+repo_purge_cache() {
+  declare desc="deletes the contents of the build cache stored in the repository"
+  local cmd="repo:purge-cache"
+  [[ -z $2 ]] && dokku_log_fail "Please specify an app to run the command on"
+  local APP="$2"; local CACHE_DIR="$DOKKU_ROOT/$APP/cache"
+  verify_app_name "$APP"
+
+  find "$CACHE_DIR" -mindepth 1 -exec rm -rf {}
```
michaelshobbs: Does this work? Check the apps `pre-delete` trigger. We mount a container to `rm` this in that case due to permissions.

michaelshobbs:
Couldn't we implement `download` with an inline `tar` pipeline?

josegonzalez:
- Yeah I think so, given that it's the repo you push to.
- I don't write tests. You do, remember?
- Probably? Not sure how that would work...

michaelshobbs:
Oh yeah that's fair. We clone a depth of one but each subsequent push will add to that history.

michaelshobbs:
Thinking further, download would either need a client or we could tar up to a directory and just echo out the path for the user to scp. Not sure how much it's worth to do though. 

Also, write some damn tests. 😜

josegonzalez:
No, u.

michaelshobbs:
You could prevent the race condition by using `flock` like we do for deploys.

josegonzalez:
I figure the `-p` is good enough.

michaelshobbs:
:+1:

josegonzalez:
How do I test the `gc` call? I know how to test the purge-cache, but don't know about the gc call. Maybe neither is necessary?

michaelshobbs:
Rebased from master and added tests

josegonzalez:
Uhh that's not a rebase ;)

michaelshobbs:
Hrm what happened there. I did `git rebase master` in this branch.

michaelshobbs:
I think that fixed it. Maybe those showed up because we're merging squashed commits when we merge PRs?

michaelshobbs:
I cherry-picked that last commit from master. Not sure why github shows it as a change since the files match....

PR #2052, opened 2016-03-30
Use upstream releases when creating deb packages. Closes #2048

### LINKED ISSUE #2048: refactor how we build dokku dependencies
Currently we build our golang dependencies from source. We should probably not do that as they both have binary releases of their own.

plugn: https://github.com/dokku/dokku/blob/v0.5.2/deb.mk#L126-L142
sigil: https://github.com/dokku/dokku/blob/v0.5.2/deb.mk#L161-L178

josegonzalez:
- drop golang dependency
- use wget to retrieve upstream dependency


michaelshobbs:
Assuming this works, I'm :+1:

josegonzalez:
I built packages locally, they seemed to work fine :)

PR #2060, opened 2016-03-31
filter restart policies from exec-app-json containers

michaelshobbs:


josegonzalez:
Is it possible to test this?

michaelshobbs:
I tested locally. You think we need a unit test?

josegonzalez:
Maybe? I vaguely recall us saying there were other args that couldn't be applied to certain phases. I'm fine with merging either way.

michaelshobbs:
We have that other filtering in place in the docker-options plugin itself. This is sort of a special edge case (aren't they all?!) that's specific to restart policies and the pre/post deploy containers. i.e. we want to include the deploy docker-options because we want that container to look as much like the actual deployed container as possible but the restart policy doesn't make sense to include.

josegonzalez:
Gotcha. Again, works for me if you feel it's good to merge :)

michaelshobbs:
I reserve the right to add a test in the future. (maintainers prerogative :wink:)

PR #2061, opened 2016-04-01
Fix ignored settings in the CHECKS file

znz:
I used settings in the `CHECKS` file.
But they are ignored after upgraded from v0.4.14 to v0.5.3.
This fixes typo and they work well again.


michaelshobbs:
Good catch. Sorry about that 😁

PR #2069, opened 2016-04-03
Fix broken migration link in documentation

RickCarlino:
The link to the migration guide was not valid. This PR points it to the correct URL.


josegonzalez:
We don't support browsing the documentation from within the docker repo, so this would actually break the link [here](http://dokku.viewdocs.io/dokku/upgrading/).

PR #2074, opened 2016-04-05
Add support for heroku's script.postdeploy hook

josegonzalez:
This hook is a bit different from the scripts.dokku.postdeploy in that it only applies to the *first* deploy, not every deploy, and is thus suitable for hooks such as:

- slack notification: user just deployed a test application [here](http://example.com).

Also adds a new hook, `post-initial-deploy`.

### DIFF
--- a/plugins/git/commands
+++ b/plugins/git/commands
@@ -8,6 +8,7 @@ git_build_app_repo() {
   declare desc="builds local git app repo for app"
   verify_app_name "$1"
   local APP="$1"; local REV="$2"
+  local DOKKU_FIRST_DEPLOY=false
 
   # clean up after ourselves
   local GIT_BUILD_APP_REPO_TMP_WORK_DIR=$(mktemp -d "/tmp/dokku_git.XXXX")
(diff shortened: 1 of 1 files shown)


josegonzalez:
Would also be great to support setting initial env vars as mentioned [here](https://devcenter.heroku.com/articles/app-json-schema#env).

plugins/git/commands
```diff
@@ -21,7 +22,8 @@ git_build_app_repo() {
   chmod 755 "$GIT_BUILD_APP_REPO_TMP_WORK_DIR"
   unset GIT_DIR GIT_WORK_TREE
   pushd "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" > /dev/null
-  [[ ! -d "$DOKKU_ROOT/$APP" ]] && apps_create "$APP"
+  [[ ! -d "$DOKKU_ROOT/$APP" ]] && DOKKU_FIRST_DEPLOY=true
```
michaelshobbs: If a user runs `apps:create` before a push then this will never be true even though that push would be the actual first deploy.

michaelshobbs:
Definitely will want a test for this

plugins/git/commands
josegonzalez: Yeah I thought about that. Not sure what the best way to deal with that is. Potentially create a sentinel file (or use the new config stuff) to track whether its new or not.

josegonzalez:
Need to get the existing tests running first :P

PR #2077, opened 2016-04-05
Migrate all dokku related configuration to filedb wrapper

josegonzalez:
This change allows us to stop polluting the environment
for deployed applications, making configuration setting/unsetting
a simple matter of calling filedb.

This change breaks compatibility with plugins that expected dokku
config to be available via `dokku config:get`, as it will also
silently migrate this configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 DOKKU_VERSION = master
 
-FILEDB_URL ?= https://raw.githubusercontent.com/josegonzalez/bash-filedb/v0.0.3/filedb
+FILEDB_URL ?= https://raw.githubusercontent.com/josegonzalez/bash-filedb/v0.3.0/filedb
 SSHCOMMAND_URL ?= https://raw.githubusercontent.com/dokku/sshcommand/v0.4.0/sshcommand
 PLUGN_URL ?= https://github.com/dokku/plugn/releases/download/v0.2.1/plugn_0.2.1_linux_x86_64.tgz
 SIGIL_URL ?= https://github.com/gliderlabs/sigil/releases/download/v0.4.0/sigil_0.4.0_Linux_x86_64.tgz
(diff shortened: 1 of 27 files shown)


michaelshobbs:
A few questions
- Why not migrate everything to filedb on install?
- Why change the call to `get_config_value` instead of just implementing it in `config_get`?
- Seems like we now have two methods to set/get config? Why?

josegonzalez:
- Because not everyone installs via the deb packaging.
- Because then we'd need to have another flag to figure out if it's internal to dokku or called by a user.
- One is for application env config, the other is for general dokku-related configuration.

michaelshobbs:
> Because not everyone installs via the deb packaging.

If you add it to the `install` trigger then it gets run on install....(see [here](https://github.com/dokku/dokku/blob/v0.5.4/plugins/checks/install#L6-L39))

>Because then we'd need to have another flag to figure out if it's internal to dokku or called by a user.

ok. docs will make that clear. might want to add that to the PR details? It's not clear when a developer might use one over the other.

>One is for application env config, the other is for general dokku-related configuration.

again, docs (of course :smile:)

josegonzalez:
Yeah, need to write docs in the plugin creation section for sure. Not sure why tests are failing yet...

michaelshobbs:
Also a meta question, any reason to add the complexity of file-db and not just move the location of config using the same interface we have today?

josegonzalez:
I wanted to be able to do something like:

```
# provides filedb binary with a different backend
apt-get install consuldb
```

And have dokku-level configuration persist outside of the config plugin. Shipped dokku images have the build-env already set, so we need a way to persist dokku config outside of that if the dokku image is run on a separate machine.

michaelshobbs:
I'll withhold further comment until you've fleshed this out. I'm still not clear why we have two methods of setting config.

PR #2099, opened 2016-04-11
Added clarification on configuration via separate files

ojosdegris:


docs/nginx.md
```diff
@@ -131,6 +131,9 @@ chown dokku:dokku /home/dokku/myapp/nginx.conf.d/upload.conf
 service nginx reload
 ```
 
+Example above shows that you need to add additional configuration files directly on the dokku host. Unlike nginx.conf.sigil file those will not be copied over from your application repo.
```
josegonzalez: comma after the `nginx.conf.sigil file` part.

docs/nginx.md
```diff
@@ -131,6 +131,9 @@ chown dokku:dokku /home/dokku/myapp/nginx.conf.d/upload.conf
...
+
+
```
josegonzalez: Remove the extra newline.

PR #2103, opened 2016-04-13
Fix plugin rename bug

crisward:
This seems to fix the bug when plugins with git in their name get changed to dokku


michaelshobbs:
Duplicate of #2102

PR #2109, opened 2016-04-14
Allow user to modify the repository and tag when deploying an app

josegonzalez:
Uses two new plugin triggers:

- deployed-app-repository
- deployed-app-image-tag

Useful when the image is a custom tagged image on a remote repository

Needs docs for the new hooks and they also should be added to the events plugin.

### DIFF
--- a/plugins/00_dokku-standard/subcommands/deploy
+++ b/plugins/00_dokku-standard/subcommands/deploy
@@ -9,7 +9,7 @@ dokku_deploy_cmd() {
   declare desc="deploy phase"
   local cmd="deploy"
   [[ -z $2 ]] && dokku_log_fail "Please specify an app to deploy"
-  local APP="$2"; local IMAGE_TAG="$3"; local IMAGE=$(get_app_image_name "$APP" "$IMAGE_TAG")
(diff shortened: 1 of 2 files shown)


plugins/00_dokku-standard/subcommands/deploy
```diff
@@ -9,7 +9,7 @@ dokku_deploy_cmd() {
   declare desc="deploy phase"
   local cmd="deploy"
   [[ -z $2 ]] && dokku_log_fail "Please specify an app to deploy"
-  local APP="$2"; local IMAGE_TAG="$3"; local IMAGE=$(get_app_image_name "$APP" "$IMAGE_TAG")
+  local APP="$2"; local IMAGE_TAG="$3"; local IMAGE=$(get_deployed_app_image_name "$APP" "$IMAGE_TAG")
```
michaelshobbs: i think this is a misnomer. reading that function name makes me think it will return the image name from an app that is already deployed. here we want to know the image name that you want to deploy, right? 

something like `get_deploying_app_image_name()` perhaps?

michaelshobbs:
Looks like the `tags:deploy` test failed

michaelshobbs:
@josegonzalez merged but just noticed no docs for new triggers

PR #2121, opened 2016-04-18
dokku.me domain updated

crisward:
wasn't sure about mentioning 10.0.0.2. I'd hope most people using dokku will know it's a local ip.


u2mejc:
Why drop the example line to add to `/etc/hosts`?

crisward:
Because it's no longer needed. Jeff just updated his dokku.me domain to point at 10.0.0.2. with a wildcard, so no more host file messing... (just discussed on slack)

PR #2127, opened 2016-04-20
added dokku require plugin

crisward:


josegonzalez:
FWIW you can add [ci skip] to documentation PRs.

crisward:
Sorry, CI systems should have an ignore system for specific folders, when it comes to writing to commit message I've pressed submit before I even realise.

josegonzalez:
This is the world we live in.

PR #2128, opened 2016-04-20
add some 'neat' plugins by michaelshobbs

michaelshobbs:


josegonzalez:
By some definition of "neat"...

michaelshobbs:
left to the reader
YMMV
etc etc

PR #2129, opened 2016-04-20
condense upgrading instructions

christiangenco:
No real content changes, just condensing information to make it easier to read.

Links were to `/dokku`, but it looks like that's been changed to `/docs`, so I fixed them.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -1,36 +1,32 @@
 # Upgrading
 
-This document covers upgrades for the 0.3.0 series and up. If upgrading from older versions, we recommend [a fresh install](/dokku/installation) on a new server.
-
-> As of 0.3.18, dokku is installed by default via a debian package. Source-based installations are still available, though not recommended.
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -1,36 +1,32 @@
 # Upgrading
 
-This document covers upgrades for the 0.3.0 series and up. If upgrading from older versions, we recommend [a fresh install](/dokku/installation) on a new server.
-
-> As of 0.3.18, dokku is installed by default via a debian package. Source-based installations are still available, though not recommended.
+If your version of dokku is pre 0.3.0 (check with `dokku version`), we recommend [a fresh install](/docs/installation) on a new server.
```
michaelshobbs: we use viewdocs.io to render our docs site and `/dokku` is expected. please revert that bit.

christiangenco: Ahh that makes sense. Reverted!

Issue #2111, opened 2016-04-15
app build-env is blank when CURL_CONNECT_TIMEOUT or CURL_TIMEOUT are not set


michaelshobbs:
When `build_env_pre_build_buildpack()` is called and `CURL_CONNECT_TIMEOUT` or `CURL_TIMEOUT` are unset globally, we call `config_set --global`. This ends up setting `$APP` to `--global` and the call to `config_export` ends up being `config_export app --global`. The end result is the app's config vars do not get injected into the build container.

michaelshobbs:
@ypadlyak I think this could be your issue. Try setting those two variables up front in the global space and see if it works for you.

PR #2147, opened 2016-04-26
small refactor to support config_set for globals

michaelshobbs:
refs #2111 #2130


josegonzalez:
I responded via chat, but this whole config plugin needs a do-over.

Changes look good, though probably worth having tests for the two broken features so a refactor doesn't completely screw the pooch.

PR #2155, opened 2016-05-02
Add the ability to run containers in detached mode

josegonzalez:
This emulates heroku's one-off containers feature.

### DIFF
--- a/docs/deployment/one-off-processes.md
+++ b/docs/deployment/one-off-processes.md
@@ -32,6 +32,13 @@ dokku --rm-container run node-js-app ls -lah
 dokku --rm run node-js-app ls -lah
 ```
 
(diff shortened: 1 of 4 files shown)


michaelshobbs:
How do we manage said container after it's been detached?

josegonzalez:
It spits out a container id and it's up to the user to do something with that container ID if necessary.

michaelshobbs:
Might be good to capture that in the docs

josegonzalez:
Updated that in docs.

michaelshobbs:
Does this work if you run it from the console? Specifically, I don't think you can use `-it` and `--detach` together, right?

josegonzalez:
It definitely spins off the container and runs the task I want on the cli.

PR #2168, opened 2016-05-09
allow zero-downtime to be completely disabled.

### LINKED ISSUE #2073: [RFC]: completely disable zero-downtime
In `0.5.0` we introduced the `checks` plugin and the ability to disable zero-downtime checks. 
i.e.: https://github.com/dokku/dokku/blob/v0.5.3/plugins/checks/check-deploy#L61-L64
```
  if [[ "$(is_app_checks_enabled "$APP")" == "false" ]]; then
    dokku_log_info2_quiet "Zero downtime checks for app ($APP) have been disabled. moving on..."
    exit 0
  fi
```

However, we don't actually disable "zero-downtime" when doing this. We still startup a new container alongside the previous one and just short circuit the execution of checks.

We would like to allow the latter to be modified on a per app basis thus that a user can define an app's container(s) be stopped prior to the new deployed container being started. This RFC is meant to cover the details on UI and implementation.

New UI:
```
$ dokku checks:help
Usage: dokku checks[:COMAND]

Manage zero-downtime checks.

Additional commands:
    checks <app>           Show zero-downtime status
    checks:disable <app> [<process-type>]    Disable zero-downtime deployment  (WARNING: this will cause downtime during deployments)
    checks:enable <app> [<process-type>]    Enable zero-downtime deployment
    checks:skip <app> [<process-type>]   Skip zero-downtime checks
```

- introduces `checks:skip` that implements the current functionality of just skipping zero-downtime checks
- changes `checks:disable` to truly disable zero-downtime deploys (for all processes or just specified ones)
  - i.e. stop old container before starting new container (implies `skip`)
- `DOKKU_CHECKS_ENABLED=0` will be migrated to `DOKKU_CHECKS_SKIPPED="_all_"`
- `DOKKU_CHECKS_DISABLED` will be a comma-separated list of process types for which we won't follow zero-downtime deployment logic; can also be set to `_all_`.
- Running `disable`, `enable`, `skip` without a process type will be the same as running with all process types. We will use a special marker (perhaps `_all_` or similar) to denote this state.

**updated based on conversation from #2125**

### LINKED ISSUE #2125: [RFC]: don't run overlapping containers for non-web proc types
Should we modify the deploy flow such that for non-web proc types we don't run a second container to provide "zero-downtime?"

This question came up in a couple of my production environments and I know I've seen this on IRC once or twice. I'm thinking it makes sense to stop non-web proc containers before starting the new ones.

Thoughts here?

michaelshobbs:
- [x] migrate old config vars
- [x] implement `checks:skip` to just skip zero-downtime checks
- [x] reimplement `checks:disable` to actually disable zero-downtime
- [x] docs
- [x] tests

```
$ dokku checks:help
Usage: dokku checks[:COMAND]

Manage zero-downtime settings.

Additional commands:
    checks <app>                             Show zero-downtime status
    checks:disable <app> [process-type(s)]   Disable zero-downtime deployment for all processes (or comma-separated process-type list) ***WARNING: this will cause downtime during deployments***
    checks:enable <app> [process-type(s)]    Enable zero-downtime deployment for all processes (or comma-separated process-type list)
    checks:skip <app> [process-type(s)]      Skip zero-downtime checks for all processes (or comma-separated process-type list)
```
closes #2073
closes #2125

### DIFF
--- a/docs/appendices/0.6.0-migration-guide.md
+++ b/docs/appendices/0.6.0-migration-guide.md
@@ -1,0 +1,9 @@
+# 0.6.0 Migration Guide
+
+## zero-downtime deployment
+- You can now **actually** disable zero-downtime deployments per-app and per-process-type
+  - Some config variables have been moved around
+    - `DOKKU_CHECKS_ENABLED` has been migrated to `DOKKU_CHECKS_SKIPPED`
+    - `DOKKU_CHECKS_DISABLED` is now a thing
+    - The values for the above can be a comma-separated list of process-types or the literal string `_all_`
+  - See the [updated `checks`](/dokku/checks-examples/) docs for more info
--- a/docs/checks-examples.md
+++ b/docs/checks-examples.md
@@ -3,9 +3,10 @@
 > New as of 0.5.0
 
 ```
-checks <app>                                                                                 Show zero-downtime status
-checks:disable <app>                                                                         Disable zero-downtime checks
-checks:enable <app>                                                                          Enable zero-downtime checks
+checks <app>                             Show zero-downtime status
+checks:disable <app> [process-type(s)]   Disable zero-downtime deployment for all processes (or comma-separated process-type list) ***WARNING: this will cause downtime during deployments***
+checks:enable <app> [process-type(s)]    Enable zero-downtime deployment for all processes (or comma-separated process-type list)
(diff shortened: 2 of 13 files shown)


benpye:
I'm curious, what's the status on this? This would be useful as with `docker-options` to set the ports to forward this could allow UDP apps under dokku in a reasonable manner. I'm currently trying to do this and my solution is to manually stop the previous application, which obviously isn't ideal.

PR #2170, opened 2016-05-10
Dont fail when adding a duplicated domain

### LINKED ISSUE #2165: Incorrect domain existence check on domains:add
In the [domains_add](https://github.com/dokku/dokku/blob/5940e43c85a36d4a8f609bb00d1d700f25ae8e96/plugins/domains/functions#L69) code, we currently only check for the existence of the first domain, and subsequently bail out if we find it. This command actually takes multiple domains.

We should instead:

- Check on each domain during the iteration [here](https://github.com/dokku/dokku/blob/5940e43c85a36d4a8f609bb00d1d700f25ae8e96/plugins/domains/functions#L75).
- Avoid bailing out and instead just continue, since the desired state already exists.

jvanbaarsen:
Currently we exit when trying to add a duplicated domain, even if we give a list
of domain names. This behaviour is not as it suppose to work.

**Expected behaviour**
When adding a domain, and the domain already exists, just skip that domain, and
continue with the world. This is no problem since we are already in the state we
want to be in.

**New behaviour**
When you try to add a domain that is already added, we notify the user of this
by showing a log message: `Skipping: example.com already added to example app`,
and then we continue with the list of domain names.

Fixes #2165

### DIFF
--- a/plugins/domains/functions
+++ b/plugins/domains/functions
@@ -66,15 +66,13 @@ domains_add() {
   verify_app_name "$1"
   local APP="$1"; local APP_VHOST_PATH="$DOKKU_ROOT/$APP/VHOST"
 
-  if [[ $(egrep -w "^${2}$" "$APP_VHOST_PATH" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
-    dokku_log_fail "$2 is already defined for $APP"
-    exit 1
-  fi
-  shift 1
-
   for DOMAIN in "$@"; do
-    echo "$DOMAIN" >> "$APP_VHOST_PATH"
-    dokku_log_info1 "Added $DOMAIN to $APP"
+    if [[ $(egrep -w "^$DOMAIN$" "$APP_VHOST_PATH" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
+      dokku_log_info1 "Skipping: $DOMAIN already added to $APP"
+    else
+      echo "$DOMAIN" >> "$APP_VHOST_PATH"
+      dokku_log_info1 "Added $DOMAIN to $APP"
+    fi
   done
 
   if [[ "$(is_app_vhost_enabled "$APP")" == "false" ]];then
--- a/plugins/domains/functions
+++ b/plugins/domains/functions
@@ -66,6 +66,8 @@ domains_add() {
   verify_app_name "$1"
   local APP="$1"; local APP_VHOST_PATH="$DOKKU_ROOT/$APP/VHOST"
 
+  shift 1
+
   for DOMAIN in "$@"; do
     if [[ $(egrep -w "^$DOMAIN$" "$APP_VHOST_PATH" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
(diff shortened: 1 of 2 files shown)


jvanbaarsen:
I think that test failure is unrelated. @josegonzalez can you restart the build?

josegonzalez:
Can you check that the docs reflect this new behavior?

plugins/domains/functions
```diff
@@ -66,15 +66,13 @@ domains_add() {
   verify_app_name "$1"
   local APP="$1"; local APP_VHOST_PATH="$DOKKU_ROOT/$APP/VHOST"
 
-  if [[ $(egrep -w "^${2}$" "$APP_VHOST_PATH" > /dev/null 2>&1; echo $?) -eq 0 ]]; then
-    dokku_log_fail "$2 is already defined for $APP"
-    exit 1
-  fi
-  shift 1
```
michaelshobbs: You'll need to include the `shift 1`. Otherwise, the first arg in `$@` is the app name. This is also why the tests are failing.

jvanbaarsen: @michaelshobbs Ah ok, thanks! Tests looked very unrelated xD (But that might be a bats issue)

jvanbaarsen:
@josegonzalez Looks like the docs don't mention anything about adding the domain, or the behaviour around this.

michaelshobbs:
I was reading a combo of bats and dokku output. See below
```
# -----> Added rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f to rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f
# -----> Added test.dokku.me to rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f
# -----> Configuring rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f...(using built-in template)
# -----> Configuring rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f.dokku.me...(using built-in template)
# -----> Configuring test.dokku.me...(using built-in template)
# -----> App rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f has not been deployed. Skipping nginx config creation
# output: http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f.dokku.me http://test.dokku.me
# status: 0
# urls: http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f.dokku.me http://test.dokku.me
# expected: http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f.dokku.me
# http://test.dokku.me
# actual:   http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f
# http://rdmtestapp-j60vt8ibg4esf91uxvhwwn0cjl8ha09f.dokku.me
# http://test.dokku.me
```

The first line is attempting to add the app name as a domain.

jvanbaarsen:
@michaelshobbs :+1: Thanks for explaining :-) I made the change!

PR #2171, opened 2016-05-10
allow all apps to be restored on boot

tobru:
When rebooting a Dokku host the init process tries to restore all apps
using `dokku ps:restore`, alphabetically. If an app is defined but not
deployed, the function `ps_start` called by `dokku ps:restore` will
exit 0 and stop trying to restore the other apps later in the alphabet.
This change fixes the restore behaviour.

### DIFF
--- a/plugins/ps/functions
+++ b/plugins/ps/functions
@@ -91,12 +91,14 @@ ps_start() {
   local APP="$1"; verify_app_name "$APP"
   local IMAGE_TAG=$(get_running_image_tag "$APP");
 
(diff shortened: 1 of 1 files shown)


jvanbaarsen:
@tobru Would it be possible to add a regression test for this?

tobru:
@jvanbaarsen I don't have any idea how to do that :cry:

jvanbaarsen:
@michaelshobbs Is a testing hero :-) He might be able to give you some pointers!

michaelshobbs:
Working on a test now

PR #2173, opened 2016-05-10
add test for ps:restore with undeployed app

michaelshobbs:
refs #2171

### DIFF
--- a/tests/unit/10_ps-herokuish.bats
+++ b/tests/unit/10_ps-herokuish.bats
@@ -8,18 +8,19 @@ setup() {
 
 teardown() {
   destroy_app
+  destroy_app 0 "$MYAPP" || true
 }
 
-# @test "(ps) herokuish" {
-#   # CI support: 'Ah. I just spoke with our Docker expert --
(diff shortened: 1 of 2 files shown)


tests/unit/10_ps-herokuish.bats
```diff
@@ -8,18 +8,19 @@ setup() {
...
-#   # looks like docker exec is built to work with docker-under-libcontainer,
-#   # but we're using docker-under-lxc. I don't have an estimated time for the fix, sorry
-#   skip "circleci does not support docker exec at the moment."
-#   run bash -c "dokku ps $TEST_APP | grep -q \"node web.js\""
-#   echo "output: "$output
-#   echo "status: "$status
-#   assert_success
-# }
+@test "(ps) herokuish" {
+  # CI support: 'Ah. I just spoke with our Docker expert --
```
josegonzalez: Is this still applicable?

michaelshobbs: Yes

PR #2175, opened 2016-05-11
show available types and ids on dokku enter error

### LINKED ISSUE #2174: Improved error messages & smarter behaviour for entering existing containers
**Error messages**
When running `dokku enter APP` without giving it a container-type/id, dokku will return an `No container id specified` which is confusing because it's easier to just pass the `container-type` like `web`.

So it at least should say "No container type specified". Even better suggest an example like `web`.

**Smarter behaviour**
If there is only 1 container running for that app, it should just enter that container without having to specify the container-type oder container-id.

michaelshobbs:
- [x] show available process-types or container-ids
- [x] auto-populate if only one process-type or container
- [x] update docs

example output:
```
$ dokku enter node-js-app
 !     No container type specified.
Available types for app (node-js-app): web worker
$ dokku enter node-js-app --container-id
 !     No container id specified.
Available ids for app (node-js-app): 95d5b4daf6e9fa0db9358588ad84f004ca6fcc828d1d7d0f75320de117fb1818 d1900a88847773250de2f00d33a4b265beb953c4a2b26c2b59c0893879422d50
```

closes #2174

### DIFF
--- a/docs/deployment/one-off-processes.md
+++ b/docs/deployment/one-off-processes.md
@@ -68,6 +68,8 @@ dokku enter node-js-app web.1
 dokku enter node-js-app --container-id ID
 ```
 
+Additionally, you can run `enter` with no container-type. If only a single container-type is defined in your app, you will be dropped into the only running container. This behavior is not supported when specifying a custom command; as described below.
+
 By default, it runs a `/bin/bash`, but can also be used to run a custom command:
 
 ```shell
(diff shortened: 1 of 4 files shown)


josegonzalez:
Should we auto-enter the first container if there is only one process type/process running?

michaelshobbs:
I'll update this shortly

josegonzalez:
Also needs a slight doc tweak :)

michaelshobbs:
Updated

PR #2190, opened 2016-05-13
Add ElasticMQ plugin to documentation

cu12:
Again ^^


michaelshobbs:
Why again?

cu12:
@michaelshobbs You merged my FakeSNS plugin yesterday ;)

michaelshobbs:
Ah ok. Curious, what do both of those plugins do?

cu12:
FakeSNS mimics the behaviour of [Amazon SNS](https://aws.amazon.com/sns/) and [ElasticMQ](https://github.com/adamw/elasticmq) does the same for [Amazon SQS](https://aws.amazon.com/sqs/).

We're using Dokku as a core component in our QA environment. Since we use SNS and SQS services, I created these to be able to test everything we need (mostly :)). I hope others will find them useful as well.

michaelshobbs:
Excellent! Thanks for the share.

PR #2196, opened 2016-05-14
Simplify vhost setting

josegonzalez:
Not sure what the difference is between the two files, but if tests pass, then surely the `HOSTNAME` isn't necessary?

### DIFF
--- a/contrib/dokku-installer.py
+++ b/contrib/dokku-installer.py
@@ -95,9 +95,6 @@ class GetHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
                 os.remove('{0}/VHOST'.format(dokku_root))
             except OSError:
                 pass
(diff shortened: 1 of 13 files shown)


michaelshobbs:
IIRC, `HOSTNAME` can be an IP but `VHOST` cannot.

josegonzalez:
When are we supposed to use one over the other?

michaelshobbs:
That's a good question. Originally VHOST was just for nginx config. If that wasn't populated, we'd use HOSTNAME for printing out app urls

josegonzalez:
So why do we sometimes refer to HOSTNAME in our codebase? Seems a bit wacky to have both...

michaelshobbs:
If I understand the original intent, `HOSTNAME` is for the host itself and `VHOST` is for the default domain for all apps. Then of course there's `VHOST` in the app directory. 

My 2 cents: I think we should leave it as is and make the docs more clear.

PR #2220, opened 2016-05-31
remove DOKKU_PROCFILE before attempting to extract it

josegonzalez:
Otherwise removing the procfile from a repo would result in that file sitting around.


josegonzalez:
@michaelshobbs I think you wanted to wait before merging, any reason why?

PR #2224, opened 2016-06-03
#user-auth page not found

shemerey:
Fix link for #user-auth plugin trigger documentation


josegonzalez:
You are meant to view our docs from the site, not the github repository. This change would break the link [here](http://dokku.viewdocs.io/dokku/deployment/user-management/).

shemerey:
I see, thx

PR #2242, opened 2016-06-12
Avoid calling dokku binary

### LINKED ISSUE #2166: Using --app flag may result in invalid dokku subcommand calls
Description of problem:

If you call a command using the remote client, the `--app APP` flag is attached to the request. Internally, dokku prepends that to the running command for core plugins, which then can be used like normal.

If a dokku command calls `dokku` internally, the second-level call will pick up the `DOKKU_APP_NAME` env var and prepend it to it's own calls, even when it already has `app` set properly. This is a problem for commands like `config:set` which will trigger a release, and therefore call `dokku release APP herokuish latest`, causing the command to be transformed into `dokku release APP APP herokuish latest`, which is going to fail as the image isn't found.

The general problem is that we're always adding `DOKKU_APP_NAME`, even when the flag wasn't set during the current run. Not sure what the best fix is, but we should probably set another flag somehow to avoid unshifting `DOKKU_APP_NAME` twice.

```
-----> uname: Linux dokku.josediazgonzalez.com 3.13.0-43-generic #72-Ubuntu SMP Mon Dec 8 19:35:06 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
-----> memory:
                 total       used       free     shared    buffers     cached
    Mem:           490        464         25          7         36        169
    -/+ buffers/cache:        258        231
    Swap:          999        491        508
-----> docker version:
    Client:
     Version:      1.10.3
     API version:  1.22
     Go version:   go1.5.3
     Git commit:   20f81dd
     Built:        Thu Mar 10 15:54:52 2016
     OS/Arch:      linux/amd64

    Server:
     Version:      1.10.3
     API version:  1.22
     Go version:   go1.5.3
     Git commit:   20f81dd
     Built:        Thu Mar 10 15:54:52 2016
     OS/Arch:      linux/amd64
-----> docker daemon info:
    Containers: 10
     Running: 5
     Paused: 0
     Stopped: 5
    Images: 336
    Server Version: 1.10.3
    Storage Driver: aufs
     Root Dir: /var/lib/docker/aufs
     Backing Filesystem: extfs
     Dirs: 451
     Dirperm1 Supported: false
    Execution Driver: native-0.2
    Logging Driver: json-file
    Plugins:
     Volume: local
     Network: bridge null host
    Kernel Version: 3.13.0-43-generic
    Operating System: Ubuntu 14.04.3 LTS
    OSType: linux
    Architecture: x86_64
    CPUs: 1
    Total Memory: 490 MiB
    Name: dokku.box.dev
    ID: RJDI:RRNQ:IRXS:4AZG:DD55:IC36:RUAV:QE3N:2KNJ:YCDF:33VZ:A3ZA
    WARNING: No swap limit support
-----> sigil version: 0.4.0
-----> herokuish version:
    herokuish: 0.3.12
    buildpacks:
      emberjs.tgz                latest
      buildpack-nginx            v5
      heroku-buildpack-clojure   v75
      heroku-buildpack-erlang    fa17af9
      heroku-buildpack-go        v34
      heroku-buildpack-gradle    v17
      heroku-buildpack-grails    v20
      heroku-buildpack-java      v44
      heroku-buildpack-multi     v1.0.0
      heroku-buildpack-nodejs    v90
      heroku-buildpack-php       v101
      heroku-buildpack-play      v26
      heroku-buildpack-python    v80
      heroku-buildpack-ruby      v146
      heroku-buildpack-scala     v70
-----> dokku version: 0.5.6
-----> dokku plugins:
    plugn: dev
      00_dokku-standard    0.5.6 enabled    dokku core standard plugin
      20_events            0.5.6 enabled    dokku core events logging plugin
      apps                 0.5.6 enabled    dokku core apps plugin
      auth                 0.4.0 enabled    dokku auth plugin
      build-env            0.5.6 enabled    dokku core build-env plugin
      certs                0.5.6 enabled    dokku core certificate management plugin
      checks               0.5.6 enabled    dokku core checks plugin
      common               0.5.6 enabled    dokku core common plugin
      config               0.5.6 enabled    dokku core config plugin
      copyfiles-to-image   1.0.0 enabled    dokku plugin to copy files from the host to the built docker image
      couchdb              1.0.0 enabled    dokku couchdb service plugin
      docker-options       0.5.6 enabled    dokku core docker-options plugin
      domains              0.5.6 enabled    dokku core domains plugin
      enter                0.5.6 enabled    dokku core enter plugin
      git                  0.5.6 enabled    dokku core git plugin
      letsencrypt          0.8.0 enabled    Automated installation of let's encrypt TLS certificates
      logs                 0.5.6 enabled    dokku core logs plugin
      mongo                1.0.0 enabled    dokku mongo service plugin
      mysql                1.0.0 enabled    dokku mysql service plugin
      named-containers     0.5.6 enabled    dokku core named containers plugin
      nginx-vhosts         0.5.6 enabled    dokku core nginx-vhosts plugin
      plugin               0.5.6 enabled    dokku core plugin plugin
      postgres             1.0.0 enabled    dokku postgres service plugin
      proxy                0.5.6 enabled    dokku core proxy plugin
      ps                   0.5.6 enabled    dokku core ps plugin
      redis                1.0.0 enabled    dokku redis service plugin
      rethinkdb            1.0.0 enabled    dokku rethinkdb service plugin
      shell                0.5.6 enabled    dokku core shell plugin
      storage              0.5.6 enabled    dokku core storage plugin
      tags                 0.5.6 enabled    dokku core tags plugin
      tar                  0.5.6 enabled    dokku core tar plugin
```

- Output of failing dokku commands after running `dokku trace on`: https://gist.github.com/josegonzalez/62c692c16331750ef4f863a89ac3b347#file-gistfile1-txt-L1389

Environment details (AWS, VirtualBox, physical, etc.): DigitalOcean image
How was dokku installed?: bootstrap script
How reproducible: Every time
Steps to Reproduce:

1. `dokku couchdb:create service-name # run locally from a dokku app`
2. `dokku couchdb:link service-name # run locally from a dokku repo`

Actual Results: Dokku command fails overall because it prepends the app name on every subsequent embedded `dokku` call.
Expected Results: Linked service.

josegonzalez:
### Problem
Whenever we call a dokku subshell, the `DOKKU_APP_NAME` arg persists. As the official remote client uses `--app` functionality, the following will break a bit:

```
dokku couchdb:link service-name # run locally from a dokku app
```

Why does it break? The above call is translated to the following:

```
dokku --app APP couchdb:link service-name
```

Which then will do weird stuff if you call a subshell (as we do for docker-options and config set) within the couchdb plugin:

```
# this is called
dokku config:set APP KEY=VALUE

# but since we have `DOKKU_APP_NAME` set, the `dokku` binary turns that into this:
dokku config:set APP APP KEY=VALUE
```

### Solution
Rather than sometimes calling the dokku binary and sometimes not, I figure we should move to calling the sourced functions directly so that we standardize on one way of doing things.

The alternative would be to always set `DOKKU_APP_NAME` to an empty string when calling a dokku sub-shell, which we'll totally forget to do.

### Implementation
- removes the following CLI commands
  - `dokku build`
  - `dokku receive`
  - `dokku release`
  - `dokku tar:build`
  - `dokku tar:build-locked`
  - `dokku git:build`
  - `dokku git:build-locked`
- implements a central locking function for use in both git and tar plugins
- calls aforementioned locking function in `receive-app` trigger in given receiving plugin

Closes #2166


plugins/00_dokku-standard/subcommands/version
```diff
@@ -4,7 +4,7 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/checks/functions"
 source "$PLUGIN_AVAILABLE_PATH/config/functions"
 source "$PLUGIN_AVAILABLE_PATH/nginx-vhosts/functions"
-source "$PLUGIN_CORE_AVAILABLE_PATH/proxy/functions"
+source "$PLUGIN_AVAILABLE_PATH/proxy/functions"
```
michaelshobbs: Should a user be able to disable the proxy meta plugin?

josegonzalez: In theory they should be able to use whatever plugin implementation instead of this proxy one. Only the common and standard plugins are not overrideable.

michaelshobbs: This plugin isn't a proxy implementation. It's the interface plugin to call the configured proxy plugin.

josegonzalez:
I have a bit of work to do here to make sure I call everything properly, so don't merge this yet (even if tests start passing).

michaelshobbs:
If you're going to move some `*_cmd()` functions to a sourceable `functions` file, why not move them all? I'm unclear as to the reason we would want git/tar cmds sourceable.

josegonzalez:
The git/tar commands need to be sourceable because things outside of those plugins need to execute them.

michaelshobbs:
Example?

plugins/git/receive-app
```diff
@@ -1,18 +1,20 @@
...
   local APP="$1"; local REV="$2"
+  local cmd="git-build"
 
   # Don't trigger git build if there is no git repository.
   if [[ ! -d "$DOKKU_ROOT/$APP/refs" ]]; then
     true
   else
     # shellcheck disable=SC2086
-    dokku git-build $APP $REV
+    git_build_cmd $cmd $APP $REV
```
josegonzalez: This needs to execute the function.

plugins/git/receive-app
```diff
@@ -1,18 +1,20 @@
...
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/git/functions"
 
-git_receive_app(){
+git_receive_app() {
   declare desc="git receive-app plugin trigger"
   local trigger="git_receive_app"
   local APP="$1"; local REV="$2"
+  local cmd="git-build"
```
michaelshobbs: I'd like to see desc/trigger/cmd at the top of the functions. Seems cleaner to me to have metadata vars at the top.

plugins/tar/functions
```diff
@@ -0,0 +1,70 @@
...
+
+  tar -x -C "$TAR_BUILD_TMP_WORK_DIR" -f "$DOKKU_ROOT/$APP/src.tar" --strip-components="$BOGUS_PARTS"
+  chmod -R u+r "$TAR_BUILD_TMP_WORK_DIR"
+
+  if [[ -f Dockerfile ]] && [[ "$([[ -f .env ]] && grep -q BUILDPACK_URL .env; echo $?)" != "0" ]] && [[ ! -f ".buildpacks" ]]; then
+    plugn trigger pre-receive-app "$APP" "dockerfile" "$TAR_BUILD_TMP_WORK_DIR"
+    dokku receive "$APP" "dockerfile" "$TAR_BUILD_TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
+  else
+    plugn trigger pre-receive-app "$APP" "herokuish" "$TAR_BUILD_TMP_WORK_DIR"
+    dokku receive "$APP" "herokuish" "$TAR_BUILD_TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
```
michaelshobbs: Are we doing anything here about the call to `dokku receive`?

plugins/git/receive-app
michaelshobbs: Is there an analog in the `tar` plugin?

plugins/tar/functions
josegonzalez: Yeah I haven't finished porting the - broken - pr I previously pushed up. Same re: the tar plugin.

josegonzalez: I made this pr mostly so tests would run.

michaelshobbs: Understood. For the record, you don't need to PR to make tests run. CI will run branch pushed to the project repo.

pltchuong:
I'd love to see this is fixed :)

josegonzalez:
I have some work to do but this will go out 0.6.x.

michaelshobbs:
@josegonzalez I still question the benefit of doing this particular refactor

michaelshobbs:
Additionally how does this relate to #2166. Can you show examples of where this is causing problems?

josegonzalez:
Whenever we call a dokku subshell, the `DOKKU_APP_NAME` arg persists. As the official remote client uses `--app` functionality, the following will break a bit:

```
dokku couchdb:link service-name # run locally from a dokku app
```

Why does it break? The above call is translated to the following:

```
dokku --app APP couchdb:link service-name
```

Which then will do weird stuff if you call a subshell (as we do for docker-options and config set) within the couchdb plugin:

```
# this is called
dokku config:set APP KEY=VALUE

# but since we have `DOKKU_APP_NAME` set, the `dokku` binary turns that into this:
dokku config:set APP APP KEY=VALUE
```

Rather than sometimes calling the dokku binary and sometimes not, I figure we should move to calling the sourced functions directly so that we standardize on one way of doing things.

The alternative would be to always set `DOKKU_APP_NAME` to an empty string when calling a dokku sub-shell.

michaelshobbs:
Updated initial comment with implementation details. I think this is g2g assuming tests pass. I'll keep an eye on the test run.

docs/development/plugin-triggers.md
```diff
@@ -164,7 +164,7 @@ esac
 ### `docker-args-build`
 
 - Description:
-- Invoked by: `dokku build`
+- Invoked by: `dokku_build`
```
josegonzalez: We should probably say these are functions within dokku if anything. I would otherwise leave them as is.

dokku
```diff
@@ -100,13 +100,10 @@ execute_dokku_cmd() {
     nginx|nginx:*)
       local PLUGIN_NAME=${PLUGIN_NAME/nginx/nginx-vhosts}
       ;;
-    receive|deploy|build|release)
+    deploy|delete|ls|run|cleanup)
```
josegonzalez: Are there cases where users may have been calling receive? Should we deprecate and *then* remove?

plugins/tar/receive-app
```diff
@@ -11,7 +12,10 @@ tar_receive_app() {
   if [[ ! -f "$DOKKU_ROOT/$APP/src.tar" ]]; then
     true
   else
-    dokku tar:build "$APP"
+    acquire_app_deploy_lock "$APP"
+    # shellcheck disable=SC2086
+    tar_build "$@"
+    release_app_deploy_lock "$APP"
```
josegonzalez: No way for us to move the lock to the function itself? Or maybe a function like `lock_app_build tar "$APP"`?

josegonzalez: Yeah, we have `tar_build_locked_cmd`. Don't know if that can be called here instead.

dokku
michaelshobbs: There was never a case to directly call `receive` AFAIK. Given this is a "minor" release, I say let's document it in the migration guide and leave it as is.

josegonzalez: Thats fair. I guess I've been thinking of our minors as majors until we do the big 1.0.0. But I don't care either way, given that nothing official has called this stuff (afaik).

plugins/tar/receive-app
michaelshobbs: That implementation only worked because we used `flock -o "$APP_BUILD_LOCK" dokku tar:build-locked "$@"`. That is not possible when calling functions.

michaelshobbs: I played around with `flock` quite a bit inside the functions. This was the cleanest and most reliable way. `flock` also seems to release the lock automatically after the process completes.

josegonzalez: Gotcha, makes sense.

PR #2246, opened 2016-06-14
[docs] fixed link to plugin-creation

MichaelSp:


josegonzalez:
You're not meant to read the docs from the github repository, and rather should read from the [dokku website](http://dokku.viewdocs.io/dokku/plugins/). Bit weird, but it's a limitation of our current documentation solution.

PR #2264, opened 2016-06-23
Use Markdown for sponsors page, so it's clickable from GitHub

troy:
I got as far as the sponsors page, but couldn't click Deis to get to their site. Other docs (README, etc) are already using Markdown, so this changes SPONSORS to do so.

AFAIK, satisfies https://github.com/dokku/dokku/blob/master/CONTRIBUTING.md#submitting-changes and the rest of that doc.


josegonzalez:
Thanks for the pull request!

PR #2270, opened 2016-06-26
Allow apps:destroy when not in a project

guillaumewuip:
Previously, when not in a git project with a `dokku` remote, `dokku apps:destroy` failed because of `git remote remove dokku` error status.

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -33,6 +33,14 @@ client_help_msg() {
   exit 20 # exit with specific status. only used in units tests for now
 }
 
+is_git_repo() {
+    git rev-parse &>/dev/null
+}
+
+has_dokku_remote() {
+    git remote show dokku &>/dev/null
+}
+
 if [[ -z $DOKKU_HOST ]]; then
   if [[ -d .git ]] || git rev-parse --git-dir > /dev/null 2>&1; then
     DOKKU_HOST=$(git remote -v 2>/dev/null | grep -Ei "^dokku" | head -n 1 | cut -f1 -d' ' | cut -f2 -d '@' | cut -f1 -d':' 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


michaelshobbs:
How does heroku behave in this situation? 

This client is intended to mimic the official Heroku behaviors and so we should definitely do what they do here.

guillaumewuip:
heroku seems to give us two commands : 
```
$ heroku help
    ...
    apps:delete                    #  permanently destroy an app
    apps:destroy --app APP         #  permanently destroy an app
    ...
```


I've just tested `heroku apps:destroy --app <app>` outside any project and it works :

capture d ecran 2016-06-26 a 17 50 14

michaelshobbs:
Does it end up leaving the remote in your git repo?

guillaumewuip:
In case I use `heroku apps:delete` in the projet repo, yes : 
capture d ecran 2016-06-26 a 18 04 57

michaelshobbs:
Ok I expect that behavior. Does it leave the remote in the repo if you delete the app from another directory?

guillaumewuip:
Yes the `heroku` is not touched when destroying the app from another directory :
capture d ecran 2016-06-26 a 18 10 45

michaelshobbs:
Excellent! Let me get these tests running again and we'll get this merged. 

Thanks for contributing!

guillaumewuip:
@michaelshobbs thanks !

michaelshobbs:
Looks like there is a bug here. Though I'm not 100% certain why based on the diff. If you have time to debug that would be great otherwise I can look at it later.

contrib/dokku_client.sh
```diff
@@ -81,7 +89,7 @@ if [[ ! -z $DOKKU_HOST ]]; then
         fi
         ;;
     apps:destroy)
-      git remote remove dokku
+      is_git_repo && has_dokku_remote && remote remove dokku
```
josegonzalez: This is missing the `git` part of the `remote remove dokku` call.

PR #2273, opened 2016-06-27
Add the ability to manually execute checks against an application

### LINKED ISSUE #2218: checks plugin should expose a command for manual checking
When implementing monitoring for applications it would be rather helpful to re-use the existing infrastructure for application checks using the `CHECKS` file. From what I gather it's not possible to execute the checks outside deploying as of Dokku 0.5.7.

A command like `dokku checks:run <app>` with the exit status matching the check result would be desirable.

josegonzalez:
Checks can be run in several ways:

- checks:run APP
- checks:run APP PROCESS_TYPE
- checks:run APP PROCESS_TYPE.CONTAINER_INDEX

This will reuse the existing checks infrastructure as defined by the checks plugin, and therefore if no checks file is defined, it will default to the normal Dokku container check.

Closes #2218

### DIFF
--- a/plugins/checks/check-deploy
+++ b/plugins/checks/check-deploy
@@ -53,7 +53,6 @@ checks_check_deploy() {
     local DOKKU_APP_CONTAINER_ID=${DOKKU_APP_CIDS[0]}
   fi
 
-
   # source global and in-app envs to get DOKKU_CHECKS_WAIT and any other necessary vars
   eval "$(config_export global)"
   eval "$(config_export app "$APP")"
--- a/docs/deployment/zero-downtime-deploys.md
+++ b/docs/deployment/zero-downtime-deploys.md
@@ -3,10 +3,11 @@
 > New as of 0.5.0
 
 ```
-checks <app>                                   # Show zero-downtime status
-checks:disable <app> [process-type(s)]         # Disable zero-downtime deployment for all processes (or comma-separated process-type list) ***WARNING: this will cause downtime during deployments***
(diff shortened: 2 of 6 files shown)


josegonzalez:
@michaelshobbs tips on how I should test this functionality?

plugins/checks/check-deploy
```diff
@@ -53,12 +53,11 @@ checks_check_deploy() {
     local DOKKU_APP_CONTAINER_ID=${DOKKU_APP_CIDS[0]}
   fi
 
-
   # source global and in-app envs to get DOKKU_CHECKS_WAIT and any other necessary vars
   eval "$(config_export global)"
   eval "$(config_export app "$APP")"
 
-  if [[ "$(is_app_proctype_checks_skipped "$APP" "$PROC_TYPE")" == "true" ]]; then
+  if [[ "$(is_app_proctype_checks_skipped "$APP" "$DOKKU_APP_CONTAINER_TYPE")" == "true" ]]; then
```
michaelshobbs: Whoopsie 😸

josegonzalez: I mean it still worked...

michaelshobbs: `is_app_proctype_checks_skipped()` will always return false without this fix

michaelshobbs:
Adding a test now

plugins/checks/check-deploy
josegonzalez: I *think* the variable will be available to sub-shells, and the thing that calls this trigger sets `PROC_TYPE`.

PR #2275, opened 2016-06-27
Prefer file import in certs:add if files given

sseemayer:
Fixes dokku/dokku-letsencrypt#67

Change the `certs:add` command to use a file import without trying to
detect a piped-in tarball if server.crt and server.key files were passed
in as arguments and those files exist.

Fixes hangs on certificate imports by filename when the check for a
STDIN pipe fails e.g. when running `dokku certs:add` or `dokku letsencrypt` remotely using a SSH command.


michaelshobbs:
Gracias!

PR #2277, opened 2016-06-28
Attempt to support strings with double-quotes

josegonzalez:
This is probably not the best way, but I'm trying it nonetheless.

Refs #1262

### DIFF
--- a/plugins/build-env/pre-build-buildpack
+++ b/plugins/build-env/pre-build-buildpack
@@ -27,7 +27,7 @@ build_env_pre_build_buildpack() {
     dokku_log_info1 "Adding BUILD_ENV to build environment..."
     # create build env files for use in buildpacks like this:
     # https://github.com/niteoweb/heroku-buildpack-buildout/blob/5879fa3418f7d8e079f1aa5816ba1adde73f4948/bin/compile#L34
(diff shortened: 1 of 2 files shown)


josegonzalez:
I was thinking about this a bit more, and perhaps a cleaner way would be to do something like the following (pseudocode btw):

```
TMP=create_tmp_env_dir_on_host
for value in app_config; do
    echo "$VALUE" > "TMP/$KEY"
done
docker run -v $TMP:/tmp/env /bin/bash -c '/bin/bash /tmp/set_env.sh'
```

Would avoid all the sed calls and potentially allow us to have an easier time of handling various types of values...

josegonzalez:
Another attempt could be to build a non-exported version of the config vars:

```
 dokku config --shell python-sample > $TMP/file
docker run -v $TMP:/tmp/mounted-env /bin/bash -c 'mkdir /tmp/env && for key in file; do WRITE_VALUE; done'
```

josegonzalez:
Ugh tests

cu12:
Without trying I was wondering if `printf %q` could be in help here? I'm using that quite successfully for complicated MySQL queries over ssh.

http://wiki.bash-hackers.org/commands/builtin/printf

Issue #2276, opened 2016-06-27
Multiple Hostnames


xadh00m:
We have a dokku server and can reach it via two distinct hostnames configured in our DNS server. It seems to be not supported yet to declare two hostnames in dokku. Although this works:
```
dokku domains:set-global "foo bar"
```
It does not create the correct hostnames for apps when calling:
```
dokku domains:disable myApp && dokku domains:enable myApp
```
The global domain is treated as one domain string. This leads to the app hostnames:
```
dokku domains myApp
=====> Global Domain Name
foo bar
=====> myApp Domain Names
myApp.foo bar
```

Is this feature something Dokku should support? AFAIK the relevant functionlity is in `dokku/plugins/enabled/domains/functions`.
Would you be interested in a PR for this? Any hints how this feature should be incorporated and which edge cases need to be considered?

Thanks,
Mirko

josegonzalez:
A few things:

- If you put quotes around strings in bash, bash will treat them as a single string.
- We do not have the ability to set more than one global hostname. That's why the method is called `domains:set-global`, not `domains:add-global`.
  - @michaelshobbs it appears that we can get more than one global hostname via `get_app_domains`, but we never use the function that way. I'm actually not sure where in the code we retrieve all the hostnames...
- If you'd like to set more than one hostname for an application, you can use the `dokku domains:add` command.

xadh00m:
Thank you for the fast reply!

Yes, `dokku domains:add`can achieve that but I would actually like to declare another global domain which applies to all deployed and future apps without setting an extra domain per app.

josegonzalez:
We don't currently support that. If you'd like to contribute a patch to add `domains:add-global` and `domains:set-global` (with tests!), that would be great.

xadh00m:
`domains:set-global`does exist already. Did you mean `domains:remove-global` instead?
Speaking of it: Does it sound more consistent to align the domains global commands with the app domains commands?

E.g.:
```
domains:add-global
domains:remove-global
domains:clear-global
domains:disable-global
domains:enable-global
```
Note: No `domains:set-global` command there for consistency.

Although this looks a bit more consistent it might be hard to achieve e.g. because [`is_global_vhost_enabled`](https://github.com/dokku/dokku/blob/master/plugins/common/functions#L824) is determined implicitly.

josegonzalez:
Yeah I meant remove.

xadh00m:
[...]

PR #2283, opened 2016-06-29
Multiple global domains

xadh00m:
Refs #2276


docs/deployment/domain-configuration.md
```diff
@@ -4,11 +4,13 @@
...
 ```shell
 domains:add <app> DOMAIN                                                                     Add a domain to app
+domains:add-global <domain>                                                                  Add global domain name
 domains [<app>]                                                                              List domains
 domains:clear <app>                                                                          Clear all domains for app
 domains:disable <app>                                                                        Disable VHOST support
 domains:enable <app>                                                                         Enable VHOST support
 domains:remove <app> DOMAIN                                                                  Remove a domain from app
+domains:remove-global <domain>                                                               Remove global domain name
 domains:set-global <domain>                                                                  Set global domain name
```
michaelshobbs: I think we'll want to drop `set-global` in favor of `add-global`/`remove-global`

plugins/common/functions
```diff
@@ -814,24 +814,24 @@ get_ipv6_regex() {
...
-  local GLOBAL_VHOST=$(get_global_vhost)
-  local GLOBAL_VHOST_ENABLED=true
+  local GLOBAL_VHOSTS=$(get_global_vhosts)
+  local GLOBAL_VHOSTS_ENABLED=true
   local RE_IPV4="$(get_ipv4_regex)"
   local RE_IPV6="$(get_ipv6_regex)"
 
-  if [[ -z "$GLOBAL_VHOST" ]] || [[ "$GLOBAL_VHOST" =~ $RE_IPV4 ]] || [[ "$GLOBAL_VHOST" =~ $RE_IPV6 ]]; then
-    local GLOBAL_VHOST_ENABLED=false
+  if [[ -z "$GLOBAL_VHOSTS" ]] || [[ "$GLOBAL_VHOSTS" =~ $RE_IPV4 ]] || [[ "$GLOBAL_VHOSTS" =~ $RE_IPV6 ]]; then
```
michaelshobbs: @josegonzalez what behavior would we expect if a user set the global host to something like
```
dokku.me
184.23.5.105
```

plugins/domains/functions
```diff
@@ -20,40 +20,43 @@ get_app_domains() {
   fi
 }
 
-get_default_vhost() {
+get_default_vhosts() {
   declare desc="return default vhost"
```
michaelshobbs: for completeness, the description should be made plural as well

plugins/common/functions
michaelshobbs: The main issue I see here is inconsistent behavior depending on global host order.

examples
- this will return `false`
```
dokku.me
184.23.5.105
```
this will return true
```
184.23.5.105 
dokku.me
```
Looks like we strip out IPs when we retrieve global vhosts. So as long as we have one valid hostname, I think this can safely return true. Of course, this function will need to be modified to do that.

docs/deployment/domain-configuration.md
josegonzalez: We could not list it, though I think we still need to have the command around (and it proxies to the add-global?). Set has different semantics from add though.

michaelshobbs: Why have both?

josegonzalez: Oh this is for next minor, nevermind. Fine to remove then.

xadh00m:
I think I´ve addressed all comments of this PR so far. Anything else I could do to land it?

josegonzalez:
You could help us work on all the other issues for 0.7.0.

josegonzalez:
@xadh00m Mind rebasing this PR on top of master?

xadh00m:
I´ve rebased all changes. Thank you!

josegonzalez:
Can you refresh your CircleCI build?

xadh00m:
The build succeeded this time.

PR #2290, opened 2016-07-02
Implement restart-policy handling

### LINKED ISSUE #1734: [RFC] Revisit docker restart policy
### RFC from @josegonzalez
---

I think the way to go here is to set a config var like `DOKKU_RESTART_POLICY`. This variable could contain the exact restart policy being applied to an application. An alternative is to use the docker-options plugin.

Valid restart policies would match the following bash regular expression:

```
^(no|unless-stopped|always|on-failure(:[1-9][0-9]*)?)$
```

If you unset that config variable - or if it did not yet exist - it would be reset to the initial default of `on-failure:10` Removing the docker-options entry would simply remove all restart-policies. Restart policies would apply to *all* processes being managed.

Potential porcelain around the plumbing:

```shell
# show the restart policy
dokku ps:restart-policy APP

# set the restart policy for a given application
dokku ps:set-restart-policy APP on-failure:10
```

### Original post

---

Is it time to look into the restart policy again with docker 1.9.x out? Add a new unless-stopped restart policy (#15348) was added in 1.9.0

See: 
#1591
#1505  

Repasting my message from https://github.com/progrium/dokku/pull/1591#issuecomment-161059631 below:

OK, I was able to test out the ```docker-option --restart=unless-stopped```.  Based on the testing that I have performed this behaves as I would desire.  

Below is a summary of my testing:
I have a test-app deployed with DOKKU_SCALE=2

```
    dokku docker-options:add test-app deploy --restart=unless-stopped
    sudo reboot
```

After setting the restart option following any reboot I would see the following behavior:
* The initial ```docker ps``` shows 4 containers up.
* After a couple of minutes it shows only the 2 expected containers up.

```docker inspect``` showed for both containers:
```json
"RestartPolicy": {
            "Name": "unless-stopped",
            "MaximumRetryCount": 0
        }
```

After causing the app to crash (process.exit(1)) each container twice:
```docker inspect``` showed for both containers:
```json 
"RestartCount": 2
```

Then ```sudo reboot```
* See reboot behavior description above.

Execute ```dokku ps:restart test-app``` 3 Times in quick succession:
* ```docker ps``` Shows 8 containers up for 1-2 minutes
* Then back down to expected 2

Then ```sudo reboot```
* See reboot behavior description above.

Then ```dokku ps:stop test-app```
* ```docker ps``` shows no containers running

Then ```sudo reboot```
* This time only two containers come up.  
* This is the same as doing a ```ps:stop``` and reboot with an app that does not have the restart policy.

Push new version of the test-app
* Behaves the same as the reboot behavior description above.
* ```docker inspect``` shows that the RestartPolicy is retained for both containers.

Test crash the test-app 
* the containers are still restarted

Then ```sudo reboot```
* Same reboot behavior described above

josegonzalez:
Applications without a restart-policy will have their policies set to `on-failure:10`. Users can completely unset the restart-policy using the `docker-options` plugin, though this will be equivalent to setting it explicitly to `no`.

Restart policies must be explicitly set, and the following are all valid:

- no
- unless-stopped
- always
- on-failure
- on-failure:NUMBER

TODO:

- [x] Tests (please save me @michaelshobbs)

Closes #1734

### DIFF
--- a/plugins/ps/commands
+++ b/plugins/ps/commands
@@ -16,6 +16,8 @@ case "$1" in
     ps:restart <app>, Restart app container(s)
     ps:restartall, Restart all deployed app containers
     ps:restore, Start previously running apps e.g. after reboot
+    ps:restart-policy <app>, Shows the restart-policy for an app
+    ps:set-restart-policy <app> <policy>, Sets app restart-policy
 help_content
     }
 
--- a/plugins/ps/functions
+++ b/plugins/ps/functions
@@ -166,14 +166,14 @@ ps_scale() {
   fi
 }
 
-get_restart_policies() {
-  declare desc="strips docker options and prints restart policies"
-  local -r phase_file_path=$1
-  [[ -r "$phase_file_path" ]] && sed -e '/^--restart=/!d' -e 's/^--restart=/    /' < "$phase_file_path"
-}
-
 get_raw_restart_policies() {
   declare desc="strips docker options and prints raw restart policies"
   local -r phase_file_path=$1
   [[ -r "$phase_file_path" ]] && sed -e '/^--restart=/!d' < "$phase_file_path"
+}
+
+get_restart_policies() {
+  declare desc="strips docker options and prints restart policies"
+  local -r phase_file_path=$1
+  get_raw_restart_policies "$phase_file_path" | sed -e 's/^--restart=//g'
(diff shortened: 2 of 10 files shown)


michaelshobbs:
tests added

PR #2299, opened 2016-07-05
url subcommand returns http even if https is configured

cu12:
I recently upgraded to the latest version and noticed that the behaviour of `dokku url APP` has been changed and returns `http` scheme instead of `https`. This unfortunately breaks the deployment for some of my applications.

I know about the `dokku urls` command, although I was wondering if this change was really intended as I couldn't find anything in release notes. Also since `http` really just redirects to `https` if certs are configured it's really not the correct address to display.

If you think this should go in, I'll add a test as well.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -991,7 +991,7 @@ get_app_urls() {
     local URLS="$(merge_dedupe_list "$port_urls $app_urls" " ")"
     case "$1" in
       url)
-        echo "$URLS" | tr ' ' '\n' | head -n1
+        echo "$URLS" | tr ' ' '\n' | tail -1
       ;;
       urls)
         echo "$URLS" | tr ' ' '\n' | sort
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -961,6 +961,7 @@ get_container_ports() {
 
 get_app_urls() {
   declare desc="print an app's available urls"
+  source "$PLUGIN_AVAILABLE_PATH/certs/functions"
   source "$PLUGIN_AVAILABLE_PATH/config/functions"
   source "$PLUGIN_AVAILABLE_PATH/domains/functions"
   source "$PLUGIN_AVAILABLE_PATH/proxy/functions"
@@ -976,7 +977,7 @@ get_app_urls() {
       local port_map app_vhost
       local app_vhosts=$(get_app_domains "$APP")
       for port_map in $DOKKU_PROXY_PORT_MAP; do
-        local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
+      local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
         local listen_port="$(awk -F ':' '{ print $2 }' <<< "$port_map")"
         for app_vhost in $app_vhosts; do
           if [[ "$listen_port" != "80" ]] && [[ "$listen_port" != "443" ]]; then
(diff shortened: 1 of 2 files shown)


josegonzalez:
Definitely needs a test.

michaelshobbs:
This was most likely a side-effect of the proxy port mapping stuff. Might be a better solution to grab all https urls if a cert is configured and pick the first or last of those instead of just inverting the logic of `dokku url APP`. My 2 cents...

josegonzalez:
@cu12 Mind updating the PR with @michaelshobbs comment?

josegonzalez:
Tests fail because dokku.me doesn't have https. Not sure what the best thing to do here is... @michaelshobbs ?

plugins/common/functions
```diff
@@ -976,7 +977,7 @@ get_app_urls() {
       local port_map app_vhost
       local app_vhosts=$(get_app_domains "$APP")
       for port_map in $DOKKU_PROXY_PORT_MAP; do
-        local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
+      local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
```
michaelshobbs: need to indent this line

tests/unit/40_core_2.bats
```diff
@@ -20,6 +20,15 @@ assert_urls() {
   assert_output < <(tr ' ' '\n' <<< "${urls}" | sort)
 }
 
+assert_url() {
```
michaelshobbs: why do we need this when we already have`assert_urls`?

michaelshobbs: nvm. i see now

michaelshobbs:
The error is due to a syntax error. Additionally the logic of the new test doesn't make sense. specifically these lines:

```
add_domain "test.dokku.me"
assert_url "https://test.dokku.me" "https://${TEST_APP}.dokku.me"
``` 

`assert_url` calls `dokku url $APP` and will only over return the very first url. i think this patch to the PR is sufficient to test the desired behavior

```
diff --git a/plugins/common/functions b/plugins/common/functions
index 041a129..5857883 100755
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -977,7 +977,7 @@ get_app_urls() {
       local port_map app_vhost
       local app_vhosts=$(get_app_domains "$APP")
       for port_map in $DOKKU_PROXY_PORT_MAP; do
-      local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
+        local scheme="$(awk -F ':' '{ print $1 }' <<< "$port_map")"
         local listen_port="$(awk -F ':' '{ print $2 }' <<< "$port_map")"
         for app_vhost in $app_vhosts; do
           if [[ "$listen_port" != "80" ]] && [[ "$listen_port" != "443" ]]; then
diff --git a/tests/unit/40_core_2.bats b/tests/unit/40_core_2.bats
index 69f5bbe..7608171 100644
--- a/tests/unit/40_core_2.bats
+++ b/tests/unit/40_core_2.bats
@@ -26,7 +26,7 @@ assert_url() {
   echo "output: "$output
   echo "status: "$status
   echo "url: ${url}"
-  assert_output < <("${url}")
+  assert_output "${url}"
 }

 build_nginx_config() {
@@ -76,8 +76,6 @@ build_nginx_config() {
   assert_url "https://dokku.me"
   build_nginx_config
   assert_url "https://${TEST_APP}.dokku.me"
-  add_domain "test.dokku.me"
-  assert_url "https://test.dokku.me" "https://${TEST_APP}.dokku.me"
 }
```

michaelshobbs:
@cu12 let me know if that makes sense or not

josegonzalez:
@cu12 mind updating your pr?

josegonzalez:
@cu12 *bump*

josegonzalez:
@cu12 *bump*

josegonzalez:
Closing in favor of #2848

Issue #2005, opened 2016-03-18
Fully qualified domain name being pushed as subdomain


samholmes1337:
I have created a docker image that I am pushing to my dokku server hosted at `example.xyz`. I have set the git remote to `dokku@example.xyz:subdomain.example.xyz`, as I host multiple different primary domains from this server.

However after going through a brief building process it says that the application has been successfully deployed on `http://subdomain.example.xyz.example.xyz`. Was wondering if this was perhaps something to do with the TLD `xyz`?

Running `Dokku 0.4.14`, `Linux example 3.13.0-77-generic #121-Ubuntu SMP Wed Jan 20 10:50:42 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`, `Docker 1.10.3`. It is currently being hosted on a 512mb DigitalOcean droplet that was setup through their app system.

samholmes1337:
I have been able to reproduce this issue when using a `herokuish` build. The following is the terminal output:

```

**Comment by josegonzalez** [comment]
Can you add *all* the information we ask for in our [ISSUE_TEMPLATE.md](https://github.com/dokku/dokku/blob/master/ISSUE_TEMPLATE.md)?

**Comment by samholmes1337** [comment]
### uname -a
```
Linux example 3.13.0-77-generic #121-Ubuntu SMP Wed Jan 20 10:50:42 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

### free -m
```
             total       used       free     shared    buffers     cached
Mem:           490        468         21          1         33        186
-/+ buffers/cache:        249        240
Swap:            0          0          0
```

### docker version
```
Client:
 Version:      1.10.1
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   9e83765
 Built:        Thu Feb 11 19:27:08 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.1
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   9e83765
 Built:        Thu Feb 11 19:27:08 2016
 OS/Arch:      linux/amd64
```

### docker -D info
```
Containers: 8
 Running: 3
 Paused: 0
 Stopped: 5
Images: 25
Server Version: 1.10.1
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 73
 Dirperm1 Supported: false
Execution Driver: native-0.2
Logging Driver: json-file
Plugins:
[...]

PR #2304, opened 2016-07-08
Add distinction for fully qualified domain name

samholmes1337:
When publishing a fully-qualified domain name (Add initial check for FQDN termination or FQDN: a DNS entry terminated by a period), it will now bypass all other checks and just publish that domain name as is.

Refs #2005.


samholmes1337:
Made another mistake, sorry guys!

PR #2316, opened 2016-07-14
use default privileged user in herokuish-0.3.17

michaelshobbs:
- adds `--env=USER=herokuishuser"` to `$DOKKU_GLOBAL_RUN_ARGS`

### DIFF
--- a/deb.mk
+++ b/deb.mk
@@ -1,6 +1,6 @@
 HEROKUISH_DESCRIPTION = 'Herokuish uses Docker and Buildpacks to build applications like Heroku'
 HEROKUISH_REPO_NAME ?= gliderlabs/herokuish
-HEROKUISH_VERSION ?= 0.3.14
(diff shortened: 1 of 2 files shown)


deb.mk
```diff
@@ -95,7 +95,7 @@ deb-herokuish:
 	@echo "sudo docker build -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
 
 	@echo "-> Cloning repository"
-	git clone -q "https://github.com/$(HEROKUISH_REPO_NAME).git" /tmp/tmp/herokuish > /dev/null
+	git clone -q "https://github.com/$(HEROKUISH_REPO_NAME).git" --branch release /tmp/tmp/herokuish > /dev/null
```
josegonzalez: Why not checkout the tag instead?

michaelshobbs: Only because that's not what we talked about....

PR #2317, opened 2016-07-15
Properly remap http port 80 mappings to https 443 when adding an ssl certificate

josegonzalez:
Rather than removing all port mappings, we simply remap port 80 mappings to 443. The previous behavior would remove custom port mappings for all applications and reset them on any certificate change, preventing automation from plugins such as letsencrypt. While this behavior doesn't matter for buildpack deploys - which only expose a single port on container port 5000 - Dockerfile deploys would frequently be affected by such a change, requiring a remapping of all custom ports.

/cc @sseemayer, as this should simplify the usage of letsencrypt with Dokku a bit :)

### DIFF
--- a/docs/configuration/domains.md
+++ b/docs/configuration/domains.md
@@ -12,6 +12,8 @@ domains:remove <app> DOMAIN              # Remove a domain from app
 domains:set-global <domain>              # Set global domain name
 ```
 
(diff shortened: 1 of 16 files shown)


josegonzalez:
Heh my patch to avoid setting state doesn't quite work, because we need to extract the port mapping to retrieve urls. I still think that us setting the port mapping in the nginx plugin (and not the proxy plugin) is poor form, but the lolzier part is that the url retrieval occurs in the common plugin. *shrug* ah well.

PR #2358, opened 2016-08-08
Guard against poodle vulnerability by default

### LINKED ISSUE #2339: POODLE vulnerability
Would it be okay for me to submit a PR disabling SSLv3? I made [this](https://gitlab.com/dokku-alt/dokku-alt/commit/eca3b1da18649931ef32bcc768467be7d2e45536#diff-0) same change in my local template and it works well, SSL Labs is not reporting the vulnerability anymore. It seems like this is an important thing to fix. The only remaining issue as reported by SSL Labs is "The server does not support Forward Secrecy with the reference browsers. Grade reduced to A-." It was originally a C until I disabled SSLv3.

josegonzalez:
Closes #2339

### DIFF
--- a/plugins/nginx-vhosts/templates/nginx.conf.sigil
+++ b/plugins/nginx-vhosts/templates/nginx.conf.sigil
@@ -47,6 +47,7 @@ server {
 
   ssl_certificate     {{ $.APP_SSL_PATH }}/server.crt;
   ssl_certificate_key {{ $.APP_SSL_PATH }}/server.key;
+  ssl_protocols       TLSv1 TLSv1.1 TLSv1.2;
 
   keepalive_timeout   70;
   {{ if eq $.SPDY_SUPPORTED "true" }}add_header          Alternate-Protocol  {{ $.NGINX_SSL_PORT }}:npn-spdy/2;{{ end }}


ninive:
Hi Jose, 
sure everything rocks! Just to ping out that added line is not there anymore, at least not in my automagical letsencrypt/dokku conf. I didn't find any other track online, so dunno if is a kind of missing or is on purpose. I've patched the sigil and got identical OP behavior. A- now.

Anyway, thanks for all the rest, hope this helps. 

ssl3

josegonzalez:
We default to *only* TLS 1.2 now as shown [here](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/templates/nginx.conf.sigil#L68). That change was made in #2678.

Are you sure you're running the latest version of Dokku? Is it possible you're running some custom local hack somewhere? We're for sure shipping the right thing as far as I am aware.

josegonzalez:
Also, please don't comment on 1+ year old closed issues or merged/closed pull requests, they are likely to get lost in terms of tracking. If you are seeing problems, the best way to get a resolution is to file a new issue, not by dredging up old work. Thanks!

ninive:
Sure, many thanks for this.  I've seen the latest repo, that's why was strange. I think something got wrong keeping up the upgrades, Peeke is a pretty old product, even if I've followed the guide during the upgrades, probably the latest wasn't upgraded correctly. 

Anyway, the fresh install is okay. I did it at the end. Thanks, didn't want to dig for real  ;)

Issue #2318, opened 2016-07-15
cannot add domains includes `--` such as IDN


znz:
Description of problem:

```
vagrant@dokku:~$ dokku domains:add ruby-rails-sample xn--r8jz45g.xn--zckzah
xn--r8jz45g.xn--zckzah is invalid. exiting...
vagrant@dokku:~$ dokku domains:add ruby-rails-sample xn--eckwd4c7cu47r2wf.jp
xn--eckwd4c7cu47r2wf.jp is invalid. exiting...
```

Output of the following commands

- `dokku version`: `v0.6.5-13-g714da78`

Environment details (AWS, VirtualBox, physical, etc.): vagrant on VirtualBox

How was dokku installed?: `vagrant up --provision`

How reproducible:
`dokku domains:add` with IDN.

Actual Results:
IDN is invalid. exiting...

Expected Results:
Added IDN to app

michaelshobbs:
Reference: https://www.icann.org/resources/pages/idn-2012-02-25-en

josegonzalez:
Closed by #2360.

Issue #2231, opened 2016-06-09
domains:add accepts invalid domains


hansmi:
`dokku domains:add` accepts strings not representing a FQDN:

```
$ dokku domains:add app7 http://domain.tld/
-----> Added http://domain.tld/ to app7
```

Since pull request #2229 they can be removed at least. Before that revision and thus in release 0.5.7 they can't even be removed without manually modifying the VHOST file.

Expected behaviour: Reject domains not in `[-_.a-zA-Z0-9]+` (the actual rules are [far more complicated](https://stackoverflow.com/a/26618995), but I guess this would be sufficient for Dokku).

PR #2360, opened 2016-08-09
Allow hyphen in TLD

### LINKED ISSUE #2359: domains:add rejects domains with hyphen in host path
Description of problem:
domains:add rejects domains that contain a hyphen in the host path e.g. dokku.my-server

How was dokku installed?: apt-get

Dokku version: 0.7.0

How reproducible: easily

Steps to Reproduce:
1. dokku domains:add <app_name> dokku.my-server

Actual Results: 
dokku.my-server is invalid. exiting...

Expected Results:
domain added

We face this issue running dokku in a private network.

xadh00m:
Fixes #2359

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -1065,7 +1065,7 @@ strip_inline_comments() {
 
 is_valid_hostname() {
   declare desc="return 0 if argument is a valid hostname; else return 1"
-  local hostname_string="${1,,}"; local hostname_regex='^([a-z0-9\-\*]+\.)*[a-z0-9\-\*]+$'
+  local hostname_string="${1,,}"; local hostname_regex='^([a-z0-9\*-]+\.)*[a-z0-9\*-]+$'
   if [[ $hostname_string =~ $hostname_regex ]]; then
     return 0
   else
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -1065,7 +1065,7 @@ strip_inline_comments() {
 
 is_valid_hostname() {
   declare desc="return 0 if argument is a valid hostname; else return 1"
-  local hostname_string="${1,,}"; local hostname_regex='^(([a-z0-9\*](-?[a-z0-9])*)\.)+([a-z0-9\*](-?[a-z0-9])*)$'
+  local hostname_string="${1,,}"; local hostname_regex='^([a-z0-9\-\*]+\.)*[a-z0-9\-\*]+$'
   if [[ $hostname_string =~ $hostname_regex ]]; then
     return 0
   else
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -1065,7 +1065,7 @@ strip_inline_comments() {
 
 is_valid_hostname() {
   declare desc="return 0 if argument is a valid hostname; else return 1"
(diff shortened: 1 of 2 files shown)


xadh00m:
This also removes the restriction that the last host part needs to consist of at least two characters. Is this an issue?

jvanbaarsen:
I think leaving the at least 2 characters for tld in there makes sense. Since it is a requirement. Also not really sure about allowing the hyphen. I went over available tld list, and no tld has a hyphen in there. /cc @josegonzalez

xadh00m:
We use dokku in a company private network where they use hyphen in the tld (like "backend.i-robot").
I think its fine to relax the validation. AFAIK the whole validation has been introduced to prevent user from entering uris like: "http://dokku.me".

How should the requirement for at least two characters look like with hyphen?

josegonzalez:
Needs tests before I'll merge this.

znz:
IDN TLD have `--`. see https://github.com/dokku/dokku/issues/2318#issuecomment-233016277
And including `---` is valid domain. for example: https://github.com/plataformatec/devise/blob/6720f93d2c54904f10b6e6d6beb51d0b5b330976/test/devise_test.rb#L97

josegonzalez:
Awesome, that works for me.

@xadh00m mind updatnig the tests to include the domain from #2318?

jvanbaarsen:
@josegonzalez / @xadh00m Aren't they talking about the domain part itself? not the tld? as in:

**Allowed**: domain--name-with-hyphen.com
**not allowed**: domain.hi--there

xadh00m:
I did not (yet) find a clear statement regarding IDN TLDs with one ore more hyphen in the ICANN specs.

@znz: How did you get the IDNs with hyphens in the TLD mentioned in your [comment](https://github.com/dokku/dokku/issues/2318#issue-165784999)?

xadh00m:
At least there seem to be no practially used [gTLDs](https://domaintyper.com/domain-names/top-level-domains/new-gTLD) with hyphens yet.

Anyway, Dokku is used in private networks with custom domain name policies in place.
Should Dokku enforce any kind of spec here? Should Dokku help the user typing the domain correctly according to the latest IDN specs (if that is even possible with a regex)?

The [initial intention](https://github.com/dokku/dokku/issues/2231) of the validation was to ensure that the user does not type "http://dokku.me" but "dokku.me" which made the API clear and prevented a specific error case.

jvanbaarsen:
@xadh00m I don't think the main use case for Dokku is to be within a private network.

xadh00m:
I agree that it might not be the main use case but its an actual use case and might be worth supporting it.

znz:
Examples are from JPNIC's web sites.
https://www.nic.ad.jp/ja/basics/terms/idntld.html
http://punycode.jp/

`IDNs non Ascii` are gTLDs with hyphens, because there are encoded as [Punycode](https://en.wikipedia.org/wiki/Punycode).
For example, `みんな` gTLD is `xn--q9jyb4c`.

josegonzalez:
@xadh00m can you update your pr to add a few tests for hyphens in the tld? That would at least be a step in the right direction, and I'd be happy to support that use case.

xadh00m:
I´ve updated the regex and added a test case accordingly.

xadh00m:
See [regexper](https://regexper.com/#%5E(%5Ba-z0-9%5C*-%5D%2B%5C.)*%5Ba-z0-9%5C*-%5D%2B%24) for details.

PR #2369, opened 2016-08-12
Fix ssh-keys:add permission error

### LINKED ISSUE #2366: Passing path to ssh-keys:add is broken
Description of problem:
Passing a path, ie `dokku ssh-keys:add <user> path/to/.ssh/key.pub` is failing if the dokku user can't read the public key. Piping isn't a problem and still works as expected. I belive CI tests are passing because they are run as the dokku user.

- `dokku version`:
0.7.0

Environment details (AWS, VirtualBox, physical, etc.):
GCE, Ubuntu 14.04 LTS

How was dokku installed?:
bootstrap script from 0.7.0 docs

How reproducible:
https://github.com/u2mejc/terraform-gce-example

Steps to Reproduce:
1. `sudo dokku ssh-keys:add justin .ssh/id_rsa.pub`

Actual Results:
`cat: .ssh/id_rsa.pub: Permission denied`

Expected Results:
Print hash of key (see expected result from piping in key)

Additional info:

u2mejc:
Resolves #2366

### DIFF
--- a/dokku
+++ b/dokku
@@ -56,14 +56,14 @@ if [[ "${args[0]}" =~ ^--.* ]]; then
 fi
 ! has_tty && DOKKU_QUIET_OUTPUT=1
 
-if [[ $(id -un) != "dokku" ]] && [[ ! $1 =~ plugin:* ]]; then
+if [[ $(id -un) != "dokku" ]] && [[ ! $1 =~ plugin:* ]] && [[ ! $1 == "ssh-keys:add" ]]; then
   export SSH_USER=$(id -un)
   sudo -u dokku -E -H "$0" "$@"
   exit $?
 fi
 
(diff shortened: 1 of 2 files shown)


josegonzalez:
Why not just verify that the file is readable and raise an error if it isn't?

josegonzalez:
As well, since this check isn't plugin-wide, it should almost assuredly be done at the plugin command level.

u2mejc:
The issue is dokku rerun's it's self as user "dokku" (except for plugin:* before this patch). 

Because the default parent dir (by default, .ssh) has 700 perms and no executable bit for everyone else, it's not possible to read files inside, even though they may have 744 permissions inside. I missed this in development I think because I was making new keys in /tmp. 

The other change in the user-auth file was simply to allow non sudo users see :help and :list.

Issue #2368, opened 2016-08-12
Running "dokku:ssh-keys add" via dokku_client.sh puts wrong key name


IlyaSemenov:
Description of problem:

Running `dokku:ssh-keys add` via dokku_client.sh silently puts the app name instead of the key name:

``` console
$ cat gitlab.pub | dokku ssh-keys:add gitlab
SHA256:CuDHNdVnVaN35nwLdaoAfEalepXJ9ispgrLEUeS24iw

root@dokku:~# tail -n1 /home/dokku/.ssh/authorized_keys
command="FINGERPRINT=SHA256:CuDHNdVnVaN35nwLdaoAfEalepXJ9ispgrLEUeS24iw NAME=\"myapp\" ... gitlab
```

Note that `NAME` is set to "myapp" instead of "gitlab".

Running the same command via SSH works fine:

``` console
$ cat gitlab.pub | ssh dokku@dokku.myapp.com ssh-keys:add gitlab
SHA256:CuDHNdVnVaN35nwLdaoAfEalepXJ9ispgrLEUeS24iw

root@dokku:~# tail -n1 /home/dokku/.ssh/authorized_keys
command="FINGERPRINT=SHA256:CuDHNdVnVaN35nwLdaoAfEalepXJ9ispgrLEUeS24iw NAME=\"gitlab\" ... gitlab
```

Output of the following commands:
- `uname -a`: Linux dokku 4.4.0-34-generic #53-Ubuntu SMP Wed Jul 27 16:06:39 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
- `sigil -v`: 0.4.0
- `dokku version`: 0.7.0
- `dokku plugin | grep ssh`: ssh-keys             0.6.4 enabled
- `dpkg -l | grep sshcommand`: ii  sshcommand                         0.5.0                           amd64

u2mejc:
This appears to be a larger regression from here: https://github.com/dokku/dokku/issues/1638

I'm pushing a patch that should fix trace, domains, etc... I don't think we have testing for this since it's inside /contrib/.

Unfortunately, I also pushed another patch earlier which clarifies that keys should be added by root or *admin* (as define by dokku ssh-keys / sshcommand). This means that dokku_client.sh isn't supported for adding keys (simialrly like plugins:* isn't) since the ssh user is always "dokku". 

[...]

PR #2370, opened 2016-08-13
Add exceptions back to dokku_client.sh

u2mejc:
Add exceptions back to dokku_client.sh for plugins that should not be passed an app name implicitly. This addresses a regression from commit https://github.com/dokku/dokku/commit/4204d210c1e78bae2280e1b76858cdbccba64df1

Discovered here (note that issue is marked wontfix): #2368

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -93,6 +93,7 @@ if [[ ! -z $DOKKU_HOST ]]; then
       ;;
     esac
 
+    [[ $1 =~ domains|help|plugins*|ps:restartall|trace|version|ssh-keys* ]] && unset appname
     [[ -n "$@" ]] && [[ -n "$appname" ]] && app_arg="--app $appname"
     # echo "ssh -o LogLevel=QUIET -p $DOKKU_PORT -t dokku@$DOKKU_HOST -- $app_arg $@"
     # shellcheck disable=SC2068,SC2086


u2mejc:
Do not merge yet, need to fix tests for non implicit app name support.

josegonzalez:
@u2mejc *bumpies* :)

u2mejc:
Sorry! I was out of town for a bit, I'll work to resolve this evening.

josegonzalez:
Note: these are the core commands that do not take an app as the first argument:

```
apps
certs
certs:chain
domains:add-global
domains:remove-global
events*
help
ls
nginx
plugin*
ps:rebuildall
ps:restartall
ps:restore
shell
ssh-keys*
storage
trace
version
```

And these are the commands where it is ambiguous.

```
config (<app>|--global)
config:get (<app>|--global) KEY
config:set (<app>|--global) KEY1=VALUE1 [KEY2=VALUE2 ...]
config:unset (<app>|--global) KEY1 [KEY2 ...]
```

It might make sense to just pass `--app APP_NAME` as the arg always, and let the `dokku` binary decide on when to apply the flag or not. This would also allow developers of other clients to take advantage of the `--app` flag support.

josegonzalez:
Supersceded by #2557, which contains this change (with a few minor tweaks).

PR #2378, opened 2016-08-19
Skip container finish processing when zero downtime is disabled

knjcode:
- Skip unnecessary container finish processing when zero downtime is disabled

Even when zero downtime is disabled, unnecessary container finish process is working.
This pull request skips this unnecessary process.

example working output:

before

```
$ dokku deploy ruby-rails-sample
-----> Attempting to run scripts.dokku.predeploy from app.json (if defined)
-----> App Procfile file found (/home/dokku/ruby-rails-sample/DOKKU_PROCFILE)
-----> DOKKU_SCALE file found (/home/dokku/ruby-rails-sample/DOKKU_SCALE)
=====> web=1
-----> zero downtime is disabled for app (ruby-rails-sample.web). stopping currently running containers
=====> stopping ruby-rails-sample.web (c38ab49331ee6460b1cf16ef266cd9215c2409ea4281821d254df34b7e8f1335)

.....

-----> Setting config vars
       DOKKU_APP_RESTORE: 1
-----> Shutting down old containers in 60 seconds
=====> c38ab49331ee6460b1cf16ef266cd9215c2409ea4281821d254df34b7e8f1335
```

after (this pull request)

```
$ dokku deploy ruby-rails-sample
-----> Attempting to run scripts.dokku.predeploy from app.json (if defined)
-----> App Procfile file found (/home/dokku/ruby-rails-sample/DOKKU_PROCFILE)
-----> DOKKU_SCALE file found (/home/dokku/ruby-rails-sample/DOKKU_SCALE)
=====> web=1
-----> zero downtime is disabled for app (ruby-rails-sample.web). stopping currently running containers
=====> stopping ruby-rails-sample.web (3fc7e1fa23cf1b3bcdde026f03b9a75b27bb72446b5b201a9118dd5d9c14bb7b)

.....

-----> Setting config vars
       DOKKU_APP_RESTORE: 1
```

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -549,6 +549,8 @@ dokku_deploy_cmd() {
         dokku_log_info2 "stopping $APP.$PROC_TYPE ($cid)"
         # shellcheck disable=SC2086
         docker stop $DOCKER_STOP_TIME_ARG "$cid" &> /dev/null
+        # remove cid from oldids to skip the old container finish processing
+        oldids="$(remove_val_from_list "$cid" "$oldids" " ")"
       done
     fi
 


josegonzalez:
Doesn't this completely skip killing old containers that are no longer included in the Procfile? As well, `$PROC_TYPE` as you use it only refers to the last process type, so this has the potential to not kill containers if only the last process type has zero-downtime disabled.

knjcode:
Thanks for your commenting.

I am not sure if I understand what you meant.
Did you mean this pull request is unnecessary or it needs to be modified?

If the current process type has zero-downtime disabled, dokku trying to stop currently running conainers in dokku_deploy_cmd. (https://github.com/dokku/dokku/blob/master/plugins/common/functions#L545-L553)
So, I guess container finish processing part (https://github.com/dokku/dokku/blob/master/plugins/common/functions#L660-L680) always trying to stop already stopped containers and display unnecessary messages. (`Shutting down old containers in 60 seconds` and already stopped container ID)

This is why I want to skip container finish processing in this pull request.

I’m sorry that I couldn’t explain and communicate with you well due to the lack of my English skill.

josegonzalez:
Lets say you have the following application:

```
web: bin/server
worker: bin/sidekiq
```

If you disable checks for the worker:

```
dokku checks:disable APP worker
```

And we merge your pull request, the last value of `PROC_TYPE` will be `worker`, so the if statement will be true, and we won't ever stop the old web processes (which still have checks enabled).

If anything, you want to *not* add the container id to `oldids` when zero-downtime was disabled for that process type. If you want to change the logic to do that, I would accept the pull request.

knjcode:
Thank you for your explanation.
I was able to understand the problem.

I change the logic.
Remove zero-downtime disabled container ids from oldids to skip the old container finish processing.

plugins/common/functions
```diff
@@ -549,6 +549,8 @@ dokku_deploy_cmd() {
         dokku_log_info2 "stopping $APP.$PROC_TYPE ($cid)"
         # shellcheck disable=SC2086
         docker stop $DOCKER_STOP_TIME_ARG "$cid" &> /dev/null
+        # remove cid from oldids to skip the old container finish processing
+        oldids="$(remove_val_from_list "$cid" "$oldids")"
```
michaelshobbs: The default delimiter in `remove_val_from_list` is a comma. However`oldids` is space-separated. I believe you will need to change this call to this
```
oldids="$(remove_val_from_list "$cid" "$oldids" " ")"
```

knjcode: Thanks. I fix it.

josegonzalez:
Thanks for the pull request!

PR #2385, opened 2016-08-26
actually merge dokku-app-user into core

michaelshobbs:
my initial pass at this was, well, lazy to say the least. this allows a user to set `DOKKU_APP_USER` as a config variable to override the default `herokuishuser`. the `pre-deploy` trigger will inspect the app's storage mounts and ensure the container user has ownership.


josegonzalez:
Needs docs.

michaelshobbs:
added docs and fixed and ensure we default to `herokuishuser` in the `pre-deploy` trigger

PR #2392, opened 2016-08-27
Use python2 in functions/get_json_*

swg:
On hosts where python3 is the system default python env, Dokku will silently fail to return the values of json keys.


josegonzalez:
What is the proper script for python-3?

swg:
It'd be `json.loads(...)` and `print(json.dumps(...)` but invoking `python2` works on all distros you list as currently supported, no matter how they're configured.

PR #2406, opened 2016-09-04
Use apps_create method when renaming an application. Closes #2252

### LINKED ISSUE #2252: dokku apps:rename old_app new_app shouldn't trigger post-delete

josegonzalez:
Since they are separate resources, triggering the normal app creation flow makes the most sense.

We still remove the cache directory because we want the correct cache permissions for the new application.


michaelshobbs:
@pltchuong can you confirm this works for your issue?

PR #2407, opened 2016-09-05
Move core post-deploy triggers to core-post-deploy

josegonzalez:
This will help ensure that user-created post-deploy triggers do not accidentally orphan containers.

Deployment tasks were moved to post-deploy as they are user-defined, and can fail for a variety of non-dokku reasons.


josegonzalez:
@michaelshobbs Should we also move the actual killing of old docker containers to `core-post-deploy` as well? Or move the user `post-deploy` trigger to after the kill call at least? We can *still* orphan containers, even with the above change, as the `docker kill` bit happens in `common/functions:dokku_deploy_cmd`.

This change simply ensures that stuff from dokku's core-post-deploy happens before community plugin (and deployment task) post-deploy.

michaelshobbs:
Funny I was just thinking about what this was actually doing. Let me noodle on this before we merge.

docs/advanced-usage/deployment-tasks.md
```diff
@@ -15,7 +15,8 @@ Dokku accomplishes this by using an `app.json` file. The format in use is simila
 However, Dokku currently only supports the nodes `scripts.dokku.predeploy` and `scripts.dokku.postdeploy`.
 For buildpack apps, simply place an `app.json` file in the root of your repository.
 For dockerfile apps, place `app.json` in the configured `WORKDIR` directory; otherwise Dokku defaults to the buildpack app behavior of looking in `/app`.
->NOTE: postdeploy changes are *NOT* committed to the app image.
```
jasonahjie: We should commit this line to the code again unless similar comment been made somewhere else.

PR #2408, opened 2016-09-05
Use $DOKKU_VHOST_ENABLE instead of $VHOST_ENABLE

jcscottiii:
Given a user wants to manually override the `vhost_enable` debian config, a user has to specify both `$DOKKU_VHOST_ENABLE` and `$VHOST_ENABLE`. This patch makes it rely on `$DOKKU_VHOST_ENABLE` like the others which use the `$DOKKU_` prefix


josegonzalez:
Good catch!

PR #2410, opened 2016-09-07
Clarify domains help, improve domains unit tests

IlyaSemenov:
1) This clarifies that `domains:add` and `domains:remove` support multiple domains, and removes `<app>` vs. ` DOMAIN` inconsistency.

2) This fixes `tests/unit/20_domains.bats` unit tests which didn't actually test anything other than the commands didn't crash. Even `refute_line` tests in `domains:remove` never worked as `domains:remove` didn't print the new domains set.

### DIFF
--- a/plugins/domains/commands
+++ b/plugins/domains/commands
@@ -8,13 +8,13 @@ case "$1" in
       declare desc="return domains plugin help content"
       cat<<help_content
     domains [<app>], List domains
-    domains:add <app> DOMAIN, Add a domain to app
-    domains:add-global DOMAIN, Add global domain name
+    domains:add <app> <domain> [<domain>...], Add domain(s) to app
+    domains:add-global <domain> [<domain>...], Add global domain name(s)
     domains:clear <app>, Clear all domains for app
     domains:disable <app>, Disable VHOST support
     domains:enable <app>, Enable VHOST support
-    domains:remove <app> DOMAIN, Remove a domain from app
-    domains:remove-global DOMAIN, Remove global domain name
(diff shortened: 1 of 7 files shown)


plugins/domains/commands
```diff
@@ -8,13 +8,13 @@ case "$1" in
...
-    domains:add-global DOMAIN, Add global domain name
+    domains:add <app> <domain> [<domain>...], Add domain(s) to app
+    domains:add-global <domain> [<domain>...], Add global domain name(s)
     domains:clear <app>, Clear all domains for app
     domains:disable <app>, Disable VHOST support
     domains:enable <app>, Enable VHOST support
-    domains:remove <app> DOMAIN, Remove a domain from app
-    domains:remove-global DOMAIN, Remove global domain name
+    domains:remove <app> <domain> [<domain>...], Remove domain(s) from app
+    domains:remove-global <domain> [<domain>...], Remove global domain name(s)
```
josegonzalez: Can you update the command descriptions for each command script, as well as the help output in the domains documentation?

IlyaSemenov:
My bad, updated everything related.

On a related note, I actually wanted to create `domains:set` command, as I personally tired of running `domains:clear` followed by `domains:add` when I just want to set or change the app domain name(s). I noticed that tests for the current `domains:*` commands were not really working and thus this PR went first. Question is, is `domains:set` something that has a chance to be accepted, or has it been specifically left out?

josegonzalez:
We had a `domains:set` before, but removed it in favor of `domains:add`. If you want to re-add it, I am not opposed, though you'll need to do it for the global domain as well.

PR #2416, opened 2016-09-14
Adds a mention of manually adding nginx entry

njaxx:
I was working on a Dockerfile deploy that was setup to listen on port 8844.  I wanted to add ssl and couldn't figure out why adding the ssl cert didn't making things 'just work'.  Since the `proxy:ports-add` step I discuss in this commit is run both `under the hood` and `sometimes`, I thought it might be worth issuing a PR for.


josegonzalez:
> Since the proxy:ports-add step I discuss in this commit is run both under the hood and sometimes, I thought it might be worth issuing a PR for.

What do you mean by "under the hood" and "sometimes"?

njaxx:
Unless I'm mistaken, an equivalent operation to `proxy:ports-add` takes place automatically when a user does `dokku certs:add`.  I say that it takes place sometimes because it's a requirement that the app listen on port 80.  When the app listens on another port, the nginx.conf file will not be updated under the hood and the ssl port must be manually exposed.  I'm somewhat out of my element, so if that doesn't seem right than please close this PR.

josegonzalez:
No thats fine, I just wanted to clarify your phrasing there.

PR #2419, opened 2016-09-19
Support for CentOS 7

### LINKED ISSUE #478: Deploy dokku to CentOS
Hi,

Can anyone advise how to deploy dokku to CentOS OS ?

Thanks,

ebeigarts:
This pull request adds support for CentOS 7 and also builds RPM packages. Fixes #478.

Currently `rpm.mk` looks very similar to `deb.mk`, maybe this could be refactored in the future to reuse code more effectively.

Installation and testing using DigitalOcean with Centos 7.2:

```
package_cloud push ebeigarts/dokku/el/7 *.rpm

wget https://raw.githubusercontent.com/ebeigarts/dokku/centos/bootstrap.sh
sed -i "s/dokku\/dokku/ebeigarts\/dokku/" bootstrap.sh
DOKKU_TAG=v0.7.2 bash bootstrap.sh

ssh-keygen -f ~/.ssh/id_rsa -t rsa -N ''

dokku apps:create node-js-sample
dokku ssh-keys:add node-js-sample ~/.ssh/id_rsa.pub
dokku domains:add node-js-sample localhost

git clone https://github.com/heroku/node-js-sample
cd node-js-sample/
git remote add dokku dokku@localhost:node-js-sample
git push dokku master

curl http://localhost/ # => Hello World!
```

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -14,10 +14,6 @@ set -eo pipefail; [[ $TRACE ]] && set -x
 ensure-environment() {
   local FREE_MEMORY
   echo "Preparing to install $DOKKU_TAG from $DOKKU_REPO..."
-  if ! command -v apt-get &>/dev/null; then
-    echo "This installation script requires apt-get. For manual installation instructions, consult http://dokku.viewdocs.io/dokku/advanced-installation/"
-    exit 1
-  fi
 
   hostname -f > /dev/null 2>&1 || {
     echo "This installation script requires that you have a hostname set for the instance. Please set a hostname for 127.0.0.1 in your /etc/hosts"
@@ -33,10 +29,15 @@ ensure-environment() {
 
 install-requirements() {
   echo "--> Ensuring we have the proper dependencies"
-  apt-get update -qq > /dev/null
-  if [[ $(lsb_release -sr) == "12.04" ]]; then
-    apt-get -qq -y install python-software-properties
-  fi
+
+  case "$DOKKU_DISTRO" in
+    debian|ubuntu)
+      apt-get update -qq > /dev/null
+      if [[ "$DOKKU_DISTRO_VERSION" == "12.04" ]]; then
+        apt-get -qq -y install python-software-properties
+      fi
+      ;;
+  esac
 }
 
 install-dokku() {
@@ -68,9 +69,14 @@ install-dokku() {
   fi
 }
 
-
 install-dokku-from-source() {
   local DOKKU_CHECKOUT="$1"
+
+  if ! command -v apt-get &>/dev/null; then
+    echo "This installation script requires apt-get. For manual installation instructions, consult http://dokku.viewdocs.io/dokku/advanced-installation/"
+    exit 1
+  fi
+
   apt-get -qq -y install git make software-properties-common
   cd /root
   if [[ ! -d /root/dokku ]]; then
@@ -84,6 +90,21 @@ install-dokku-from-source() {
 }
 
 install-dokku-from-package() {
+  case "$DOKKU_DISTRO" in
+    debian|ubuntu)
+      install-dokku-from-deb-package "$@"
+      ;;
+    centos)
+      install-dokku-from-rpm-package "$@"
+      ;;
+    *)
+      echo "Unsupported Linux distribution. For manual installation instructions, consult http://dokku.viewdocs.io/dokku/advanced-installation/"
+      exit 1
+      ;;
+  esac
+}
+
+install-dokku-from-deb-package() {
   local DOKKU_CHECKOUT="$1"
   local NO_INSTALL_RECOMMENDS=${DOKKU_NO_INSTALL_RECOMMENDS:=""}
 
@@ -124,7 +145,39 @@ install-dokku-from-package() {
   fi
 }
 
+install-dokku-from-rpm-package() {
+  local DOKKU_CHECKOUT="$1"
+
+  if [[ "$DOKKU_DISTRO_VERSION" != "7" ]]; then
+    echo "Only CentOS version 7 is supported."
+    exit 1
+  fi
+
+  echo "--> Installing docker"
+  curl -fsSL https://get.docker.com/ | sh
+
+  echo "--> Installing epel for nginx packages to be available"
+  yum install -y epel-release
+
+  echo "--> Installing herokuish and dokku"
+  curl -s https://packagecloud.io/install/repositories/dokku/dokku/script.rpm.sh | bash
+  if [[ -n $DOKKU_CHECKOUT ]]; then
+    yum -y install herokuish "dokku-$DOKKU_CHECKOUT"
+  else
+    yum -y install herokuish dokku
+  fi
+
+  echo "--> Starting nginx"
+  systemctl start nginx
+}
+
 main() {
+  export DOKKU_DISTRO DOKKU_DISTRO_VERSION
+  # shellcheck disable=SC1091
+  DOKKU_DISTRO=$(. /etc/os-release && echo "$ID")
+  # shellcheck disable=SC1091
+  DOKKU_DISTRO_VERSION=$(. /etc/os-release && echo "$VERSION_ID")
+
   export DEBIAN_FRONTEND=noninteractive
   export DOKKU_REPO=${DOKKU_REPO:-"https://github.com/dokku/dokku.git"}
 
--- a/docs/getting-started/install/rpm.md
+++ b/docs/getting-started/install/rpm.md
@@ -1,5 +1,7 @@
 # RPM Package Installation Notes
 
+> New as of 0.8.0
+
 Dokku defaults to being installed via RPM package on CentOS 7. While certain hosts may require extra work to get running, you may optionally wish to automate the installation of Dokku without the use of our `bootstrap.sh` bash script. The following are the steps run by said script:
 
 ```shell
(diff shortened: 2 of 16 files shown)


josegonzalez:
I'll discuss this changeset with @michaelshobbs this week, but good work so far! The only thing I can think of is that the `bootstrap.sh` should be modified to handle Centos7 (and no other versions of course).

josegonzalez:
Without reviewing the code, a few things:

- None of the core maintainers use CentOS - nor do we have any desire to.
- We'll bring you - @ebeigarts - up in any issues referencing CentOS support.
- These packages will not be pushed to any official CentOS/RHEL repository, and any such packages will be considered unofficial and unsupported. All packages will be available in our packagecloud repository, as we do for debian packages.
- This integration would be considered experimental and not supported by the core team.
- Should this integration become a hassle to upkeep, we'll drop it from the core.
- Needs a few more touch-ups to the release process documentation.

@ebeigarts Assuming this integration goes well, we'll have you as point-person for the package, and likely eventually commit bit.

plugins/nginx-vhosts/install
```diff
@@ -17,6 +17,11 @@ case "$DOKKU_DISTRO" in
   arch)
     echo "%dokku ALL=(ALL) NOPASSWD:/usr/bin/systemctl reload nginx, /usr/sbin/nginx -t" > /etc/sudoers.d/dokku-nginx
     ;;
+
+  centos)
+    echo "%dokku ALL=(ALL) NOPASSWD:/usr/bin/systemctl reload nginx, /usr/sbin/nginx -t" > /etc/sudoers.d/dokku-nginx
+    echo "Defaults:dokku !requiretty" >> /etc/sudoers.d/dokku-nginx
```
josegonzalez: Why this?

rpm.mk
```diff
@@ -0,0 +1,171 @@
...
+	git clone -q "https://github.com/$(HEROKUISH_REPO_NAME).git" \
+		--branch "v$(HEROKUISH_VERSION)" /tmp/tmp/herokuish > /dev/null
+	rm -rf /tmp/tmp/herokuish/.git /tmp/tmp/herokuish/.gitignore
+
+	@echo "-> Copying files into place"
+	mkdir -p "/tmp/build/var/lib"
+	cp -rf /tmp/tmp/herokuish /tmp/build/var/lib/herokuish
+
+	@echo "-> Creating $(HEROKUISH_RPM_PACKAGE_NAME)"
+	sudo fpm -t rpm -s dir -C /tmp/build -n herokuish \
```
josegonzalez: Can fpm for rpms be run under ubuntu?

rpm.mk
```diff
@@ -0,0 +1,171 @@
...
+		-v $(HEROKUISH_VERSION) \
+		-a $(RPM_ARCHITECTURE) \
+		-p $(HEROKUISH_RPM_PACKAGE_NAME) \
+		--depends 'docker-engine >= 1.9.1' \
+		--depends 'sudo' \
```
josegonzalez: This is a debianism, but worth keeping in case CentOs screws the pooch or there is a user out there who thinks they are slick.

rpm.mk
```diff
@@ -0,0 +1,171 @@
...
+		--depends 'git' \
+		--depends 'make' \
+		--depends 'curl' \
+		--depends 'gcc' \
+		--depends 'man-db' \
+		--depends 'sshcommand' \
+		--depends 'gliderlabs-sigil' \
+		--depends 'docker-engine >= 1.9.1' \
+		--depends 'bind-utils' \
+		--depends 'nginx' \
```
josegonzalez: What version of nginx comes with CentOs 7?

rpm/dokku.postinst
```diff
@@ -0,0 +1,60 @@
...
+    /usr/local/bin/sshcommand "$@"
+  elif [[ -x /usr/bin/sshcommand ]]; then
+    /usr/bin/sshcommand "$@"
+  else
+    echo "Unable to find sshcommand binary" 1>&2
+    exit 1
+  fi
+}
+
+echo "Setting up dokku user"
```
josegonzalez: Any reason why this section isn't all in a function at least? The debian one isn't because of the diff, though this all should be.

rpm/herokuish.postinst
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set +o posix # rpm uses /bin/sh that forces bash in posix mode
+
+echo 'Starting docker'
+systemctl start docker
+
+sleep 5
+
+count=`sudo docker images | grep gliderlabs/herokuish | wc -l`
```
josegonzalez: Can we use `$()` instead of backticks everywhere?

rpm/dokku.postinst
```diff
@@ -0,0 +1,60 @@
...
+if [[ -f .ssh/authorized_keys ]]; then
+  sed -i.bak 's#/usr/local/bin/dokku#/usr/bin/dokku#' "${DOKKU_ROOT}/.ssh/authorized_keys"
+  rm "${DOKKU_ROOT}/.ssh/authorized_keys"
+fi
+
+echo "Install all core plugins"
+dokku plugin:install --core
+
+rm -f ${DOKKU_ROOT}/VERSION
+cp ${DOKKU_LIB_ROOT}/STABLE_VERSION ${DOKKU_ROOT}/VERSION
```
josegonzalez: Please quote all variables and make these files pass shellcheck.

michaelshobbs:
This should probably be in the contributing doc if it isn't already but we try to follow this style guide as best we can. 

https://github.com/progrium/bashstyle

plugins/nginx-vhosts/install
ebeigarts: CentOS has `requiretty` by default in `/etc/sudoers` and requires that `sudo` [command has tty](https://www.shell-tips.com/2014/09/08/sudo-sorry-you-must-have-a-tty-to-run-sudo/). I was getting `sudo: sorry, you must have a tty to run sudo` error when using `git push`, dokku couldn't restart the nginx with sudo.

rpm.mk
ebeigarts: Yes, it just requires `apt-get install rpm` which I added in `rpm-setup`.

rpm.mk
ebeigarts: none. [EPEL](https://fedoraproject.org/wiki/EPEL) repository includes nginx 1.10.1. That is why I had to `yum install -y epel-release`

rpm/dokku.postinst
ebeigarts: This is mostly copy/paste from debian/postinst.

rpm/dokku.postinst
ebeigarts: Do we need to quote this also in `debian/postinst`?

rpm/herokuish.postinst
ebeigarts: This was taken from https://github.com/dokku/dokku/blob/master/deb.mk#L87-L95. Do we need to change that also in deb packages?

ebeigarts:
Everythin else sounds good, I will change `bootstrap.sh` to support centos7

rpm/dokku.postinst
josegonzalez: I'm pretty sure it is, if not, we should, but separate pr please.

rpm/herokuish.postinst
josegonzalez: This file isnt a Makefile.

ebeigarts:
I have added support for CentOS in `bootstrap.sh` and also improved code using shellcheck suggestions. I also switched to using `/etc/os-release` that is available on Ubuntu/Debian/Arch/CentOS and others by default (`yum install redhat-lsb-core` is not required this way)

bootstrap.sh
```diff
@@ -33,10 +29,15 @@ ensure-environment() {
...
-  apt-get update -qq > /dev/null
-  if [[ $(lsb_release -sr) == "12.04" ]]; then
-    apt-get -qq -y install python-software-properties
-  fi
+
+  case "$DOKKU_DISTRO" in
+    debian|ubuntu)
+      apt-get update -qq > /dev/null
+      if [[ "$DOKKU_DISTRO_VERSION" == "12.04" ]]; then
+        apt-get -qq -y install python-software-properties
```
ebeigarts: Is this still needed? I saw that 12.04 is not supported anymore.

josegonzalez: Yes, I would leave this here for now.

ebeigarts:
@josegonzalez what do you mean by rpm docs page? Something like http://dokku.viewdocs.io/dokku/getting-started/install/debian/ or release instructions? Release instructions are already documented - rpms are built using `vagrant up build`.

josegonzalez:
Yes, like that page.

ebeigarts:
@josegonzalez I have added [rpm docs](https://github.com/ebeigarts/dokku/blob/centos/docs/getting-started/install/rpm.md). anything else?

docs/getting-started/install/rpm.md
```diff
@@ -0,0 +1,16 @@
+# RPM Package Installation Notes
+
```
josegonzalez: Can you add a blockquote like so:

```
> New as of 0.8
```

ebeigarts: done

ebeigarts:
@josegonzalez is there anything holding this back from merging?

josegonzalez:
@ebeigarts it'll go in the next minor.

PR #2430, opened 2016-09-24
Fixed link to "deployment tasks" in HISTORY.md

matthisk:


josegonzalez:
Does the old link not redirect?

matthisk:
Yes it does, that is weird I was sure the other day it didn't redirect me and I got a 404.

Issue #2424, opened 2016-09-21
Document the `logs` plugin


josegonzalez:
This is currently undocumented.

Needs:

- Link on the sidebar after [this](https://github.com/dokku/dokku/blob/master/docs/template.html#L200) line
- A section at the top with the help output for the log output like [this](https://github.com/dokku/dokku/blob/master/docs/deployment/application-management.md)
- A section on behavioral modifiers [like so](https://github.com/dokku/dokku/blob/master/docs/deployment/remote-commands.md#behavioral-modifiers)
- It needs to be a new document called `docs/deployment/logs.md`

cynicaldevil:
I'll take a stab at this.

josegonzalez:
Yep closing!

PR #2438, opened 2016-09-28
Update event-logs.md

vishnubhagwan:


vishnubhagwan:
#2424

josegonzalez:
This doesn't actually document the logs plugin, so I cannot accept it as is. I was thinking it would actually be a new logs.md file in [this](https://github.com/dokku/dokku/tree/master/docs/deployment) section.

PR #2451, opened 2016-10-07
Document the logs plugin

mbtamuli:
Solves issue #2424.

> - [x] Link on the sidebar after [this](https://github.com/dokku/dokku/blob/master/docs/template.html#L200) line
> - [x] A section at the top with the help output for the log output like [this](https://github.com/dokku/dokku/blob/master/docs/deployment/application-management.md)
> - [x] A section on behavioral modifiers [like so](https://github.com/dokku/dokku/blob/master/docs/deployment/remote-commands.md#behavioral-modifiers)
> - [x] It needs to be a new document called `docs/deployment/logs.md`

Suggestions are welcome if further changes are required.

### DIFF
--- a/docs/deployment/logs.md
+++ b/docs/deployment/logs.md
@@ -1,0 +1,38 @@
+# Log Management
+
+```
+logs <app>           # display recent log output
+```
+
+## Usage
+
+### Application Logs
+You can easily get logs of an application using the `logs` command:
+
+```shell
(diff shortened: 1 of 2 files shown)


josegonzalez:
❤️

mbtamuli:
Wow. That was fast! :+1:  Thanks.

PR #2452, opened 2016-10-08
Solves SSH Key problem when admin user already exists

### LINKED ISSUE #2445: Web SSH Key not added when key is already attached to server
As both the debian package and the web ui use the `admin` user, the web ui will fail to add the ssh key, resulting in some users not having access to dokku push.

We should instead use `admin-web` as the admin user from the web ui if we detect that there is an existing `admin` user.

[Here](https://github.com/dokku/dokku/blob/master/contrib/dokku-installer.py#L103) is the line of code that executes this. What I'd do is list users, check if `admin` is in the list, and if so, change the username there from `admin` to `admin-web`.

mbtamuli:
This PR resolves #2445 

Suggestions are welcome if further changes are required.

### DIFF
--- a/contrib/dokku-installer.py
+++ b/contrib/dokku-installer.py
@@ -101,11 +101,20 @@ class GetHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
         with open('{0}/HOSTNAME'.format(dokku_root), 'w') as f:
             f.write(params['hostname'].value)
 
-        user = 'admin'
-        if self.admin_user_exists():
-            user = 'web-admin'
-        command = ['sshcommand', 'acl-add', 'dokku', user]
-        for key in params['keys'].value.split("\n"):
+        for (index, key) in enumerate(params['keys'].value.splitlines(), 1):
+            user = 'admin'
+            if self.admin_user_exists() is not None:
+                user = 'web-admin'
+                if self.web_admin_user_exists() is not None:
+                    index = int(self.web_admin_user_exists()) + 1
+                elif self.web_admin_user_exists() is None:
+                    index = 1
+            elif self.admin_user_exists() is None:
+                pass
+            else:
+                index = int(self.admin_user_exists()) + 1
+            user = user + str(index)
+            command = ['sshcommand', 'acl-add', 'dokku', user]
             proc = subprocess.Popen(command, stdin=subprocess.PIPE)
             proc.stdin.write(key)
             proc.stdin.close()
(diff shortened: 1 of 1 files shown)


contrib/dokku-installer.py
```diff
@@ -119,6 +123,16 @@ def do_POST(self):
         self.end_headers()
         self.wfile.write(json.dumps({'status': 'ok'}))
 
+    def admin_user_exists(self):
+        command = 'dokku ssh-keys:list'
+        pattern = re.compile(r'NAME="admin"')
+        proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
+        for line in proc.stdout:
+            if pattern.search(line):
```
josegonzalez: You can just do `return pattern.search(line)`

contrib/dokku-installer.py
```diff
@@ -100,7 +101,10 @@ def do_POST(self):
...
             f.write(params['hostname'].value)
 
-        command = ['sshcommand', 'acl-add', 'dokku', 'admin']
+        user = 'admin'
+        if self.admin_user_exists():
+            user = 'web-admin'
+        command = ['sshcommand', 'acl-add', 'dokku', user]
         for key in params['keys'].value.split("\n"):
             proc = subprocess.Popen(command, stdin=subprocess.PIPE)
             proc.stdin.write(key)
```
josegonzalez: Each key will need a new username...

josegonzalez: Right but the web ui allows splitting the keys via newline, and as new sshcommand doesn't allow duplicate usernames, we'll need to modify the user for each successive write.

mbtamuli: Okay, so the usernames should be asked though the form(web UI) or the usernames should be generated? Like `web-admin1`, `web-admin2`, etc?

josegonzalez: Yep, that works for me :)

contrib/dokku-installer.py
```diff
@@ -123,15 +132,28 @@ def do_POST(self):
...
-        proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
-        for line in proc.stdout:
-            if pattern.search(line):
-                return True
-            else:
-                return False
+    return self.user_exists('admin(\d+)')
+
+    def user_exists(self, name):
+    command = 'dokku ssh-keys:list'
```
josegonzalez: This isn't properly indented.

contrib/dokku-installer.py
```diff
@@ -123,15 +132,28 @@ def do_POST(self):
         self.end_headers()
         self.wfile.write(json.dumps({'status': 'ok'}))
 
+    def web_admin_user_exists(self):
+    return self.user_exists('web-admin(\d+)')
```
josegonzalez: Not properly indented.

PR #2461, opened 2016-10-13
Update functions

prakhar987:


jvanbaarsen:
What is the reason for this change?

jvanbaarsen:
@josegonzalez I think we can close this.

PR #2462, opened 2016-10-13
Update dokku-installer.py

prakhar987:


josegonzalez:
This doesn't actually fix anything, and someone else is working on it in #2452.

Issue #2446, opened 2016-10-03
Adding a file with two ssh-keys can result in an error


josegonzalez:
We should validate that the input file only has one string. As well, the output of the `ssh-keys:add` should include a 0 exit code _only_ if the `authorized_keys` file is still valid.

arr-dev:
I could try fixing this for hacktoberfest

josegonzalez:
:)

polettix:
@arr-dev did you start working on this? I'd like to give it a try too but I don't want to "steal" anything.

arr-dev:
Hi @polettix I've started on it but have been busy and haven't committed anything.
I hope to have time over the weekend to finish it.

polettix:
NP happy coding!

2016-10-14 10:45 GMT+02:00 Nenad <notifications@github.com>:

arr-dev:
Hey @polettix, work stuff got their better of me over the weekend too,
so hack away on this if you can.

josegonzalez:
Closing as a pr was merged!

PR #2465, opened 2016-10-16
Patch GH #2446 "Adding a file with two ssh-key..."

polettix:
This is a proposed change for addressing GH #2446 "Adding a file with two ssh-keys can result in an error".

The following checks are performed:

- validity of key contents when read from a file
- key contents only contains one single line/key
- authorized_keys file is ultimately valid

The last test is an extension over the previous implementation in that all lines in the file are checked.

The commit also includes extensions to the relevant test file to excercise the new checks. It also takes care to save and then restore the authorized_keys file.


josegonzalez:
Thanks for the pull request!

PR #2476, opened 2016-10-30
Non-zero exit code when ENV file missing/empty

### LINKED ISSUE #2447: Incorrect config:get exit code
Currently `dokku config:get <param>`:

- returns `0` exit code if there is no ENV file
- returns `1` exit code if there is an ENV file, but the variable is not set.

Can we change [config:get](https://github.com/dokku/dokku/blob/master/plugins/config/functions#L153) to return `1` if there is no ENV file or this is some kind of special case?

Example:

```
$ dokku config --global
no config vars for global

$ dokku config:get --global DOKKU_LETSENCRYPT_GRACEPERIOD
$ echo $?
0

$ dokku config:set --global DUMMY=1
-----> Setting config vars
       DUMMY: 1

$ dokku config:get --global DOKKU_LETSENCRYPT_GRACEPERIOD
$ echo $?
1
```

This introduces probem in dokku-letsencrypt and maybe other plugins also:

```
$ dokku letsencrypt:ls
-----> App name           Certificate Expiry        Time before expiry        Time before renewal
/var/lib/dokku/plugins/available/letsencrypt/functions: line 33: [: : integer expression expected
myapp                   2016-10-20 10:07:00       15d, 18h, 14m, 5s
```

amaltson:
This should fix #2447.

### DIFF
--- a/plugins/config/functions
+++ b/plugins/config/functions
@@ -150,7 +150,8 @@ config_get() {
 
   config_create "$ENV_FILE"
   if [[ ! -s $ENV_FILE ]]; then
-    return 0
+    echo "ENV file empty or new."
+    return 1
   fi
 
   local KEY="${*: -1}"


josegonzalez:
Already closed by #2454. Thanks though!

amaltson:
@josegonzalez whoops! No worries!

PR #2477, opened 2016-10-31
Fix dokku-redeploy systemd script to start only after docker

ebeigarts:
According to https://docs.docker.com/engine/admin/host_integration/#/systemd
it should use `After=docker.service` and it works at least on CentOS 7.


josegonzalez:
Seems legit.

PR #2486, opened 2016-11-15
Fixing typo & broken link in docs

OmarShehata:
The last line of the docs on this page:

http://dokku.viewdocs.io/dokku/advanced-usage/docker-options/

Have a typo "Move" instead of "More" and a broken link.


josegonzalez:
Seems legit.

PR #2489, opened 2016-11-17
add plugin uninstall trigger

joshmanders:
It'd be nice if plugins were able to clean up after themselves when uninstalled. Remove docker containers, config files, etc.

### DIFF
--- a/docs/development/plugin-triggers.md
+++ b/docs/development/plugin-triggers.md
@@ -867,6 +867,7 @@ some code to remove a docker hub tag because it's not implemented in the CLI....
 ```
 
 ### `uninstall`
+
  - Description: Used to cleanup after itself.
  - Invoked by: `dokku plugin:uninstall`
  - Arguments: `$PLUGIN`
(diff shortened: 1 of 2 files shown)


plugins/plugin/functions
```diff
@@ -70,6 +70,7 @@ uninstall_plugin() {
   local PLUGIN="$1"
   [[ -e $PLUGIN_CORE_AVAILABLE_PATH/$PLUGIN ]] && dokku_log_fail "Cannot uninstall a core plugin"
   [[ ! -e $PLUGIN_AVAILABLE_PATH/$PLUGIN ]] && dokku_log_fail "Plugin ($PLUGIN) is not currently installed"
+  plugn trigger uninstall
```
josegonzalez: This will trigger uninstall for all plugins, not just the one being installed.

docs/development/plugin-triggers.md
```diff
@@ -866,6 +866,23 @@ APP="$1"; IMAGE_TAG="$2"
 some code to remove a docker hub tag because it's not implemented in the CLI....
 ```
 
+### `uninstall`
+ - Description: Used to cleanup after itself.
```
josegonzalez: Newline before this one. Also add a warning to check the plugin name or people are going to do silly stuff and blame Dokku when their app's break.

Also, add the following to the commit to skip tests since its just a documentation change:

```
[ci skip]
```

PR #2490, opened 2016-11-18
configure dokku user with DOKKU_USER env var

joshmanders:
This allows you to set what user you would like dokku to run as.

It affects all plugins and how you interact with dokku.

If you install with `DOKKU_USER` env var set to say `git` then you will no longer add remotes as `dokku@example.com` but as `git@example.com`.

Plugins now will have to check for `$DOKKU_USER` to do such things as chown directories, or run things under the user used.

This is still a WIP and not 100% yet, but we can go over it and make changes as needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ copyfiles:
 		PLUGIN_PATH=${PLUGINS_PATH} plugn enable $$plugin ;\
 		done
 	chown ${DOKKU_USER}:${DOKKU_USER} -R ${PLUGINS_PATH} ${CORE_PLUGINS_PATH} || true
+	echo "${DOKKU_USER}" > /home/.dokku-user 2> /dev/null
 	$(MAKE) addman
 
 addman: help2man
--- a/dokku
+++ b/dokku
@@ -2,7 +2,7 @@
 set -eo pipefail
 shopt -s nullglob
 
-export DOKKU_USER=${DOKKU_USER:=$USER}
+export DOKKU_USER=${DOKKU_USER:=$(cat "/home/.dokku-user")}
 export DOKKU_ROOT=${DOKKU_ROOT:=$(getent passwd "$DOKKU_USER" | cut -d: -f6)}
 [[ -f $DOKKU_ROOT/dokkurc ]] && source "$DOKKU_ROOT/dokkurc"
 [[ -d $DOKKU_ROOT/.dokkurc ]] && for f in $DOKKU_ROOT/.dokkurc/*; do source "$f"; done


josegonzalez:
This seems reasonable. Have you tested via building a debian package and installing that way?

Needs docs. Maybe an advanced section titled `Whitelabeling`.

 @michaelshobbs thoughts?

joshmanders:
I have no tested building a debian package.

josegonzalez:
Our release process for building a deb is documented [here](http://dokku.viewdocs.io/dokku/development/release-process/), if you can take that for a spin and verify nothing breaks, awesome.

You'll probably also want to somehow make the dokku user configurable.

Makefile
```diff
@@ -63,11 +65,12 @@ copyfiles:
...
 		cp -R plugins/$$plugin ${CORE_PLUGINS_PATH}/available && \
 		ln -s ${CORE_PLUGINS_PATH}/available/$$plugin ${PLUGINS_PATH}/available; \
-		find /var/lib/dokku/ -xtype l -delete;\
+		find ${DOKKU_LIB_ROOT} -xtype l -delete;\
 		PLUGIN_PATH=${CORE_PLUGINS_PATH} plugn enable $$plugin ;\
 		PLUGIN_PATH=${PLUGINS_PATH} plugn enable $$plugin ;\
 		done
-	chown dokku:dokku -R ${PLUGINS_PATH} ${CORE_PLUGINS_PATH} || true
+	chown ${DOKKU_USER}:${DOKKU_USER} -R ${PLUGINS_PATH} ${CORE_PLUGINS_PATH} || true
+	echo "${DOKKU_USER}" > /home/.dokku-user 2> /dev/null
```
josegonzalez: We should place this in the `DOKKU_LIB_ROOT` instead.

dokku
```diff
@@ -2,7 +2,8 @@
 set -eo pipefail
 shopt -s nullglob
 
-export DOKKU_ROOT=${DOKKU_ROOT:=~dokku}
+export DOKKU_USER=${DOKKU_USER:=$(cat "/home/.dokku-user")}
```
josegonzalez: What happens if this file doesn't exist?

joshmanders: Yeah so this I wanted to talk with you guys about. I can't think of a good way to go about this in a clean manner. What do you suggest?

joshmanders: Ah yeah, good suggestion.

Makefile
```diff
@@ -5,6 +5,8 @@ PLUGN_URL ?= https://github.com/dokku/plugn/releases/download/v0.2.2/plugn_0.2.2
 SIGIL_URL ?= https://github.com/gliderlabs/sigil/releases/download/v0.4.0/sigil_0.4.0_Linux_x86_64.tgz
 STACK_URL ?= https://github.com/gliderlabs/herokuish.git
 PREBUILT_STACK_URL ?= gliderlabs/herokuish:latest
+DOKKU_USER ?= dokku
+DOKKU_ROOT ?= /home/${DOKKU_USER}
```
michaelshobbs: Edge case nitpick: What if `/home` doesn't exist? For example, what if I use `/usr/homes` for my home dir root or something like that?

Makefile
michaelshobbs: Yeah let's not pollute `/home`. Also see my previous statement about `/home`.

Vagrantfile
```diff
@@ -6,11 +6,12 @@ BOX_CPUS = ENV["BOX_CPUS"] || "1"
 BOX_MEMORY = ENV["BOX_MEMORY"] || "1024"
 DOKKU_DOMAIN = ENV["DOKKU_DOMAIN"] || "dokku.me"
 DOKKU_IP = ENV["DOKKU_IP"] || "10.0.0.2"
+DOKKU_USER = ENV["DOKKU_USER"] || "dokku"
 FORWARDED_PORT = (ENV["FORWARDED_PORT"] || '8080').to_i
 PREBUILT_STACK_URL = File.exist?("#{File.dirname(__FILE__)}/stack.tgz") ? 'file:///root/dokku/stack.tgz' : nil
 PUBLIC_KEY_PATH = "#{Dir.home}/.ssh/id_rsa.pub"
 
-make_cmd = "DEBIAN_FRONTEND=noninteractive make -e install"
+make_cmd = "DEBIAN_FRONTEND=noninteractive make -e install DOKKU_USER=#{DOKKU_USER}"
```
michaelshobbs: I expect this will need to go in front of the `make` command similar to `DEBIAN_FRONTEND`. If it indeed works this way, then you can downgrade this to a style nitpick. :smile:

Vagrantfile
```diff
@@ -48,7 +49,7 @@ Vagrant::configure("2") do |config|
     end
 
     vm.vm.provision :shell, :inline => "export DEBIAN_FRONTEND=noninteractive && apt-get update > /dev/null && apt-get -qq -y install git > /dev/null && cd /root/dokku && #{make_cmd}"
-    vm.vm.provision :shell, :inline => "cd /root/dokku && make dokku-installer"
+    vm.vm.provision :shell, :inline => "cd /root/dokku && make dokku-installer DOKKU_USER=#{DOKKU_USER}"
```
michaelshobbs: See similar comment on env vars and `make`

plugins/20_events/install
```diff
@@ -13,30 +13,30 @@ flag_rsyslog_needs_restart=n
...
+      chown "syslog:$DOKKU_USER" "$DOKKU_EVENTS_LOGFILE"
       ;;
   esac
   chmod 664 "$DOKKU_EVENTS_LOGFILE"
 fi
 
 if [[ -d "$DOKKU_RSYSLOG_FILTER_DIR" && ! -f  "$DOKKU_RSYSLOG_FILTER" ]]; then
   cat >"$DOKKU_RSYSLOG_FILTER" <<EOF
-:syslogtag, contains, "dokku" $DOKKU_EVENTS_LOGFILE
+:syslogtag, contains, "$DOKKU_USER" $DOKKU_EVENTS_LOGFILE
```
michaelshobbs: If I'm recalling rsyslog correctly, I think that has the potential to route other logs to this file if the configured user owns other processes.

josegonzalez:
@joshmanders needs rebase and *bump* :)

joshmanders:
Sounds good, hopefully soon I'll get some free time to jump back on this.

josegonzalez:
@joshmanders sounds like this isn't something you want to pickup again, is that correct?

joshmanders:
@josegonzalez if someone else wants to, they can, otherwise I can when I get some time, this feature hasn't been a priority for me anymore.

PR #2500, opened 2016-11-25
Suppress output unless the `git submodule update` call fails

### LINKED ISSUE #2479: git submodule update --init failure needs a better error message
Description of problem:
Pushing a repo that causes `git submodule update --init` to fail gives a generic error message from git that doesn't point to the real problem.  It would be nice to say that there was a problem checking out a submodule.

In my case, this was because I had bundled some dependencies into my repo and the dependency manger had cloned a repo or two into that bundle.

- `uname -a`: Linux podcastinet 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u2 (2016-10-19) x86_64 GNU/Linux

- `free -m`:
```
             total       used       free     shared    buffers     cached
Mem:          1000        845        155          0         64        584
-/+ buffers/cache:        195        804
Swap:            0          0          0
```
- `docker version`: Docker version 1.12.3, build 6b644ec
- `docker -D info`:
- `sigil -v`:
- `docker run -ti gliderlabs/herokuish:latest herokuish version`:
- `dokku version`: 0.7.2

Environment details (AWS, VirtualBox, physical, etc.): Digial Ocean

How was Dokku installed?: curl | bash

How reproducible: Every time

Steps to Reproduce:
1. Bundle up a common lisp webapp with [fukamachi/qlot](https://github.com/fukamachi/qlot)
2. Save the `quicklisp/` directory to the app repo
3. Attempt to push to dokku

Actual Results: https://gist.github.com/fisxoj/a4f5784e0398d13375cae96a3ee6d906
```
To podcastinet.com:podcastinet
error: failed to push some refs to 'dokku@podcastinet.com:podcastinet'
```

Expected Results:

Additional info:

josegonzalez:
This will give a bit more warning to developers as to what might be going on, allowing them to respond appropriately.

Closes #2479

> This was not tested on a real server, so don't merge until I do that.

### DIFF
--- a/plugins/git/functions
+++ b/plugins/git/functions
@@ -28,7 +28,7 @@ git_build_app_repo() {
   git remote add origin "$DOKKU_ROOT/$APP" &> /dev/null
   git fetch --depth=1 origin "refs/tags/$TMP_TAG" &> /dev/null
   git reset --hard FETCH_HEAD &> /dev/null
-  git submodule update --init --recursive &> /dev/null
+  suppress_output git submodule update --init --recursive
   GIT_DIR="$DOKKU_ROOT/$APP" git tag -d "$TMP_TAG" &> /dev/null || true
   find -name .git -prune -exec rm -rf {} \; > /dev/null
 
@@ -40,6 +40,21 @@ git_build_app_repo() {
     dokku_receive "$APP" "herokuish" "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" | sed -u "s/^/"$'\e[1G'"/"
   fi
 }
+
+suppress_output() {
+  declare desc="suppress all output from a given command unless there is an error"
+  local TMP_COMMAND_OUTPUT
+  TMP_COMMAND_OUTPUT=$(mktemp "/tmp/${FUNCNAME[0]}.XXXX")
+  trap 'rm -rf "$TMP_COMMAND_OUTPUT" > /dev/null' RETURN INT TERM EXIT
+
+  $* 2>&1 > "$TMP_COMMAND_OUTPUT" || {
+    local exit_code="$?"
+    cat "$TMP_COMMAND_OUTPUT"
+    return "$exit_code"
+  }
+  return 0
+}
+
 
 git_hook_cmd() {
   declare desc="kick off receive-app trigger from git prereceive hook"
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
I would say it'd actually be nice to see what's happening when submodules are being fetched. My projects can have submodules that take a long time to sync, and it'd be nice to know what's actually running

plugins/git/functions
```diff
@@ -41,6 +41,21 @@ git_build_app_repo() {
...
+
+  "$@" > "$TMP_COMMAND_OUTPUT" 2>&1 || {
+    local exit_code="$?"
+    cat "$TMP_COMMAND_OUTPUT"
+    return "$exit_code"
+  }
+  return 0
+}
+
+
```
michaelshobbs: extraneous newline

plugins/git/functions
```diff
@@ -41,6 +41,21 @@ git_build_app_repo() {
   fi
 }
 
+suppress_output() {
```
michaelshobbs: any reason not to place this in common?

plugins/git/functions
```diff
@@ -41,6 +41,21 @@ git_build_app_repo() {
...
+  declare desc="suppress all output from a given command unless there is an error"
+  local TMP_COMMAND_OUTPUT
+  TMP_COMMAND_OUTPUT=$(mktemp "/tmp/${FUNCNAME[0]}.XXXX")
+  trap 'rm -rf "$TMP_COMMAND_OUTPUT" > /dev/null' RETURN INT TERM EXIT
+
+  "$@" > "$TMP_COMMAND_OUTPUT" 2>&1 || {
```
michaelshobbs: clever! does this work as expected?

michaelshobbs:
as @RealOrangeOne brings up, perhaps we shouldn't hide the output at all?

plugins/git/functions
josegonzalez: It's only used here.

plugins/git/functions
josegonzalez: Yep, though I recommend you add a failing test for me ;)

josegonzalez:
Maybe. I think the submodule output is kinda verbose, so turning it on doesn't have real benefit for most people.

I also think submodules are gross as hell, and you shouldn't ever use them, but thats a separate matter.

plugins/git/functions
michaelshobbs: Academic argument: it's generic enough that it probably should exist in common. regardless, this sort of thing is an option in the golang `DokkuCmd` struct

plugins/git/functions
michaelshobbs: I will when this gets ported to golang 😝

plugins/git/functions
josegonzalez: Cool great. Going forward, outputting all logs to a file and the error output from a failing command to stdout seems reasonable.

PR #2501, opened 2016-11-25
wait for /var/lib/dokku and /home/dokku on boot

javawizard:
fixes containers not starting back up when /var/lib/dokku is mounted somewhere else

### DIFF
--- a/plugins/00_dokku-standard/install
+++ b/plugins/00_dokku-standard/install
@@ -31,6 +31,7 @@ if [[ $(systemctl 2> /dev/null) =~ -\.mount ]]; then
 [Unit]
 Description=Dokku app redeploy service
 Requires=docker.service
+RequiresMountsFor=$DOKKU_ROOT $DOKKU_LIB_ROOT
 After=docker.service
 
 [Service]


josegonzalez:
What happens when you don't have those as mounts?

javawizard:
Just tested, still works fine. Doesn't work if /var/lib/dokku is a symlink to a different volume though... Gonna do a bit more research and see if I can find something that works with that as well

javawizard:
Not finding anything obvious, so this is good to go as is.

josegonzalez:
My concern here is that someone doesn't mount those directories but instead a subdirectory - like the `storage` dir. What happens then? As well, we make `DOKKU_LIB_ROOT` and `DOKKU_ROOT` modifiable. Would be useful to at least respect those environment variables.

Ideally we can use debconf or something to configure this so that a user can modify it to wait on whatever they need to. Would you mind modifying the pr to do that?

josegonzalez:
@javawizard Mind responding to my comments? I can't merge this as is, and otherwise will close the pr.

javawizard:
I probably won't have time to make it debconf configurable in the near future, but I can change it to use `DOKKU_ROOT` and `DOKKU_LIB_ROOT`. Would that work for now?

josegonzalez:
should be fine

javawizard:
I have no idea where `DOKKU_LIB_ROOT` is being set. Will it be set when `install` runs?

josegonzalez:
Both are specified here: https://github.com/dokku/dokku/blob/c3f39fe512b84d65322bcf1c9b066f70bec95ff8/debian/postinst#L10

josegonzalez:
Closing. There hasn't been work on this in a while and placing these two items on external disk isn't standard, so should be handled by configuration management.

Issue #2356, opened 2016-08-07
[RFC] Create `report` subcommand for all plugins


josegonzalez:
It would be cool if we could have commands such as:
- [x] apps:report
- [x] certs:report
- [x] checks:report
- [x] docker-options:report
- [x] domains:report
- [ ] ~~nginx:report~~
- [x] proxy:report
- [x] ps:report
- [x] storage:report

This could help fill in a command like `dokku report APP` and allow users to see debug information for their entire system as well as an app that is not behaving well.

Thoughts? Would be useful because in many cases, we might not need to see a stacktrace to diagnose a problem, but all the reporting information for various bits of dokku would certainly help. Could also be useful for those diagnosing issues in their own applications.

mbtamuli:
I would like to work on this.  Could you please give example of what would `apps:report` or `ps:report` contain so that I can start with that.

josegonzalez:
Heh oops, totally missed your comment.

`apps:report` would have output like the following:


```shell
=====> App Information
       App dir:          /var/lib/dokku/services/mysql/playground/config
       Linked Services:
       Git Sha:       -
       Status:                  running
```

Something like that. We could add extra data as time goes on for anything specific to the `app` plugin (ditto for the rest).

josegonzalez:
Note: this can be worked on plugin by plugin, no need to do all of them at once. If anyone would like sample info we want for each plugin, feel free to comment and say you're taking that specific plugin :)

mbtamuli:
I will start with the apps plugin. (`apps:report`)

josegonzalez:
`ps:report` output. I think each process that is "scaled" should have an entry here as to whether it is started/stopped/etc. Not sure if the way I have for representing container IDs is necessarily the best one, but open to suggestions.

```
=====> node-js-app ps:report
       Processes:        3
       Deployed:         true
       Running:          true
       Restart Policy:   on-failure:3
       Status web.1:     running (CID: 12345)
       Status worker.1:  stopped
       Status worker.2:  stopped
```

josegonzalez:
Closing as pull requests are available for all plugins.

I decided against `nginx:report` as it's not yet clear what data would be useful there.

PR #2504, opened 2016-11-26
Adds the `report` subcommand for all plugin

mbtamuli:
- [x] apps:report
- [ ] checks:report
- [ ] docker-options:report
- [ ] domains:report
- [ ] nginx:report
- [ ] proxy:report
- [x] ps:report
- [ ] storage:report

This is one PR for #2356


mbtamuli:
I realized later after pushing this, that this PR is incomplete. I have to edit the `apps/command` file and also edit the docs.

plugins/apps/subcommands/report
```diff
@@ -0,0 +1,73 @@
...
+
+  if [[ -z "$INFO_FLAG" ]]; then
+    dokku_log_info2 "$APP"
+    if (is_deployed "$APP"); then
+      for flag in "${flag_map[@]}"; do
+        key="$(echo "${flag#--}" | cut -f1 -d' ' | tr - ' ')"
+        dokku_log_verbose "$(printf "%-20s %-25s" "${key^}" "${flag#*: }")"
+      done
+    else
+      echo "not deployed"
```
josegonzalez: Should use our error helpers. If there isnt one that doesnt output to stderr without exiting, create one :)

mbtamuli: Done. Should I squash the commits?

mbtamuli:
@josegonzalez Can you also guide me to write tests for these?

josegonzalez:
You can just check the specific output of commands, and also that the flags all work properly. Here is an [example](https://github.com/dokku/dokku-redis/blob/master/tests/service_info.bats#L36).

josegonzalez:
@mbtamuli Which one do you want to tackle next?

mbtamuli:
@josegonzalez I don't know if there's any order to it. Maybe `proxy:report` if there's no specific order?

About `dokku report APP` or `dokku report`, to get a complete report of everything, it will have to be done after completing the report subcommand for all the plugins mentioned.

josegonzalez:
@mbtamuli follow on with `proxy:report` :)

josegonzalez:
For the `dokku report APP` etc., yes I agree. Once the other plugins have a `report` subcommand, will be fairly trivial to do this those.

mbtamuli:
I will try to follow up on `proxy:report` soon.

PR #2508, opened 2016-11-28
Add default functions for all commands

OmarShehata:
### Problem:

When you run `dokku`, it defaults to the help, which lists all the commands. Something like:

```
...
shell
ssh-keys
storage
...
```

If you then run `dokku ssh-keys` you get the mysterious error:

```
 !     `ssh-keys` is not a dokku command.
 !     See `dokku help` for a list of available commands.
```
Which is confusing for a beginner, since `dokku help` does list `ssh-keys` as a valid command, but trying it says it's not a valid command. 

### My Solution:

I've added a default for all commands that had this problem, which are:

- certs
- nginx 
- ssh-keys
- storage

It outputs the same help message you get as if you had ran `dokku COMMAND:help`. To make sure the code wasn't duplicated, I moved the help function into the `functions` file. I was a little confused why the help outputted by the `commands` file seem to have a formatted and an unformatted version. So I just left that intact by having two help functions in the `functions` file. I don't know if this was the best way to do it, I'm open to feedback/criticism!

### Alternative solutions:

Most other commands seem to default to a "list", it would be good to stay consistent, but that's not always possible. The `nginx` command has no list option for example.

In any case, I think every option _should_ have a default. Saying something isn't a command when it clearly is can cause unnecessary confusion.


josegonzalez:
I would rather we change this a little:

- `certs:default` should be an alias for `certs:info`
- `ssh-keys:default` should be an alias for `ssh-keys:list`
- `storage:default` should be an alias for `storage:list`

Thoughts?

OmarShehata:
I think that makes sense to me, @josegonzalez. That'd be consistent with the rest of the commands too.

Do you have any comments/feedback on how I created the aliases or is this fine? I can add a few more commits to swap the aliases to list/info.

josegonzalez:
I'd revert most of this changeset (minus the nginx one) and then move any extra logic from the subcommands into the `functions` file for that given plugin, then just call that.

OmarShehata:
Ok, I've reverted everything except nginx, and added the new ones. Let me know if you'd like me to do anything else on this one!

josegonzalez:
Heh I'm actually rethinking this, and maybe think the default should always be help. In cases where we do a `list` for the default but have no `list` command, we should make a list command. Something to clean up for the 1.0 I guess.

PR #2510, opened 2016-11-28
Add domains:set and domains:set-global commands

IlyaSemenov:
As discussed in #2410, this adds `domains:set` and `domains:set-global`, along with tests and documentation updates.

I am indeed missing this feature. 90% of the time we install the new app and we immediately get rid of *app.dokku.me* domain name in favour of *app.com* or *staging.app.com*. Currently that involves running three commands every time instead of one.

### DIFF
--- a/docs/configuration/domains.md
+++ b/docs/configuration/domains.md
@@ -3,14 +3,16 @@
 > New as of 0.3.10
 
 ```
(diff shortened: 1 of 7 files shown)


josegonzalez:
Thanks for the pull request!

PR #2523, opened 2016-12-07
Allowing specifying a git remote

andyjeffries:
This allows you to create multiple Dokku git remotes (e.g. `production` and `staging`) for the same project, then manage them using:

```
dokku production ps:restart
dokku staging ps:restart
```

It maintains backward compatibility so if one isn't specified (or the specified one doesn't exist), it uses `dokku` and passes the argument in as usual, so this also works:

```
dokku ps:restart
```

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -38,12 +38,12 @@ is_git_repo() {
 }
 
 has_dokku_remote() {
-    git remote show | grep dokku
+    git remote show | grep $DOOKU_GIT_REMOTE
 }
 
 if [[ -z $DOKKU_HOST ]]; then
   if [[ -d .git ]] || git rev-parse --git-dir > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


andyjeffries:
Is there anything else I can do to get this merged?

josegonzalez:
Can we instead do:

```shell
dokku --remote production cmd
```
?

josegonzalez:
Closing this as it hasn't been updated to do the `--remote` version, which I am implementing in a separate pull request. Thanks for the idea though, @andyjeffries :)

andyjeffries:
OK, that's great that you're doing it - thanks @josegonzalez.

andyjeffries:
Did you also see https://github.com/dokku/dokku-maintenance/issues/6 - a bug report so that plugin will work nicely with dokku_client.sh?

josegonzalez:
I did not, but I'll take a look at it.

devniel:
Thanks, I have just used the `--remote` flag, it seems that is necessary to update the docs to mention it, if possible, I'll take that task.

Thanks, again.

Issue #2532, opened 2016-12-16
Node JS app not working


facundomedica:
Description of problem:

I've installed Dokku using the one click app in Digital Ocean, in Ubuntu 16.04. I've tried running the Node JS sample of Heroku and it builds just fine and it outputs a url (Which is my server ip with the port). The problem is that when I put this url on my browser, it tries to load it for a long time but it does nothing. I think is something with related with ports, maybe?

(I'll try to complete most of the things below)

Output of the following commands

- `uname -a`:
Linux XsomenameX-dokku 4.4.0-53-generic #74-Ubuntu SMP Fri Dec 2 15:59:10 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

- `free -m`:

```
              total        used        free      shared  buff/cache   available
Mem:            488          77          65           5         345         371
Swap:             0           0           0
```

- `docker version`:

Client:
 Version:      1.12.4
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   1564f02
 Built:        Tue Dec 13 00:08:34 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.4
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   1564f02
 Built:        Tue Dec 13 00:08:34 2016
 OS/Arch:      linux/amd64

- `docker -D info`:

Containers: 8 **_--> THIS LOOKS ODD, I ONLY HAVE ONE APP_**
 Running: 1
 Paused: 0
 Stopped: 7
Images: 17
Server Version: 1.12.4
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 36
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: host bridge overlay null
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Security Options: apparmor seccomp
Kernel Version: 4.4.0-53-generic
Operating System: Ubuntu 16.04.1 LTS
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 488.5 MiB
Name: XsomenameX-dokku
ID: LJ2H:FYCT:5DKP:AMOL:R6IL:T5F6:G2GN:OSXH:EASL:IXMS:JPQB:6J2A
Docker Root Dir: /var/lib/docker
Debug Mode (client): true
Debug Mode (server): false
[...]

PR #2533, opened 2016-12-16
Solution to problem related to firewall

facundomedica:
Resume: Ubuntu 16.04 seems to have the firewall enabled by default, which prohibits the access to apps.
Original issue: https://github.com/dokku/dokku/issues/2532

### DIFF
--- a/docs/getting-started/troubleshooting.md
+++ b/docs/getting-started/troubleshooting.md
@@ -174,3 +174,16 @@ echo vm.swappiness = 10 | sudo tee -a /etc/sysctl.conf
 ```
 
 ***
+
+__Symptom:__ I successfully deployed my application with no deployment errors but I'm receiving Connection Timeout when attempting to access the application.
+
+__Solution:__
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/troubleshooting.md
```diff
@@ -174,3 +174,16 @@ echo vm.swappiness = 10 | sudo tee -a /etc/sysctl.conf
 ```
 
 ***
+
+__Symptom:__ I successfully deployed my application with no deployment errors but I'm receiving Connection Timeout when attempting to access the application.
+
+__Solution:__
+
+This can occur if Dokku is running on a system with a firewall enabled (some OS versions like Ubuntu 16.04 have this enabled by default). The follow script will disable the firewall.
```
josegonzalez: Can you note that the following command is only for ufw, and how to check if ufw is enabled?

facundomedica: Okay!

PR #2539, opened 2016-12-28
Fix dokku-installer.service removal

ebeigarts:


debian/postrm
```diff
@@ -11,13 +11,13 @@ readonly DOKKU_LIB_ROOT="${DOKKU_LIB_PATH:-/var/lib/dokku}"
 readonly DOKKU_LOGS_DIR="${DOKKU_LOGS_DIR:="/var/log/dokku"}"
 
 main() {
-  if [[ -f /etc/systemd/system/dokku-installer.conf ]] || [[ -f /etc/init/dokku-installer.conf ]]; then
+  if [[ -f /etc/systemd/system/dokku-installer.service ]] || [[ -f /etc/init/dokku-installer.conf ]]; then
```
joshmanders: Still have a `.conf` that could be changed to `.service` here too.

ebeigarts: `/etc/init/dokku-installer.conf` is for upstart and I should stay `.conf`

joshmanders: Ah my bad

PR #2544, opened 2017-01-03
update herokuish to 0.3.29

michaelshobbs:


joshmanders:
I'm confused. This says updating to `0.3.29` but the commit has `0.3.24`

josegonzalez:
fixed!

joshmanders:
That's one way to do it. Was just not sure if it meant to be 0.3.24 or 0.3.29. :D

michaelshobbs:
doh! Supposed to be 0.3.24. Was doing too many things at once. Unfortunately the git commit comment is wrong 😢

PR #2546, opened 2017-01-04
convert repo plugin to golang

michaelshobbs:
NOTE: This is meant to be a proof of concept and is wide open for discussion. 

- [x] converts repo plugin to golang
- [x] includes basic framework for building core plugins from golang
  - each plugin will need to provide a `Makefile` and manage it's own vendored dependencies
  - [x] official and community plugins can implement their compiling steps in the `install` plugin trigger
- [x] includes test execution for golang tests
- [x] includes port of some common functions
  - can be included in plugins via `import github.com/dokku/dokku/plugins/common`
- [x] uses glide for golang dependency mgmt (glide is not required. any golang 1.5+ dep tool is fine)
- [x] docs when we've finalized the approach

Thoughts on supporting bash plugins going forward:
- a plugin should expose (as desired) it's functions for use in other plugins via an executable (perhaps named `go_functions` and a lightweight bash function wrapper.
- ex. in `plugins/config/functions`
```
config_get() {
    "$PLUGIN_AVAILABLE_PATH/config/go_functions config_get "$@"
}
```


michaelshobbs:
@mattaitchison you had some comments on this, yea?

josegonzalez:
Looks pretty decent imo. How does this affect installs from source?

ebeigarts:
Does it mean that dokku contributors will need to learn and install go? Does go provide something better than bash for the repo plugin?

josegonzalez:
Dokku was never meant to always be bash. You could always implement plugins in any language you wanted to - python, erlang, node, brainfuck, whitespace. As long as Dokku could communicate with a binary, then it should work fine.

As far as the core is concerned, yes, you'd have to learn Golang. It's actually quite easy to read, so should be easy enough to follow/contribute to.

Why are we moving away from bash?

- People are afraid to contribute to bash. Somehow, it's hard to pickup (I feel as though this is just people coming up with excuses for not contributing).
- We've reached the limits of bash. We've done some crazy shit, and it's getting harder to both test/continue developing with the limited syntax afforded to us.
- Some people didn't want to use Dokku *because* it is written in bash. Crazy as it seems, they see it as either non-performant, not-secure, or just plain old not-well architected. Getting away from this perception is good for our own mental health, as no one wants to work on something that others denigrate.
- Neither @michaelshobbs nor I really want to keep writing bash.

@michaelshobbs actually worked on a python port that sort of worked, but it got lost as we kept iterating on the bash version. An iterative approach in this repo is probably our best bet, so this is where we are.

---

@michaelshobbs we'll need to update the development docs.

joshmanders:
I'm 👍  on going to Golang. I don't know it, but it's something I wanted to learn, but had nothing to learn with. Dokku advanced my bash knowledge, and it's really the only bash I work with, so switching to golang is a reason for me to learn go.

patricktruebe:
I think this opens dokku for a much broader base of contributors. E.g. I am able to read and understand almost every line in dokku. But writing clean and sane bash is much harder. Tried and failed. So many simple things like profiling, writing tests and debugging seem cumbersome in bash to me.

I second that this shift away from bash should be announced in a central place (like contribution docs).

josegonzalez:
One other thing: worth scrubbing any references to bash in our repo. It's a cute selling point, but honestly the language a tool like ours is written in shouldn't really affect the end user.

One thing I can think of that will be annoying is having people use patches locally to test changes.

plugins/common/common.go
```diff
@@ -0,0 +1,149 @@
...
+
+// LogFail is the failure log formatter
+// prints text to stderr and exits with status 1
+func LogFail(text string) {
+	fmt.Fprintln(os.Stderr, fmt.Sprintf("FAILED: %s", text))
+	os.Exit(1)
+}
+
+// GetDeployingAppImageName returns deploying image identifier for a given app, tag tuple. validate if tag is presented
+func GetDeployingAppImageName(args ...string) (imageName string) {
```
michaelshobbs: @mattaitchison in the current bash version, this function can get called with 1 or more args. is this implementation as reasonable way to handle that going forward? if not, suggestions?

josegonzalez: Lets only cover current use cases and expand as necessary. Maybe like `GetDeployingAppImageNameFromApp`?

michaelshobbs: @mattaitchison @josegonzalez i'm curious what's wrong with the `GetDeployingAppImageName(args ...string)` implementation. can either of you expound a bit beyond "it's not idiomatic golang"? 😄

josegonzalez: You'd need to check and cast individual parts of the incoming `args`, which isn't fun.

michaelshobbs: Yes but that is the idiomatic way to do variadic function in golang, correct?

josegonzalez: They are just silly though.

michaelshobbs:
updated

michaelshobbs:
Added docs and converted the [smoke-test-plugin](https://github.com/dokku/smoke-test-plugin) to golang as an example implementation.

michaelshobbs:
refactored to build individual binaries

michaelshobbs:
@josegonzalez @mattaitchison ping

josegonzalez:
@alexquick ping assuming you want to review some golang

Makefile
```diff
@@ -9,6 +9,9 @@ DOKKU_LIB_ROOT ?= /var/lib/dokku
 PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/plugins
 CORE_PLUGINS_PATH ?= ${DOKKU_LIB_ROOT}/core-plugins
 
+export GO_REPO_ROOT := /go/src/github.com/dokku/dokku
+export BUILD_IMAGE := golang:1.7.1
```
josegonzalez: 1.7.4 is latest.

docs/development/plugin-creation.md
```diff
@@ -9,6 +11,23 @@ If you create your own plugin:
 5. Subscribe to the [dokku development blog](http://progrium.com) to be notified about API changes and releases
 
 
+### Compilable Plugins (i.e. golang, java(?), c, etc.)
+The plugin developer is required to implement the `install` trigger such that it will output your built executable(s) in the correct directory structure to implement the plugin's desired command and/or trigger API. See [smoke-test-plugin](https://github.com/dokku/smoke-test-plugin) for an example.
```
josegonzalez: Can we provide guidance on what the best method is for various use cases?

plugins/common/common.go
```diff
@@ -0,0 +1,141 @@
...
+// GetAppImageRepo is the central definition of a dokku image repo pattern
+func GetAppImageRepo(appName string) string {
+	return strings.Join([]string{"dokku", appName}, "/")
+}
+
+// VerifyImage returns true if docker image exists in local repo
+func VerifyImage(image string) bool {
+	imageCmd := NewDokkuCmd(fmt.Sprintf("docker inspect %s", image))
+	imageCmd.ShowOutput = false
+	if imageCmd.Execute() {
```
josegonzalez: I'd like to log all output in a failed run. How would that work?

plugins/common/glide.lock
```diff
@@ -0,0 +1,8 @@
+hash: e438ca8cf1bdca3e9c984bb1ee70c0d58be9da52cbe13a079a7c6cbc14210a37
```
josegonzalez: Need dev docs on glide.

plugins/repo/Makefile
```diff
@@ -0,0 +1,26 @@
+.PHONY: build-in-docker build clean
```
josegonzalez: Can we include this from a common.mk in the common plugin? I'd imagine this will get repetitive.

alexquick:
@josegonzalez I don't have a lot to say. It's way more fun than bash.

I'm not a super experienced go dev but I can talk about what I ran into already using it feel free to ignore anything that's really dumb.

* Are you open to having the subcommands target built by some find magic? I found that copying and pasting those subtargets was easy but irritating. I can make said find magic if you're open to it.

* common: in the config branch I've added a helper in common for triggering plugn. The other obvious omission was logging in the same format as the rest of dokku. I can add that if it's alright.

* In my branch I have a variadic partner to NewDokkuCmd. How do you feel about moving NewDokkuCmd itself to being variadic? `NewDokkuCmd("docker", "inspect", "image")` feels better than `NewDokkuCmd(fmt.Sprintf("docker inspect %s", image))`, especially when an infinitely devious user somehow gets a space into an imagename (somehow?).

* I'm still trying to wrap my head around the vendored dependencies. Should gomega be vendored into common? If I need to use the common package, I don't vendor it nor its dependencies do I? If I share dependencies with common and use it do I need to vendor those, or do I get them for free?

shulcsm:
Great. I really wish this happened a lot earlier. Have you considered usign native go dockerclient so we can minimize calls to shell and all the fragile string manipulations?

One thing i'm concerned about is exposing required primitives and environment to plugins in other languages.

plugins/common/common.go
michaelshobbs: That's not covered here. Suggestions welcomed.

plugins/common/glide.lock
michaelshobbs: https://github.com/Masterminds/glide/blob/master/README.md

plugins/repo/Makefile
michaelshobbs: sure

plugins/common/glide.lock
josegonzalez: I mean we need to look to them in our docs

michaelshobbs:
@shulcsm: the go dockerclient has been considered and I can take a close look at it in a future PR. When attempted to port dokku to python late last year, the python client was limiting to say the least. I'm assuming the go client is much more flexible but I have not attempted to use it here just yet. Can you expound on your concerns regarding exposing things to plugins in other languages?

@alexquick: 
- For sure. Can you show an example of said "find magic"?
- +1 on making `NewDokkuCmd` variadic. What do you mean by "On the other hand repo:purge-cache is gonna get ugly"?
- The idea is to never call the `dokku` binary from within dokku. Instead a plugin should be using exposed functions
- As each plugin is compiled independently, the plugin (author) is responsible for providing it's own dependencies and can use whatever go 1.5+ compliant tool they want. See: `plugins/common/vendor` and `plugins/repo/src/vendor`

plugins/common/glide.lock
michaelshobbs: Glide is not a requirement. It's just the tool I chose. Any golang 1.5+ dep tool is fine

docs/development/plugin-creation.md
michaelshobbs: Best method for what? I linked to the smoke-test-plugin as an example instead of cluttering up the docs with code examples.

alexquick:
@michaelshobbs You can get those subcommands into a pattern rule like so: https://github.com/dokku/dokku/compare/mh-repo-golang...alexquick:feature-makefile-refactor?expand=1 I don't know if the complication is worth it though. 

* If `ShellCmd` were variadic, I assume it wouldn't do any further tokenization on its arguments. I I meant the call in `repo:purge-cache` (https://github.com/dokku/dokku/pull/2546/files#diff-9b75cf42f39af3ee67746b2d69d61375R27) is pretty complicated, so tokenizing that ahead of time would be tedious. 

* I agree that calling the dokku binary again is not great, but there will be a mix of plugins of various languages. A golang plugin can't call the functions exposed in say `plugins/ps/functions`, so what is the best way to share  behavior from bash to golang plugins? In the other direction you can really easily write a bash wrapper around a binary, so that part is solved as you said.

* The dependencies clicked for me finally, thanks.

josegonzalez:
I think we can use go-basher to share from bash to golang, but the idea here is that the Dokku core will never do that. As we rewrite to golang, all bits we need will be ported as well.

michaelshobbs:
@alexquick: 
- Regarding the find magic, I agree on the complexity point. As a plugin author, you will be explicit in defining your subcommands and I think that explicitness should probably extend to the `Makefile` as well. Regardless of my opinion or how we implement it in core, it will be up to each plugin author to implement their own `Makefile`. That being said, I'd like to see core plugins use this explicit pattern.
- calls to `docker`, specifically, should be implemented with the go-dockerclient if at all possible. I'll look into translating that one soon
- as @josegonzalez mentioned, we might be able to use go-basher to bridge the gap as we port core plugins....not sure it's necessary but that really depends on how much change we can stomach at once. 😄

shulcsm:
@michaelshobbs i don't think go client is any better compared to python since they both cover only things that are implemented by docker remote api and it's lower level than shell client. The difference is mostly sugar in form of run and exec.
My concerns regarding exposing things is that i don't want to re-implement `common` if i do plugin in other language.

michaelshobbs:
@shulcsm Makes sense. dokku core is responsible for maintaining `common` functions. As we progress down the path of porting to go, `common` functions will come along as well.

josegonzalez:
I'd like to make a 0.8.1 and then we can release this in 0.8.2. Fine by you, @michaelshobbs ?

weyert:
Great, looks like a good reason to learn Go. Now I just need to come up with a nice plug-in for Dokku. Can you also make a sub-plugin for a plugin? E.g. add support for backing up database backups to Azure instead of S3 for `dokku-postgres`?

josegonzalez:
Weyert: no, thats currently not possible, and also not the goal of this refactor.

PR #2548, opened 2017-01-05
Update images/tar deployment method links

Firfi:


joshmanders:
Not for sure why this needs to be. If you look at it on the page, http://dokku.viewdocs.io/dokku/getting-started/upgrading/ those links work, so no need to hardcode them in to dokku.viewdocs.io

Firfi:
Make sense. I thought it's git-only version. Sorry for disturbance.

joshmanders:
No worries. Thanks for the PR though!

PR #2555, opened 2017-01-09
Skip failing applications when running ps:restore on boot

### LINKED ISSUE #2463: Dokku doesn't restore all apps after system reboot
I have around 30 apps on one host and after a reboot only half of my apps were started, because one app failed to start and `ps:restore` subcommand has `set -e` which exits the script and doesn't try to start all the other apps. Also it is not very efficient to restore 30+ apps using just single CPU.

I am currently using `dokku apps | grep -v '=====>' | parallel -v -v dokku ps:start` to restart all the apps and use all 8 CPU cores and speeding the process 8x.

One solution would be to get rid of `set -e` in `ps:restore` command.
Another solution would be to use GNU parallel, but that's a new dependency, which is available using `apt-get install parallel`.

@josegonzalez WDYT?

josegonzalez:
Also adds support for using parallel for restoring apps.

@ebeigarts mind checking this out to see if it works for you? My limited testing shows that it should work.

@michaelshobbs had to use the `dokku` bin directly instead of `ps_start` as `parallel` starts it's own shell and doesn't have access to all functions (exporting them all would be a pita).

Closes #2463

### DIFF
--- a/plugins/ps/subcommands/restore
+++ b/plugins/ps/subcommands/restore
@@ -11,7 +11,11 @@ ps_restore_cmd() {
     local DOKKU_APP_RESTORE=$(config_get "$app" DOKKU_APP_RESTORE || true)
     if [[ $DOKKU_APP_RESTORE != 0 ]]; then
       echo "Restoring app $app ..."
-      ps_start "$app"
+      if ps_start "$app"; then
+        continue
+      fi
+
+      dokku_log_warn "dokku ps:restore ${app} failed"
     fi
   done
 }
--- a/plugins/ps/subcommands/restore
+++ b/plugins/ps/subcommands/restore
@@ -7,6 +7,12 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 ps_restore_cmd() {
   declare desc="starts all apps with DOKKU_APP_RESTORE not set to 0 via command line"
   local cmd="ps:restore"
+
+  if which parallel > /dev/null 2>&1; then
+    dokku_apps | parallel dokku ps:start
+    return
+  fi
+
   for app in $(dokku_apps); do
     local DOKKU_APP_RESTORE=$(config_get "$app" DOKKU_APP_RESTORE || true)
     if [[ $DOKKU_APP_RESTORE != 0 ]]; then
--- a/plugins/ps/subcommands/restore
+++ b/plugins/ps/subcommands/restore
@@ -20,7 +20,6 @@ ps_restore_cmd() {
       if ps_start "$app"; then
         continue
       fi
-
       dokku_log_warn "dokku ps:restore ${app} failed"
     fi
   done


ebeigarts:
@josegonzalez parallel works, maybe we need to add `Recommends: parallel` to https://github.com/dokku/dokku/blob/master/debian/control#L7?

plugins/ps/subcommands/restore
```diff
@@ -7,11 +7,21 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
...
+
   for app in $(dokku_apps); do
     local DOKKU_APP_RESTORE=$(config_get "$app" DOKKU_APP_RESTORE || true)
     if [[ $DOKKU_APP_RESTORE != 0 ]]; then
       echo "Restoring app $app ..."
-      ps_start "$app"
+      if ps_start "$app"; then
+        continue
+      fi
+
```
michaelshobbs: nitpick: :-1 on extra space

josegonzalez:
@ebeigarts mind making a PR for that?

plugins/ps/subcommands/restore
```diff
@@ -7,11 +7,20 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 ps_restore_cmd() {
   declare desc="starts all apps with DOKKU_APP_RESTORE not set to 0 via command line"
   local cmd="ps:restore"
+
+  if which parallel > /dev/null 2>&1; then
+    dokku_apps | parallel dokku ps:start
+    return
```
vincentfretin: you don't seem to check DOKKU_APP_RESTORE here for each app like it's done below

vincentfretin:
I'm still using an old dokku 0.4.13 and I actually modified directly /var/lib/dokku/plugins/available/ps/commands some months ago to use the sem command (from the parallel package) like this:

```      
      ps:restartall)
        for app in $(dokku_apps); do
          sem -j+0 "dokku ps:restart $app"
        done
        sem --wait
        ;;                                                                      
                                                                                
      ps:restore)
        for app in $(dokku_apps); do
          DOKKU_APP_RESTORE=$(dokku config:get $app DOKKU_APP_RESTORE || true)
          if [[ $DOKKU_APP_RESTORE != 0 ]]; then
            echo "Restoring app $app ..."
            sem -j+0 "dokku ps:start $app"
          fi
        done
        sem --wait
      ;;
```

plugins/ps/subcommands/restore
josegonzalez: @vincentfretin how do you recommend we do that?

vincentfretin: @josegonzalez I attached the code below in the PR #2555, it adds a hard dependency on parallel though.

josegonzalez: Mind making a pr?

vincentfretin: @josegonzalez one year later, here a PR that fixes it for current master #3023

PR #2557, opened 2017-01-09
Dokku cli improvements

josegonzalez:
Lots of general CLI cleanup:

- General CLI tool cleanup
- Add the ability to specify a dokku remote via `--remote REMOTE_NAME`. Note that setting `DOKKU_GIT_REMOTE` will override any detection set via this flag, and `DOKKU_HOST` will override the extracted host from the specified remote.
-  #2370: Add exceptions back to dokku_client.sh
- Properly handle --global flag
- Properly handle passed flags in the official bash client
- Add support for specifying the --app flag in the official bash client


dokku
```diff
@@ -35,10 +35,12 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 
 parse_args "$@"
 args=("$@")
+skip_arg=false
```
michaelshobbs: can you explain this flow?

josegonzalez: If we have a flag that takes an argument, we want to also skip that argument. In this case, the `--app` flag is used with an arg (the app name) so we want to skip it.

If we don't skip it, as soon as we hit that argument, dokku will assume it's part of the command, so something like:

```
dokku --app APP --force apps:destroy
```

Ends up being the following:

```
APP --force apps:destroy
```

And since there is no `APP` command, we won't actually the destroy the app.

Instead, we skip the `APP` parameter, then *also* skip `--force`, resulting in the command being:

```
apps:destroy
```

I made a similar fix to the common/functions bit.

PR #2569, opened 2017-01-16
Reload nginx after adding default vhost file

OmgImAlexis:


docs/configuration/domains.md
```diff
@@ -92,6 +92,8 @@ server {
 }
 ```
 
+Make sure to restart nginx after creating this file by running `service nginx reload`.
```
josegonzalez: Should say `reload`, not `restart`.

PR #2583, opened 2017-01-23
feat: implement application cloning

josegonzalez:
Application cloning can be used to setup review applications based on an existing application.

This is useful in CI/CD pipelines where a developer may wish to verify code in a custom environment without overwriting a shared staging environment.

### DIFF
--- a/plugins/apps/subcommands/clone
+++ b/plugins/apps/subcommands/clone
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/apps/functions"
+source "$PLUGIN_AVAILABLE_PATH/ps/functions"
+
+apps_clone_cmd() {
+  declare desc="renames an app"
+  declare OLD_APP="$2" NEW_APP="$3"
+  local cmd="apps:rename"
+  local SKIP_REBUILD=false
+
+  if [[ "$2" == "--skip-deploy" ]]; then
+    SKIP_REBUILD=true
+    OLD_APP="$3"
+    NEW_APP="$4"
+  fi
+
+  [[ -z "$OLD_APP" ]] && dokku_log_fail "Please specify an app to run the command on"
+  [[ -d "$DOKKU_ROOT/$NEW_APP" ]] && dokku_log_fail "Name is already taken"
+  local NEW_CACHE_DIR="$DOKKU_ROOT/$NEW_APP/cache"
+
+  apps_create "$NEW_APP"
+  cp -a "$DOKKU_ROOT/$OLD_APP/." "$DOKKU_ROOT/$NEW_APP"
+
+  if [[ -d "$NEW_CACHE_DIR" ]] && ! rmdir "$NEW_CACHE_DIR"; then
+    docker run "$DOKKU_GLOBAL_RUN_ARGS" --rm -v "$NEW_CACHE_DIR:/cache" "dokku/$OLD_APP" chmod 777 -R /cache
+  fi
+  rm -rf "$NEW_CACHE_DIR"
+
+  [[ -f "$DOKKU_ROOT/$NEW_APP/URLS" ]] && sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/URLS"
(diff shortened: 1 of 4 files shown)


plugins/apps/subcommands/clone
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/apps/functions"
+source "$PLUGIN_AVAILABLE_PATH/ps/functions"
+
+apps_clone_cmd() {
+  declare desc="renames an app"
```
znz: rename → clone ?

plugins/apps/subcommands/clone
```diff
@@ -0,0 +1,39 @@
...
+  declare OLD_APP="$2" NEW_APP="$3"
+  local cmd="apps:rename"
```
znz: ditto.

plugins/apps/subcommands/clone
```diff
@@ -0,0 +1,39 @@
...
+  if [[ -d "$NEW_CACHE_DIR" ]] && ! rmdir "$NEW_CACHE_DIR"; then
+    docker run "$DOKKU_GLOBAL_RUN_ARGS" --rm -v "$NEW_CACHE_DIR:/cache" "dokku/$OLD_APP" chmod 777 -R /cache
+  fi
+  rm -rf "$NEW_CACHE_DIR"
+
+  [[ -f "$DOKKU_ROOT/$NEW_APP/URLS" ]] && sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/URLS"
+  [[ -f "$DOKKU_ROOT/$NEW_APP/VHOST" ]] && sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/VHOST"
+  [[ -f "$DOKKU_ROOT/$NEW_APP/hooks/pre-receive" ]] && sed -i -e "s/git-hook $OLD_APP/git-hook $NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/hooks/pre-receive"
+  [[ "$SKIP_REBUILD" == "false" ]] || ps_rebuild "$NEW_APP"
+  plugn trigger post-app-clone "$OLD_APP" "$NEW_APP"
```
rvalyi: great that you thought about adding such a hook!

plugins/apps/subcommands/clone
```diff
@@ -0,0 +1,39 @@
...
+  apps_create "$NEW_APP"
+  cp -a "$DOKKU_ROOT/$OLD_APP/." "$DOKKU_ROOT/$NEW_APP"
+
+  if [[ -d "$NEW_CACHE_DIR" ]] && ! rmdir "$NEW_CACHE_DIR"; then
+    docker run "$DOKKU_GLOBAL_RUN_ARGS" --rm -v "$NEW_CACHE_DIR:/cache" "dokku/$OLD_APP" chmod 777 -R /cache
+  fi
+  rm -rf "$NEW_CACHE_DIR"
+
+  [[ -f "$DOKKU_ROOT/$NEW_APP/URLS" ]] && sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/URLS"
+  [[ -f "$DOKKU_ROOT/$NEW_APP/VHOST" ]] && sed -i -e "s/$OLD_APP/$NEW_APP/g" "$DOKKU_ROOT/$NEW_APP/VHOST"
```
michaelshobbs: Won't think break the nginx config in certain scenarios with more generic vhosts? i.e. `www.mydomain.com` or `staging.mydomain.com`? Might be good to note that in the docs.

josegonzalez: Good catch.

josegonzalez:
@michaelshobbs can I get an amen?

tests/unit/10_apps.bats
```diff
@@ -90,3 +90,36 @@ teardown () {
...
+  run bash -c "dokku --force apps:destroy great-test-name"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+
+  run bash -c "dokku apps:clone --skip-deploy $TEST_APP great-test-name"
+  echo "output: "$output
+  echo "status: "$status
+  assert_success
+  run bash -c "curl --silent --write-out '%{http_code}\n' `dokku url great-test-name` | grep 404"
```
michaelshobbs: Not clear as to why both `curl` tests should return 404. Seems like the one without `--skip-deploy` should return actual app output.

josegonzalez: For the first, we're replicating the `apps:rename` test. In this one, we assert a failure.

michaelshobbs: Gah! No reviews before coffee!

PR #2586, opened 2017-01-25
Behind-proxy support for herokuish install

bevand10:
Inject http proxy envvars to docker build command to allow package installation behind corporate proxies.

These envvars (http_proxy and https_proxy) are standard, and usually configured via system-wide /etc/profile.d files.

In cases where these envvars are not configured, the additions appear to be mute and non-destructive.

### DIFF
--- a/deb.mk
+++ b/deb.mk
@@ -92,7 +92,7 @@ deb-herokuish:
 	@echo "  sudo docker rmi gliderlabs/herokuish" >> /tmp/tmp/post-install
 	@echo "fi" >> /tmp/tmp/post-install
 	@echo "echo 'Importing herokuish into docker (around 5 minutes)'" >> /tmp/tmp/post-install
-	@echo "sudo docker build -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
+	@echo "sudo docker build --build-args http_proxy=$http_proxy --build-args https_proxy=$https_proxy -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
 
 	@echo "-> Cloning repository"
 	git clone -q "https://github.com/$(HEROKUISH_REPO_NAME).git" --branch "v$(HEROKUISH_VERSION)" /tmp/tmp/herokuish > /dev/null
--- a/deb.mk
+++ b/deb.mk
@@ -92,7 +92,10 @@ deb-herokuish:
 	@echo "  sudo docker rmi gliderlabs/herokuish" >> /tmp/tmp/post-install
 	@echo "fi" >> /tmp/tmp/post-install
 	@echo "echo 'Importing herokuish into docker (around 5 minutes)'" >> /tmp/tmp/post-install
-	@echo "sudo docker build --build-args http_proxy=$http_proxy --build-args https_proxy=$https_proxy -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${http_proxy+x} ]; then BUILDARGS="--build-args http_proxy=$http_proxy"; fi' >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${https_proxy+x} ]; then BUILDARGS="$BUILDARGS --build-args https_proxy=$https_proxy"; fi' >> /tmp/tmp/post-install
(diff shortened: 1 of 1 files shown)


deb.mk
```diff
@@ -92,7 +92,7 @@ deb-herokuish:
 	@echo "  sudo docker rmi gliderlabs/herokuish" >> /tmp/tmp/post-install
 	@echo "fi" >> /tmp/tmp/post-install
 	@echo "echo 'Importing herokuish into docker (around 5 minutes)'" >> /tmp/tmp/post-install
-	@echo "sudo docker build -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
+	@echo "sudo docker build --build-args http_proxy=$http_proxy --build-args https_proxy=$https_proxy -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
```
josegonzalez: Can we instead wrap this in an if statement so we don't add them when they don't exist?

deb.mk
```diff
@@ -92,7 +92,10 @@ deb-herokuish:
 	@echo "  sudo docker rmi gliderlabs/herokuish" >> /tmp/tmp/post-install
 	@echo "fi" >> /tmp/tmp/post-install
 	@echo "echo 'Importing herokuish into docker (around 5 minutes)'" >> /tmp/tmp/post-install
-	@echo "sudo docker build --build-args http_proxy=$http_proxy --build-args https_proxy=$https_proxy -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${http_proxy+x} ]; then BUILDARGS="--build-args http_proxy=$http_proxy"; fi' >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${https_proxy+x} ]; then BUILDARGS="$BUILDARGS --build-args https_proxy=$https_proxy"; fi' >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${BUILDARGS+x} ]; then echo Adding proxy settings to docker build: $BUILDARGS; fi' >> /tmp/tmp/post-install
```
josegonzalez: The message should be quoted.

deb.mk
```diff
@@ -92,7 +92,10 @@ deb-herokuish:
 	@echo "  sudo docker rmi gliderlabs/herokuish" >> /tmp/tmp/post-install
 	@echo "fi" >> /tmp/tmp/post-install
 	@echo "echo 'Importing herokuish into docker (around 5 minutes)'" >> /tmp/tmp/post-install
-	@echo "sudo docker build --build-args http_proxy=$http_proxy --build-args https_proxy=$https_proxy -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${http_proxy+x} ]; then BUILDARGS="--build-args http_proxy=$http_proxy"; fi' >> /tmp/tmp/post-install
```
josegonzalez: Can you set a default for `BUILD_ARGS` before this? Also, please rename it `BUILD_ARGS.`

josegonzalez: Since we're using bash for the `postinst`, we can use double brackets for our guards, `[[ ]]`.

deb.mk
```diff
@@ -92,7 +92,10 @@ deb-herokuish:
...
+	@echo 'if [ ! -z ${https_proxy+x} ]; then BUILDARGS="$BUILDARGS --build-args https_proxy=$https_proxy"; fi' >> /tmp/tmp/post-install
+	@echo 'if [ ! -z ${BUILDARGS+x} ]; then echo Adding proxy settings to docker build: $BUILDARGS; fi' >> /tmp/tmp/post-install
+	@echo "sudo docker build $BUILDARGS -t gliderlabs/herokuish /var/lib/herokuish 1> /dev/null" >> /tmp/tmp/post-install
```
josegonzalez: You need to escape the dollar sign or use single quotes.

josegonzalez:
Is there any reason you don't just specify these in your global docker config?

bevand10:
@josegonzalez  - because it seems that while the 'docker run' commands obey any envvars (or particularly in the case of systemd setups such as Ubuntu 16, an override file defined following the advice at https://docs.docker.com/engine/admin/systemd/ - specifically the "HTTP proxy" section), the 'docker build' command does not. See https://github.com/docker/docker/issues/24697 for some discussions about the bug.

PR #2587, opened 2017-01-26
Allow specifying the deploy branch via DOKKU_DEPLOY_BRANCH

### LINKED ISSUE #2494: Allow configuring the default push branch
By default, we only allow triggering a build when the [master](https://github.com/dokku/dokku/blob/c3f39fe512b84d65322bcf1c9b066f70bec95ff8/plugins/git/functions#L53) branch is pushed. This is pretty common for PaaS systems, but you might want to configure it for a variety of reasons:

- You hate the word master, and whatever connotations it brings up (there are plenty of reasons, lets not get into that).
- You'd like to force a certain app to build for a specific branch only (like, say, a staging app, or only allow prod branch to build).
- You need to enforce that the production server only builds the production branch (which is somehow protected in your build process).
- You are weird and just like configuring all the knobs.

We currently allow alternative branches to build via the following methods:

- `git push OTHER_BRANCH:master`
- A custom plugin (blech)

Would be cool to allow configuring what branch is deployable by default. We can do this via something like `DOKKU_DEPLOY_BRANCH`. The default would be `master`, as that is least likely to break existing workflows. This can be specified globally or on an app-by-app basis:

```shell
dokku config:set --global DOKKU_DEPLOY_BRANCH=other-branch
dokku config:set APP DOKKU_DEPLOY_BRANCH=other-branch
```

Only a single push branch is specifiable, and we should bork deploys if that specified branch isn't a valid git identifier (is there a regex for that somewhere?).

You'll almost certainly want to do this *before* the application is first deployed, or you're gonna have a bad time (I don't think we want to pre-fill that branch with whatever is currently deployed).

This change *must* have documentation accompanying it [here](http://dokku.viewdocs.io/dokku/deployment/application-deployment/#deploying-non-master-branch), and should probably also have a test associated with it. The test can just do the following:

```shell
dokku config:set APP DOKKU_DEPLOY_BRANCH=other-branch
git push dokku master:other-branch
```

While this change might have limited global use-case, I think it's fairly small, lets us be a bit more flexible, doesn't break existing use cases, and therefore seems like a reasonable change to implement. If someone wants to contribute to the core, this is a pretty good issue to poke at :).

josegonzalez:
Closes #2494

### DIFF
--- a/plugins/git/functions
+++ b/plugins/git/functions
@@ -55,16 +55,39 @@ suppress_output() {
   return 0
 }
 
+git_deploy_branch() {
+  declare desc="retrieve the deploy branch for a given application"
+  local cmd="git-hook"
+  local APP="$1"
+
+  local DOKKU_DEPLOY_BRANCH="$(config_get "$APP" DOKKU_DEPLOY_BRANCH || true)"
+  local DOKKU_GLOBAL_DEPLOY_BRANCH="$(config_get --global DOKKU_DEPLOY_BRANCH || true)"
+  if [[ -n "$DOKKU_DEPLOY_BRANCH" ]]; then
+    echo "$DOKKU_DEPLOY_BRANCH"
+  elif [[ -n "$DOKKU_GLOBAL_DEPLOY_BRANCH" ]]; then
+    echo "$DOKKU_GLOBAL_DEPLOY_BRANCH"
+  else
+    echo "master"
+  fi
+}
+
 git_hook_cmd() {
   declare desc="kick off receive-app trigger from git prereceive hook"
   local cmd="git-hook"
-  local APP="$2"
+  local APP="$2"  
+  local DOKKU_DEPLOY_BRANCH="$(git_deploy_branch "$APP")"
+
+  if ! git check-ref-format --branch "$DOKKU_DEPLOY_BRANCH" 2> /dev/null; then
+    echo $'\e[1G\e[K'"-----> WARNING: Invalid branch name '$DOKKU_DEPLOY_BRANCH' specified via DOKKU_DEPLOY_BRANCH."
+    echo $'\e[1G\e[K'"-----> For more details, please see the man page for 'git-check-ref-format.'"
+    return
+  fi
 
   local oldrev newrev refname
   while read -r oldrev newrev refname; do
     # Only run this script for the master branch. You can remove this
     # if block if you wish to run it for others as well.
-    if [[ $refname = "refs/heads/master" ]]; then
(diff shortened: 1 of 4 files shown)


plugins/git/functions
```diff
@@ -55,16 +55,39 @@ suppress_output() {
...
+  declare desc="retrieve the deploy branch for a given application"
+  local cmd="git-hook"
+  local APP="$1"
+
+  local DOKKU_DEPLOY_BRANCH="$(config_get "$APP" DOKKU_DEPLOY_BRANCH || true)"
+  local DOKKU_GLOBAL_DEPLOY_BRANCH="$(config_get --global DOKKU_DEPLOY_BRANCH || true)"
+  if [[ -n "$DOKKU_DEPLOY_BRANCH" ]]; then
+    echo "$DOKKU_DEPLOY_BRANCH"
+  elif [[ -n "$DOKKU_GLOBAL_DEPLOY_BRANCH" ]]; then
+    echo "$DOKKU_GLOBAL_DEPLOY_BRANCH"
```
michaelshobbs: Is a "global" deploy branch necessary?

tests/unit/20_config.bats
```diff
@@ -145,3 +145,17 @@ teardown() {
   echo "status: "$status
   assert_success
 }
+
+@test "(config) deploy specific DOKKU_DEPLOY_BRANCH" {
+  run ssh dokku@dokku.me config:set --global DOKKU_DEPLOY_BRANCH=global-branch
```
michaelshobbs: This isn't testing anything new AFAICT. If anything this should be a deploy test.

plugins/git/functions
josegonzalez: Just wanted to make it possible to modify this per app as well as globally.

tests/unit/20_config.bats
josegonzalez: What do you mean, its not testing anything new? We test whether the deploy of the `global-branch` succeeds successfully. Is there a better way to do this?

michaelshobbs: Oh I missed the call to deploy_app on the line below. Nothing to see here.

plugins/git/functions
michaelshobbs: I can't see the use case of a global deploy branch but what do I know?

josegonzalez: The global bit is inspired by a tweet from some users that didn't want to push `master` because of the connotations of the word. I figure being able to adjust it globally makes most sense.

Also, this makes it so we can specify a `production` box :)

PR #2603, opened 2017-02-11
use temporary file for `ssh-keygen -lf` commands

sebble:
See also #1076. Fails lint SC2064 (https://circleci.com/gh/sebble/dokku/2) deliberately.

Early expansion of the `$tmp_key_file` is required for cleanup to work outside of the scope of this function.

### DIFF
--- a/plugins/ssh-keys/functions
+++ b/plugins/ssh-keys/functions
@@ -13,10 +13,12 @@ list_ssh_keys() {
 verify_ssh_key_file() {
   declare desc="Test that public key is valid"
   [[ -s ${DOKKU_ROOT}/.ssh/authorized_keys ]] || dokku_log_fail "No public keys found."
-  local key line=0
+  local key line=0 tmp_key_file="$(mktemp)"
+  trap 'rm -rf "$tmp_key_file" > /dev/null' RETURN INT TERM EXIT
   while read -r key; do
     line=$((line + 1))
-    ssh-keygen -l -f /dev/stdin <<<"$key" &>/dev/null || dokku_log_fail "${DOKKU_ROOT}/.ssh/authorized_keys line $line failed ssh-keygen check."
+    echo "$key" >"$tmp_key_file"
+    ssh-keygen -l -f "$tmp_key_file" &>/dev/null || dokku_log_fail "${DOKKU_ROOT}/.ssh/authorized_keys line $line failed ssh-keygen check."
   done <"${DOKKU_ROOT}/.ssh/authorized_keys"
 }
 
--- a/plugins/ssh-keys/subcommands/add
+++ b/plugins/ssh-keys/subcommands/add
@@ -8,14 +8,16 @@ add_keys() {
   declare desc="add a new key via sshcommand"
   local cmd="ssh-keys:add"
   shift
-  local name="$1" key_file="$2" key_contents key_from_pipe
+  local name="$1" key_file="$2" key_contents key_from_pipe tmp_key_file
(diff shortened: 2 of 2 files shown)


plugins/ssh-keys/functions
```diff
@@ -12,10 +12,12 @@ list_ssh_keys() {
 verify_ssh_key_file() {
   declare desc="Test that public key is valid"
   [[ -s ${DOKKU_ROOT}/.ssh/authorized_keys ]] || dokku_log_fail "No public keys found."
-  local key line=0
+  local key line=0 tmp_key_file="$(mktemp)"
+  trap "rm \"$tmp_key_file\"" EXIT
```
josegonzalez: I *think* you can avoid the extra quotes and it'll work as desired, right @michaelshobbs ?

sebble: In many cases, yes. But a quick test shows that white-space in the `$tmp_key_file` will be interpreted incorrectly:

```
main() {
    t="/tmp/tmp file"
    touch "$t"
    trap "rm ""$t" EXIT
}
main
```
results in :
```
rm: cannot remove '/tmp/tmp': No such file or directory
rm: cannot remove 'file': No such file or directory
```

`"rm ""$t"` did not help and the lint fails with various single quotes as well, while `trap "rm "\""$t"\" EXIT` produces `SC2140: The double quotes around this do nothing. Remove or escape them.`

The only solution may be to define the trap clean as another function.

Thanks for your help.

michaelshobbs: What about this?
```
trap "rm \'$t\' EXIT
```

michaelshobbs: Also if a solution works and it's on the edge case of the lint rule, we can disable it for that particular line.

sebble:
I have added an exception by way of an inline comment:

```
# shellcheck disable=2064
```

This seems to be unavoidable as using either single or double quotes inside another and "SC2064" still flags it as an error.

```
In ./plugins/ssh-keys/functions line 16:
  trap  "rm \'$tmp_key_file\'" EXIT
              ^-- SC2064: Use single quotes, otherwise this expands now rather than when signalled.
```

I also tried using a predefined function to remove the temp file, but the variable must still be expanded at initialisation of the trap (to avoid storing extra global variables) with exactly the same linting issues.

It is a shame that bash heredoc cannot be used in all cases.

For full disclosure, the original issue happens only inside Debian and CentOS containers, not Ubuntu.  I was curious about the source of this issue but so far have only found was closed and fixed a long time ago (https://github.com/docker/docker/issues/2067) and I have not yet found any other information.

Unfortunately, the CircleCI tests are still failing with `not ok 3 (apps) apps:rename`. I'm not sure if this is my fault as the same failure occurs on the commit immediately before any modification by myself (https://circleci.com/gh/sebble/dokku/1).  I have had difficulty replicating the proper test environment locally.

I have also merged the current master with no change to the test success.  Maybe I need to configure my CircleCI account?

josegonzalez:
@joshmanders Mind taking a look at the failed deploy and seeing how we can fix this PR? Would love to get it merged :)

joshmanders:
looks like it's failing because the ssh-key plugin is trying to delete the temp file after importing it, but the changes don't utilize the file anymore or something.

josegonzalez:
Closing as there is an updated pull request. Apologies for this being so late to merge, hopefully I can resolve the issues with shellcheck in the new PR.

Thanks for the initial filing!

PR #2622, opened 2017-02-23
Allow file names with multiple dots in certs:add command

raphaklaus:
Certificate files like mydomain.com.crt and mydomain.com.key will be better interpreted.


josegonzalez:
Can you update the tests for this? Just so we don't accidentally break it later.

raphaklaus:
Okay, added this scenario too.

tests/unit/10_certs.bats
```diff
@@ -33,6 +33,13 @@ teardown() {
   assert_success
 }
 
+@test "(certs) certs:add with multiple dots in the filename" {
+  run bash -c "dokku certs:add $TEST_APP $BATS_TMPDIR/tls/domain.com.crt $BATS_TMPDIR/tls/domain.com.key"
```
josegonzalez: Haven't checked, but these files do exist, right?

raphaklaus: Where does files like server.crt and server.key are in this repository, GitHub search sucks :\

josegonzalez: They are from [here](https://github.com/raphaklaus/dokku/tree/b0777a44cb7e1a4d8d30934798f95cd500a9d7ba/tests/unit) and the tar file is extracted [here](https://github.com/dokku/dokku/pull/2622/files#diff-3d75297c33bf18c21085b512d18b3d7dR5).

raphaklaus: Done. But something broke the tests... Not sure if is something I did. Will check.

raphaklaus: Okay, it's done @josegonzalez

PR #2641, opened 2017-03-06
Added better default nginx error

chiedo:
This adds a much nicer nginx error page if the proxy has issues for whatever reason...

The downside is that I had to verbosely list all the possible errors.

The other downside is that I couldn't find a way to return the same error that was initially caused so the error page defaults to a 500 page... That's not ideal... I'm hoping you have input on how to return the same error.

### DIFF
--- a/docs/deployment/user-management.md
+++ b/docs/deployment/user-management.md
@@ -73,3 +73,16 @@ dokku ssh-keys:remove KEY_NAME
 ## Scoping commands to specific users
 
 Keys are given unique names, which can be used in conjunction with the [user-auth](/dokku/development/plugin-triggers/#user-auth) plugin trigger to handle command authorization. Please see the documentation on that trigger for more information.
+
+## Granting other Unix user accounts Dokku access
+
+Any Unix user account which belongs to the 'sudo' Unix group can run Dokku.  However, you may want to give them Dokku access but not full sudo privileges.
+
+To allow other Unix user accounts to be able to run Dokku commands, without giving them full sudo access, modify your sudoers configuration.
+
+Use `visudo /etc/sudoers.d/dokku-users`, or `visudo /etc/sudoers` to add the following line:
+
+```
+%dokku ALL=(ALL:ALL) NOPASSWD:SETENV: /usr/bin/dokku
+```
+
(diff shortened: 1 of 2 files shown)


josegonzalez:
Screenshot?

> The other downside is that I couldn't find a way to return the same error that was initially caused so the error page defaults to a 500 page... That's not ideal... I'm hoping you have input on how to return the same error.

What do you mean by that?

chiedo:
Good question. It's not something a screenshot can communicate. What I mean is that, I have to explicitly return a 500 error regardless of whether it was a 404, 502, etc.

See this line
https://github.com/dokku/dokku/pull/2641/commits/bc26370267772f08ada8f262578cd3dfde45ae3f#diff-3459fed174d925e80eeab34ce741fc8cR40

josegonzalez:
No, I meant can I get a screenshot of what it looks like.

chiedo:
screen shot 2017-03-06 at 2 36 24 pm

josegonzalez:
Some comments:

- Can we use a common font-stack instead of google?
- That message should be slightly more professional than "Uh-oh"
- Can we template out the error pages to somewhere in `/etc/nginx/dokku-errors`, one for each error, and then reference that?

chiedo:
**Can we use a common font-stack instead of google?** - Yep. Just let me know which one you'd like to be used.

**That message should be slightly more professional than "Uh-oh"** - Let me know what you'd like it to be instead and I'll update it!

**Can we template out the error pages to somewhere in /etc/nginx/dokku-errors, one for each error, and then reference that?** - I'm not familiar with how the dokku install process works from a technical perspective. Let me know where to add the files in the git repo and I'll do so. I'll create a blanket 500 level error page, a blanket 400 level error page, and then a 404 error page.

josegonzalez:
- Whatever looks good, I'll leave that to you.
- Ummm, idk, but it definitely shouldn't say "Uh-oh".
- A plugin can have an `install` plugin trigger that can write files anywhere - they are invoked as a sudo user. In this case, probably [here](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/install).

josegonzalez:
@chiedo why did you close the PR?

chiedo:
See https://github.com/dokku/dokku/pull/2642 wanted to keep the commits clean for you and squashed

josegonzalez:
You can force push to your existing branch, and I also don't fret too much about PR history.

chiedo:
Awesome. I was on master. Mistake number 1. lol

PR #2652, opened 2017-03-14
Installing key

R3ido101:
I have made this pull request to fix the key issue for ubuntu os.

Thanks

R3ido101

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -125,10 +125,7 @@ install-dokku-from-deb-package() {
   fi
   echo "Installing Docker Key! (If on Ubuntu)"
   curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-  sudo add-apt-repository \
-   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
-   $(lsb_release -cs) \
-   stable"
+  wget -nv -O - https://get.docker.com/ | sh
 
   echo "--> Installing dokku"
   wget -nv -O - https://packagecloud.io/gpg.key | apt-key add -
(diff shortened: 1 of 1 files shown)


josegonzalez:
I'm not familiar with this being a bug. The docker bash installer should install the key on its own. If anything, this is a bug with docker, not with Dokku.

Can you show log output that describes this issue?

R3ido101:
taken straight out of build 3600 


Fetched 9,208 kB in 2min 0s (76.2 kB/s)
W: GPG error: https://apt.dockerproject.org ubuntu-precise Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY F76221572C52609D
W: Failed to fetch http://toolbelt.heroku.com/ubuntu/./en_US  Could not connect to heroku-toolbelt.s3.amazonaws.com:80 (52.216.224.48), connection timed out [IP: 54.243.165.7 80]

josegonzalez:
We don't support Ubuntu Precise 12.04, just Trusty and Up.

R3ido101:
Yes I know but you will still need that.

josegonzalez:
You got that error on a precise box, not a trusty box, so I don't think thats true at all.

R3ido101:
well i ran it on 16.04 and got the same exact error so ya and please read! https://docs.docker.com/engine/installation/linux/ubuntu/#prerequisites

josegonzalez:
I just ran our bootstrap script on an Ubuntu 14.04 box and did not receive any gpg error. Running on a 16.04 box now.

josegonzalez:
Wasn't able to replicate this on 16.04 either.

A few things:

- The error you are referencing in our build output is transient. CircleCI has it's own issues with it's infra, so sometimes running apt-get update fails. I typically just rebuild and move on.
- We do not run the bootstrap.sh file in our automated testing, so this doesn't do anything.
- I can read, and even if this was an issue in our setup - it's not - this would be an issue with the docker bootstrap file - it isn't.

Closing this pull request as it doesn't actually fix anything. Thanks for your contribution though.

PR #2677, opened 2017-03-23
Prefer HTTP2 to SPDY in nginx-vhosts

callahad:
1.  SPDY has been deprecated and removed from modern browsers:
    http://caniuse.com/#feat=spdy

2.  HTTP/2 is present in all modern browsers:
    http://caniuse.com/#feat=http2

3.  SPDY and HTTP/2 are mutually exclusive in port-based `listen` directives:
    http://nginx.org/en/docs/http/ngx_http_core_module.html#listen

Caniuse suggests that 74% of global users have a browser that supports HTTP/2, while only 27% have a browser that supports SPDY. Only IE11 and Safari support both.

Therefore, when both are available, Dokku should choose HTTP/2 to maximize the benefit to users.

### DIFF
--- a/plugins/nginx-vhosts/templates/nginx.conf.sigil
+++ b/plugins/nginx-vhosts/templates/nginx.conf.sigil
@@ -38,8 +38,8 @@ server {
 }
 {{ else if eq $scheme "https"}}
 server {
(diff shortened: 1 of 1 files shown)


callahad:
Re-pushing to kick circle-ci. Seems like one of the containers died during build.

PR #2678, opened 2017-03-23
Update the default cipher suite

callahad:
This is the Modern preset from: https://mozilla.github.io/server-side-tls/ssl-config-generator/

I dropped the `ssl_session_tickets off;` directive from the recommendations since it's only supported by Nginx 1.5.9 and later, and Ubuntu 14.04 LTS shipped with Nginx 1.4.6. If we don't support versions of Nginx that old, then we should add that directive back into the configuration.

The other settings and cipher suites should all be supported on the versions of Nginx and OpenSSL shipping with Ubuntu 16.04 LTS, Debian 8.2, and CentOS 7.

Similarly, the cipher suites should be available on all platforms going back to IE11 on Windows 7, Safari 9, and Android 5 (Lollipop).


callahad:
I'm not quite sure what to do to ensure that users actually use the new cipher suites, since `/etc/nginx/conf.d/dokku` is only written if it doesn't exist. It's never re-initialized or updated. 😕

josegonzalez:
Note: this is the cipher suite you need to use to be PCI compliant. Not sure if we want to default for that, as it drops a ton of older IE stuff (like IE10 iirc):

```
  ssl_ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:RSA+AESGCM:RSA+AES:!aNULL:!MD5:!DSS;
```

josegonzalez:
I modified this PR a bit to get us to be PCI compliant by default, as well as require Nginx 1.8.0+. That should do it as far as getting everything stable and ready.

josegonzalez:
Note: the `install` hook is performed on [upgrade](https://github.com/dokku/dokku/blob/master/debian/postinst#L76), so this should work as expected.

PR #2679, opened 2017-03-23
Remove unnecessary Linode-specific instructions

### LINKED ISSUE #2675: Deprecate special Linode instructions?
As of version 1.12, Docker will happily use Linux 4.x's `overlayfs` filesystem with its [overlay2 driver](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/), which is evidently a natural successor to AUFS. Linode's kernel supports `overlayfs`.

As such, the warnings at [docs/getting-started/install/linode.md](https://github.com/dokku/dokku/blob/9ee6c7aab630eaccf20d9858b10302e79e467519/docs/getting-started/install/linode.md) and inside [bootstrap.sh](https://github.com/dokku/dokku/blob/ab51c3b4b451e750432954c6dfc20cb5c04665e5/bootstrap.sh#L120-L125) can probably be removed.

callahad:
Modern versions of Docker work just fine on Linode's default kernel thanks to the `overlay2` storage driver.

Closes #2675


josegonzalez:
Note: for documentation updates, you can just add `[ci skip]` to the message to skip tests.

PR #2680, opened 2017-03-24
missing `dokku` on `shell command`

flashios09:


docs/deployment/user-management.md
```diff
@@ -53,7 +53,7 @@ b7:76:27:4f:30:90:21:ae:d4:1e:70:20:35:3f:06:d6
 Admin users and root can also add keys remotely:
 
 ```shell
-cat ~/.ssh/id_rsa.pub | ssh dokku@dokku.me ssh-keys:add KEY_NAME
+cat ~/.ssh/id_rsa.pub | ssh dokku@dokku.me dokku ssh-keys:add KEY_NAME
```
josegonzalez: This shouldn't be necessary as `sshcommand` will autoprefix anything sent to the dokku user via ssh with the `dokku` bin.

flashios09: unfortunately this doesn't work for me
```fish
fish: Unknown command 'ssh-keys:add flashios09-iMac'
fish: ssh-keys:add flashios09-iMac
```

josegonzalez: Does it work with the prefix? What version of dokku are you using?

flashios09: this command works for me:
```
cat ~/.ssh/id_rsa.pub | ssh dokku@dokku.me dokku ssh-keys:add KEY_NAME
```
I'm using dokku 0.8.2

josegonzalez: @michaelshobbs this shouldn't be possible, right?

michaelshobbs: depends on what his `authorized_keys` file already looks like

josegonzalez: @flashios09 whats in your `authorized_keys` file?

josegonzalez:
This was because the user for ssh wasn't `dokku`, but their root user.

PR #2684, opened 2017-03-25
feat: Allow triggering a release via the `contrib/release` script

josegonzalez:
To run it:

```
# where RELEASE_TYPE can be [major, minor, patch, betafish]
contrib/release RELEASE_TYPE
```

- Requires docker, git, and package_cloud gems
- Requires the PACKAGECLOUD_API_TOKEN environment variable
- Builds both a deb and rpm, pushing both to packagecloud
- Upgrades the golang build environment to 1.7.5
- Combines both build Dockerfiles into a single, generic env
- Fixes issue where internal-functions file for tar plugin was not executable
- Ensures that built plugins have a callback to clean up their source files before packaging
- Adds support for “betafish” releases, which are untested versions of Dokku that are available via a separate packagecloud repository.

> This doesn't yet create a tagged release with release notes, but I'll work on that in a future iteration.


josegonzalez:
Long term, we can use this to generate a deb package for all pull requests and potentially install dokku within a docker container, to validate that the deb packages can in fact still be built and contain the functionality we want. The betafish repository will serve to hold these HIGHLY unsupported packages.

HISTORY.md
```diff
@@ -1091,7 +1091,7 @@ Thanks to the *many* contributors for making this release our best release so fa
 - #1473: @josegonzalez Handle crashing containers by using restart=on-failure policy
 - #1476: @michaelshobbs Support static nginx port when deploying without an application VHOST
 - #1476: nginx proxy without VHOST
-- #1477: @arthurschreiber Support removing config variables that contain `\n`.
+- #1477: @arthurschreiber Support removing config variables containing newlines.
```
michaelshobbs: Curious why this change was necessary.

contrib/build.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+RUN wget -qO /tmp/go${GOLANG_VERSION}.linux-amd64.tar.gz https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz \
+    && tar -C /usr/local -xzf /tmp/go${GOLANG_VERSION}.linux-amd64.tar.gz
+
+ARG WORKDIR=/go/src/github.com/dokku/dokku
+
+WORKDIR ${WORKDIR}
+
+COPY Makefile ${WORKDIR}/
+
```
michaelshobbs: nitpick: group same directives. i.e. no extra whitespace

deb.mk
```diff
@@ -140,16 +140,16 @@ endif
...
+	cp debian/lintian-overrides /tmp/build/usr/share/lintian/overrides/dokku
 ifdef DOKKU_GIT_REV
 	echo "$(DOKKU_GIT_REV)" > /tmp/build/var/lib/dokku/GIT_REV
 else
 	git rev-parse HEAD > /tmp/build/var/lib/dokku/GIT_REV
 endif
-	sed -i "s/^Version: .*/Version: `cat /tmp/build/var/lib/dokku/STABLE_VERSION`/g" /tmp/build/DEBIAN/control
-	dpkg-deb --build /tmp/build "/tmp/dokku_`cat /tmp/build/var/lib/dokku/STABLE_VERSION`_$(DOKKU_ARCHITECTURE).deb"
-	lintian "/tmp/dokku_`cat /tmp/build/var/lib/dokku/STABLE_VERSION`_$(DOKKU_ARCHITECTURE).deb"
+	sed -i.bak "s/^Version: .*/Version: `cat /tmp/build/var/lib/dokku/STABLE_VERSION`/g" /tmp/build/DEBIAN/control && rm /tmp/build/DEBIAN/control.bak
```
michaelshobbs: Not sure if you actually wanted to save the `.bak` for troubleshooting but if not, you can use `-i ""`

HISTORY.md
josegonzalez: The thing that writes out history was rewriting these as actual newlines.

deb.mk
josegonzalez: Was testing this on OS X, which doesn't have the raw `-i` flag.

michaelshobbs: Right. I'm pretty sure you can use `-i ""` (note: the double quotes) to avoid the backup file being written.

josegonzalez: *shrug* this is the pattern we've followed elsewhere.

michaelshobbs: That's not necessarily a reason to leave it as is though. In any case, if you're comfortable with maintaining the pattern, then that's cool with me.

josegonzalez: Yep I'm happy with it.

PR #2691, opened 2017-03-31
Fix package building when golang binaries are available

josegonzalez:


plugins/repo/Makefile
```diff
@@ -23,4 +23,4 @@ clean:
 	rm -rf commands subcommands
 
 src-clean:
-	rm -rf .gitignore src vendor
+	rm -rf .gitignore src vendor Makefile
```
michaelshobbs: this is all happening in a tmp dir, correct?

deb.mk
```diff
@@ -124,8 +124,12 @@ deb-dokku:
 	cp dokku /tmp/build/usr/bin
 	cp LICENSE /tmp/build/usr/share/doc/dokku/copyright
 	find . -name ".DS_Store" -depth -exec rm {} \;
+	$(MAKE) go-build
+	cp common.mk /tmp/build/var/lib/dokku/core-plugins/common.mk
 	cp -r plugins/* /tmp/build/var/lib/dokku/core-plugins/available
+	find plugins/ -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | while read plugin; do cd /tmp/build/var/lib/dokku/core-plugins/available/$$plugin && if [ -e Makefile ]; then $(MAKE) src-clean; fi; done
```
josegonzalez: @michaelshobbs Yeah it happens here, in the `/tmp` dir.

PR #2694, opened 2017-04-03
Add documentation note re: git-pre-pull vs. auth

scjody:
As discussed with @josegonzalez on Slack - it's tempting but incorrect to use `git-pre-pull` for authenticating git access. Instead `user-auth` should be used, so document this.

### DIFF
--- a/docs/development/plugin-triggers.md
+++ b/docs/development/plugin-triggers.md
@@ -321,6 +321,10 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 # TODO
 ```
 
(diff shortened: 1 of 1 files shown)


docs/development/plugin-triggers.md
```diff
@@ -321,6 +321,10 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 # TODO
 ```
 
+Note that the `git-pre-pull` trigger should _not_ be used for authentication
```
josegonzalez: Should be:

```
> WARNING: MESSAGE_HERE
```

So a blockquote etc.

scjody:
@josegonzalez Is this formatting OK?

PR #2697, opened 2017-04-04
Restore installer note regarding AUFS on Linode

callahad:
See discussion in: https://github.com/dokku/dokku/issues/2675
Upstream Docker bug: https://github.com/docker/docker/issues/32361

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -117,6 +117,12 @@ install-dokku-from-deb-package() {
   apt-get -qq -y install apt-transport-https
 
   echo "--> Installing docker"
+  if uname -r | grep -q linode; then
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -117,6 +117,12 @@ install-dokku-from-deb-package() {
   apt-get -qq -y install apt-transport-https
 
   echo "--> Installing docker"
+  if uname -r | grep -q linode; then
+    echo "--> NOTE: Using Linode? Docker might complain about missing AUFS support."
```
josegonzalez: might => may

callahad: Pfft, you let me get away with that [two years ago](https://github.com/callahad/dokku/commit/81523b3bd01910d485b20dba255e298d7cff0a33) ;)

PR #2699, opened 2017-04-05
Use git worktree in git_build_app_repo() for git >= 2.11

### LINKED ISSUE #2512: git version compatibility
Description of problem:
We're running into issues with the newly released Git version (1:2.11.0-0ppa3~ubuntu14.04.1)
The first push of a project seems to incorrectly create the tag in the repo, and afterwards every other push will fail with the omnimous `! [remote rejected] master -> master (pre-receive hook declined)`

This is probably going to be related to the #2263 issue... (we are running the latest 16.10 after inline upgrades from 14.04 to 16.04 to 16.10) and only started getting these errors with the very latest git version... 

As i couldn't find anyone else with these types of issues...I'm posting it here.. :)

If someone is running into this aswell the way to `resolve` this is to downgrade to a GIT version of (in our case) 1:1.9.1-1

mbreit:
Git 2.11.0 breaks the git clone approach in git_build_app_repo() in
https://github.com/git/git/commit/25ab004c53cdcfea485e5bf437aeaa74df47196d

git worktree is a much simpler approach but only works for git >= 2.5
so it is not available in Ubuntu 14.04.

Fixes #2512.

### DIFF
--- a/plugins/git/functions
+++ b/plugins/git/functions
@@ -4,6 +4,23 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
 source "$PLUGIN_AVAILABLE_PATH/apps/functions"
 source "$PLUGIN_AVAILABLE_PATH/config/functions"
 
+use_git_worktree() {
+    declare desc="detects whether to use git worktree"
+    local GIT_VERSION MAJOR_VERSION MINOR_VERSION
+
+    GIT_VERSION=$(git --version | awk '{split($0,a," "); print a[3]}')
+    MAJOR_VERSION=$(echo "$GIT_VERSION" | awk '{split($0,a,"."); print a[1]}')
+    MINOR_VERSION=$(echo "$GIT_VERSION" | awk '{split($0,a,"."); print a[2]}')
+
+    if [[ "$MAJOR_VERSION" -ge "3" ]]; then
+        return 0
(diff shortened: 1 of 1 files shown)


plugins/git/functions
```diff
@@ -4,6 +4,23 @@ source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
...
+    declare desc="detects whether to use git worktree"
+    local GIT_VERSION MAJOR_VERSION MINOR_VERSION
+
+    GIT_VERSION=$(git --version | awk '{split($0,a," "); print a[3]}')
+    MAJOR_VERSION=$(echo "$GIT_VERSION" | awk '{split($0,a,"."); print a[1]}')
+    MINOR_VERSION=$(echo "$GIT_VERSION" | awk '{split($0,a,"."); print a[2]}')
+
+    if [[ "$MAJOR_VERSION" -ge "3" ]]; then
+        return 0
+    elif [[ "$MAJOR_VERSION" -eq "2" ]] && [[ "$MINOR_VERSION" -ge "11" ]]; then
```
josegonzalez: Should be used for `2.5+`, right?

mbreit: I did not have the time to test this with git 2.5, so I concentrated on git 2.11.  This way I am confident that we don't break any previously working git versions.  And up to git 2.10 should work with the old clone approach.
I can change it to 2.5 of course, but then I would recommend some more testing...

PR #2710, opened 2017-04-10
Quiet output for git-related commands

josegonzalez:


plugins/git/functions
```diff
@@ -60,7 +60,7 @@ git_build_app_repo() {
   suppress_output git submodule update --init --recursive
   find . -name .git -prune -exec rm -rf {} \; > /dev/null
 
-  plugn trigger post-extract "$APP" "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" "$REV"
+  plugn trigger post-extract "$APP" "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" "$REV" | sed -u "s/^/"$'\e[1G'"/"
```
josegonzalez: This removes the `remote:` prefix that we normally remove when it's just triggered by `pre-receive-app`.

plugins/git/functions
```diff
@@ -107,7 +107,7 @@ git_hook_cmd() {
   local APP="$2"  
   local DOKKU_DEPLOY_BRANCH="$(git_deploy_branch "$APP")"
 
-  if ! git check-ref-format --branch "$DOKKU_DEPLOY_BRANCH" 2> /dev/null; then
+  if ! git check-ref-format --branch "$DOKKU_DEPLOY_BRANCH" > /dev/null 2>&1; then
```
josegonzalez: This outputs the branch name, though we actually don't care about the output, so long as the command returns zero (meaning its a valid branch name).

PR #2715, opened 2017-04-14
Use urls that are linkable on github

josegonzalez:
This should relieve us from getting random pull requests on doc links because people don't use the docs site.

I made the requisite changes in viewdocs to support this functionality. https://github.com/progrium/viewdocs/commit/96358856003bb364738b01139cd44712b85f8c94


xuanduc987:
Links is now unusable at http://dokku.viewdocs.io

josegonzalez:
@xuanduc987 fixed the issue (was upstream in viewdocs, I missed a case where we didn't cleanup doc links).

PR #2729, opened 2017-04-18
Extract some scheduler-related functionality into a separate plugin

josegonzalez:
I'm not sure how to make it so we can "inherit" schedulers. For instance, you'll probably want anything build-related to run locally, but maybe you don't in some cases. Thats why it has some weird logic around checking if you should use a particular scheduler.


michaelshobbs:
let me know when you get tests passing and i'll review

josegonzalez:
Closing this for now. I think with the networking changes, this was maybe a bit premature to start working on.

Issue #2711, opened 2017-04-11
apps:clone fails when trying to copy the cache dir


elia:
**Description of problem:**

`dokku apps:clone` fails when trying to copy the cache dir

output with trace on:

```
⤑ bin/dokku apps:clone master master-copy
Creating master-copy... done
✂️
+ local phase_file_path=/home/dokku/master-copy2/DOCKER_OPTIONS_DEPLOY
+ [[ -f /home/dokku/master-copy2/DOCKER_OPTIONS_DEPLOY ]]
+ touch /home/dokku/master-copy2/DOCKER_OPTIONS_DEPLOY
+ echo --restart=on-failure:10
+ local all_phase_options=--restart=on-failure:10
+ echo -e --restart=on-failure:10
+ sed '/^$/d'
+ sort -u
+ cp -a /home/dokku/master/. /home/dokku/master-copy2
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/cache': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/specifications': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/bin': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/extensions': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/build_info': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/doc': Permission denied
cp: cannot access '/home/dokku/master/./cache/cedar-14/vendor/bundle/ruby/2.3.0/gems': Permission denied
Connection to XXXXXXXXX closed.
```

Let me add the app is ruby on rails and having it clone the cache dir would save significant time avoiding reinstalling all dependencies.

---

**dokku report**

```
~# dokku report
-----> uname: Linux XXXXXHOSTNAMEXXXX 3.13.0-57-generic #95-Ubuntu SMP Fri Jun 19 09:28:15 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
-----> memory: 
             total       used       free     shared    buffers     cached
Mem:         16048      14698       1350        676        442       3394
-/+ buffers/cache:      10861       5187
Swap:            0          0          0
[...]

PR #2730, opened 2017-04-19
Fix ticket #2711

weyert:
Potential fix ticket #2711

### DIFF
--- a/plugins/apps/subcommands/clone
+++ b/plugins/apps/subcommands/clone
@@ -21,7 +21,9 @@ apps_clone_cmd() {
   local NEW_CACHE_DIR="$DOKKU_ROOT/$NEW_APP/cache"
 
   apps_create "$NEW_APP"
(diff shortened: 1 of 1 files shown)


plugins/apps/subcommands/clone
```diff
@@ -21,7 +21,9 @@ apps_clone_cmd() {
   local NEW_CACHE_DIR="$DOKKU_ROOT/$NEW_APP/cache"
 
   apps_create "$NEW_APP"
-  cp -a "$DOKKU_ROOT/$OLD_APP/." "$DOKKU_ROOT/$NEW_APP"
+  cd "$DOKKU_ROOT/$OLD_APP/."
```
josegonzalez: Use pushd and popd instead of cd.

weyert:
Hmm, why is it failing?

josegonzalez:
Check how to run tests locally, it's documented. You just need shellcheck.

weyert:
I will have a closer look tomorrow :)

weyert:
Who is that the fix? I was still looking into it :)

PR #2736, opened 2017-04-21
Implement Network Plugin

### LINKED ISSUE #2403: [RFC] Network Plugin
There are a few shortcomings in network support at the moment:

- Docker [introduced a change](https://github.com/moby/moby/issues/2801) which causes it to sometimes change the internal IP/proxied port of a container on reboot of the host (and sometimes on container restart).
- When a container is restarted by docker - for instance, on server reboot - there is no way outside of a `ps:rebuild` to update the IPs associated with an application.
- Enabling the proxy plugin forces the application to run on an internal network ip. The reverse is also true.
- In the case where an application is using a network overlay, we do not grab the correct ip address - mostly because docker doesn't provide a straightforward way of retrieving the information.

What would fixing the above allow?
- Fix issues where an application might have stale IPs attached because _docker_.
- We could use a network overlay, allowing users to do custom routing.
- Allow the proxy handling to continue to occur properly, while also allowing stuff like consul to report on the state of the actual container as opposed to the proxied app.

What are we not going to fix here?
- Full swarm support. This would help us get there from a networking perspective, but that is not the goal.
### Proposal

I'm proposing a new `network` plugin to handle all of these issues. Things it would do:
- provide `functions/network_ip_for_container` for retrieving the current ip associated with a container. This should also retrieve the `deploy` docker options for an application and parse out the relevant `--network` argument to figure out if we should retrieve an ip from a specific overlay when binding to the "internal" interface.
- provide `functions/network_write_container_ips` for writing the relevant ip to disk for a given container
- `ps:restore` needs to call `network:rebuild-ips` for each app.

New commands:
- `network:set APP key value`: set a property to a value for the network plugin. Properties include:
  - `bind-all`: Whether we should bind to all interfaces or not. Valid options:
    - `internal`: `default`, meaning lets use the "internal" docker ip (existing `DOKKU_IS_APP_PROXY_ENABLED` functionality).
    - `external`: Tell our internal networking that we want to bind ports on the host's `0.0.0.0`
- `network:rebuild-ips APP`: rebuilds the ip mapping at any point in time
- `network:report APP`: report the currently written and _actual_ ips associated to a given application, as well as generic networking information

We'd slightly refactor [this bit of code](https://github.com/dokku/dokku/blob/master/plugins/common/functions#L590-L603) to use the new methodology.

> I'd also like to say we could move the port handling here, though my gut feeling says that it belongs more closely with the `proxy` plugin.
### Relevant issues
- #2085: overlay network, nginx listening to the wrong container ip
- #2400: Dokku doesn't restore apps on server reboots
- #2356: [RFC] Create `report` subcommand for all plugins

---

Thoughts? I think the above should handle a few different issues and potentially increase the flexibility of Dokku going forward.

josegonzalez:
Todo (lord knows if it even passes current tests):

- [x] `network:rebuild`: Implement (also implement `rebuildall`)
- [x] Integrate with existing deploy setup
- [x] `ps:restore` `nginx:build-config`: Rebuild networking config when these are called
- [x] `network:set`: Add support for setting arbitrary network properties (`bind-all-interfaces` will be the only supported one at the moment).
- [x] `network:report`: Add a reporting command
- [x] Remove direct access to port and ip files from everything but this plugin.
- [ ] ~~Call this from an in-resident docker container via [docker-plugins](https://github.com/progrium/docker-plugins) and the dokku-daemon package in order to cycle the network config as appropriate (a faster solution than ps:rebuild).~~ People can do this on their own if they wish.
- [x] Tests: thanks in advance @michaelshobbs 
- [x] Rebuild in golang? I intentionally used plugin triggers as the interface to keep this clean.
- [x] Verify that docs are up to date and useful

Closes #2403.


plugins/network/commands
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
michaelshobbs: New core plugins should probably be implemented in golang, no?

josegonzalez: >  Rebuild in golang? I intentionally used plugin triggers as the interface to keep this clean.
 Once tests pass, I'll rewrite.

michaelshobbs: Yup, totally cool with the calls to plugin triggers. I was specifically referring to the plugin only code. e.g. commands and such

josegonzalez: Yeah, just wanted to write the integration as a proof of concept before I got down and dirty with golang.

plugins/network/functions
```diff
@@ -0,0 +1,147 @@
...
+
+      CONTAINER_ID="$(cat "$DOKKU_CONTAINER_ID_FILE")"
+      [[ -n "$CONTAINER_ID" ]] || continue
+      CONTAINER_RUNNING="$(docker inspect -f '{{.State.Running}}' "$CONTAINER_ID" 2>/dev/null || true)"
+      [[ "$CONTAINER_RUNNING" == "true" ]] || continue
+
+      ipaddr=$(network-get-container-ipaddr "$APP" "$PROC_TYPE" "$IS_HEROKUISH_CONTAINER" "$CONTAINER_ID")
+      port=$(network-get-container-port "$APP" "$PROC_TYPE" "$IS_HEROKUISH_CONTAINER" "$CONTAINER_ID")
+
+      [[ -n "$ipaddr" ]] && echo "$ipaddr" > "$DOKKU_IP_FILE"
```
znz: `plugn trigger network-write-ipaddr`?

plugins/network/functions
```diff
@@ -0,0 +1,147 @@
...
+      [[ -n "$port" ]] && echo "$port" > "$DOKKU_PORT_FILE"
```
znz: `plugn trigger network-write-port`?

plugins/network/functions
```diff
@@ -0,0 +1,147 @@
...
+  if [[ "$PROC_TYPE" != "web" ]]; then
+    return
+  fi
+
+  if [[ "$IS_HEROKUISH_CONTAINER" == "false" ]]; then
+    local DOKKU_DOCKERFILE_PORTS=($(config_get "$APP" DOKKU_DOCKERFILE_PORTS || true))
+  fi
+
+  if [[ -n "${DOKKU_DOCKERFILE_PORTS[*]}" ]]; then
+    for p in ${DOKKU_DOCKERFILE_PORTS[*]};do
```
znz: `p` should be local?

plugins/network/functions
```diff
@@ -0,0 +1,147 @@
...
+      if [[ ! "$p" =~ .*udp.* ]]; then
+        # set port to first non-udp port
+        p=${p//\/tcp}
+        PORT=${port:="$p"}
```
znz: `port` set as a global variable and unused?

plugins/nginx-vhosts/functions
```diff
@@ -217,13 +217,7 @@ nginx_build_config() {
   if [[ -z "$DOKKU_DISABLE_PROXY" ]]; then
     if [[ -z "$DOKKU_APP_LISTEN_PORT" ]] && [[ -z "$DOKKU_APP_LISTEN_IP" ]]; then
       shopt -s nullglob
```
znz: unused nullglob

plugins/common/common.go
```diff
@@ -134,8 +153,213 @@ func GetAppImageRepo(appName string) string {
...
+func FileExists(filePath string) bool {
+	fi, err := os.Stat(filePath)
+	if err != nil {
+		return false
+	}
+
+	return fi.Mode().IsRegular()
+}
+
+// GetAppImageName returnS image identifier for a given app, tag tuple. validate if tag is presented
```
znz: `returnS` → `returns`

plugins/common/common.go
```diff
@@ -134,8 +153,213 @@ func GetAppImageRepo(appName string) string {
...
+	dockerCmd.ShowOutput = false
+	return dockerCmd.Execute()
+}
+
+// LogInfo1 is the info1 header formatter
+func LogInfo1(text string) {
+	fmt.Fprintln(os.Stdout, fmt.Sprintf("-----> %s", text))
+}
+
+// LogInfo2Quiet is the info1 header formatter (with quiet option)
```
znz: 2 → 1

plugins/network/network.go
```diff
@@ -0,0 +1,165 @@
...
+		"docker",
+		"inspect",
+		"--format",
+		"'{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'",
+		containerId,
+	}, " "))
+
+	b, err := imageCmd.Output()
+	if err != nil || len(b) == 0 {
+		// docker < .19 compatibility
```
znz: .19 → 1.9

plugins/network/network.go
```diff
@@ -0,0 +1,165 @@
...
+	lines, err := common.FileToSlice(scaleFile)
+	if err != nil {
+		return
+	}
+	for _, line := range lines {
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		procParts := strings.SplitAfterN(line, ":", 2)
```
znz: `SplitAfterN` → `SplitN` and `":"` → `"="`

plugins/network/network.go
```diff
@@ -0,0 +1,165 @@
...
+			continue
+		}
+		procType := procParts[0]
+		procCount, err := strconv.Atoi(procParts[1])
+		if err != nil {
+			continue
+		}
+
+		containerIndex := 0
+		for containerIndex <= procCount {
```
znz: `<=` → `<`

plugins/network/src/triggers/network-write-ipaddr/network-write-ipaddr.go
```diff
@@ -0,0 +1,40 @@
...
+		common.LogFail(err.Error())
+	}
+
+	appRoot := strings.Join([]string{common.MustGetEnv("DOKKU_ROOT"), appName}, "/")
+	filename := fmt.Sprintf("%v/IP.%v.%v", appRoot, procType, containerIndex)
+	f, err := os.Create(filename)
+	if err != nil {
+		common.LogFail(err.Error())
+	}
+
```
znz: missing `defer f.Close()`

plugins/network/src/triggers/network-write-port/network-write-port.go
```diff
@@ -0,0 +1,40 @@
...
+		common.LogFail(err.Error())
+	}
+
+	appRoot := strings.Join([]string{common.MustGetEnv("DOKKU_ROOT"), appName}, "/")
+	filename := fmt.Sprintf("%v/PORT.%v.%v", appRoot, procType, containerIndex)
+	f, err := os.Create(filename)
+	if err != nil {
+		common.LogFail(err.Error())
+	}
+
```
znz: missing `defer f.Close()`

plugins/common/common.go
```diff
@@ -134,8 +153,213 @@ func GetAppImageRepo(appName string) string {
...
+	return imageCmd.Execute()
+}
+
+// ContainerIsRunning checks to see if a container is running
+func ContainerIsRunning(containerId string) bool {
+	b, err := sh.Command("docker", "inspect", "--format", "'{{.State.Running}}'", containerId).Output()
+	if err != nil {
+		return false
+	}
+	return string(b[:]) == "true"
```
znz: `ContainerIsRunning` always returns `false` because `b` is `"true\n"` instead of `"true"`.

josegonzalez:
@znz while I appreciate the review, this PR doesn't even pass tests yet, and isn't ready for general review.

plugins/config/config.go
```diff
@@ -0,0 +1,30 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	common "github.com/dokku/dokku/plugins/common"
+)
+
+func GetWithDefault(appName string, key string, defaultValue string) string {
```
alexquick: @josegonzalez when I replace this should it fall back on the global config before the default or must this be only app config or default?

josegonzalez: Only app config. We can have an alternative method for retrieving the global config version of any value.

aleclarson:
What are the recommended Docker and Dokku versions to roll back to while I wait for this PR to be merged? I'm trying to avoid the IP changing when a server auto-restarts.

josegonzalez:
I would look at [this issue](https://github.com/moby/moby/issues/2801), which might have information as to what docker version was pushed out that changed. As I recall, its a pretty old version (relative to whats currently out there).

Benjamin-Dobell:
@aleclarson No need to roll back Docker, you can simply tell current versions of Docker to exhibit the old behaviour. Depends which os/initial system you're using. But you want something like:

```
DOCKER_OPTS="--bip=172.17.42.1/24"
```

Mind you, this won't help you reasssociate containers when the system reboots. Instead it's just useful in certain circumstances to be able to listen/bind to certain address ranges (e.g. fell your DB to listen on that subnet so it's not exposed externally).

One thing to consider is startup order. That subnet is not immediately available on boot. Even more annoying it's _not_ available once dockerd has launched, as dockerd forks and initiatives systems think it has finished launching. It's easiest just to have your startup scripts retry if they can't bind. There was an issue on Docker's Github concerning firing an event when the network has been initialised but I can't remember the outcome. Nonetheless, that's pretty difficult to setup with systemd etc.

josegonzalez:
The following fix involves finishing this PR and having a listener reassociate container info on restart with dokku (via the new plugin commands mounted by a docker/dokku socket).

Even single-container paas systems are hard :(.

ebeigarts:
I'm using [this workaround] until this is fixed.

```
sudo curl -L -o /etc/systemd/system/dokku-redeploy-fix.service https://gist.github.com/ebeigarts/e3256021986f625acec62d9add04dfc3/raw/d85afd4e8beba843e273f4c8b7237c375a7eabea/dokku-redeploy-fix.service &&
sudo curl -L -o /usr/bin/dokku-redeploy-fix https://gist.github.com/ebeigarts/e3256021986f625acec62d9add04dfc3/raw/b4b3be7f3fdf51ba4e7716ac8cf04f40b97a762b/dokku-redeploy-fix.sh &&
sudo chmod +x /usr/bin/dokku-redeploy-fix &&
sudo systemctl enable dokku-redeploy-fix
```

aleclarson:
That works perfectly, @ebeigarts. Thank you for the help!

josegonzalez:
Shame all the deleted comments arent collapsed. Ah well.

.editorconfig
```diff
@@ -17,5 +17,5 @@ indent_size = 4
 
 [*.go]
 insert_final_newline = true
-indent_style = space
+indent_style = tab
```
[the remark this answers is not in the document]
josegonzalez: Golang indents with tabs, not spaces. You probably want to checkout [gofmt](https://golang.org/cmd/gofmt/).

plugins/network/src/triggers/install/install.go
```diff
@@ -0,0 +1,33 @@
...
+// runs the install step for the network plugin
+func main() {
+	err := common.PropertySetup("network")
+	if err != nil {
+		common.LogFail(fmt.Sprintf("Unable to install the network plugin: %s", err.Error()))
+	}
+
+	apps, err := common.DokkuApps()
+	if err != nil {
+		return
```
michaelshobbs: what do we want to do here though? seems like we should panic or return some error msg, yea?

josegonzalez: I think it only errors if there are no apps? If thats the case, just ignore...

.editorconfig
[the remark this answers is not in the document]
michaelshobbs: oh i read that backwards. i'm dum....

josegonzalez:
>  Call this from an in-resident docker container via docker-plugins and the dokku-daemon package in order to cycle the network config as appropriate (a faster solution than ps:rebuild).

This doesn't need to happen on this PR, and is probably a separate project we can create if there is a need.

ebeigarts:
👏 🎉

rvalyi:
congratulations!!

PR #2751, opened 2017-05-01
Move config plugin to golang

alexquick:
I wanted to start getting feedback on this because it's quite a lot of changes.

This resolves https://github.com/dokku/dokku/issues/1262

To Do:
- [x] Document new export flags
- [x] Change existing documentation to no longer suggest backslash-escaping spaces and newlines. For newlines it never really worked anyway. This will be a breaking change for dokku 1.0 if that's okay.
- [x] Add logging back in. Should I make some go helpers in common for this?
- [x] We need this: https://github.com/gliderlabs/herokuish/pull/252 and a new herokuish build
- [x] There is a ShellCmd out to `ps:restart` I gather should be using `go-basher` for this but I'm lost as to how to do that. Can you point me in the right direction?
 - [x] Remaining Bugs
   - [x] Empty env vars for Dockerfile-based apps choke
   - [x] cURL env defaults are missing (CURL_CONNECT_TIMEOUT, CURL_TIMEOUT)
- [x] get godotenv PR merged
- [x] linting boyscouting

Usage:
```
dokku config:set --global key='value with spaces'
dokku config:set --global key='value with spaces
and newline'
dokku config:set --encoded --global key="$(echo "something" | base64)" # to allow setting complicated config values over ssh
dokku config:get --quoted --global key # export quoted for bash export/eval
dokku config:export --bundle # export tarfile of herokuish BUILD_ENV directory
dokku config:export --merged APP_NAME # export app config merged with global config 
```


plugins/config/src/configenv/configenv.go
```diff
@@ -0,0 +1,194 @@
...
+	filename       string
+	EscapeNewlines bool
+}
+
+func (e *Env) String() string {
+	return e.EnvfileString()
+}
+
+//EnvfileString returns the contents of this Env in ENVFILE format
+func (e *Env) EnvfileString() string {
```
josegonzalez: For env parsing, could we instead lean on something like [godotenv](https://github.com/joho/godotenv). Dotenv file parsing is hard, and now that we're at a higher-level language, might do some good to offload the munging to a proven library?

plugins/config/src/subcommands/set/set.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if changed {
+		//log
+		env.Write()
+		args := append([]string{appName, "set"}, entries...)
+		common.PlugnTrigger("post-config-update", args...)
+	}
+
+	if shouldRestart && env.GetBoolDefault("DOKKU_APP_RESTORE", true) {
+		cmd := common.NewTokenizedShellCmd("dokku", "ps:restart", appName)
```
josegonzalez: We could instead introduce a new plugin trigger. I've been relying on those instead of sourcing stuff in the new networking plugin. Not sure how I feel about a new plugin trigger in general, but its on the table. /cc @michaelshobbs .

josegonzalez:
I added a few shell helpers [here](https://github.com/dokku/dokku/pull/2736/files#diff-1a45b709cf4d8def775f1bf1109a4f01R314). If we overlap, I can handle the merge conflicts.

plugins/config/src/configenv/configenv.go
alexquick: I like that idea. Right now they only support reading, not writing. I'll see if they are interested in a PR for writing out .env files too. Even if not I agree it's worth the move.

alexquick:
@josegonzalez Do you have any idea why the tests would fail in CI but not in the dev vagrant? Do you have any tips on how to get my local testing rig closer to the CI environment?

josegonzalez:
Not sure, they should be equivalent. Checkout our circle.yml file for any changes. Pretty sure both are trusty...

plugins/config/src/subcommands/set/set.go
michaelshobbs: Plugin trigger for what?

josegonzalez: For triggering an app restart. Not sure if this is the correct way to do it instead.

michaelshobbs: I view plugin triggers as integration points; i.e. places where we want to allow something else to take over. I definitely see where this might be useful in the future but I'm not convinced now is the right time to do that. I'd rather not convolute the golang port effort with a shift in how core functionality interacts. I got in some trouble doing that on the python port. I'd be :+1: for a follow up PR that does that perhaps.

plugins/config/config.go
```diff
@@ -17,6 +17,20 @@ func GetWithDefault(appName string, key string, defaultValue string) string {
 	return env.GetDefault(key, defaultValue)
 }
 
+//HasKey determines if the config given by appName has a value for the given key
+func HasKey(appName string, key string) bool {
+	env, err := loadConfig(appName)
+	if err != nil {
+		return false
+	}
+	for _, v := range env.Keys() {
```
josegonzalez: Could just be

```go
if val, ok := env[key]; ok {
  return true
}
return false
```

Assuming env is a map of course.

alexquick: :+1: great point, thanks. I’ve gotta brush up on my go.

alexquick:
@josegonzalez @michaelshobbs This is getting pretty close. Can you think of anything else I should test?

docs/configuration/environment-variables.md
```diff
@@ -35,10 +40,10 @@ When setting or unsetting environment variables, you may wish to avoid an applic
 dokku config:set --no-restart node-js-app ENV=prod
 ```
 
-If you wish to have the variables output in an `eval`-compatible form, you can use the `--export` flag:
+If you wish to have the variables output in an `eval`-compatible form, you can use the `config:export` command
```
josegonzalez: Does the export flag no longer exist?

josegonzalez:
Not much else I can think of, this seems to cover all the bases pretty well. I'll give it a final review tomorrow.

docs/configuration/environment-variables.md
alexquick: @josegonzalez yeah, good point. I saw that we are trying to move away from bare `dokku blah` commands and instead favor `dokku blah:subcommand` commands, right? Would we like to preserve the old functionality? I not what should happen on bare `dokku config`?

josegonzalez: Thats fine, just want to make sure that was the intention.

alexquick: There were some small issues with the docs that I just fixed. I also added `--shell` and `--export` back to `dokku config`, albeit marked as deprecated.

josegonzalez:
Lets see if my rebase works...

alexquick:
@josegonzalez I'll fix codacy and the merge issues in the next few days. Do you want me to squash/rebase or are you okay with this kinda messy history?

josegonzalez:
I'm fine either way, so whatever you prefer is cool.

josegonzalez:
@alexquick merge issues fixed, now all thats left is figuring out and fixing whatever `codacy` is complaining about.

plugins/common/common.go
```diff
@@ -25,11 +25,16 @@ type ShellCmd struct {
 // NewShellCmd returns a new ShellCmd struct
 func NewShellCmd(command string) *ShellCmd {
 	items := strings.Split(command, " ")
-	cmd := items[0]
-	args := items[1:]
+	return NewTokenizedShellCmd(items...)
+}
+
+// NewTokenizedShellCmd creates a new ShellCmd
+func NewTokenizedShellCmd(command ...string) *ShellCmd {
```
michaelshobbs: if `NewShellCmd` just calls `NewTokenizedShellCmd` why not just update `NewShellCmd`

plugins/common/common.go
```diff
@@ -343,3 +348,14 @@ func VerifyImage(image string) bool {
 	imageCmd.ShowOutput = false
 	return imageCmd.Execute()
 }
+
+//PlugnTrigger fire the given plugn trigger with the given args
+func PlugnTrigger(triggerName string, args ...string) (string, error) {
+	shellArgs := make([]interface{}, 0, len(args))
```
michaelshobbs: i have some questions about this function
- why is `shellArgs` a `[]interface{}` and not `[]string`?
- why are we setting the len to 0 and cap to length and not just setting the len to `len(args) + 2` and using the index from range to set shellArgs?

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/dokku/dokku/plugins/common"
+	configenv "github.com/dokku/dokku/plugins/config/src/configenv"
```
michaelshobbs: no need to alias

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
+	columnize "github.com/ryanuber/columnize"
```
michaelshobbs: no need to alias

plugins/config/config.go
michaelshobbs: also thinking maybe this should go at the root of the package. not sure why it's a subpackage

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
+	if err != nil {
+		return defaultValue
+	}
+	return env.GetDefault(key, defaultValue)
+}
+
 // GetWithDefault returns the value set for a given key, returning defaultValue if none found
-func GetWithDefault(appName string, key string, defaultValue string) (value string) {
-	value = defaultValue
+// func GetWithDefault(appName string, key string, defaultValue string) (value string) {
```
michaelshobbs: remove commented out version?

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
+	}
+	_, ok := env.Get(key)
+	return ok
+}
+
+//SetMany variables in the environment. If appName is empty the global config is used. If restart is true the app is restarted.
+func SetMany(appName string, entries map[string]string, restart bool) {
+	global := appName == ""
+	env := GetConfig(appName, false)
+
```
michaelshobbs: spurious newline

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
+	keys := make([]string, 0, len(entries))
+
```
michaelshobbs: spurious newline

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
-		if !strings.HasPrefix(line, prefix) {
-			continue
+
+	if !global && restart && env.GetBoolDefault("DOKKU_APP_RESTORE", true) {
+		Restart(appName)
+	}
+}
+
+//PrettyPrintEnvEntries in columns
+func PrettyPrintEnvEntries(prefix string, entries map[string]string) string {
```
michaelshobbs: does this need to be exported?

plugins/config/config.go
```diff
@@ -2,29 +2,156 @@ package config
...
+	colConfig.Delim = "\x00"
+	lines := make([]string, 0, len(entries))
+	for k, v := range entries {
+		lines = append(lines, fmt.Sprintf("%s:\x00%s", k, v))
+	}
+	return columnize.Format(lines, colConfig)
+}
+
+//GetCommonArgs extracts common positional args (appName and keys)
+func GetCommonArgs(global bool, args []string) (string, []string) {
```
michaelshobbs: i'd like to see us use named return vars where we can

plugins/ps/app-restart
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
michaelshobbs: why did this change? seems unrelated to config....

tests.mk
```diff
@@ -105,6 +105,16 @@ else
...
 
+unit-tests-config:
+	@echo running config unit tests...
+	@$(QUIET) bats tests/unit/20_config.bats
+
+unit-tests-config-oddities:
+	@echo running config oddities unit tests...
+	@$(QUIET) bats tests/unit/30_config-oddities.bats
+
+
```
michaelshobbs: spurious newlines

tests.mk
```diff
@@ -105,6 +105,16 @@ else
 	@$(QUIET) ./tests/ci/unit_test_runner.sh $$UNIT_TEST_BATCH
 endif
 
+unit-tests-config:
```
michaelshobbs: do we need this target?

tests.mk
```diff
@@ -105,6 +105,16 @@ else
...
+	@echo running config unit tests...
+	@$(QUIET) bats tests/unit/20_config.bats
+
+unit-tests-config-oddities:
```
michaelshobbs: do we need this target?

tests/apps/config/check_deploy
```diff
@@ -12,4 +12,5 @@ test_url () {
 
 base_url="$1"
 test_url / "config-test"
-test_url /hello "Hello world"
+test_url /hello "Hello
+ 'world'"
```
michaelshobbs: need newline at EOF

tests/apps/dockerfile/app.json
```diff
@@ -23,6 +23,10 @@
     "WEB_CONCURRENCY": {
       "description": "The number of processes to run.",
       "generator": "echo 5"
+    },
+    "FROM_APP_JSON": {
```
michaelshobbs: why this change?

tests/apps/dockerfile/check_deploy
```diff
@@ -1,2 +1,15 @@
...
+  expected_output="$2"
+  url="${base_url}:3000${path}"
+  output="$(curl -s -S "$url")"
+  test "$output" == "$expected_output"
+}
+
+base_url="$1"
+test_url / "Hello World!"
+test_url /env/HELLO "Hello
+ 'world' "
```
michaelshobbs: missing newline at EOF

tests/unit/30_config-oddities.bats
```diff
@@ -0,0 +1,47 @@
...
+  line two'
+  run dokku config:set --global double_quotes="$multiline" 
+  echo "output: "$output
+  echo "status: "$status
+  run dokku config:get --global double_quotes
+  echo "output: "$output
+  echo "status: "$status
+  assert_output "$multiline"
+  assert_success
+}
```
michaelshobbs: missing newline at EOF

plugins/config/src/configenv/configenv_test.go
```diff
@@ -0,0 +1,66 @@
+package configenv
```
michaelshobbs: As this is a heavily used plugin i'd like to see more test coverage here.

michaelshobbs:
@alexquick i'm happy to take a pass at cleaning up some of my concerns here if you're short on time. lemme know

plugins/ps/app-restart
alexquick: The old config plugin used a direct execution of `ps_restart` if an app's config had been changed and `--no-restart` wasn't set. The easiest way I could find to get to this functionality from the new go-based config plugin is through a plugn trigger.

plugins/common/common.go
alexquick: 1) sh.Command is declared as `Command(name string, a ...interface{})` and I couldn't find a good way to convert `[]string{}` to `[]interface{}` and didn't want to leak `[]interface{}` into callers. Any advice?

2)I figured append may be more readable and less likely to mess up indexing if edited down the line, but reading it again yeah switching to indexing is a better call.

plugins/config/config.go
alexquick: I was trying to isolate it from the dokku-specific stuff but I can move it to the root

alexquick:
@michaelshobbs thanks for the review. It's a big PR with a lot of changes across much of dokku so I'm sure it wasn't that fun to read.

I'd like to resolve them if that's alright, I should have time this week. I left 2-3 inline comments on your review for clarification, but the rest makes sense.

plugins/config/config.go
michaelshobbs: Perhaps this could be moved to an independent repo and included as a vendored dependency. @josegonzalez thoughts?

josegonzalez: Whats this dependency do exactly?

alexquick: The idea was that it was a config type that wasn't linked to dokku internals while the config package had the flat helpers which mimic what was in the old `functions` file.

I'm all for eliminating the subpackage, it has a dumb name and I was on the fence about it anyhow. It was more relevant when it handled dotenv parsing/serialization itself but I got the changes into `godotenv` we needed so it doesn't do that anymore.

plugins/common/common.go
alexquick: I'm gonna just get rid of this. It was vestigial from when I was making some calls to `dokku` rather than `plugn`. I might open something similar as another PR in the future.

plugins/config/config.go
alexquick: Bare `dokku config` uses it to print out the environment right now.

alexquick:
@michaelshobbs @josegonzalez I've applied your feedback and refactored a little bit to reduce the API surface area of config and remove the subpackage.

plugins/common/common.go
alexquick: @josegonzalez Is it necessary that plugn goes through `sh.Command` and not the `os/exec`package? If it's moved to use `os/exec` it will get simpler because the `NewShellCmd` stuff can get used. Does it rely on actually being executed in a shell?

plugins/config/environment.go
```diff
@@ -0,0 +1,274 @@
...
+	name     string
+	filename string
+	env      map[string]string
+}
+
+//newEnvFromString creates an env from the given ENVFILE contents representation
+func newEnvFromString(rep string) (env *Env, err error) {
+	envMap, err := godotenv.Unmarshal(rep)
+	env = &Env{
+		"<unknown>",
```
michaelshobbs: in struct literals, please use field names. i.e.:
```
env = &Env{
 	name: "<unknown>",
 	filename: "",
 	env: envMap,
 }
```

plugins/config/environment.go
```diff
@@ -0,0 +1,274 @@
...
+	return columnize.Format(lines, colConfig)
+}
+
+func loadFromFile(name string, filename string) (env *Env, err error) {
+	envMap := make(map[string]string)
+	if _, err := os.Stat(filename); err == nil {
+		envMap, err = godotenv.Read(filename)
+	}
+
+	env = &Env{
```
michaelshobbs: same comment regarding struct literals and field names

josegonzalez:
🎉

PR #2753, opened 2017-05-02
Use short-hand method for shutting down all applications in upgrade docs

abrkn:
Add short-hand method for shutting down all apps. Useful for users with large number of apps.


josegonzalez:
Thanks for the pull request!

PR #2754, opened 2017-05-02
add support for set DOKKU_IMAGE per app

### LINKED ISSUE #2745: Ability to specify base image in app or plugin
Hello, in some cases it could be handy if the application itself or eventually a plugin could set some custom base image before the buildpacks are applied.

Today one can do that with the dokkurc file. But this is system wide then and this can be a bit annoying. I looked at the current code if I could change the DOKKU_IMAGE in a plugin hook such as post-extract (so based on some app file analysis I would like to eventually switch the base image) but I couldn't figure out anything. If I try to export DOKKU_IMAGE again it doesn't override the value where it is called.

Currently, the DOKKU_IMAGE is set by sourcing dokkurc or setting its default value in the dokku executable and it is used in:
```
plugins/common/functions:      local id=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d -v $CACHE_DIR:/cache -e CACHE_PATH=/cache $DOCKER_ARGS $IMAGE /build)
```

Is there any way to override it? Any simple refactor I could contribute to allow it? Thanks.

fzerorubigd:
fixes #2745

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -443,6 +443,7 @@ dokku_build() {
 
   case "$IMAGE_SOURCE_TYPE" in
     herokuish)
+      DOKKU_IMAGE="$(config_get $APP DOKKU_IMAGE || echo $DOKKU_IMAGE)"
       local id=$(tar -c . | docker run "$DOKKU_GLOBAL_RUN_ARGS" -i -a stdin "$DOKKU_IMAGE" /bin/bash -c "mkdir -p /app && tar -xC /app")
       test "$(docker wait "$id")" -eq 0
       docker commit "$id" "$IMAGE" > /dev/null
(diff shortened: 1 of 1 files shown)


josegonzalez:
Needs a test of some sort.

fzerorubigd:
I can not find a way to check for the base image. a way is to create an image with specific file in it and the check for that specific file after build.

josegonzalez:
@fzerorubigd that works for me. Can you fork herokuish and add some change to make it possible to test? We can then probably merge that into a custom branch, make a tag of it on docker hub, and then use that for the test here.

fzerorubigd:
@josegonzalez is that really required? why not simply run herokuish, create a file in it and then tag it as new image and not adding an external dependency to an image to pull?
Am I missing something?

josegonzalez:
Sure thats fine.

rvalyi:
many tanks for this feature!

PR #2755, opened 2017-05-02
Fixed `getting-started`-document

weyert:
Fixed `getting-started`-document so it doesn't refer to deprecated calls


josegonzalez:
They are only recently deprecated, so the new call doesn't work on old systems. As the old call redirects the warning to stderr, this is fine.

weyert:
But shouldn't doc reflect the state of `master`?

josegonzalez:
Yes and no. In this case, developers are likely coming in to upgrade _to_ master. Once we hit a few more minor releases (or a major) then we can consider removing the deprecated call imo.

weyert:
Okay :) I will keep the branch for later reference then

PR #2763, opened 2017-05-10
Update vagrant.md

ZiadSalah:
Included instructions to run vagrant up dokku-windows if on a windows host

### DIFF
--- a/docs/getting-started/install/vagrant.md
+++ b/docs/getting-started/install/vagrant.md
@@ -21,6 +21,7 @@
     cd path/to/dokku
     vagrant up
     ```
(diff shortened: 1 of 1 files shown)


ZiadSalah:
Hi, it took me a while to realise I need to specify dokku-windows when setting up on a windows host. I figured the lack of instructions might turn people away, or that at the very least this might save someone some time.

joshmanders:
Thanks @ZiadSalah! Once tests finish we can merge this in. For future reference, when doing docs changes, you can put `[ci skip]` in the body of your PR and that will tell CircleCI to not run the test suite.

PR #2764, opened 2017-05-10
Create PULL_REQUEST_TEMPLATE.md

joshmanders:
Puts note about [ci skip] so that CI doesn't trigger when making PR's about docs.


josegonzalez:
That has to be in the body of the commit, not the body of the PR, unfortunately. But still a worth contribution!

joshmanders:
Hmmm damn. Maybe edit it to say "In the body of the `HEAD` commit"?

PR #2769, opened 2017-05-12
Fix SlackButton for mobile devices.

lucianopf:

Just adding a `inherit` to it in order to make it fit nicely when working with smaller devices.


josegonzalez:
Yay i've been meaning to look into this but never had the time :)

PR #2781, opened 2017-05-19
Adds docker.io support

### LINKED ISSUE #2775: Support for docker.io packages maintained by Ubuntu
At the moment, dokku requires one of four different docker packages to be installed: `docker-engine-cs (>= 1.9.1) | docker-engine (>= 1.9.1) | docker-ce | docker-ee` (see https://github.com/dokku/dokku/blob/master/debian/control)

Nevertheless, there exists a further, quite importan one: `docker.io` (see http://packages.ubuntu.com/de/xenial/docker.io)
It is maintained by the Ubuntu community and available in the official package sources. There are a few use cases, where you have to use this version, for example, if you use docker in a lxc/lxd container.

Since this package depends on the offical docker sources, and the version is recent enough to contain all features used by dokku, I don't see any problems to add this package as a fifth variant for the docker dependency.

fruitl00p:
This (small) PR adds the docker-io package from Ubuntu Xenial as supported docker version

fixes #2775

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Version: 0.9.4
 Section: web
 Priority: optional
 Architecture: amd64
-Depends: locales, git, make, curl, gcc, man-db, sshcommand (>= 0.6.0), gliderlabs-sigil, docker-engine-cs (>= 1.9.1) | docker-engine (>= 1.9.1) | docker-io (>= 1.10.3)  | docker-ce | docker-ee, software-properties-common, python-software-properties | python3-software-properties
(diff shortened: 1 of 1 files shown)


fruitl00p:
I'm too unfarmiliar with CircleCI to include a test that makes sure the environment matching the supported docker versions are also used in the runs... 

@josegonzalez This might need some better testing as you'd want to run the entire suite of tests in the various supported environments... I expect this is already done, but as said, I don't have enough CircleCi experience to include an environment with only docker-io installed already and going from there..

PR #2790, opened 2017-05-29
Add entry for insecure connection issue in Rails

flyinggrizzly:
If an application is configured to enforce secure connections and SSL isn't configured for Dokku or the application, browsers will raise an Invalid Connection error.

Entry describes issue, a code resolution for Rails, and points users to the SSL docs.

[ci skip]

### DIFF
--- a/docs/getting-started/troubleshooting.md
+++ b/docs/getting-started/troubleshooting.md
@@ -195,3 +195,12 @@ sudo ufw disable
 
 ***
 
+__Symptom:__ I can't connect to my application because the server is sending an invalid response, or can't provide a secure connection.
+
+__Solution:__
+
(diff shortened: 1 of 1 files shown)


josegonzalez:
*sigh*

Merging as I'm sure someone else will blame Dokku, though I hate framework/language-specific stuff in our docs :(

Thanks for the pull request!

joshmanders:
I think it's good info to have, but I wouldn't include the Rails example. The first half is enough to give the user a reason to make sure that's the issue.

flyinggrizzly:
Any time--I debated including the example, but decided to go for it to match the Node example earlier in the 502 error issue... which I suppose makes it worse now that it's multiple languages frameworks?

Regardless, enjoying Dokku. Thank you!

PR #2815, opened 2017-06-28
Add stickler-ci configuration.

markstory:
In talking with @josegonzalez, stickler-ci.com helps reduce the amount of time maintainers spend reviewing code by automatically giving style feedback as pull request comments.

I've added linters for shell scripts and go files based on the linting I could find in the existing make files.

If you merge this, you will also need to log in to https://stickler-ci.com and enable webhooks for the repository by enabling the appropriate checkbox.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,0 +1,9 @@
+linters:
+  shellcheck:
+    shell: bash
+    exclude: SC2034
+  golint:
+
+files:
+    ignore:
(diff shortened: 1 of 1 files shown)


josegonzalez:
Added the stickler webhook, any way you can bump this PR to see the output?

markstory:
This pull request doesn't modify any shell or go files so stickler wouldn't leave any feedback.

PR #2823, opened 2017-07-03
migrate env for NGINX_* to PROXY_*

OmgImAlexis:
This switches the env for `DOKKU_NGINX_*` to `DOKKU_PROXY_*` so we don't need different env settings for different proxy types.

### DIFF
--- a/docs/deployment/application-management.md
+++ b/docs/deployment/application-management.md
@@ -8,8 +8,10 @@ apps:create <app>                              # Create a new app
 apps:destroy <app>                             # Permanently destroy an app
 apps:exists <app>                              # Checks if an app exists
 apps:list                                      # List your apps
+apps:lock                                      # Creates a '.deploy.lock' file in an application's repo
 apps:rename <old-app> <new-app>                # Rename an app
 apps:report [<app>] [<flag>]                   # Display report about an app
+apps:unlock                                    # Removes the '.deploy.lock' file from an application's repo
 ```
 
 ## Usage
@@ -134,7 +136,7 @@ This will copy all of your app's contents into a new app directory with the name
 
 ### Cloning an existing app
 
-> New as of 0.8.1
+> New as of 0.11.5
 
 You can clone an existing app using the `apps:clone` command.  Note that the application *must* have been deployed at least once, or cloning will not complete successfully:
 
@@ -156,6 +158,39 @@ By default, Dokku will deploy this new application, though you can skip the depl
 dokku apps:clone --skip-deploy node-js-app io-js-app
 ```
 
+### Locking app deploys
+
+> New as of 0.11.6
+
+If you wish to disable deploying for a period of time, this can be done via deploy locks. Normally, deploy locks exist only for the duration of a deploy so as to avoid deploys from colliding, but a deploy lock can be created by running the `apps:lock` subcommand.
+
+
+```shell
+dokku apps:lock node-js-app
+```
+
+```
+-----> Deploy lock created
+```
+
+### Unlocking app deploys
+
+> New as of 0.11.6
+
+In some cases, it may be necessary to remove an existing deploy lock. This can be performed via the `apps:unlock` subcommand.
+
+> Warning: Removing the deploy lock _will not_ stop in progress deploys. At this time, in progress deploys will need to be manually terminated by someone with server access.
+
+```shell
+dokku apps:unlock node-js-app
+```
+
+```
+ !     A deploy may be in progress.
+ !     Removing the application lock will not stop in progress deploys.
+-----> Deploy lock removed.
+```
+
 ### Displaying reports about an app
 
 > New as of 0.8.1
@@ -171,12 +206,14 @@ dokku apps:report
        App dir:             /home/dokku/node-js-app
        Git sha:             dbddc3f
        Deploy method:       git
+       Locked:              false
 =====> python-sample
 not deployed
 =====> ruby-sample
        App dir:             /home/dokku/ruby-sample
        Git sha:             a2d477c
        Deploy method:       git
+       Locked:              false
 ```
 
 You can run the command for a specific app also.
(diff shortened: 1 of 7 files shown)


plugins/nginx-vhosts/install
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/config/functions"
```
josegonzalez: These two need to be sourced in the functions file, not here.

plugins/nginx-vhosts/install
```diff
@@ -66,6 +68,8 @@ cp "${PLUGIN_CORE_AVAILABLE_PATH}/nginx-vhosts/templates/500-error.html" "${DOKK
 # patch broken nginx 1.8.0 logrotate
 [[ -f /etc/logrotate.d/nginx ]] && sed -i -e 's/invoke-rc.d/service/g' /etc/logrotate.d/nginx
 
+migrate_nginx_to_proxy
```
josegonzalez: You didn't source the nginx-vhosts/functions file here, so it'll break.

plugins/nginx-vhosts/subcommands/migrate
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/config/functions"
+source "$PLUGIN_AVAILABLE_PATH/nginx-vhosts/functions"
+
+nginx_migrate_cmd() {
+  declare desc="converts DOKKU_NGINX_* to DOKKI_PROXY_*"
+  local cmd="nginx:migrate"
+  migrate_nginx_to_proxy
```
josegonzalez: I really don't want this to be a thing. For anyone installing from source, the makefile will trigger the install hook, and packaged installs also trigger it as well.

plugins/nginx-vhosts/templates/nginx.conf.sigil
```diff
@@ -12,7 +12,7 @@ server {
   access_log  /var/log/nginx/{{ $.APP }}-access.log;
   error_log   /var/log/nginx/{{ $.APP }}-error.log;
 {{ if (and (eq $listen_port "80") ($.SSL_INUSE)) }}
-  return 301 https://$host:{{ $.NGINX_SSL_PORT }}$request_uri;
+  return 301 https://$host:{{ $.PROXY_SSL_PORT }}$request_uri;
```
josegonzalez: This value is set [here](https://github.com/OmgImAlexis/dokku/blob/master/plugins/nginx-vhosts/functions#L298) so you also need to modify that...

plugins/nginx-vhosts/subcommands/migrate
OmgImAlexis: I'll remove it, it was mainly for testing it.

plugins/nginx-vhosts/templates/nginx.conf.sigil
OmgImAlexis: I think I got all of them in the new commit.

josegonzalez: This *still* has to be set in [this map](https://github.com/dokku/dokku/pull/2823/files#diff-a3f91ae60990139f055466b7c67a446bR289). `sigil` doesn't expand the environment into templates.

And I would say we should support the old versions as well.

OmgImAlexis: Since the install script will run at startup it should convert over all existing containers and rebuild the config, no?

josegonzalez: It's for custom `nginx.conf.sigil` files.

OmgImAlexis:
@josegonzalez could you please have a look over this as I'd like to get Caddy support added ASAP.

plugins/nginx-vhosts/functions
```diff
@@ -295,7 +295,9 @@ nginx_build_config() {
           HTTP2_SUPPORTED="$HTTP2_SUPPORTED"
           DOKKU_APP_LISTEN_PORT="$DOKKU_APP_LISTEN_PORT" DOKKU_APP_LISTEN_IP="$DOKKU_APP_LISTEN_IP"
           APP_SSL_PATH="$APP_SSL_PATH" SSL_INUSE="$SSL_INUSE" SSL_SERVER_NAME="$SSL_SERVER_NAME"
-          NGINX_PORT="$NGINX_PORT" NGINX_SSL_PORT="$NGINX_SSL_PORT" RAW_TCP_PORTS="$RAW_TCP_PORTS"
+          # @TODO: Remove this after a few versions
+          NGINX_PORT="$NGINX_PORT" NGINX_SSL_PORT="$NGINX_SSL_PORT"
```
josegonzalez: The values here are now empty, you'll need to change them to `$PROXY_PORT` and `$PROXY_SSL_PORT`.

OmgImAlexis: Done.

OmgImAlexis:
@josegonzalez what needs to happen for this to get merged?

OmgImAlexis:
@josegonzalez what needs to happen for this to get merged?

josegonzalez:
Tests don't pass, so there's that

josegonzalez:
Note: this is a breaking change and will need to go out in a minor release. Can't wait for people to complain about it 🙃

josegonzalez:
I kinda screwed the pooch here. Will re-do this in a separate pr. Sorry about that.

Issue #2835, opened 2017-07-11
Cannon install dokku 0.10.3 on Debian 8.8


MSchmidt:
Debian packages nginx 1.6.2-5+deb8u4 but dokku requires nginx >= 1.8.0.

```
The following packages have unmet dependencies:
 dokku : PreDepends: nginx (>= 1.8.0) or
                     openresty but it is not installable
```

josegonzalez:
You'll need to setup backports to enable 1.10.0+. Nginx 1.6 is too old for us to support. Sorry.

MSchmidt:
That sounds like I could also upgrade to Debian 9 (which I'd like to do) but was hesitant because dokku lists 8.2 specifically as supported.

josegonzalez:
I guess whoever tested 8.2 support - wasn't me, I use Ubuntu 14.04 as thats what we run CI against - had backports enabled. We can certainly update our docs to state that backports are necessary, but I am *not* going to lower the required nginx version, as we need a newer version for certain functionality.

MSchmidt:
Oh sure. Don't lower anything :) I'm fully supportive of newer versions as requirements. I was just stuck in between versions for a second because herokuish updated and dokku didn't, then a build failed and I thought I need the newest dokku. It's all good now. I'm going to upgrade a test host to Debian 9 and will try the greatest and latest dokku there.

fishnux:
I've tested this on Debian 8 and Ubuntu 14.04 and hit the same error. By default, it seems to have nginx 1.4.x (Ubuntu 14.04):
https://packages.ubuntu.com/trusty/nginx

Might be worth to add a warning about that as well.

josegonzalez:
Mind making a doc pull request?

PR #2836, opened 2017-07-12
Added a note on documentation (issue #2835)

fishnux:
Added a small note that points to issue #2835, since some Linux distros appear to install an old nginx version by default.

[ci skip]


joshmanders:
Instead of linking to the issue would it make more sense to explain what's going on and how to fix it?

fishnux:
Should be good now. Open to more suggestions :)

PR #2843, opened 2017-07-18
Starting an FAQ for known pitfalls

stevemacn:
Digital Ocean is a very common platform and the current docs don't work on it. Starting an FAQ at the bottom may help people get this example to work. Each time an issue comes up about this example we should add the garnered insight to this FAQ. [ci skip]


josegonzalez:
Maybe just add to our troubleshooting doc?

stevemacn:
@josegonzalez it is already on the troubleshooting page! 🤦‍♂️

PR #2850, opened 2017-07-26
add optional PROC_TYPE and CONTAINER_INDEX to docker-args-deploy plugn trigger

michaelshobbs:
this will allow plugin developers more flexibility in crafting docker-args for the deploy phase. ex: `--log-opt tag=${APP}.${PROC_TYPE}.${CONTAINER_INDEX}`


josegonzalez:
Nifty. What's an example use case?

michaelshobbs:
ex: `--log-opt tag=${APP}.${PROC_TYPE}.${CONTAINER_INDEX}`

Issue #2857, opened 2017-08-04
nginx incorrectly configured when deploying Docker image with multiple EXPOSE ports


cstroe:
# Description of problem:

When deploying an application straight from a Docker image, if the Docker image contains multiple EXPOSE ports, only the first will be correctly exposed.

# dokku report

```
root@dokku:/home/dokku/gogs# dokku postgres:info gogs-db
=====> Container Information
       Config dir:          /var/lib/dokku/services/postgres/gogs-db/config
       Data dir:            /var/lib/dokku/services/postgres/gogs-db/data
       Dsn:                 postgres://postgres:644e267523f5caa6e497ae749207da0a@dokku-postgres-gogs-db:5432/gogs_db
       Exposed ports:       -                        
       Id:                  1acc7eee87477041ad520e6b5853420f787383044a803911615a46f49b87a04b
       Internal ip:         172.17.0.2               
       Links:               gogs                     
       Service root:        /var/lib/dokku/services/postgres/gogs-db
       Status:              running                  
       Version:             postgres:9.6.1           
root@dokku:/home/dokku/gogs# 
root@dokku:/home/dokku/gogs# 
root@dokku:/home/dokku/gogs# dokku report
-----> uname: Linux dokku 4.4.0-87-generic #110-Ubuntu SMP Tue Jul 18 12:55:35 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
-----> memory: 
              total        used        free      shared  buff/cache   available
Mem:           2000         154         106          17        1739        1620
Swap:             0           0           0
-----> docker version: 
Client:
 Version:      17.06.0-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:23:31 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.06.0-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:19:04 2017
 OS/Arch:      linux/amd64
 Experimental: false
-----> docker daemon info: 
Containers: 3
 Running: 3
 Paused: 0
 Stopped: 0
Images: 11
Server Version: 17.06.0-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 44
 Dirperm1 Supported: true
Logging Driver: json-file
[...]

Issue #2676, opened 2017-03-23
Unused $listener_port in nginx.conf.sigil


callahad:
The nginx.conf.sigil template [defines upstream blocks](https://github.com/dokku/dokku/blob/9ee6c7aab630eaccf20d9858b10302e79e467519/plugins/nginx-vhosts/templates/nginx.conf.sigil#L84) as:

(Reindented for clarity)

```go
{{ if $.DOKKU_APP_LISTENERS }}
  {{ range $upstream_port := $.PROXY_UPSTREAM_PORTS | split " " }}

    upstream {{ $.APP }}-{{ $upstream_port }} {
      {{ range $listeners := $.DOKKU_APP_LISTENERS | split " " }}
        {{ $listener_list := $listeners | split ":" }}
        {{ $listener_ip := index $listener_list 0 }}
        {{ $listener_port := index $listener_list 1 }}
        server {{ $listener_ip }}:{{ $upstream_port }};
      {{ end }}
    }

  {{ end }}
{{ end }}
```

The `$listener_port` variable is unused, and `$upstream_port` is used on the `server` line. Is that correct?

callahad:
It doesn't *seem* right, since `$upstream_port` is an array instead of a single value, but it's been in there for [9 months](https://github.com/dokku/dokku/commit/6c1d6225f9162e33e5af13c30a6ac664374313ab), so maybe it is?

josegonzalez:
Next release should have a fix for this.

max2three:
Before this, I was able to change the default container port from 5000 to 4000 using

`dokku proxy:ports-add node-js-app http:80:4000`.

After upgraded to 0.10.2, my app doesn't work due to #2739 and I need to manually update the `upstream`.

josegonzalez:
@max2three what does it change the port to since 0.10.2?

max2three:
@josegonzalez It uses the `$listener_port` which is 5000.

josegonzalez:
@max2three can you gist the logs of `dokku nginx:build-config` after running `dokku trace on`?

max2three:
@josegonzalez Sorry for the wait, here is the [gist](https://gist.github.com/max2three/6065afbcd75e3767ced51a9109ece6ea) for `dokku nginx:build-config`.

josegonzalez:
Wait why is your `node-js-app` app listening on port `4000`? Is this actually a herokuish app? For those, we expect you to respect the upstream port, which is set to `5000`.

[...]

PR #2858, opened 2017-08-04
Use correct port number for the upstream.

cstroe:
This is a quick fix for #2857 

Investigating how to write tests around it.

### DIFF
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -177,8 +177,7 @@ upstream {{ $.APP }}-{{ $upstream_port }} {
 {{ range $listeners := $.DOKKU_APP_LISTENERS | split " " }}
 {{ $listener_list := $listeners | split ":" }}
 {{ $listener_ip := index $listener_list 0 }}
-{{ $listener_port := index $listener_list 1 }}
-  server {{ $listener_ip }}:{{ $listener_port }};{{ end }}
+  server {{ $listener_ip }}:{{ $upstream_port }};{{ end }}
 }
 {{ end }}
 ```
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -174,8 +174,11 @@ server {
 
 {{ range $upstream_port := $.PROXY_UPSTREAM_PORTS | split " " }}
 upstream {{ $.APP }}-{{ $upstream_port }} {
-{{ range $listener_ip := $.DOKKU_APP_LISTENERS | split " " }}
-  server {{ $listener_ip }}:{{ $upstream_port }};{{ end }}
+{{ range $listeners := $.DOKKU_APP_LISTENERS | split " " }}
+{{ $listener_list := $listeners | split ":" }}
+{{ $listener_ip := index $listener_list 0 }}
+{{ $listener_port := index $listener_list 1 }}
+  server {{ $listener_ip }}:{{ $listener_port }};{{ end }}
 }
 {{ end }}
 ```
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -174,11 +174,8 @@ server {
 
 {{ range $upstream_port := $.PROXY_UPSTREAM_PORTS | split " " }}
 upstream {{ $.APP }}-{{ $upstream_port }} {
-{{ range $listeners := $.DOKKU_APP_LISTENERS | split " " }}
-{{ $listener_list := $listeners | split ":" }}
-{{ $listener_ip := index $listener_list 0 }}
-{{ $listener_port := index $listener_list 1 }}
-  server {{ $listener_ip }}:{{ $listener_port }};{{ end }}
+{{ range $listener_ip := $.DOKKU_APP_LISTENERS | split " " }}
+  server {{ $listener_ip }}:{{ $upstream_port }};{{ end }}
 }
 {{ end }}
 ```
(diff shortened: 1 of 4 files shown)


josegonzalez:
This reverts a previous change. We should instead look at why we're not outputting the correct exposed port in the `DOKKU_APP_LISTENERS`.

cstroe:
The `DOKKU_APP_LISTENERS` variable is constructed [here](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/functions#L221).

I enabled line-by-line debugging for bash and got the following:

```
root@dokku:/var/lib/dokku/plugins/enabled/nginx-vhosts# dokku proxy:ports-add gogs http:333:333
-----> Setting config vars
       DOKKU_PROXY_PORT_MAP: http:2222:22 http:3000:3000 http:333:333
+ [[ -z '' ]]
+ [[ -z '' ]]
+ shopt -s nullglob
+ local DOKKU_APP_IP_FILE
+ for DOKKU_APP_IP_FILE in '$DOKKU_ROOT/$APP/IP.web.*'
+ local DOKKU_APP_PORT_FILE=/home/dokku/gogs/PORT.web.1
+ local DOKKU_APP_LISTENER_IP=172.17.0.5
+ local DOKKU_APP_LISTENER_PORT=22
+ local 'DOKKU_APP_LISTENERS+=172.17.0.5:22 '
++ echo '172.17.0.5:22 '
++ xargs
+ local DOKKU_APP_LISTENERS=172.17.0.5:22
+ shopt -u nullglob
+ set +x
-----> Configuring gogs.dokku.lan...(using built-in template)
       Reloading nginx
```

So, because there's only one file that matches `'$DOKKU_ROOT/$APP/IP.web.*'`, the second port never gets detected.

In the app directory there is indeed only one port detected:
```
root@dokku:/home/dokku/gogs# ls -la
total 60
drwxr-xr-x 2 dokku dokku 4096 Aug  4 04:34 .
drwxr-xr-x 5 dokku dokku 4096 Aug  4 02:43 ..
-rw-r--r-- 1 dokku dokku   65 Aug  4 03:14 CONTAINER.web.1
-rw-r--r-- 1 dokku dokku  208 Aug  4 03:14 DOCKERFILE_ENV_FILE
-rw-r--r-- 1 dokku dokku   53 Aug  4 02:37 DOCKER_OPTIONS_BUILD
-rw-r--r-- 1 dokku dokku  124 Aug  4 02:38 DOCKER_OPTIONS_DEPLOY
-rw-r--r-- 1 dokku dokku  100 Aug  4 02:38 DOCKER_OPTIONS_RUN
-rw-r--r-- 1 dokku dokku    6 Aug  4 02:41 DOKKU_SCALE
-rw------- 1 dokku dokku  259 Aug  4 04:34 ENV
-rw-r--r-- 1 dokku dokku   11 Aug  4 03:14 IP.web.1
-rw------- 1 dokku dokku 4922 Aug  4 04:34 nginx.conf
-rw-r--r-- 1 dokku dokku    3 Aug  4 03:14 PORT.web.1
-rw-r--r-- 1 dokku dokku  104 Aug  4 04:34 URLS
-rw-r--r-- 1 dokku dokku   15 Aug  4 02:41 VHOST
```

cstroe:
After more debugging it seems that there is a 1:1 relationship between an `IP.web.` file and a container running an app.  Since I'd never need to scale Gogs to more than one instance, only `IP.web.1` will ever exist.

So `DOKKU_APP_LISTENERS` will only ever contain one entry per container, even though each container might `EXPOSE` more than one port.

Since we're configuring the app with proxy commands in the first place, it might make more sense to use `DOKKU_PROXY_PORT_MAP`:

```
root@dokku:/home/dokku# dokku config:get gogs DOKKU_PROXY_PORT_MAP
http:2222:22 http:3000:3000
```

And this is the `sigil` command:

```
+ sigil -f /var/lib/dokku/plugins/available/nginx-vhosts/templates/nginx.conf.sigil 
  APP=gogs DOKKU_ROOT=/home/dokku NOSSL_SERVER_NAME= 
  DOKKU_APP_LISTENERS=172.17.0.3:22 DOKKU_LIB_ROOT=/var/lib/dokku 
  PASSED_LISTEN_IP_PORT= SPDY_SUPPORTED=false HTTP2_SUPPORTED=false 
  DOKKU_APP_LISTEN_PORT= DOKKU_APP_LISTEN_IP= APP_SSL_PATH=/home/dokku/gogs/tls 
  SSL_INUSE= SSL_SERVER_NAME= NGINX_PORT= NGINX_SSL_PORT= 'RAW_TCP_PORTS=22 3000' 
  'PROXY_PORT_MAP=http:2222:22 http:3000:3000' 'PROXY_UPSTREAM_PORTS=22 3000'
```

# Possible approaches
1. Use `PROXY_UPSTREAM_PORTS` to build the upstream section of the `nginx.conf` and stop using `DOKKU_APP_LISTENERS` there.
2. Something else?

cstroe:
This issue was caused by the fix from https://github.com/dokku/dokku/issues/2676

In #2676, it seems that rather than removing the unused variable `$listener_port`, it was decided to make it used again.  But since the `DOKKU_APP_LISTENERS` variable has a different meaning (it's only reliable to extract only the IP of each upstream container, not the upstream ports), the port should still be configured via `$upstream_port`, which is correctly parsed from `PROXY_UPSTREAM_PORTS`.

cstroe:
@josegonzalez I made the fix to `nginx.conf.sigil` and related code in `test_helper.bash`.

I'm going to try to write a test around this.

gastonmorixe:
Thank you.

PR #2861, opened 2017-08-04
Use non-deprecated apps command

adelq:
[ci skip]

### DIFF
--- a/docs/getting-started/upgrading.md
+++ b/docs/getting-started/upgrading.md
@@ -24,7 +24,7 @@ If Dokku was installed via `apt-get install dokku` or `bootstrap.sh` (most commo
 
 ```shell
 sudo apt-get update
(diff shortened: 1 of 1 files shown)


josegonzalez:
I think apps:list is fairly new so users of older versions will have trouble with this change. Might be worth mentioning that instead.

wiredmind:
This commit applies to the latest version, doesn't it? In which case this should not cause confusion, but perhaps there should be a note for users of older versions to check the documentation for their version.

josegonzalez:
> I think apps:list is fairly new so users of older versions will have trouble with this change. Might be worth mentioning that instead.

josegonzalez:
I clarified the docs since no one seemed interested in doing so. Thanks for the initial contribution though!

PR #2872, opened 2017-08-15
(re)Added SSL protocols TLSv1, TLSv1.1

facundomedica:
Not so old Androids are having problems connecting. (Android 4.2.2 and Android 4.3).

More info: I work with mobile apps APIs, so I'm seeing this failing quite often

### DIFF
--- a/plugins/nginx-vhosts/templates/nginx.conf.sigil
+++ b/plugins/nginx-vhosts/templates/nginx.conf.sigil
@@ -65,7 +65,7 @@ server {
 
   ssl_certificate     {{ $.APP_SSL_PATH }}/server.crt;
   ssl_certificate_key {{ $.APP_SSL_PATH }}/server.key;
-  ssl_protocols       TLSv1.2;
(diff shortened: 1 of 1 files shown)


facundomedica:
This commit only is not enough, apparently, and I've seen another pull request that was merged talking about PCI compliance, so I may need to do a downgrade

josegonzalez:
I'm closing this as we're trying to ensure that Dokku is secure by default. I do understand that older browsers and mobile devices are affected - I went through this migration a few years ago - but as the proportion of older browser/device issues decrease, we feel as though this is a good time to make this tradeoff.

As documented in the [migration guide](http://dokku.viewdocs.io/dokku/appendices/0.10.0-migration-guide/), you can just vendor the `nginx.conf.sigil` file in your repository. Note that you may also need to set custom ssl ciphers as well in order to support certain devices, and that not all devices will play nice with newer SSL certificates - Samsung is the big offender here...

PR #2875, opened 2017-08-16
Fix deprecated command invocation

wiredmind:
Invoking `dokku apps` has been deprecated, use `dokku apps:list`

[ci skip]


josegonzalez:
Duplicate of #2861, so reposting this comment:

> I think apps:list is fairly new so users of older versions will have trouble with this change. Might be worth mentioning that instead.

wiredmind:
Duplicate of #2861

PR #2890, opened 2017-08-29
use circleci 2.0

michaelshobbs:
last build time was 20:53. down from the last master build time of 1:47:35


.circleci/config.yml
```diff
@@ -0,0 +1,29 @@
...
+    parallelism: 4
+    steps:
+      - checkout
+      - run: |
+          make ci-dependencies
+      - run: |
+          ./tests/ci/setup.sh
+      - run: |
+          sudo apt-get remove -qq -y git
+          sudo apt-get install -qq -y git=1:1.9.1-1ubuntu0.6 git-man=1:1.9.1-1ubuntu0.6
```
josegonzalez: Woof but I get it.

tests/ci/setup.sh
```diff
@@ -0,0 +1,27 @@
...
+  sudo -E CI=true make -e sshcommand
+  # need to add the dokku user to the docker group
+  sudo usermod -G docker dokku
+  [[ "$1" == "buildstack" ]] && BUILD_STACK=true make -e stack
+  sudo -E CI=true make -e install
+  sudo -E make -e setup-deploy-tests
+  bash --version
+  docker version
+  lsb_release -a
+  # setup .dokkurc
```
josegonzalez: Wait does dokku now require this directory? I thought it was optional?

.circleci/config.yml
michaelshobbs: Yup....

tests/ci/setup.sh
michaelshobbs: This has always been there. We place some CI specific settings there in `.circleci/config.yml`

PR #2900, opened 2017-09-07
refactor: split-up build process into "builder" functionality

josegonzalez:
This will allow users the ability to customize the build process by replacing herokuish/dockerfile builds with something custom, such as a lambda function builder.

Totally possible I don't merge this in.

@michaelshobbs this is just a POC, was wondering if you had thoughts on whats going on here.


michaelshobbs:
@josegonzalez what exactly is your question?

josegonzalez:
Assuming I can get tests to pass - thats a big if - do you have concerns about the way this is headed?

michaelshobbs:
I'm 👍 on modularizing the build process

PR #2906, opened 2017-09-18
fix: Check the location of nginx in a central way

manuel-colmenero:
- Now there's only a "single source of truth" for the location of the nginx
  binary.
- It is checked if it's there before it's needed.

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -6,14 +6,29 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
 source "$PLUGIN_AVAILABLE_PATH/domains/functions"
 source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
+get_nginx_location() {
+  declare desc="check that nginx is at the expected location and return it"
+  local NGINX_LOCATION="/usr/sbin/nginx"
+  if ! [ -x "$NGINX_LOCATION" ]; then
+    dokku_log_fail "Could not find nginx binary at ${NGINX_LOCATION}."
+    exit 1;
+  fi
+  echo "$NGINX_LOCATION"
+}
+
 validate_nginx() {
   declare desc="validate entire nginx config"
+  local NGINX_LOCATION=$(get_nginx_location)
+  if [ -z "$NGINX_LOCATION" ]; then
+    exit 1;
+  fi
+
   set +e
-  sudo /usr/sbin/nginx -t > /dev/null 2>&1
+  sudo "$NGINX_LOCATION" -t > /dev/null 2>&1
   local exit_code=$?
   set -e
   if [[ "$exit_code" -ne "0" ]]; then
-    sudo /usr/sbin/nginx -t
+    sudo "$NGINX_LOCATION" -t
     shopt -s nullglob
     local conf_file
     for conf_file in $DOKKU_ROOT/*/nginx.conf; do
@@ -152,11 +167,6 @@ is_spdy_enabled() {
   local MAJOR_VERSION MINOR_VERSION PATCH_VERSION
   local HAS_SUPPORT=true
 
-  if ! which nginx > /dev/null 2>&1; then
-    echo $HAS_SUPPORT
-    return
-  fi
-
   MAJOR_VERSION=$(echo "$NGINX_VERSION" | awk '{split($0,a,"."); print a[1]}')
   MINOR_VERSION=$(echo "$NGINX_VERSION" | awk '{split($0,a,"."); print a[2]}')
   PATCH_VERSION=$(echo "$NGINX_VERSION" | awk '{split($0,a,"."); print a[3]}')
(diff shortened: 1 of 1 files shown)


plugins/nginx-vhosts/functions
```diff
@@ -6,14 +6,29 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
 source "$PLUGIN_AVAILABLE_PATH/domains/functions"
 source "$PLUGIN_AVAILABLE_PATH/ps/functions"
 
+get_nginx_location() {
+  declare desc="check that nginx is at the expected location and return it"
+  local NGINX_LOCATION="/usr/sbin/nginx"
```
josegonzalez: This breaks it for people who package their own nginx. Why isn't `nginx` in the `$PATH` for you? It is for me.

plugins/nginx-vhosts/functions
```diff
@@ -6,14 +6,29 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
...
+  if ! [ -x "$NGINX_LOCATION" ]; then
```
josegonzalez: This should be `if [[ -z "$NGINX_LOCATION" ]]; then`

plugins/nginx-vhosts/functions
```diff
@@ -6,14 +6,29 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
...
+    dokku_log_fail "Could not find nginx binary at ${NGINX_LOCATION}."
+    exit 1;
+  fi
+  echo "$NGINX_LOCATION"
+}
+
 validate_nginx() {
   declare desc="validate entire nginx config"
+  local NGINX_LOCATION=$(get_nginx_location)
+  if [ -z "$NGINX_LOCATION" ]; then
```
josegonzalez: Double-quotes.

plugins/nginx-vhosts/functions
```diff
@@ -281,7 +286,11 @@ nginx_build_config() {
       done
     fi
 
-    local NGINX_VERSION="$(nginx -v 2>&1 | cut -d'/' -f 2)"
+    local NGINX_LOCATION=$(get_nginx_location)
+    if [ -z "$NGINX_LOCATION" ]; then
```
josegonzalez: Double-quotes.

plugins/nginx-vhosts/functions
manuel-colmenero: Well it was already broken for them, then. In the function `validate_nginx`, the sbin path is hardcoded in mainline, which means their config can never be valid.

plugins/nginx-vhosts/functions
manuel-colmenero: How could `$NGINX_LOCATION` empty at that point? It's set right above that. 

Also, what I'm doing here is checking that `niginx` is executable for the current user, not finding out if the variable is empty.

plugins/nginx-vhosts/functions
josegonzalez: Actually not really. I have nginx installed in `/usr/local/bin` - custom compiled from openresty - and it works fine. Hardcoding to the `sbin` path also invalidates any special `$PATH` setting.

manuel-colmenero: Are you sure? If you run `which nginx` as the `dokku` user inside the Vagrant box, it does find it? I checked both on my server and in a Vagrant box, and that was not the case for me.

This is because `/sbin` and `/usr/sbin` are not in the path of regular user (not administrators) by default in Debian, AFAIK.

plugins/nginx-vhosts/functions
manuel-colmenero: Fair enough.

plugins/nginx-vhosts/functions
manuel-colmenero: It's hardcoded upstream right here:
https://github.com/dokku/dokku/blob/3a81c8f279ff823cde1ae7ce0ac4af994d1fd981/plugins/nginx-vhosts/functions#L12

As for the custom path, I think the best option here would  be for `get_nginx_location` both search in path with `which` and check the "standard" location even if it's not in path as a backup. Thoughts?

josegonzalez: Fancy, I guess I must also symlink it.

We should probably check the `$PATH` first and then the `/usr/sbin/nginx` location.

plugins/nginx-vhosts/functions
```diff
@@ -8,18 +8,22 @@ source "$PLUGIN_AVAILABLE_PATH/ps/functions"
...
   declare desc="check that nginx is at the expected location and return it"
-  local NGINX_LOCATION="/usr/sbin/nginx"
-  if ! [ -x "$NGINX_LOCATION" ]; then
-    dokku_log_fail "Could not find nginx binary at ${NGINX_LOCATION}."
-    exit 1;
+  local NGINX_LOCATION=$(which nginx 2>/dev/null)
+
+  if [[ -z "$NGINX_LOCATION" ]]; then
+    NGINX_LOCATION="/usr/sbin/nginx"
+    if ! [[ -x "$NGINX_LOCATION" ]]; then
```
josegonzalez: You might want to move this out of the containing if statement, and also turn it into:

```
if [[ ! -x "$NGINX_LOCATION" ]]; then
```

manuel-colmenero: Done and done.

josegonzalez:
Please fix all the stickler-ci issues.

manuel-colmenero:
There you go, the linters are happy again.

My bad, I overlooked the `CONTRIBUTING.md` file. It might be helpful for new contributors if it was linked in the `README.md` under "Contribution".

josegonzalez:
Pull requests welcome.

PR #2928, opened 2017-10-02
fix: do not overwrite the VHOST file during installation if it exists

### LINKED ISSUE #2927: On Ubuntu `apt-get upgrade dokku` system overwrites /home/dokku/VHOST
On Ubuntu 16.04 `apt-get upgrade dokku` system command overwrites /home/dokku/VHOST so I need to re-do `dokku domains:set-global mydomain.tld`

silverfix:
Fix to prevent dokku upgrade on debian systems overwrites ${DOKKU_ROOT}/VHOST

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -109,7 +109,7 @@ case "$1" in
           rm -f "${DOKKU_ROOT}/VHOST"
       fi
     fi
-    
+
     db_get "dokku/hostname"
     echo "$RET" > "${DOKKU_ROOT}/HOSTNAME"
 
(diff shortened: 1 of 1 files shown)


debian/postinst
```diff
@@ -100,14 +100,16 @@ case "$1" in
       exit 0
     fi
 
-    db_get "dokku/vhost_enable"
-    if [ "$RET" = "true" ]; then
-        db_get "dokku/hostname"
-        echo "$RET" > "${DOKKU_ROOT}/VHOST"
-    else
-        rm -f "${DOKKU_ROOT}/VHOST"
+    if [ ! -f "${DOKKU_ROOT}/VHOST" ]; then
```
josegonzalez: This

PR #2935, opened 2017-10-09
docs: WIP on explaining optional herokuish

jcrben:
Per our conversation on Slack @josegonzalez - were you perhaps referring to the NO_INSTALL_RECOMMENDS in https://github.com/dokku/dokku/blob/3f50097619b3de2930b198d914b8dc1605816097/bootstrap.sh#L106 since herokuish is just a [recommended package](https://github.com/dokku/dokku/blob/3f50097619b3de2930b198d914b8dc1605816097/debian/control#L7)? Seems less optional in centos and also not sure the effect of not installing the other recommended package (parallel).

Seems that maybe it'd be best to point to this trick about about a `-` suffix https://serverfault.com/a/663803/309584 and provide a way to pass it through? Don't want people to skip recommends just to skip herokuish as the list of recommends should be decoupled.

Not real sure about `if [[ -n $DOKKU_DOCKERFILE ]]; then` in bootstrap.sh - haven't found where that's documented or how the name relates to what it seems to do (skipping recommends).

[ci skip]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.rpm
 .DS_Store
 .ruby-version
+.idea/
 .vagrant
 data/
 stack.tgz
--- a/docs/getting-started/advanced-installation.md
+++ b/docs/getting-started/advanced-installation.md
@@ -44,7 +44,9 @@ cd dokku
 sudo BUILD_STACK=true STACK_URL=https://github.com/gliderlabs/herokuish.git make install
 ```
 
-Herokuish is recommended but not required. For Debian, `sudo DOKKU_NO_INSTALL_RECOMMENDS=true bash bootstrap.sh` is one way to skip the dependency, although it will also skip other recommended dependencies.
+## Skipping herokuish installation
+
+The `herokuish` package is recommended but not required if not using Heroku Buildpacks for deployment. Debian-based OS users can run the bootstrap installer via `sudo DOKKU_NO_INSTALL_RECOMMENDS=true bash bootstrap.sh` to skip the dependency. Please note that this will _also_ skip installation of other recommended dependencies.
 
 ## Configuring
 
--- a/docs/getting-started/advanced-installation.md
+++ b/docs/getting-started/advanced-installation.md
@@ -44,6 +44,8 @@ cd dokku
 sudo BUILD_STACK=true STACK_URL=https://github.com/gliderlabs/herokuish.git make install
 ```
 
+Herokuish is recommended but not required. For Debian, `sudo DOKKU_NO_INSTALL_RECOMMENDS=true bash bootstrap.sh` is one way to skip the dependency, although it will also skip other recommended dependencies.
+
 ## Configuring
 
 Once Dokku is installed, if you are not using the web-installer, you'll want to configure a the virtualhost setup as well as the push user. If you do not, your installation will be considered incomplete and you will not be able to deploy applications.


josegonzalez:
I don't really like adding this to these docs, and would rather have a note in the advanced debian installation notes.

jcrben:
That makes sense. But how about the approach? Are you thinking `--no-install-recommends` or should a way to pass the `-` suffix be exposed to specifically exclude herokuish?

Maybe there could be a table explaining the parameters that `bootstrap.sh` can take and which systems they apply towards - or perhaps listing the options at the top of the file with an explanation for each. `bootstrap.sh` is a lot like a function.

josegonzalez:
There is no way to exclude just herokuish. Everything we recommend is just that - a recommendation.

To be quite fair, the bootstrap installation is "easy mode" installation. You shouldn't skip something in that mode, and anyone who wants more control over what gets installed - and any options around it - should defer to using a config management tool that installs the repository/packages separately.

The `DOKKU_DOCKERFILE` bit was added to make the initial work towards installing Dokku _within_ a docker container easier, but no one has - to my knowledge - publicly continued on that work.

jcrben:
I moved the note over to the advanced install. Haven't had the chance to actually test that suggested command. What do you think?

josegonzalez:
I forget why we want to document this. If you can't install herokuish from source, how would you install Dokku?

jcrben:
I wanted to skip herokuish is because I'm  never going to use it. I use docker builds only. Herokuish seems a bit weird - not all that clear on how it works, plus it's pretty weak. In comparison, docker is well-documented, widely understood and extremely powerful.

One prompt was seeing the message that installation of herokuish will take 5 minutes.

It sort of sounds like you consider herokuish to be a required dependency? I can close this if you'd  like.

josegonzalez:
Herokuish allows detection and application of the correct heroku buildpacks. I _do_ consider it optional, but I don't think the 80% use case is to not install herokuish.

I'm curious as to what you think isn't well documented or is "weak" about herokuish. I will admit that path isn't for everyone, but I can see where you're coming from.

@michaelshobbs didnt we change it to just pull the image recently?

jcrben:
As far as the 80% use case, this is in the advanced install notes...

> I'm curious as to what you think isn't well documented or is "weak" about herokuish.

When I first set up dokku a couple years ago (June 2015), I couldn't figure out how to make them flexible. I can't remember all the details but I wrote in [a blog post at the time](https://web.archive.org/web/20161024144140/http://bencreasy.com/dokku-docker-and-deployment/) that "After some maneuvering, I determined that the buildpacks were simply too inflexible and moved to a Dockerfile. The buildpack method results in a Docker container, but as this post there is no official way to reverse engineer a container or image into a Dockerfile".

It seems that [the deployment doc](https://github.com/dokku/dokku/blob/3684c10a78d21beec73b7a4982ee67c706c20fa4/docs/deployment/application-deployment.md#deploy-the-app) links out to a [heroku page on buildpacks](https://devcenter.heroku.com/articles/buildpacks) these days, but it seems to be a loose dependency since dokku links to an unversioned page and there's little explanation on the dokku side that I can find about how buildpacks work and why I should use them, except perhaps the herokuish readme which doesn't say much.

In general, I try to learn and use things that are reusable - I use docker at work - and also try to lock dependencies down very tightly and eliminate all dependencies on proprietary moving targets. I'm sort of open to learning new things but I kinda need motivation. I also anticipate that using dockerfiles will make it easier to switch out into multi-host architecture down the line.

josegonzalez:
I'm merging this, but after thinking for a while, I think it is a bit silly. Installation adds ~1GB of disk usage and doesn't change your overall usage.

One funny thing about this change is that should you use `herokuish` later - whether by accident or on purpose - it will pull latest master (thanks Docker!), and not necessarily whatever was tested working with your current release of Dokku. That's definitely a quirk of how we reference herokuish - allows users to update their base image out of release band - but something I figured was worth mentioning.

jcrben:
OK. If you'd rather remove it later, feel free - you're the one putting in the work.

To be fair, there is some documentation on buildpacks at https://github.com/dokku/dokku/blob/9fcd9b00659fdc54385b830bfc22299baea3c74b/docs/deployment/methods/buildpacks.md but the overall meta discussion of the costs and benefits of buildpacks versus Dockerfiles still seems to be missing.

I suppose a large part of it depends on whether or not someone knows more about heroku or docker. I'm also guessing that people using buildpacks are using local or vagrant-based dev environments

josegonzalez:
I don't think thats necessarily true. A large part of the value-add of something like Heroku is that a developer doesn't need to think about server dependencies. Just like many developers never get past working on small changes to wordpress plugins - if that - many developers will also never get past buildpacks which automatically detect your application's requirements.

The primary value-add of Dokku is that you get the Heroku workflow but do not need to pay the Heroku pricing.

As per whether we have a meta discussion on the advantages and disadvantages of either system, we mention in our dockerfile docs:

> Dockerfile support is considered a Power User feature. By using Dockerfile-based deployment, you agree that you will not have the same comfort as that enjoyed by Buildpack users, and Dokku features may work differently. Differences between the two systems will be documented here.

Personally, I think thats good enough for our documentation. There are many advantages and disadvantages to Buildpack vs Dockerfile usage, and the goal of the project is to be flexible and target the 80% case by default. Buildpacks get our users there, Dockerfiles do not. And while ideally every developer knows exactly how their application is built and run, my professional experience tells me most simply want to work on their code and be done with it.

Having a meta discussion about something like this is better suited to a blog post than to docs. I'd rather we have sane defaults for the majority of our users, and then let power users build on top of the system if they wish.

Hope that makes sense, and happy to discuss further via slack.

PR #2951, opened 2017-10-21
fix: handle case where the app directory is a symlink

### LINKED ISSUE #2943: Place apps on diffrent harddrives
My current setup has 2 harddrives, on which the first holds the OS, Dokku and all apps and the other being empty. I'm wondering if Dokku supports placing the app on custom locations,  something like `apps:create fooApp -l /some/path` would be super handy.

In general, is there any possibility to specify a custom location or move an existing app to another folder/drive?

josegonzalez:
Closes #2943


josegonzalez:
Ugh.

PR #2958, opened 2017-10-27
Add tutorial for running Dokku on an external volume

silverfix:
[ci skip]


docs/advanced-usage/run-on-external-volume.md
```diff
@@ -0,0 +1,75 @@
...
+the following is an easy tutorial to achieve Dokku runs on them.
+
+## Tutorial
+The following is intended to be executed on the dokku host machine as `root`.
+Say, _for instance_, that our volume is mapped into the systems as `/dev/vdb1`.
+
+Stop docker daemon
+
+```shell
+services docker stop
```
josegonzalez: This ties us to upstart, and most/all distributions are moving to systemd.

silverfix:
OK @josegonzalez thank you for the feedback. Now solved in the last commit. Any other issues?

josegonzalez:
That doesn't actually solve the change. What I'm saying is that we have different init systems that Dokku supports, and this post assumes quite a bit about your installation.

josegonzalez:
Note: thats probably fine, but this almost feels like we should have a "tutorials" section, or that this is better suited for a blog post (official or not).

silverfix:
Oh.. ok..  I'm sorry for that. I don't know, if you want you can reject the
PR without concerns :)

Thank you for your feedback.
Andrea

josegonzalez:
No its fine, I'm just thinking aloud.

My reasoning for not having this in our default docs - even under advanced installation - is that I don't think this is a very common thing to do at all.

A better way around this would be for us to support sourcing an `/etc/defaults/dokku` file around [here](https://github.com/dokku/dokku/blob/master/dokku#L4), [here](https://github.com/dokku/dokku/blob/master/debian/postinst#L8) and [here](https://github.com/dokku/dokku/blob/master/rpm/dokku.postinst#L53). You could imagine the following being set to something within your mounted volume:

- `DOKKU_ROOT`: usually `~dokku`
- `DOKKU_LIB_ROOT`: usually `/var/lib/dokku`

*That* could go under our advanced installation docs, and I think would require quite a bit less dokku-specific documentation. You'd then really just be documenting this workflow for Docker, and I'm hopeful that *that* would be better documented on their site :)

josegonzalez:
Thanks for the contribution!

silverfix:
Thank you guys for your awesome project!

Issue #2984, opened 2017-11-14
Weak Diffie-Hellman key exchange parameters


bitmand:
Apps using the Lets Encrypt Dokku plugin suffer from weak Diffie-Hellman (DH) key exchange parameters. A SSL Labs test results in grade B just because the server is setup using common DH primes:

<img width="969" alt="ssllabs" src="https://user-images.githubusercontent.com/344214/32777779-cdcb83ee-c937-11e7-9455-a4e32911598b.png">

As far as I can tell, the Lets Encrypt plugins does not handle nginx ssl settings, but instead relies on the Certs plugin to manage those. Although I haven't tested with the certs plugin, I would expect it to have the same problem.

This is on Ubuntu 16.04.

An example on how to fix, could be to once build a global custom dh parameters files with:

    openssl dhparam -out /etc/nginx/dhparam.pem 2048

And once include it in the default dokku nginx file `/etc/nginx/conf.d/dokku.conf` by adding this line:

    ssl_dhparam /etc/nginx/dhparam.pem;

I'll happily test, provide more info or anything else if needed.

dokku report results:
```
-----> memory:
              total        used        free      shared  buff/cache   available
Mem:           7980         477        2709          50        4793        7235
Swap:           511           0         511
-----> docker version:
Client:
 Version:      17.10.0-ce
 API version:  1.33
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:04:16 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.10.0-ce
 API version:  1.33 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:02:56 2017
 OS/Arch:      linux/amd64
 Experimental: false
-----> docker daemon info:
Containers: 19
 Running: 6
 Paused: 0
 Stopped: 13
Images: 32
Server Version: 17.10.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
[...]

PR #2985, opened 2017-11-14
Build a custom dhparam file once for nginx and include it as default

bitmand:
MR to solve #2984 

Please be aware that this is mostly harmless, but:

* It depends on the `openssl` command
* It will take a little bit of time and generate some output the first time this is run
* Also, I have no idea if this is actually the preferred Dokku way of solving this :)

### DIFF
--- a/plugins/nginx-vhosts/install
+++ b/plugins/nginx-vhosts/install
@@ -26,6 +26,12 @@ esac
 
 chmod 0440 /etc/sudoers.d/dokku-nginx
 
+# if dhparam.pem has not been created, create it the first time
+if [[ ! -f /etc/nginx/dhparam.pem ]]; then
+  openssl dhparam -out /etc/nginx/dhparam.pem 2048
+  chown root:root /etc/nginx/dhparam.pem
+fi
+
 # if dokku.conf has not been created, create it
 if [[ ! -f /etc/nginx/conf.d/dokku.conf ]]; then
   mkdir -p /etc/nginx/conf.d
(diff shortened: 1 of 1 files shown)


bitmand:
I checked the failed circleci build, but it does not seem related to my changes.

josegonzalez:
Does this break existing SSL certificates?

bitmand:
@josegonzalez No it does not.

The actual SSL Labs test is here, with more information:
https://www.ssllabs.com/ssltest/analyze.html?d=gorm.longboat.io&s=139.162.162.128

The security issue explained: https://weakdh.org/

Also some information on how to fix: https://weakdh.org/sysadmin.html - specifically:

> Use a Strong, Diffie Hellman Group. A few 1024-bit groups are used by millions of servers, which makes them an optimal target for precomputation, and potential eavesdropping. Administrators should use 2048-bit or stronger Diffie-Hellman groups with "safe" primes.

..that's why if Dokku just generates this file randomly per installation and includes it in the nginx config, it fixes the problem :)

josegonzalez:
Right but if we add a new file, does that break anyone's already provisioned ssl certificates? And how does that impact people overriding this one-off at a different level? Want to make sure we have that well documented.

bitmand:
@josegonzalez 

> Right but if we add a new file, does that break anyone's already provisioned ssl certificates?

No it doesn't. This isn't related to the certificates, neither already installed or in the future.
It's a setting for the TLS handshake and the encryption of the TLS connection, that fixes a security issue.

> And how does that impact people overriding this one-off at a different level? Want to make sure we have that well documented.

Are you are thinking of the cases where people have already fixed this with custom nginx config? It should be able to co-exist (both in the http and server settings), but I can test this, if you want me to?

josegonzalez:
Please test

PR #3002, opened 2017-12-12
fix: omit redirection of docker build output

josegonzalez:
@michaelshobbs should we just make this be a `docker pull` of the correct image?


michaelshobbs:
> @michaelshobbs should we just make this be a docker pull of the correct image?

 what is "the correct image"?

josegonzalez:
Yeah

michaelshobbs:
Yes. I think that makes sense to do.

PR #3011, opened 2017-12-15
[config] forward output from plugn triggers to user

alexquick:
The config plugin was suppressing the output from `post-config-update` and `app-restart`

### DIFF
--- a/plugins/common/common.go
+++ b/plugins/common/common.go
@@ -345,13 +345,12 @@ func VerifyImage(image string) bool {
 }
 
 //PlugnTrigger fire the given plugn trigger with the given args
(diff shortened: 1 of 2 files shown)


josegonzalez:
Guess there isn't a good way to stream the output to stdout/stderr?

alexquick:
As it is now, yeah there's no way to do it.  This is the only file that uses `PlugnTrigger`, so I can change the implementation to always stream stderr/stdout if that's preferable.

josegonzalez:
If you have a moment, that would be great.

alexquick:
@josegonzalez how's this look?

josegonzalez:
Thanks for the pull request!

PR #3023, opened 2017-12-28
restore only apps with DOKKU_APP_RESTORE=1 when parallel is installed

vincentfretin:
Properly check DOKKU_APP_RESTORE=1 when restoring all apps in parallel, like it's done in sequencial mode.

`sem` is one of the commands installed by the `parallel` package. `j+0` means number of CPU cores.


josegonzalez:
Seems like this doesn't quite pass tests :( Mind taking a look at that/

vincentfretin:
I don't quite understand what is failing here.
How do I execute the tests on my machine?

vincentfretin:
From the circleci output, do you understand what test is failing, which file?

josegonzalez:
Says the `ps:restore` tests are failing. There should be instructions in our docs on running tests locally so you can try that test out in a VM :)

josegonzalez:
Closed by #3028.

Issue #2247, opened 2016-06-14
Remove the web installer in favor of setting up via cli commands


u2mejc:
Description of problem:
- Default install of dokku exposes an insecure dokku admin access setup page on all interfaces.

How was dokku installed?:
- boostrap.sh

How reproducible:
- Fresh bootstrap.sh install

Additional info:
- From http://dokku.viewdocs.io/dokku/installation/:

> If you don't complete setup via the web installer (even if you set up SSH keys and virtual hosts otherwise) your dokku installation will remain vulnerable to anyone finding the setup page and inserting their key.

Allot of users use dokku on public facing VMs, the default install doesn't listen for a vhost and instead responds to any request for `/` on port 80. Should we add a secure option for the "attended" installs? Some new users ~~may~~ appear not to be aware if they haven't pushed an app yet and are only interacting with the CLI.

Possible changes:
- Add scary warning to CLI, everytime dokku is execute from CLI
- Randomly generated password on CLI for basic http auth during install (reset with a reinstall)
- Add CLI prompts for key / vhost settings (CLI or with the Ubuntu `dialog` command)
- Don't start the webserver until it's activated by a command inside the dokku CLI
- Add documentation around not starting dokku setup and instead document how to manually add ssh key / set hostname preference.

josegonzalez:
👍 

@u2mejc want to start on a PR for these set of changes?

u2mejc:
@josegonzalez happy to, which ones would you like to see in it?

josegonzalez:
> Add scary warning to CLI, everytime dokku is execute from CLI
> Add CLI prompts for key / vhost settings (CLI or with the Ubuntu dialog command)


@michaelshobbs thoughts on dropping the web installer completely and making people set it up on the cli?

michaelshobbs:
I'm 👍 on dropping the web installer and wrapping setup in a CLI

josegonzalez:
I wonder what the backlash will be on doing that...

u2mejc:
> I wonder what the backlash will be on doing that...

I believe non-existent. Dokku is a cli tool, but for some reason is includes like two settings you have to switch away from to cli into a website? If there is backlash, I expect it to be from the camp that says this is taking us further away from having a web gui someday, but I disagree that this would even be a building stone. I'll whip up a PR.

u2mejc:
Checklist till completion: 
- [x] Update sshcommand to support listing - PR: https://github.com/dokku/sshcommand/pull/29
[...]

PR #3041, opened 2018-01-21
Point to unattended install instructions

jcrben:
[ci skip]

> +*You should also stop and disable the `dokku-installer` service to remove public access to adding SSH keys.*

Could we get more information (link out if verbose) on the best way to do this?

> Otherwise you have to create the file manually and set it to your preferred domain

Hmm, instructions which tell people to write a file don't seem that that durable - it seems like newcomers could easily run into weird permissions errors and whatnot. Does `dokku:domains set-global` help here? Should we have a cross-platform way to set this?

related: https://github.com/dokku/dokku/issues/2247


josegonzalez:
I'll take care of the followups separately :)

Issue #3036, opened 2018-01-17
Unable to unset invalid config


prichey:
This is a pretty weird edge case which may or may not be an actual bug, I'm just not sure what to do about it. My temporary solution is to downgrade to 0.10.5, which works ok, but probably not a long-term solution.

In upgrading to 0.11.3 (I think from 0.10.5 but not totally sure), one of my app builds started breaking like so:

```
root@prichey:~# dokku ps:rebuild appname
-----> Cleaning up...
-----> Building appname from herokuish...
/var/lib/dokku/core-plugins/available/common/functions: line 448: export: --: invalid option
export: usage: export [-fn] [name[=value] ...] or export -p
```

After enabling trace, I found this:

```
++ /var/lib/dokku/plugins/available/config/subcommands/export config:export appname
++ return 0
+ eval 'export --no-restart='\''--no-restart'\''
export DOKKU_APP_RESTORE='\''1'\''
export DOKKU_APP_TYPE='\''herokuish'\''
export DOKKU_NGINX_PORT='\''80'\''
export DOKKU_NGINX_SSL_PORT='\''443'\''
export DOKKU_PROXY_PORT_MAP='\''http:80:5000 https:443:5000'\''
++ export --no-restart=--no-restart
/var/lib/dokku/core-plugins/available/common/functions: line 448: export: --: invalid option
export: usage: export [-fn] [name[=value] ...] or export -p
```

I checked my config and sure enough:

```
root@prichey:~# dokku config appname
=====> appname env vars
DOKKU_NGINX_PORT:      80
DOKKU_PROXY_PORT_MAP:  http:80:5000 https:443:5000
--no-restart:          --no-restart
```

Somehow I had been able to set an env var `--no-restart`.

I tried unsetting it, but:

```
root@prichey:~# dokku config:unset appname --no-restart
FAILED: Invalid key name: '--no-restart'
root@prichey:~# dokku config:unset appname "--no-restart"
FAILED: Invalid key name: '--no-restart'
root@prichey:~# dokku config:unset appname \--no-restart
FAILED: Invalid key name: '--no-restart'
```

So, that's where I'm at. The app won't build because I have a config with an invalid key name, but I can't unset that var because its name is invalid.

Any ideas?

Here's a link to my `dokku report` if pertinent: https://gist.github.com/prichey/0e8563cee603bd17447538bc9e1e2cdb.

josegonzalez:
Can you downgrade and unset, then re-upgrade?

[...]

PR #3051, opened 2018-01-28
skip validity check for config:unset

alexquick:
Sometimes <0.11 was placing invalid keys in app/global envfiles. In 0.11+ they were causing havoc and the user was unable to unset them. This allows the keys to at least be unset.

In the cases we've seen, 0.10 didn't actually display or have a value for that invalid key, it was just lurking in the envfile. Since it doesn't seem it was ever possible to use the invalid key, should we just filter them automatically when 11.0 loads the envfile?

resolves: #3036


josegonzalez:
Yeah filtering them out seems like a reasonable thing to do.

PR #3052, opened 2018-01-28
Remove bad config keys on load from app/global envfiles

alexquick:
Sometimes <0.11 was placing invalid keys in app/global envfiles. In 0.11+ they were causing havoc and the user was unable to unset them.

Now we delete them from the envfiles when we encounter them.

resolves: #3036


josegonzalez:
Thanks for the fix!

Issue #2159, opened 2016-05-04
Dokku CHECKS file and Django ALLOWED_HOSTS


lolorama:
After adding CHECKS file to prodject I have en error during deployment in my django app.

`django.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: '172.17.0.4:5000'. You may need to add '172.17.0.4' to ALLOWED_HOSTS.`

It occurs because in production we don't have 172.17.0.4 in ALLOWED_HOSTS.

I assume correct way is to get address from environment variables but it seems that there is no predefined variable for that purpose?

What is correct way to add dokku local host ip to ALLOWED_HOSTS for django app?

josegonzalez:
We can't set that as an environment variable, and I'm guessing something else is at play here.

Can you include *all* of the information we ask for in our [issue template](https://github.com/dokku/dokku/blob/master/ISSUE_TEMPLATE.md)? It is helpful for debugging problems. Thanks.

michaelshobbs:
I had the same issue in my django apps. I'm mobile right now but I ended up inspecting the network interface to add the dynamic ip to `ALLOWED_HOSTS`

michaelshobbs:
in `settings/production.py`
```
# retrieve first ip address for use in dokku deployments
ALLOWED_HOSTS.append(socket.getaddrinfo(socket.gethostname(), 'http')[0][4][0])
```

josegonzalez:
That works.

lolorama:
@michaelshobbs @josegonzalez, thanks guys!

lolorama:
Buy the way. 

Why not to add HTTP_HOST header with dokku app domain for all requests from CHECKS file?

I assume it's more efficient way for acceptance testing to ensure that container is ready for production deployment, isn't it?

michaelshobbs:
What would we set `HTTP_HOST` to? An app might have 1 or 10 or 100 domains attached to it....

lolorama:
Yeah.. probably bad idea.

josegonzalez:
@michaelshobbs we could just use the first configured domain.

michaelshobbs:
Yeah I didn't mean to imply it was a bad idea just that we would have to arbitrarily use a domain name. I'm also not clear on what the benefit would be by doing this.

lwm:
[...]

PR #3059, opened 2018-02-03
Add note for runtime host configuration for checks.

lwm:
Please see https://github.com/dokku/dokku/issues/2159 for motivation. It was mentioned in https://github.com/dokku/dokku/issues/2159#issuecomment-361097917 that the note should not be specific to Django. I tried to generalise it but still give the code that is provided in #2159. Hope that works!


josegonzalez:
I'm going to rewrite this. Not every application framework requires host checking - even in non-production modes - and as such it still feels like too application-specific. I realize Django is pretty big, but I'd like to ensure we don't become a dumping ground for random oddities users will see when deploying any kind of app.

josegonzalez:
Thanks for the original contribution/impetus to clarify our docs!

lwm:
Great, thanks for taking it in @josegonzalez, that all makes sense to me 👍

PR #3062, opened 2018-02-12
Change help run command be more explicit

shannara:
I purpose to change help run command to more explicit, infact, suggested that command is running inside current application and not that new container based upon application image.


josegonzalez:
Thanks for the pull request!

PR #3076, opened 2018-02-22
Prune worktree after build

Kjwon15:
Since #2973, Dokku add worktree when build app, And that worktree was remained after build.
This PR will prune that worktree after build.

FYI: try `git worktree list` on `/home/dokku/APPNAME`
```
kjwon15@shion:~dokku/armbot % git worktree list                                                                                                                                                                   
/home/dokku/armbot   (bare)
/tmp/dokku_git.1vgF  c9e98f9 (detached HEAD)
/tmp/dokku_git.78np  0dee744 (detached HEAD)
/tmp/dokku_git.7gJ1  c9e98f9 (detached HEAD)
/tmp/dokku_git.GNJj  5fb35cb (detached HEAD)
/tmp/dokku_git.Gtyb  ef758aa (detached HEAD)
/tmp/dokku_git.LD3N  c5b816d (detached HEAD)
/tmp/dokku_git.YDwg  ce093b7 (detached HEAD)
/tmp/dokku_git.YT7i  77ec789 (detached HEAD)
```

### DIFF
--- a/plugins/git/functions
+++ b/plugins/git/functions
@@ -41,7 +41,6 @@ git_build_app_repo() {
     # unset the git quarantine path to allow us to use 2.13.0+
     # See this issue for more information: https://github.com/dokku/dokku/issues/2796
     env -u GIT_QUARANTINE_PATH git worktree add "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" "$REV" > /dev/null
-    trap 'rm -rf "$GIT_BUILD_APP_REPO_TMP_WORK_DIR"; git worktree prune > /dev/null' RETURN INT TERM EXIT
     popd > /dev/null
     pushd "$GIT_BUILD_APP_REPO_TMP_WORK_DIR" > /dev/null
   else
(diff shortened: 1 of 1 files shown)


josegonzalez:
Good catch!

plugins/git/functions
```diff
@@ -64,6 +64,12 @@ git_build_app_repo() {
   suppress_output git submodule update --init --recursive
   find . -name .git -prune -exec rm -rf {} \; > /dev/null
 
+  if use_git_worktree; then
```
josegonzalez: Doesn't this delete the worktree before it gets used below?

Kjwon15: After deleting `.git` directory, `git worktree prune` command removes this worktree on original(`$DOKKU_ROOT/$APP`) repository. But actual directory(`/tmp/dokku_git.XXXX` which will mounted on `/tmp/app` in docker container) is remained.
I tested on my local machine, Build was fine.

josegonzalez: Ah! I see what you mean now, I thought this meant directories were staying in `/tmp` :)

PR #3095, opened 2018-03-06
Migrate env for NGINX_* to PROXY_*

josegonzalez:
This switches the env for `DOKKU_NGINX_*` to `DOKKU_PROXY_*` so we don't need different env settings for different proxy types.

Refs #2823


josegonzalez:
I'll need to review this before merge so I'm sure it does what we expect :)

PR #3102, opened 2018-03-08
Implement repo copy-files plugin

### LINKED ISSUE #2312: [RFC] Repository Copy Files Plugin
I think this was sort of mentioned in #2138, but it was brushed over because that wasn't the primary focus of that issue.

Basically, what I'd like is for the contents of `nginx.conf.d/` in my git repository to be copied to the corresponding place in the dokku application directory, so that I don't have to manually scp up files to the dokku server after I've deployed the application. (further, but outside the scope of this issue I believe that _ALL_ application config should live in the git repo, so that essentially the dokku server is a throwaway worker for running the web applications - this would make redeploying / moving between dokku servers very easy)

I know that I could add a sigil file to my root, but copying an 86 line template file and just adding one line (in this case, adding the `client_max_body_size` parameter) seems a bit suboptimal, and I might miss out on exciting changes to the default template when upgrading.

If this is something that could work, I'd be happy to have a bash (pun intended) at making it work but would probably need somebody to point me in the right direction (I had a look at a bunch of files but couldn't work out where this might go in a way that would work for git push / tar file etc).

Thanks!
Aaron

josegonzalez:
This will allow users to copy files from a repo subdirectory to the root of the directory, as well as from a repo subdirectory to the app directory on the host.

Closes #2312


josegonzalez:
Closing this. With recent changes to Dokku works (app.json, monorepo) I don't think this is necessary anymore.

PR #3104, opened 2018-03-08
Add dokku-limit to plugins documentation

sarendsen:
[ci skip]

### DIFF
--- a/docs/community/plugins.md
+++ b/docs/community/plugins.md
@@ -125,6 +125,7 @@ Note: The following plugins have been supplied by our community and may not have
 [wmluke]: https://github.com/wmluke
 [Zenedith]: https://github.com/Zenedith
 [fteychene]: https://github.com/fteychene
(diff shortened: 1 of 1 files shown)


josegonzalez:
Will merge this once https://github.com/sarendsen/dokku-limit/issues/1 is fixed ;)

josegonzalez:
Also this is something I'd love to see in the core... I even wrote a blog post about how this might work :)

sarendsen:
Done, yes hopefully this can be build out to something more integrated. Though first some stuff in core needs some love like docker-options supporting processes and maybe some service hooks so we can limit service containers aswell.

josegonzalez:
I think service hooks are a bit different from this, and the underlying plugins should implement that themselves.

We can at least support the current method of applying limits across a service, which is what your plugin does.

sarendsen:
Thats true, ill seperate resource logic and app logic soon so its more reusable outside of the apps score.

PR #3118, opened 2018-03-14
docs: CHECKS only matches start of content

Tyilo:
Right?

[ci skip]

### DIFF
--- a/docs/deployment/zero-downtime-deploys.md
+++ b/docs/deployment/zero-downtime-deploys.md
@@ -127,7 +127,7 @@ To specify checks, add a `CHECKS` file to the root of your project directory. Th
 
 ### Check Instructions
 
-The format of a check instruction is a path or relative URL, optionally followed by the expected content:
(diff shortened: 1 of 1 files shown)


josegonzalez:
Thanks for the pull request!

ilyapoz:
Wait a minute, what do you mean by start?

josegonzalez:
Actually didn't read this closely. I think it just greps all the output for what you specified. Not just the first but of text in the output. Not sure what @Tylio meant now upon rereading.

Tyilo:
Hmm, I thought that it worked that way. Maybe my testing was wrong.

We should still probably document exactly how the matching (it must be a substring of the page or is it a regex?)

ilyapoz:
looks like this is the check fwiw https://github.com/dokku/dokku/blob/2fbe5a418ff4f93d1c3b66e9cf9fa257728cbc9c/plugins/checks/check-deploy#L211

josegonzalez:
Thats a regex check.

PR #3119, opened 2018-03-15
Link through to the application deployment page after the installation

TomasHubelbauer:


TomasHubelbauer:
I am not sure to the body of *what* am I supposed to put `[ci skip]`? The PR comment? The note does that for me I think.

PR #3132, opened 2018-03-28
Updated help text on config:set to have commands in working order

marykatefain:
The `config:set` command was not working in the order the help text provided suggested. It works if you put the app name after the options for `encoded` and `no-restart`, so I rearranged the help text to make this clear.


josegonzalez:
Good catch! I'm mobile right now, but mind checking if the docs also have it in the wrong order and fixing if so?

joshmanders:
Since @josegonzalez is on mobile, just update this PR with any docs changes if you see, and ping me and I'll review and merge.

marykatefain:
@joshmanders The docs look correct to me, too! Thanks :)

PR #3153, opened 2018-04-09
Allow specifying a separate DOKKU_HOST_ROOT

scjody:
Fixes https://github.com/plotly/streambed/issues/10723

* [x] Automated tests 🐅
* [x] Manual test to make sure it fixes our problem 🐈 

Add a `DOKKU_ROOT_ON_HOST` variable and use it in place of of `DOKKU_ROOT` when we need to refer to a path on the system where the Docker daemon is being run. (If `dokku` is running in a container this may be different from the regular`DOKKU_ROOT`.)


joshmanders:
What's the details of that issue? I get 404 when trying to visit it.

scjody:
@joshmanders The issue is that we're running `dokku` inside a container, and the paths in this container are different from those on the host.

Inside the container Dokku's files live at `/home/dokku`. Outside, they can be anywhere (depending on how the container is configured), such as `/data/dokku`. (We map the appropriate directory into the container that runs Dokku.)

Without this change Dokku appears to work fine, except whenever the cache is used (during app build), the cache ends up under `/home/dokku/APPNAME/cache` on the host rather than in the actual Dokku root (`/data/dokku/APPNAME/cache` in the example above). This causes problems if, for example, `/home` is on a filesystem with limited disk space.

See also: https://github.com/dokku/dokku-redis/pull/103

josegonzalez:
This seems like a reasonable solution. I updated the code to cleanup the naming, and will merge once tests pass.

scjody:
@josegonzalez I added tests of the cache bind option during app deployment. I realize this doesn't cover all the additions but I feel like it's enough. I'm happy to add more tests if you disagree - let me know!

josegonzalez:
This is good enough for our use-case.

PR #3159, opened 2018-04-12
Improve docs for default site with HTTPS

zuck:
Provide a working solution of default site configuration for both HTTP and HTTPS requests.

### DIFF
--- a/docs/configuration/domains.md
+++ b/docs/configuration/domains.md
@@ -137,15 +137,20 @@ server {
     return 410;
 }
 
-# For HTTPS requests you can enable this section:
-# (You need a SSL certificate in /etc/nginx/ssl/)
-server {
-#    listen 443 ssl;
-#    server_name _;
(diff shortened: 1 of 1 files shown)


josegonzalez:
I don't think that will work all the way. Pretty sure browsers will see the catch-all and try to render with that ssl certificate, but fail because it won't be valid for the arbitrary domain.

Have you tested this to ensure it works as expected?

zuck:
Well, on my production server it works with wildcard cert. Actually I'm returning 404 instead of 410.

josegonzalez:
Okay so that makes sense, but it wont work for any arbitrary server - many of our users do not have a wildcard cert.

Mind updating the docs to make that caveat clear? This will only work when the ssl certificate is valid for the domain being requested.

zuck:
I've expanded docs a bit, hoping it's more clear now.

joshmanders:
Yeah but to use wildcard certs with the letsencrypt plugin isn't an easy task as it required DNS validation.

PR #3215, opened 2018-06-11
[ci skip] install with cloud-init

pynixwang:
add cloud-init config file to install when create server on cloud such as EC2.

[ci skip]


josegonzalez:
This should actually not point out cloud-init, but maybe say "for other installation instructions, see here" where `here` is a link to the rest of our install docs. I think keeping the readme short and sweet is important.

PR #3279, opened 2018-10-05
Make sure the universe repo is loaded into APT

### LINKED ISSUE #3278: Install fail on Ubuntu server 18.04.1
### Steps to Reproduce

```
root@server:~# wget https://raw.githubusercontent.com/dokku/dokku/v0.12.13/bootstrap.sh
--2018-10-05 16:47:11--  https://raw.githubusercontent.com/dokku/dokku/v0.12.13/bootstrap.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.128.133, 151.101.64.133, 151.101.0.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.128.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 7654 (7.5K) [text/plain]
Saving to: ‘bootstrap.sh’

bootstrap.sh                                                        100%[=================================================================================================================================================================>]   7.47K  --.-KB/s    in 0.001s  

2018-10-05 16:47:11 (6.03 MB/s) - ‘bootstrap.sh’ saved [7654/7654]

root@server:~# DOKKU_TAG=v0.12.13 bash bootstrap.sh
```

### Actual Results

```
Preparing to install v0.12.13 from https://github.com/dokku/dokku.git...
--> Ensuring we have the proper dependencies
--> Initial apt-get update
--> Installing docker
2018-10-05 16:47:39 URL:https://get.docker.com/ [13847/13847] -> "-" [1]
# Executing docker install script, commit: 36b78b2
+ sh -c apt-get update -qq >/dev/null
+ sh -c apt-get install -y -qq apt-transport-https ca-certificates curl >/dev/null
+ sh -c curl -fsSL "https://download.docker.com/linux/ubuntu/gpg" | apt-key add -qq - >/dev/null
Warning: apt-key output should not be parsed (stdout is not a terminal)
+ sh -c echo "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic edge" > /etc/apt/sources.list.d/docker.list
+ [ ubuntu = debian ]
+ sh -c apt-get update -qq >/dev/null
+ sh -c apt-get install -y -qq --no-install-recommends docker-ce >/dev/null
+ sh -c docker version
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Tue Aug 21 17:24:51 2018
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Tue Aug 21 17:23:15 2018
  OS/Arch:          linux/amd64
  Experimental:     false
If you would like to use Docker as a non-root user, you should now consider
adding your user to the "docker" group with something like:

  sudo usermod -aG docker your-user

Remember that you will have to log out and back in for this to take effect!

WARNING: Adding a user to the "docker" group will grant the ability to run
         containers which can be used to obtain root privileges on the
         docker host.
         Refer to https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface
         for more information.
--> Installing dokku
2018-10-05 16:47:55 URL:https://packagecloud.io/gpg.key [3889/3889] -> "-" [1]
OK
deb https://packagecloud.io/dokku/dokku/ubuntu/ bionic main
E: Unable to correct problems, you have held broken packages.

```

### Environment Information

```
# lsb_release -a

No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

### Other info

```
root@server:~# apt install dokku
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 dokku : PreDepends: cgroupfs-mount but it is not installable or
                     cgroup-lite but it is not installable
         Recommends: herokuish (>= 0.3.4) but it is not going to be installed
         Recommends: parallel but it is not installable
         Recommends: dokku-update but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

fruitl00p:
this should fix #3278

I know its a small change, but this won't hurt if universe is already enabled but will fix the issue mentioned.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -32,6 +32,11 @@ install-requirements() {
 
   case "$DOKKU_DISTRO" in
     debian|ubuntu)
+      if ! dpkg -l | grep -q software-properties-common; then
+        apt-get update -qq > /dev/null
+        apt-get -qq -y install software-properties-common
+      fi
+
       add-apt-repository universe > /dev/null
       apt-get update -qq > /dev/null
       ;;
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -32,6 +32,7 @@ install-requirements() {
 
   case "$DOKKU_DISTRO" in
     debian|ubuntu)
+      add-apt-repository universe > /dev/null
       apt-get update -qq > /dev/null
       ;;
   esac


bootstrap.sh
```diff
@@ -32,6 +32,7 @@ install-requirements() {
 
   case "$DOKKU_DISTRO" in
     debian|ubuntu)
+      add-apt-repository universe > /dev/null
```
josegonzalez: What happens if universe is already enabled? I also wonder we should be sending output to `/dev/null`.

josegonzalez:
Actually, I don't know if this will work on a from-scratch installation of Ubuntu. `add-apt-repostiroy` comes from `software-properties-common`, which isn't installed by default iirc. Can you test this script first on a new repository?

bootstrap.sh
fruitl00p: I've tested this locally and doesnt seem to effect my installation if already enabled... (and the > /dev/null stuff is just there to follow suit as the `apt-get update -qq > /dev/null` also has this. Since this is a minor addition, it wouldn't hurt... (i thought)

fruitl00p:
@josegonzalez i did a quick check and, you're correct. It does need the software properties
```
docker run --rm -it ubuntu:18.10 bash

apt-get update -qq && apt-get -qq -y install software-properties-common && add-apt-repository universe                     
```
(which resulted in a `'universe' distribution component is already enabled for all sources.`)

Thus this might even mean that it might even be more of a documentation issue (or a difference between ubuntu versions?)

IrregularShed:
Sorry to leap on an old PR, but this breaks the installer on Debian - there isn't a `universe` repository outside of Ubuntu, I believe. I've been trying to get dokku to install for a couple of hours and it kept silently dying; I've now removed all the `> /dev/null` parts in the script and found that out!

I'm new to the whole project - good job I'm eager, otherwise I'd have given up ages ago ;)

josegonzalez:
https://github.com/dokku/dokku/pull/3415 <-- Pull request here, releasing now.

josegonzalez:
Released as 0.14.4.

Issue #3293, opened 2018-10-26
dokku config no longer sorts by alpha


joshco:
This is a small issue, but I noticed it when upgrading from old old dokku like 0.3 to 0.12.13.

Running dokku config <app>  used to return the environment variables sorted by alpha A-Z.
With version 0.12.13 it returns them in random order that is different every time the command is run.

Is there a reason why this is?

josegonzalez:
No reason, that just wasn’t a thing we tested for.

@alexquick is this something you can look into?

PR #3295, opened 2018-10-27
[config] sort config:show by key name

alexquick:
resolves: #3293

Output of config:show now is sorted by key name to match the old config plugin (and common sense).

I checked and the rest of the things we do with outputting/saving are sorted as expected.


joshmanders:
LGTM 👍

PR #3305, opened 2018-11-05
Remove ~dokku/VERSION

t0xicCode:
As per #1558, remove `~dokku/VERSION` and move it to `/var/lib/dokku`.

Note: this will become the PR.


contrib/release
```diff
@@ -118,8 +118,7 @@ fn-version-next() {
   elif [[ "$RELEASE" == "patch" ]]; then
     patch=$((patch + 1))
   else
-    short_object=$(git log --oneline | tail -n 1 | cut -d' ' -f1)
-    echo "${CURRENT_VERSION}build+$(echo "$DOKKU_GIT_REV" | cut -c1-${#short_object})"
+    echo "$(git describe --tags | cut -c2-)"
```
josegonzalez: This is necessary for us to have build releases. I would revert.

Makefile
```diff
@@ -103,13 +103,6 @@ addman: help2man man-db
 	help2man -Nh help -v version -n "configure and get information from your dokku installation" -o /usr/local/share/man/man1/dokku.1 dokku
 	mandb
 
-version:
-ifeq ($(DOKKU_VERSION),master)
-	git describe --tags > ~dokku/VERSION  2> /dev/null || echo '~${DOKKU_VERSION} ($(shell date -uIminutes))' > ~dokku/VERSION
-else
-	echo $(DOKKU_VERSION) > ~dokku/VERSION
```
josegonzalez: Can we keep this and have it output to `/var/lib/dokku/version`?

dokku
```diff
@@ -225,6 +227,9 @@ case "$1" in
       fi
     fi
     ;;
+  version)
```
josegonzalez: Could we add support for `-v` and `--version` flags as well? That would be very user-friendly :)

dokku
```diff
@@ -55,6 +55,8 @@ export DOKKU_EVENTS_LOGFILE=${DOKKU_EVENTS_LOGFILE:="$DOKKU_LOGS_DIR/events.log"
 export DOKKU_CONTAINER_LABEL=dokku
 export DOKKU_GLOBAL_RUN_ARGS="--label=$DOKKU_CONTAINER_LABEL"
 
+export DOKKU_VERSION="0.12.13"
```
josegonzalez: We should grab this from the file, not hardcode it here.

contrib/release
t0xicCode: `git describe --tags` gives output like `v0.12.13-46-ge7f7cb8` so that doesn't prevent the build releases. Does the word _build_ matter in this, or was it just the existing convention?

dokku
t0xicCode: :+1: done!

dokku
josegonzalez: Otherwise you need to handle a sed call to get this updated in the `contrib/release` process. I don't personally want to do that right now - maybe in the future if we turn the `dokku` bin into golang, at which point it can be done at build-time.

contrib/release
josegonzalez: Yeah, it kinda does. Deb packaging is weird.

dokku
t0xicCode: I've already updated the replacement :) -> https://github.com/dokku/dokku/pull/3305/files#diff-b9f5212d14fb13cff4e50a60368ff8a2R198

I agree that go does is beautifully with compile type definitions though.

Makefile
t0xicCode: Done!

Makefile
```diff
@@ -104,10 +104,11 @@ addman: help2man man-db
 	mandb
 
 version:
+	mkdir -p /var/lib/dokku
```
josegonzalez: This should use the `DOKKU_LIB_ROOT` variable (same elsewhere in the makefile).

arch.mk
```diff
@@ -23,10 +23,9 @@ arch-setup:
 arch-dokku: arch-setup
 	echo "-> Update package sums, create metadata file and test the build of the package"
 ifeq ($(DOKKU_VERSION),master)
-	git describe --tags > /tmp/VERSION
+	git describe --tags > /tmp/version
 else
-	echo $(DOKKU_VERSION) > /tmp/VERSION
+	echo $(DOKKU_VERSION) > /tmp/version
```
josegonzalez: Lets avoid changing the case. Keep it upper-case for now.

arch.mk
```diff
@@ -23,10 +23,9 @@ arch-setup:
...
 endif
-	cat /tmp/VERSION | cut -d '-' -f 1 | cut -d 'v' -f 2 > /tmp/STABLE_VERSION
```
josegonzalez: We likely still need this `cut` sequence, though not sure how you want to handle it without creating a temp file and overwriting.

contrib/dokku-update
```diff
@@ -67,8 +67,8 @@ main() {
 
   if [[ "$COMMAND" == "version" ]] || [[ "$COMMAND" == "-v" ]]; then
     VERSION=UNRELEASED
-    if [[ -f "/var/lib/dokku-update/VERSION" ]]; then
-      VERSION="$(cat /var/lib/dokku-update/VERSION)"
+    if [[ -f "/var/lib/dokku-update/version" ]]; then
+      VERSION="$(cat /var/lib/dokku-update/version)"
```
josegonzalez: Keep uppercase please.

dokku
josegonzalez: I would really rather we don't do this for now. Can you revert this part of the change (and anything related)?

josegonzalez:
Generally speaking, the fewer/smaller changes, the easier it is for me to figure out how this will impact the release process. I'd be in favor of doing something like:

- Move `~dokku/VERSION` to `$DOKKU_LIB_ROOT/VERSION`
- Remove `$DOKKU_LIB_ROOT/STABLE_VERSION`

contrib/release
t0xicCode: reverted

dokku
t0xicCode: reverted

arch.mk
t0xicCode: reverted

contrib/dokku-update
t0xicCode: reverted

Makefile
t0xicCode: updated

t0xicCode:
@josegonzalez All requested changes have been made.

Makefile
```diff
@@ -93,7 +93,7 @@ endif
 		cp -R plugins/$(PLUGIN_NAME) ${CORE_PLUGINS_PATH}/available && \
 		rm -rf ${CORE_PLUGINS_PATH}/available/$(PLUGIN_NAME)/src && \
 		ln -s ${CORE_PLUGINS_PATH}/available/$(PLUGIN_NAME) ${PLUGINS_PATH}/available; \
-		find /var/lib/dokku/ -xtype l -delete;\
+		find ${DOKKU_LIB_ROOT} -xtype l -delete;\
```
josegonzalez: This might need a trailing slash :)

rpm.mk
```diff
@@ -79,18 +79,9 @@ ifeq ($(DOKKU_VERSION),master)
...
-endif
-ifdef DOKKU_GIT_REV
-	echo "$(DOKKU_GIT_REV)" > /tmp/build/var/lib/dokku/GIT_REV
-else
-	git rev-parse HEAD > /tmp/build/var/lib/dokku/GIT_REV
-endif
 
 	@echo "-> Creating rpm package"
-	VERSION=$$(cat /tmp/build/var/lib/dokku/STABLE_VERSION); \
+	VERSION=$$(cat /tmp/build/var/lib/dokku/VERSION); \
```
josegonzalez: IIRC a valid deb version is not a valid rpm version. That was the driver around having a `STABLE_VERSION` and a `VERSION` file (though I could be wrong).

Makefile
t0xicCode: Done! Although I think `find` does recurse into the directory automatically.

rpm.mk
t0xicCode: Should we instead do `git describe --tags` for the version, as those are valid for "upstream" and rpm, and transform them into valid deb version? It is trivial to go from a valid upstream/rpm version to a valid debian one, but a tad harder the other way around.

josegonzalez: Sounds like a good strategy.

deb.mk
```diff
@@ -54,6 +54,12 @@ ifndef IS_RELEASE
 	IS_RELEASE = true
 endif
 
+ifeq ($(IS_RELEASE),true)
+	DOKKU_DEBIAN_VERSION_CMD = `cat /tmp/build/var/lib/dokku/VERSION`
+else
+	DOKKU_DEBIAN_VERSION_CMD = `cat /tmp/build/var/lib/dokku/VERSION | awk -F- '{print $$1 "+build" $$2 "." $$3}'`
+endif
```
josegonzalez: IIRC, This will error during make runs if the file does not exist.

josegonzalez:
Can you rebase this onto master? I've made quite a few changes to the release process that will allow us to generate builds based on arbitrary commits, will be useful in testing this.

t0xicCode:
Sure thing, I'll work on it shortly and will rebase.

Cheers

josegonzalez:
Going to close this for now, as it is a bit stale, but also rebasing it is a bit tough. I'll try it from scratch, but huge thanks for taking an initial stab at this and providing ideas :)

PR #3304, opened 2018-11-05
[PROPOSAL] Remove the src directory from the go plugins

t0xicCode:
The de-facto [go project layout](https://github.com/golang-standards/project-layout#src) recommends against using `src` in the path.

I'd like to propose to move all `src` directories in _plugins_ to something more go-like, such as `cmd`.


t0xicCode:
@josegonzalez issue is now a pull request and ready for review!

t0xicCode:
Seems that this might require a rework of the build mechanics. Instead of building the source in place, moving the lot to the installation area, then removing the source, I think the build system should instead build the source cleanly in a separate location (maybe a `build` folder, and them move that folder to the installation area).

I'm closing this in the meantime.

PR #3313, opened 2018-11-15
Indicate to user to specify hostname

morenoh149:
The deployment docs suggest using the hostname `dokku.me`. Instead put a placeholder prompting the user to specify the correct hostname.

[ci skip]

### DIFF
--- a/docs/deployment/application-deployment.md
+++ b/docs/deployment/application-deployment.md
@@ -1,5 +1,8 @@
 # Deploying to Dokku
 
+> Note: This document uses the hostname `dokku.me` in commands. For your server please
+> substitute your server's hostname instead.
+
 ## Deploy tutorial
 
 Once Dokku has been configured with at least one user, applications can be deployed via a `git push` command. To quickly see Dokku deployment in action, you can use the Heroku Ruby on Rails example app.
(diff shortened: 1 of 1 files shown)


josegonzalez:
I would rather we put a note at the top of the page stating that `dokku.me` is a placeholder for your domain name. We use `dokku.me` as that pretty much everywhere in our docs for standardization - and also because it makes it so if you setup the vagrant vm locally, the instructions just work.

docs/deployment/application-deployment.md
```diff
@@ -55,14 +55,16 @@ Now you can deploy the `ruby-rails-sample` app to your Dokku server. All you hav
...
+git remote add dokku dokku@<your hostname>:ruby-rails-sample
 git push dokku master
 ```
 
 > Note: Some tools may not support the short-upstream syntax referenced above, and you may need to prefix
-> the upstream with the scheme `ssh://` like so: `ssh://dokku@dokku.me:ruby-rails-sample`
+> the upstream with the scheme `ssh://` like so: `ssh://dokku@<your hostname>:ruby-rails-sample`
 > Please see the [Git](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) documentation for more details.
 
+> Note: Your private key needs to be registered with ssh-agent. If you get a permission denied error when pushing you can register your private key by running `ssh-add -k ~/<your private key>`.
```
josegonzalez: Add a note that this is to be done locally, not on the server.

morenoh149:
squashed and updated

josegonzalez:
Great work! Thanks for the pull request :)

PR #3314, opened 2018-11-16
Remove reference to non-existent DigitalOcean hosting plan

royklutman:
DigitalOcean no longer offers a 512MB plan, so I changed the example to Vultr, which do offer a small 512MB plan.

[ci skip]


josegonzalez:
I'd rather not include provider-specific instructions if we can avoid it. I'll update your PR. Thanks for the note :)

PR #3341, opened 2018-12-23
Select only default dokku network IP

### LINKED ISSUE #3336: Problem with DOKKU_APP_LISTENERS within nginx configuration
Hello,

Within the nginx configuration of a dokku application, I use the `DOKKU_APP_LISTENERS`.

```
upstream {{ .APP }} {
{{ range .DOKKU_APP_LISTENERS | split " " }}
  server {{ . }};
{{ end }}
}
```

This worked fine but recently I got an error when deploying an application (when Dokku tries to validate the nginx configuration)

It seems that after the server keyword, I have two IP addresses used with a port at the end:

```
upstream my app {
  server 172.10.34.6172.10.34.4:5000
}
```

How to use `dokku proxy` commands to fix that?

How to change this?
Thanks for your help.
Therry

baikunz:
When a container is connected to multiple networks (i.e. the default dokku `bridge` and a custom one) the `GetContainerIpaddress` from the network plugin returns a concatenation of those IPs :
i.e. `172.10.34.6172.10.34.4` instead of just `172.10.34.6`
I fixed the `GetContainerIpaddress` function to only return the default dokku network (named `bridge`) IP, however I have no idea how it might affect the rest of the codebase.

It might fix #3336

### DIFF
--- a/plugins/network/network.go
+++ b/plugins/network/network.go
@@ -95,7 +95,7 @@ func GetContainerIpaddress(appName, procType, containerID string) (ipAddr string
 		return
 	}
 
-	b, err := common.DockerInspect(containerID, "'{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'")
+	b, err := common.DockerInspect(containerID, "'{{.NetworkSettings.Networks.bridge.IPAddress}}'")
 	if err != nil || len(b) == 0 {
 		// docker < 1.9 compatibility
 		b, err = common.DockerInspect(containerID, "'{{ .NetworkSettings.IPAddress }}'")
--- a/tests/unit/20_nginx-vhosts_1.bats
+++ b/tests/unit/20_nginx-vhosts_1.bats
@@ -10,6 +10,7 @@ setup() {
 }
 
 teardown() {
+  detach_delete_network
   destroy_app
   [[ -f "$DOKKU_ROOT/VHOST.bak" ]] && mv "$DOKKU_ROOT/VHOST.bak" "$DOKKU_ROOT/VHOST" && chown dokku:dokku "$DOKKU_ROOT/VHOST"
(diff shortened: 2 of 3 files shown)


josegonzalez:
By the way, this is nice :)

baikunz:
Well, I just created a plugin that executes a script on the dokku host after the application has been deployed.
In this script I created a docker network (if it did not already existed) and connected my container to that network.
The container was then connected to two different networks, `bridge`(dokku network for every app) and my custom one.
The problem showed up when building the nginx config which is fetching the IPaddress of the container. The fetching function `GetContainerIpaddress` is assuming that there is only one network the container is connected to (the docker 1.9 compatibility part of the code shows exactly that).

> What should we do when we need to support an alternate network?

I'm sorry I don't know everything about dokku right now. Is there a way to specify a network that dokku should use instead of using the default `bridge` one ? If that's the case, then we might just have to pass it's name to the `GetContainerIpaddress` function.

josegonzalez:
So existing tests pass. Any chance you can add a new test to verify this works after any future modifications? Probably add a network to a running app and then call the network-get-listeners hook or something :)

baikunz:
Yup I'm gonna look into it. I'm not used to bash scripts / testing nor Golang so I might need a few hints :D

josegonzalez:
I think probably something like:

- call `deploy_app` to deploy a new app
- use `$TEST_APP` to figure out the name of the running container
- create the new network
- modiy the container using `${TEST_APP}.web.1` to add the new network
- invoke something that will trigger a new nginx config - which will trigger the code you changed to get the listeners and write it to the the nginx.conf.sigil
- call `sudo nginx -t` or something to validate the nginx config (we should probably have this as an `nginx-vhosts` subcommand, like `nginx-vhosts:validate`, but if you want to add that in a separate PR, that would be awesome)

baikunz:
Thanks for the hints, I'll look into it ASAP ;-)

baikunz:
Well I took some time to work on the test. Looks like it passed ;-)

josegonzalez:
Nice work! This is great, I'll get it merged now (it'll land in the next major release, just fyi).

baikunz:
Hi ! Thanks mate awesome ;-)
I wasn't expecting it to be merged so soon, I would have squashed my commits and rebased if I knew :D

PR #3342, opened 2018-12-24
Add section to nginx docs about manually updating dokku.conf

chriswhong:


josegonzalez:
I just realized that the reason we don't always overwrite is because of #544. That change is no longer necessary as we do not support global tls as of #2058.

Given our stance on always being secure, I'd rather us always overwrite this file instead of telling people to go back and verify that things are done correctly. I'm going to close this in favor of doing that removal.

PR #3352, opened 2018-12-27
Increase security of default SSL setup

### LINKED ISSUE #3343: ssl_ciphers in dokku.conf includes some that ssllabs flags as 'weak'
I updated `dokku.conf` to include the [latest configuration from v0.13.4](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/install#L52), because ssllabs.com was indicating that my server was using some "weak" ciphers.

After updating the file and restarting nginx, I still see some 'weak' ciphers according to ssllabs (see screenshot).

ssl_server_test__planninglabs_nyc__powered_by_qualys_ssl_labs_

Per @josegonzalez on Slack: "Hmmm CBC isn't on our list.  Try restarting your nginx".  Tried that, same results on ssllabs.

Is there a better, more current `ssl_ciphers` configuration that will satisfy ssllabs and should be added to dokku?

josegonzalez:
- Switch to move secure ssl ciphers
- Increase RSA bit keys to 4096

Closes #3343


chriswhong:
Confirming that this list of ciphers no longer shows any as "weak" when using ssllabs!

IlyaSemenov:
Is there any way to disable explicit `ssl_cipher` so that it doesn't come back on Dokku update (on Ubuntu)? This just hit us today as the whole bunch of legacy clients were unable to connect after Dokkku upgrade.

I realise that I probably can create `/etc/nginx/conf.d/99-dokku-overrides.conf` but then I'll need to manually list ciphers while I prefer to just use whatever defaults nginx has.

josegonzalez:
This change was made [here](https://github.com/dokku/dokku/pull/3346). I'd prefer we just do this correctly by default, with _maybe_ an option to disable the security settings. Please file a new issue to track that, as a closed pull request is not the best place.

IlyaSemenov:
Yeah, I realise all that and I actually agree that the new defaults are correct and overwriting dokku.conf is a good thing. This is a mere question of supporting a (intentionally non-secure) setup.

I didn't open a new issue because I don't fully understand what I want from Dokku for that matter ;) Perhaps the current way of things is the least evil.

fruitl00p:
@crystianwendel Please also refer to the [mozilla configurator](https://mozilla.github.io/server-side-tls/ssl-config-generator/) for this config...

crystianwendel:
@fruitl00p i just wrote the ciphers used before this commit, so apps that were working before it could work again.

Issue #3387, opened 2019-01-10
CORS problem in Firefox for web-based installer


heyarne:
## Description of problem

I just tried using the web-based installer in Firefox 64. The scripts are not loaded and the web page is unresponsive. I disabled my ad blockers and the problem persists.

![2019-01-10-160112_900x320_scrot](https://user-images.githubusercontent.com/525780/50977842-6411b000-14f3-11e9-9a7f-7c887b64aa8c.png)

Interestingly enough Chromium 71 doesn't have this problem. The domain I used for setting up did not have a certificate at that time which might have caused the problem.

josegonzalez:
Can you manually check the sha256 for the file and verify its incorrect? I got that directly from the jQuery site, but happy to update it if necessary.

josegonzalez:
Ah I see what you mean. That's annoying. I wonder if cdnjs supports http and https?

heyarne:
A quick check tells me they're auto-redirecting. The problem with migration to https-only would also stay. I could also try to remove the jQuery dependency alltogether; the code seems very straightforward and cross-browser-compatible js is quite bearable to write and maintain these days.

josegonzalez:
Closing as there is a pull-request open.

Thanks for working on this!

PR #3389, opened 2019-01-10
Remove jQuery from web-based installer

rrrnld:
Removes the jQuery dependency to fix the problem described in #3387.

### DIFF
--- a/contrib/dokku-installer.py
+++ b/contrib/dokku-installer.py
@@ -218,9 +218,9 @@ def main():
 PAGE = """
 <html>
 <head>
+  <meta charset="utf-8" />
   <title>Dokku Setup</title>
   <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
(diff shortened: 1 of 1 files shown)


heyarne:
I could unfortunately only test this via `python2 dokku-installer.py`. More thorough testing (i.e. a complete setup process) would be required. The validation logic seems to work fine, I'm not sure about the server response handling.

The downside of this is that jQuery offers greater browser support. The most modern feature I used is [`fetch`](https://caniuse.com/#search=fetch) which is supported by all modern browsers.

josegonzalez:
Seems fine. Can you try this out on a local vagrant-vm setup, as per our [docs](http://dokku.viewdocs.io/dokku/getting-started/install/vagrant/)? Use your branch and run `vagrant up` to get your changes going.

contrib/dokku-installer.py
```diff
@@ -340,7 +340,8 @@ def main():
             window.location.href = "http://dokku.viewdocs.io/dokku~{VERSION}/deployment/application-deployment/";
           }, 3000);
         })
-        .catch(function () {
+        .catch(function (error) {
+          console.error(error)
```
josegonzalez: We should avoid having any console logging.

josegonzalez: Please do :)

heyarne:
To debugs this. I need to the server's log ouput; there doesn't seem to be anything relevant in `/var/logs`, can you tell me where to find it?

josegonzalez:
It doesn't log anything anywhere. I'd just run it locally and make sure the SSH keys etc. are setup properly.

heyarne:
Sooo... this took a bit longer than expected, but there you go. This is now tested inside vagrant and works, hope this helps! Additionally I added a `<meta charset="utf-8" />` to remove the warning about a missing charset declaration. The setup now loads without any issues or warnings inside the Firefox console.

PR #3409, opened 2019-01-20
Build golang binaries with higher concurrency

josegonzalez:
This should allow us to shave off a few minutes of build time, both locally and on CI


josegonzalez:
This didnt have the effect I expected on our CI, but works nicer for local builds!

PR #3419, opened 2019-01-24
Fix Dokku installer checkbox for WebKit and Edge browsers

jayjun:
The `input` event does not fire for checkboxes on WebKit and Edge browsers (see [Can I Use](https://caniuse.com/#feat=input-event)).

Use `change` for cross browser support.


josegonzalez:
Released in 0.14.5.

jayjun:
That was real quick, thanks @josegonzalez!

PR #3420, opened 2019-01-24
Add reference to external post-deploy-script plugin

baikunz:
[ci skip]

Referenced a plugin I made to execute scripts after a deploy.

### DIFF
--- a/docs/community/plugins.md
+++ b/docs/community/plugins.md
@@ -126,6 +126,7 @@ The following plugins are available and provided by Dokku maintainers. Where not
 [Zenedith]: https://github.com/Zenedith
 [fteychene]: https://github.com/fteychene
 [sarendsen]: https://github.com/sarendsen
+[baikunz]: https://github.com/baikunz
 
 ### Datastores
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Just a note, you shouldn't store things in the git directory, and instead use either a config/data directory like:

```
/var/lib/dokku/data/PLUGIN_NAME/APP_NAME/whatever
```

Dokku will be a refactoring the app directory to remove it's own mis-usage of it in the future.

baikunz:
Well I thought about what you said regarding not storing things in the git directory. Is there any reason for that ?
Because I thought about using your `dokku-supply-config` plugin to copy the `POST_DEPLOY_SCRIPT` I'm using on my plugin.
Is it planned that you update your plugin to support this new folder structure ?

josegonzalez:
Yeah I don't think thats the correct pattern, as it makes it hard for us to version configuration or store it in a different datastore. See #1558 for more details.

baikunz:
Thanks for pointing it out. It makes more sense now. I was not aware of that since I just started using dokku.
Regarding the plugin, I will change the location of the file according to your suggestion.
However until the `dokku-supply-config` plugin is updated, we won't be able to copy this file in it's new location automatically. So it will have to be created manually.

josegonzalez:
Would be a good pr to that plugin :)

baikunz:
Yea sure ;-)
But there is quite some work to be done, because the files will now be copied into multiple directories.

PR #3422, opened 2019-01-26
Start long running containers with --init

jayjun:
Processes not designed to run as PID 1 can leave zombie processes and not handle SIGTERM properly.

Start `docker run` with `--init` for deploys (long running containers) so child processes are properly reaped and the main process can shutdown gracefully.

See,
- [Official docs](https://docs.docker.com/engine/reference/run/#specify-an-init-process) for `--init`
- https://github.com/krallin/tini/issues/8#issuecomment-146135930 Great explanation of Tini, which powers `--init`
- [Best Practices for Building Containers](https://cloud.google.com/solutions/best-practices-for-building-containers#signal-handling) by Google

### DIFF
--- a/plugins/scheduler-docker-local/scheduler-deploy
+++ b/plugins/scheduler-docker-local/scheduler-deploy
@@ -91,16 +91,16 @@ scheduler-docker-local-scheduler-deploy() {
         START_CMD=$(fn-scheduler-docker-local-extract-start-cmd "$APP" "$PROC_TYPE" "$START_CMD" "$DOKKU_HEROKUISH" "$DOKKU_PORT")
         if [[ "$DOKKU_NETWORK_BIND_ALL" == "false" ]]; then
           # shellcheck disable=SC2086
-          cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
+          cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d --init -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
         else
           # shellcheck disable=SC2086
-          cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d $DOKKU_DOCKER_PORT_ARGS -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
+          cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d --init $DOKKU_DOCKER_PORT_ARGS -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
         fi
       else
         START_CMD=$(fn-scheduler-docker-local-extract-start-cmd "$APP" "$PROC_TYPE" "$START_CMD" "$DOKKU_HEROKUISH")
 
         # shellcheck disable=SC2086
-        cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
+        cid=$(docker run $DOKKU_GLOBAL_RUN_ARGS -d --init "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
       fi
 
       ipaddr=$(plugn trigger network-get-ipaddr "$APP" "$PROC_TYPE" "$cid")


josegonzalez:
This doesn't sound safe... otherwise it'd be the default. Do you have any other literature for or against the change?

jayjun:
**For**
* [Docker and the PID 1 zombie reaping problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/)
Classic article by Phusion with outdated solution (predates `--init`).
* [Introducing dumb-init, an init system for Docker containers](https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html)
dumb-init is Yelp’s solution (also predates `--init`).
* [Issues with running as PID 1 in a Docker container](http://blog.dscpl.com.au/2015/12/issues-with-running-as-pid-1-in-docker.html)
Blog post recommends Tini, included as `--init` since Docker 1.13.
* [Docker and Node.js Best Practices: Handling Kernel Signals](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals)
Node.js officially recommends `--init`.

**Against**
* [What should be PID 1 in a container?](https://www.youtube.com/watch?v=np5WFv-FWvE&t=20m12s)
A Senior Red Hat Engineer suggests running systemd as PID 1 is better sometimes.
* kubernetes/kubernetes#48937
GlusterFS’s official container image starts systemd, so _expects_ PID 1 to be free.

`--init` isn’t the default because it adds a shim that only matters when child processes are spawned (and assumed to be reaped by someone else) or signals are required for proper shutdown.

For one-off processes, orphaned processes are reaped after exit anyway and signals probably don’t matter much. The same cannot be said for long running containers, much less for apps you didn’t write (see jenkinsci/docker#54, which prompted the creation of Tini).

Regardless, I only added `--init` to scheduler-docker-local because it seems to be the scheduler’s responsibility. Kubernetes has pods, so it starts a “pause” container as PID 1 in every pod (see [The Almighty Pause Container](https://www.ianlewis.org/en/almighty-pause-container)). Docker added `--init` to swarm mode (see docker/cli#479).

I think it’s a sensible default for Dokku, unless containers start their own Tini or systemd. Maybe we can add an option to disable `--init` instead?

josegonzalez:
Are there any reported cases where not having this has caused issues for Dokku? I'm not aware of anything, and am not sure of what would happen if we add this in.

josegonzalez:
I'd like someone to test the following:

```
dokku docker-options:add APP run,deploy --init
```

I want to know what happens for both Dockerfile and buildpack apps, and also would like to see this be the case for when we have known-good apps, and ones that don't work as well. Please include ones that already have a good init like `tini`. I'm wary of adding a new global flag that isn't well tested.

gyfis:
@josegonzalez I did test what you suggested for the mentioned Ruby on Rails + Sidekiq + Selenium + Chromedriver + Chrome setup, and it did help! No `chrome <defunct>` processes left over, everything deleted as expected.

The buildpacks we use (stored in the `.buildpacks` file):
```
https://github.com/bundler/heroku-buildpack-bundler2.git
https://github.com/heroku/heroku-buildpack-google-chrome.git
https://github.com/heroku/heroku-buildpack-chromedriver.git
```

The orphaning and reaping is also reproducible via `dokku run APP rails c` and creating a chrome driver from Selenium there; not necessary to use Sidekiq workers for it.

josegonzalez:
Can someone confirm this for Dockerfile apps? I'd also like a test for this.

gyfis:
It seems that this is a major thing that could break stuff; maybe mentioning the command that does this via docker options in documentation would be sufficient?
```dokku docker-options:add APP run,deploy --init```

josegonzalez:
We could safely require this for buildpack apps - @michaelshobbs maybe we just build tini _into_ herokuish? - but Dockerfile apps seems... iffy.

josegonzalez:
Note: I'm not opposed to this, it just seems like a massive change that could break some workflows...

gyfis:
@michaelshobbs Yeah, I think it was stock. I don't know of any change to the config apart from the `.buildpacks` file.

josegonzalez:
I think for me to accept this change, I'll need some tests for dockerfile-tini/dockerfile-no-tini/buildpack. I don't want to introduce this change without actually testing that it resolves a problem and doesn't introduce new ones.

What docker version does circle run? 🤔

josegonzalez:
Closing. I think users can add this to their apps quite easily, and there haven't been any tests proving that this functionality is necessary (and works as expected).

FWIW, I'd be happy to scope this to just buildpack-deploys - I'm assuming that folks using a `Dockerfile` for deployment know what they are doing... - with the requisite testing. Feel free to comment if/when you have to work on that and I'll re-open.

jayjun:
@josegonzalez I do leak zombies, otherwise I wouldn’t have sent a pull request.

Sorry, I thought you wanted more feedback from Dockerfile deployments first. I’m happy to scope this to buildpack ones only and write tests since that’s what I’m using.

Testing zombie child processes doesn’t sound like an easy thing to do. I’ll appreciate if you can spare a little time to discuss how Dokku tests work, how CI works, testing strategies and so forth here or on Slack before I get started.

josegonzalez:
I think we need a unit test for both dockerfile and buildpack deploys. Something where we can say "this fails, and the code change makes it pass now". I don't think its necessarily a bad change, but adding new flags to the container running is fairly invasive and likely to lead to lots of support issues :(

josegonzalez:
Feel free to hit me up on chat if you have questions (I'm `savant`).

PR #3427, opened 2019-01-30
clarification about the location of Procfile

ker2x:
[ci skip]

somehow i misunderstood the documentation. I expected Dokku to automagically search in /app if the Procfile couldn't be found in WORKDIR. But it's not the case. I hope my tiny modification make it more clear.


josegonzalez:
This won't be true in an upcoming version of Dokku - once #3449 is merged - but I'll fix the docs in that PR as well.

Regardless, thanks for the pull request!

Issue #3418, opened 2019-01-24
`dokku apps:clone` can't work on CI because it always exits 1.


tamanobi:
## Description of problem
When I try to do the dokku tutorial(http://dokku.viewdocs.io/dokku/community/tutorials/deploying-with-gitlab-ci/#review-applications), I noticed that exit status of `dokku apps:clone` will be 1. But cloned app will work well.

I got Permission Denied because find walks cache (group id: 32767) directory.

I focus just exit status.

### How reproducible
Everytime.

### Steps to Reproduce

1. dokku apps:clone old-app new-app

#### Actual Results
```
$ dokku apps:clone like-dler like-dler2
-----> Creating like-dler2... done
find: ‘./cache/.heroku/python/lib’: Permission denied
find: ‘./cache/.heroku/python/include’: Permission denied
$ echo $?
1
```

#### Expected Results
`dokku apps:clone like-dler like-dler2` returns 0

## Environment Information
- dokku version: 0.14.2
- OS: 18.04.1 LTS (Bionic Beaver)
- Shell: GNU bash, version 4.4.19(1)-release (x86_64-pc-linux-gnu)
- find version: find (GNU findutils) 4.7.0-git

### `dokku report APP_NAME` output
Full log is here: https://gist.github.com/tamanobi/582499375dd9a319d83afd8dace2be51

### How (deb/make/rpm) and where (AWS, VirtualBox, physical, etc.) was Dokku installed?:
VPS(Sakura VPS)

### Additional information
I created patch to solve this issue and patch my `dokku`. The patch is here: https://github.com/tamanobi/dokku/pull/1/files
After this discussion, I will make a PR.

P.S. I created patch(https://github.com/dokku/dokku/pull/3434).

josegonzalez:
Can you show the output of:

```
dokku --trace apps:clone like-dler like-dler3
```

tamanobi:
@josegonzalez Thanks for your reply. The log is shown below.

```
$ dokku --trace apps:clone like-dler like-dler3
+ export DOKKU_HOST_ROOT=/home/dokku
+ DOKKU_HOST_ROOT=/home/dokku
+ export DOKKU_DISTRO
++ . /etc/os-release
++ echo ubuntu
+ DOKKU_DISTRO=ubuntu
+ export DOKKU_IMAGE=gliderlabs/herokuish
+ DOKKU_IMAGE=gliderlabs/herokuish
[...]

PR #3434, opened 2019-02-02
ignore cache directories when clone

tamanobi:
fix: https://github.com/dokku/dokku/issues/3418


josegonzalez:
Thanks for the pull request!

Issue #3291, opened 2018-10-24
Feature Request: Support passing environment variables to `dokku run`


ankane:
Hey, thanks for this awesome project. It's incredible!! I had trouble figuring out what to do with feature requests from the Contributing Guide, so let me know if this is the wrong place.

Currently, passing environment variables to commands like this:

```sh
dokku run HELLO=world rails console
```

Throws an error

```sh
setuidgid: fatal: unable to run HELLO=world: file does not exist
```

It'd be great if this set the `HELLO` variable and ran the command. This is the behavior on Heroku.

josegonzalez:
I dont see where in their docs they say they support this: https://devcenter.heroku.com/articles/one-off-dynos

ankane:
Hey @josegonzalez, I don't think it's explicitly documented, but think they try to have it behave as similar to a normal shell as possible. If you need an app to test it on, you can use [this link](https://heroku.com/deploy?template=https://github.com/pghero/pghero) for one click deploy.

There is a `-e` option that's documented in the CLI.

```sh
heroku help run
...
  -e, --env=env        environment variables to set (use ';' to split multiple vars)
```

which allows you to do

```sh
heroku run rails console -e "VAR1=value1;VAR2=value2"
```

josegonzalez:
Ah okay, looks like its a flag to heroku run. Can you show the full output of `heroku help run`?

ankane:
Yeah, here's the full output for the latest version (7.18.3):

```
run a one-off process inside a heroku dyno

USAGE
  $ heroku run

OPTIONS
  -a, --app=app        (required) app to run command against
  -e, --env=env        environment variables to set (use ';' to split multiple vars)
  -r, --remote=remote  git remote of app to use
  -s, --size=size      dyno size
  -x, --exit-code      passthrough the exit code of the remote command
  --no-notify          disables notification when dyno is up (alternatively use HEROKU_NOTIFICATIONS=0)
  --no-tty             force the command to not run in a tty
  --type=type          process type

DESCRIPTION
  Shows a notification if the dyno takes more than 20 seconds to start.

EXAMPLES
  $ heroku run bash
  Running bash on app.... up, run.1
  ~ $

  $ heroku run -s hobby -- myscript.sh -a arg1 -s arg2
  Running myscript.sh -a arg1 -s arg2 on app.... up, run.1

COMMANDS
  run:detached  run a detached dyno, where output is sent to your logs
```

alexquick:
[...]

PR #3477, opened 2019-03-23
Allow specifying environment variables for dokku run

alexquick:
This adds support for --env and -e
```
# dokku run [ --env KEY=VALUE | -e KEY=VALUE ] app command
dokku run --env NODE_ENV=development node-app npm run mytask
```

__note__  this changes the scheduler-run trigger to expect the following arguments:
`scheduler_name, app_name, env_count, env_0, ..., env_n, app_arg_0, ..., app_arg_n`.

__note__  as a courtesy this also supports `--env=KEY=VALUE` and `-e=KEY=VALUE` but not `-eKEY=VALUE`. I don't think it is worth refactoring argument parsing for dokku-standard or reimplementing in go at the moment. Now that so much of the logic that _was_ in dokku-standard is offloaded to the scheduler plugin it shouldn't be that hard to handle in the future.

resolves: dokku/dokku#3291


plugins/00_dokku-standard/subcommands/run
```diff
@@ -6,12 +6,38 @@ source "$PLUGIN_AVAILABLE_PATH/config/functions"
...
+        ;;
+    esac
+  done
+
   verify_app_name "$APP"
-  shift 2
 
   local DOKKU_SCHEDULER=$(get_app_scheduler "$APP")
-  plugn trigger scheduler-run "$DOKKU_SCHEDULER" "$APP" "$@"
+  plugn trigger scheduler-run "$DOKKU_SCHEDULER" "$APP" "${#RUN_ENV[@]}" "${RUN_ENV[@]}" "$@"
```
alexquick: Is there a better way to get this information to the scheduler? This is the best I could come up with that meets my goals of minimizing changes and processing inside the scheduler, but I have a feeling there's a better way.

josegonzalez: Given that there is only one scheduler in existence atm - docker-local - it's fine for us to change the interface in the next minor.

Maybe we pass a json blob of data for the run-env?

I do wonder what other types of info we'd pass to the `scheduler-run` trigger, and whether there is a better way generally to structure it. Assuming the docker-local plugin was in a higher-level language, do you have any suggestions on how this might change?

alexquick: Hmm, yeah I feel like json would could work. Any language triggering plugins or the scheduler itself would be written in would have robust support for json at the very least. I can foresee in the maybe future separating the app environment from the scheduler environment (or parts thereof), so it might be nice to use a format that has room for that kind of growth. I'm excited for the time when more and more things get added to it and someone finally blows through ARG_MAX :-p

On the other hand, this is really one call site and one implementation, so worrying about it now might be premature. Maybe leave it as it and revisit it when another scheduler is added or the next time new arguments need to be added to the `scheduler-run` trigger?

josegonzalez: If it works as is, thats fine by me.

Aside: http://dokku.viewdocs.io/dokku/advanced-usage/deployment-tasks/

josegonzalez:
Is this ready for review?

plugins/00_dokku-standard/subcommands/run
alexquick: I'm not quite following the relationship between deployment tasks and this. I do see that there are a number of JSON-in-bash utilities in dokku already that are exercised in that plugin, so threading this info through the trigger as JSON would be easier than I thought. Is there something else to deployment-tasks I should be thinking about?

alexquick:
@josegonzalez yup, this should be ready pending that clarification re: deployment-tasks

plugins/00_dokku-standard/subcommands/run
josegonzalez: Ah, I meant to link to the [kubernetes](https://github.com/dokku/dokku-scheduler-kubernetes) plugin - we also have a [nomad](https://github.com/dokku/dokku-scheduler-nomad) one - as scheduler implementations.

josegonzalez: In the context of having alternative schedulers already implemented, is that something that would change the implementation? If not, that's fine.

PR #3482, opened 2019-04-05
rails-database -> railsdatabase

multikatt:
Naming the database rails-database throws
```Please specify a valid name for the service. Valid characters are: [A-Za-z0-9_]+```

[ci skip]


josegonzalez:
Thanks for the pull request!

PR #3506, opened 2019-04-17
Update application-deployment.md

vanastassiou:
Clarity and style edits

[ci skip]

### DIFF
--- a/docs/deployment/application-deployment.md
+++ b/docs/deployment/application-deployment.md
@@ -1,11 +1,10 @@
 # Deploying to Dokku
 
-> Note: This document uses the hostname `dokku.me` in commands. For your server, please
-> substitute your server's hostname instead.
+> Note: This walkthrough uses the hostname `dokku.me` in commands. When deploying to your own server, you should set the `DOKKU_DOMAIN` value in the `Vagrantfile` before you initialize the Vagrant VM.
 
 ## Deploy tutorial
 
-Once Dokku has been configured with at least one user, applications can be deployed via a `git push` command. To quickly see Dokku deployment in action, you can use the Heroku Ruby on Rails example app.
+Once you have configured Dokku with at least one user, you can deploy applications using `git push`. To quickly see Dokku deployment in action, try using [the Heroku Ruby on Rails "Getting Started" app](https://github.com/heroku/ruby-getting-started).
 
 ```shell
 # from your local machine
@@ -15,7 +14,7 @@ git clone git@github.com:heroku/ruby-getting-started.git
 
 ### Create the app
 
-Create the application on the Dokku host. You will need to SSH onto the host to run this command.
+SSH into the Dokku host and create the application as follows:
 
 ```shell
 # on the Dokku host
@@ -24,7 +23,9 @@ dokku apps:create ruby-getting-started
 
 ### Create the backing services
 
-When you create a new app, Dokku by default *does not* provide any datastores such as MySQL or PostgreSQL. You will need to install plugins to handle that, but fortunately [Dokku has official plugins](/docs/community/plugins.md#official-plugins-beta) for common datastores. Our sample app requires a PostgreSQL service:
+Dokku by default **does not** provide datastores (e.g. MySQL, PostgreSQL) on a newly created app. You can add datastore support by installing plugins, and the Dokku project [provides official plugins](/docs/community/plugins.md#official-plugins-beta) for common datastores.
+
+The Getting Started app requires a PostgreSQL service, so install the plugin and create the related service as follows:
 
 ```shell
 # on the Dokku host
@@ -36,11 +37,11 @@ sudo dokku plugin:install https://github.com/dokku/dokku-postgres.git
 dokku postgres:create railsdatabase
 ```
 
-> Each service may take a few moments to create.
+Each service may take a few moments to create.
 
 ### Linking backing services to applications
 
-Once the service creation is complete, set the `DATABASE_URL` environment variable by linking the service.
+Once the services have been created, you then set the `DATABASE_URL` environment variable by linking the service, as follows:
 
 ```shell
 # on the Dokku host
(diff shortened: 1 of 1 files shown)


docs/deployment/application-deployment.md
```diff
@@ -1,77 +1,77 @@
 # Deploying to Dokku
 
-> Note: This document uses the hostname `dokku.me` in commands. For your server, please
-> substitute your server's hostname instead.
+**Note:** This walkthrough uses the hostname `dokku.me` in commands. When deploying to your own server, you should set the `DOKKU_DOMAIN` value in the `Vagrantfile` before you initialize the Vagrant VM.
```
josegonzalez: This should stay as a blockquote.

docs/deployment/application-deployment.md
```diff
@@ -1,77 +1,77 @@
...
+git remote add dokku dokku@dokku.me:rails-sample-app
 git push dokku master
 ```
 
-> Note: Some tools may not support the short-upstream syntax referenced above, and you may need to prefix
-> the upstream with the scheme `ssh://` like so: `ssh://dokku@dokku.me:ruby-rails-sample`
-> Please see the [Git](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) documentation for more details.
+**Note on `push` syntax:** Some tools may not support the short-upstream syntax referenced above, and you may need to prefix the upstream with `ssh://` (e.g. `ssh://dokku@dokku.me:ruby-sample-app`). (See the [Git](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) documentation for more information.)
+
+**Note on SSH keys:** Your private key should be registered with `ssh-agent` in your local development environment. If you get a `permission denied` error when pushing, you can register your private key as follows: `ssh-add -k ~/<your private key>`.
```
josegonzalez: These two should stay as block quotes.

docs/deployment/application-deployment.md
```diff
@@ -134,25 +134,25 @@ ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
 ssh -T git@github.com
 ```
 
-Note that if the buildpack or Dockerfile build process require SSH key access for other reasons, the above may not always apply.
+**Exception:** if the buildpack or Dockerfile build process require SSH key access for other reasons, the above may not always apply.
```
josegonzalez: Lets turn this into a blockquote.

docs/deployment/application-deployment.md
```diff
@@ -182,27 +182,27 @@ git push dokku master
...
 
-As of 0.5.x, this function removes all containers with the label `dokku` where the status is either `dead` or `exited`, as well as all `dangling` images. Previous versions would remove `dead` or `exited` containers, regardless of their label.
+One thing to be aware of (see [issue #1220](https://github.com/dokku/dokku/issues/1220)) is that Dokku executes a cleanup function prior to every deployment. As of 0.5.x, the cleanup removes all containers with the `dokku` label where the status is either `dead` or `exited` (previous versions would remove _all_ `dead` or `exited` containers). The cleanup function also removes all images with `dangling` status.
 
 ## Adding deploy users
 
 See the [user management documentation](/docs/deployment/user-management.md).
 
-## Default vhost
+## Setting a default vhost
```
josegonzalez: Mind reverting all these headers? They are essentially here as redirects for users who would have otherwise linked to the header.

docs/deployment/application-deployment.md
```diff
@@ -109,23 +109,23 @@ If you only want to rebuild and tag a container, you can skip the deployment pha
...
+dokku ps:rebuild rails-sample-app
 ```
 
 See the [process scaling documentation](/docs/deployment/process-management.md) for more information.
 
-### Deploying with private git submodules
+### Deploying with private `git` submodules
 
-Dokku uses Git locally (i.e. not a Docker image) to build its own copy of your app repo, including submodules. This is done as the `dokku` user. Therefore, in order to deploy private Git submodules, you'll need to drop your deploy key in `/home/dokku/.ssh/` and potentially add `github.com` (or your VCS host key) into `/home/dokku/.ssh/known_hosts`. The following test should help confirm you've done it correctly.
+Dokku uses `git` locally (i.e. not a Docker image) to build its own copy of your app repo, including submodules, as the `dokku` user. This means that, in order to deploy private `git` submodules, you need to put your deploy key in `/home/dokku/.ssh/` and potentially add `github.com` (or your VCS host key) into `/home/dokku/.ssh/known_hosts`. You can use the following test to confirm your setup is correct:
```
josegonzalez: We're referring to Git the project, not `git` the tool, in this paragraph, so please revert to `Git` instead of the code version :)

docs/deployment/application-deployment.md
```diff
@@ -1,77 +1,77 @@
...
+
+The Getting Started app requires a PostgreSQL service, so install the plugin and create the related service as follows:
 
 ```shell
-# on the Dokku host
-# install the postgres plugin
-# plugin installation requires root, hence the user change
+# On the Dokku host
+# Install the postgres plugin
+# Plugin installation requires root, hence the user change
```
josegonzalez: For code comments, the style in the documentation is not to use uppercase for the first character, so this should be reverted everywhere.

docs/deployment/application-deployment.md
```diff
@@ -182,27 +184,27 @@ git push dokku master
 
 ## Dokku/Docker container management compatibility
 
-Dokku is, at its core, a Docker container manager. Thus, it does not necessarily play well with other out-of-band processes interacting with the Docker daemon. One thing to note as in [issue #1220](https://github.com/dokku/dokku/issues/1220), Dokku executes a cleanup function prior to every deployment.
+Dokku is, at its core, a Docker container manager. Thus, it does not necessarily play well with other out-of-band processes interacting with the Docker daemon.
 
-As of 0.5.x, this function removes all containers with the label `dokku` where the status is either `dead` or `exited`, as well as all `dangling` images. Previous versions would remove `dead` or `exited` containers, regardless of their label.
+One thing to be aware of (see [issue #1220](https://github.com/dokku/dokku/issues/1220)) is that Dokku executes a cleanup function prior to every deployment. As of 0.5.x, the cleanup removes all containers with the `dokku` label where the status is either `dead` or `exited` (previous versions would remove _all_ `dead` or `exited` containers). The cleanup function also removes all images with `dangling` status.
```
josegonzalez: I try not to reference exact tickets in documentation - it's possible that we move issue trackers in the future, invaliding some of these links. Instead, documentation should be... self-referencing?

josegonzalez:
In the interest of having PRs not lay around and then subsequent ones conflicting - which is what happened here :( - I updated your PR and will merge it. Apologies if you were going to come back to it, but thanks for the PR nonetheless!

PR #3512, opened 2019-04-22
docs: fix the deploy tutorial

Mayeu:
Heroku's [ruby-rails-sample](https://github.com/heroku/ruby-rails-sample) does not run anymore on Dokku (or Heroku), and has been archived and replaced by [ruby-getting-started](https://github.com/heroku/ruby-getting-started).

This commit replace all the occurrences of `ruby-rails-sample` by `ruby-getting-started`.

[ci skip]


josegonzalez:
Some of this came in via #3506 but in the interest of keeping the docs up to date, I'll merge this in now.

PR #3513, opened 2019-04-22
Support installing plugins via file:// scheme

Mayeu:
This commit make it possible to install plugins via local git repositories that are accessed with the `file:` scheme.

In that case one has to point to the `.git` folder:
```
dokku plugin:install file:///path/to/the/repo/.git --name my-best-plugin
```


josegonzalez:
Can you add a test-case for this? Probably just clone our smoke-test plugin locally, and then reference the path for the install.

Mayeu:
Sure :+1: I will try to do that this week.

Mayeu:
I added a test based on the one using the remote git repo and adding the needed clone and cleaning in the setup and teardown.

josegonzalez:
Love it! Thanks for that :)

PR #3518, opened 2019-04-23
Add GitLab GIT_STRATEGY for stop_preview_app

renestalder:
Use `GIT_STRATEGY: none` in `stop_review_app` to make job work after successful
merges and when the branch was deleted.

Without this, the application is not stopped when the branch is deleted on merge as the runner pulls the git repository by default, but cannot find the tree in this moment.


josegonzalez:
Good catch, thanks!

PR #3523, opened 2019-04-27
Add an example how to specify a Dockerfile for deployment

MarcDiethelm:
[ci skip]


josegonzalez:
I think this is a bit complicated. You'll still _absolutely_ need a `Dockerfile` file with that exact name in the base of the repo - just adding this flag won't force a dockerfile deploy.

Mind updating the doc to note that? Maybe in a separate code-block with it's own explanatory text as opposed to the existing one.

PR #3528, opened 2019-05-06
feat: add support for quiet ps:scale output

### LINKED ISSUE #3520: Feature request: `--quiet` option for ps:scale
Would be great to have the `--quiet` option available for:

```
dokku ps:scale appname
```
to list the `web` and `worker` processes.

josegonzalez:
The previous output required too much parsing to be used in an automated fashion. This change makes that a bit easier, without sacrificing usability.

Note that the output is quite a bit different from heroku. This is purposeful as we do not have the concept of dynos.

Closes #3520


josegonzalez:
Output without quiet:

```
-----> Scaling for nodejs-express
proctype                  qty
--------                  ---
worker:                   0
web :                     0
```

Output with quiet:

```
worker:                   0
web :                     0
```

PR #3556, opened 2019-05-14
Add missing domains:clear-global command

josegonzalez:


josegonzalez:
I forgot to add this code in a previous commit...

PR #3585, opened 2019-06-09
Allow running Dokku in Docker

### LINKED ISSUE #1076: Docker hosted dokku

josegonzalez:
This PR allows users to run a released version of Dokku within a Docker container, making installs easy, regardless of base operating system.

Closes #1076


Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+ARG DOKKU_WEB_CONFIG=false
+ARG DOKKU_HOSTNAME=dokku.invalid
+ARG DOKKU_VHOST_ENABLE=false
+ARG DOKKU_SKIP_KEY_FILE=true
+
+RUN addgroup --gid $DOKKU_GID dokku && \
+    adduser --uid $DOKKU_UID --gid $DOKKU_GID --disabled-password --gecos "" "dokku"
+
+RUN mkdir /etc/nginx/
+COPY tests/dhparam.pem /etc/nginx/dhparam.pem
```
michaelshobbs: Why does this live in the `tests` subdir? Looks a bit out of place to me if it’s intended to be used in a production build.

michaelshobbs:
Couple questions:

What’s the final size of this image?
Thoughts on running the test suite in this built image?

Dockerfile
josegonzalez: Removed, it wasn't actually necessary.

josegonzalez:
The resulting image is 923MB.

The test suite could probably be run within the built image if we installed the extra dependencies for testing, but it would be a bit more difficult to do so for the purpose of CI. Maybe local testing would work?

michaelshobbs:
The `phusion/baseimage` currently sits at 209MB. I think it would be worth the time to pare down our image size OR at least, document the bloat in this PR and create some follow up tasks to get that size down. i was actually hoping we could use alpine or some other sort of super small base.

re: testing
it would be beneficial to have an integration test image that run the bats tests against this docker image prior to a release. i know that will make the test suite runtime longer but i think a guarantee on a working (at least integration test passing) image will be good for the community and for our (really your) sanity.

josegonzalez:
Seems like the installation of Dokku pulls in quite a few things ([gcc](https://github.com/dokku/dokku/commit/b6767f8efc10bbf4df2ab99b9f1e97378aea45fd) was introduced as a dependency really early on in our packaging lifecycle). We can almost certainly drop a few dependencies, but I don't think its worth spending too much time on this at this moment.

Dokku itself is around 70mb, as we haven't deduplicated the golang binaries.

I think that switching to alpine is feasible in the future if someone can get around to making a version that uses runit, though I'd like to see us have an apk package for Dokku + deps when we do that.

Regarding testing, would a simple smoke test be fine? I'm not sure what running bats would buy us, as the integration test is testing whether the built deb package is "good", and we already use those built packages at the moment for the CI install.

michaelshobbs:
> We can almost certainly drop a few dependencies, but I don't think its worth spending too much time on this at this moment.

ok. this still needs to be captured as a follow up task/issue

> Regarding testing, would a simple smoke test be fine? I'm not sure what running bats would buy us, as the integration test is testing whether the built deb package is "good", and we already use those built packages at the moment for the CI install.

i get what you're saying about the fact we are testing the deb package in CI currently. there's nothing explicitly testing that the image is functional however. i actually think we should consider managing the docker file/image in another repo. then in that repo CI would build and test the image

josegonzalez:
Mind filing an issue to drop gcc/make as dependencies?

---

I'm not so sure it should live in another repository. We don't do this for our deb or rpm packaging, and only do so for the arch packaging - which has made it easy to forget to update it and it's now woefully out of date.

The release process is currently:

```
contrib/release-dokku $RELEASE_TYPE
```

Keeping it in the same repository and building the image on release seems logical to me, otherwise we're looking at lag for updating that image when there doesn't need to be any.

What benefits would there be in separating the dockerfile from the main repository?

docker/etc/my_init.d/10_dokku_init
```diff
@@ -0,0 +1,57 @@
...
+log-fail() {
+  declare desc="Log fail formatter"
+  echo "!   $*" 1>&2
+  exit 1
+}
+
+main() {
+  if [[ ! -d /mount ]]; then
+    log-info "Creating missing /mount"
+    mkdir -p /mount
```
michaelshobbs: any reason not to use the more standardized `/mnt` directory?

josegonzalez: None other than forgot about it :P

docker/etc/my_init.d/10_dokku_init
```diff
@@ -0,0 +1,57 @@
...
+  fi
+
+  directories=("/etc/ssh" "/home/dokku" "/var/lib/dokku/config" "/var/lib/dokku/data")
```
michaelshobbs: putting my functional programmer hat on, i think i'd prefer to see these sections broken up into their own functions with descriptive names. i think this helps in the self document code aspect of maintainability but will defer to you on whether you think it's worth it

docker/etc/my_init.d/10_dokku_init
josegonzalez: Inside the container, I'll use `/mnt/dokku`. That seems the most appropriate, given LSB docs concerning what goes into `/mnt`.

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN addgroup --gid $DOKKU_GID dokku && \
+    adduser --uid $DOKKU_UID --gid $DOKKU_GID --disabled-password --gecos "" "dokku"
+
+RUN echo "dokku dokku/hostname string $DOKKU_HOSTNAME" | debconf-set-selections && \
+    echo "dokku dokku/skip_key_file boolean $DOKKU_SKIP_KEY_FILE" | debconf-set-selections && \
+    echo "dokku dokku/vhost_enable boolean $DOKKU_VHOST_ENABLE" | debconf-set-selections && \
+    echo "dokku dokku/web_config boolean $DOKKU_WEB_CONFIG" | debconf-set-selections && \
+    curl -sSL https://packagecloud.io/dokku/dokku/gpgkey | apt-key add - && \
+    echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ bionic main" | tee /etc/apt/sources.list.d/dokku.list && \
+    apt-get update && apt-get -y upgrade && apt-get --no-install-recommends -qq -y install \
```
michaelshobbs: minor nit:
for readability, i think this should be broken up thusly:
```
apt-get update && apt-get -y upgrade && \
  apt-get --no-install-recommends -qq -y install dokku rsync && \
```

josegonzalez:
@michaelshobbs as discussed via chat:

- You're taking over the PR. Feel free to rebase/force push on it
- The dockerfile will be modified to use a local deb package from the `build` directory. So testing the dockerfile would mean at least running `contrib/release-dokku` in some form, which is totally fine as we're going to stick with manual releases for now.
- The `setup.sh` and test harness will be modified to run the docker container and then run tests against the container.

michaelshobbs:
closing and re-opening to get codacy to analyze

PR #3587, opened 2019-06-10
feat: drop make and gcc as dependencies

josegonzalez:
They aren't actually used within Dokku.


josegonzalez:
Those are handled in a different repository, and haven't been updated since we started the move to golang. Arch packages are basically "install from source", which doesn't quite work for half-compiled apps. I'd like to eventually update our release process such that the compiled source is attached to the GitHub release, and then arch can use that to build packages.

PR #3589, opened 2019-06-11
Allow running Dokku in Docker

### LINKED ISSUE #1076: Docker hosted dokku

michaelshobbs:
Closes #1076
Replaces #3585


michaelshobbs:
@josegonzalez ping

tests.mk
```diff
@@ -61,49 +60,39 @@ endif
...
-	echo "Host dokku.me \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
+	echo "Host dokku.me \\r\\n Port 3022 \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
 	echo "Host 127.0.0.1 \\r\\n Port 22333 \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
 else ifeq ($(shell grep dokku.me /root/.ssh/config),)
-	echo "Host dokku.me \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
+	echo "Host dokku.me \\r\\n Port 3022 \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
 	echo "Host 127.0.0.1 \\r\\n Port 22333 \\r\\n RequestTTY yes \\r\\n IdentityFile /root/.ssh/dokku_test_rsa" >> /root/.ssh/config
 endif
 
-ifneq ($(wildcard /etc/ssh/sshd_config),)
```
josegonzalez: Do we care that the tests will default to testing in docker and not a regular host? In theory it's the same since we don't - currently - have any daemons, but just wanted to make sure you're thinking along the same lines.

tests/ci/setup.sh
```diff
@@ -25,54 +27,78 @@ install_dependencies() {
...
+  docker build -t dokku/dokku:test .
+}
+
+run_dokku_container() {
+  echo "=====> run_dokku_container on CIRCLE_NODE_INDEX: $CIRCLE_NODE_INDEX"
+  docker run -d \
+    --env DOKKU_HOSTNAME=dokku.me \
+    --name dokku \
+    --publish 3022:22 \
+    --publish 22333:22 \
```
josegonzalez: Why two ports?

tests.mk
michaelshobbs: Ostensibly this change is testing all the same code and is only bypassing the hosts ssh config changes.

tests/ci/setup.sh
```diff
@@ -25,54 +27,78 @@ install_dependencies() {
...
+
+  check_container
+}
 
-  sudo dpkg -i "${ROOT_DIR}/build/$HEROKUISH_PACKAGE_NAME"
-  sudo dpkg -i "${ROOT_DIR}/build/$PLUGN_PACKAGE_NAME"
-  sudo dpkg -i "${ROOT_DIR}/build/$SSHCOMMAND_PACKAGE_NAME"
-  sudo dpkg -i "${ROOT_DIR}/build/$SIGIL_PACKAGE_NAME"
-  sudo dpkg -i "${ROOT_DIR}/build/$PROCFILE_UTIL_PACKAGE_NAME"
+check_container() {
```
josegonzalez: Shame there isn't a `SIGREADY` or some generic way to wait for a process to signal as such...

tests/ci/setup.sh
```diff
@@ -25,54 +27,78 @@ install_dependencies() {
...
-  # need to add the dokku user to the docker group
-  sudo usermod -G docker dokku
-  [[ "$1" == "buildstack" ]] && BUILD_STACK=true make -e stack
 
   install_dependencies
   install_dokku
 
   sudo -E make -e setup-deploy-tests
-  bash --version
-  docker version
```
josegonzalez: Should we have this output from the built container?

tests/ci/setup.sh
michaelshobbs: https://github.com/dokku/dokku/blob/48484732a0bdc2b6e384b5e331196e2f1c6afa45/tests/unit/40_core_2.bats#L106-L118

tests/ci/setup.sh
michaelshobbs: i don't know that's useful these days. we can always add it back if we find a need

tests.mk
```diff
@@ -42,13 +42,12 @@ else
...
 
-ci-dependencies: bats shellcheck shfmt xmlstarlet
+ci-dependencies: bats shellcheck xmlstarlet
 
 setup-deploy-tests:
-	mkdir -p /home/dokku
 ifdef ENABLE_DOKKU_TRACE
 	echo "-----> Enabling tracing"
-	echo "export DOKKU_TRACE=1" >> /home/dokku/dokkurc
+	docker exec -ti dokku bash -c 'echo "export DOKKU_TRACE=1" >> /home/dokku/dokkurc'
```
josegonzalez: Should just be `dokku trace:on`, no? (The subcommand was added recently)

tests/ci/setup.sh
josegonzalez: Fair. I don't find it useful, so we can leave it off for now.

michaelshobbs:
i've reverted the changes made to test against the docker image. the plan is to do a minimal set of deploy-only tests against the docker image

michaelshobbs:
@josegonzalez this is g2g

tests.mk
```diff
@@ -47,7 +48,7 @@ ci-dependencies: bats shellcheck xmlstarlet
 setup-deploy-tests:
 ifdef ENABLE_DOKKU_TRACE
 	echo "-----> Enabling tracing"
-	docker exec -ti dokku bash -c 'echo "export DOKKU_TRACE=1" >> /home/dokku/dokkurc'
+	dokku trace:on
```
josegonzalez: Why this revert? I thought you wanted to test the built artifact?

michaelshobbs: https://github.com/dokku/dokku/pull/3589#issuecomment-510973346

we unit test just the installed version and test deploys against both

josegonzalez:
Rebasing because there were test changes in master regarding CI upgrading to 16.04 recently.

PR #3602, opened 2019-07-01
Create a 502 error page to automatically refresh if backend status changes

alexymik:
If a container goes down and the gateway timeout returns a 502, periodically check the backend using a HEAD request. If the 502 is resolved (container reboots, etc.) then automatically refresh the page. Also uses a backoff to increase the time between checks.

### DIFF
--- a/plugins/nginx-vhosts/templates/502-error.html
+++ b/plugins/nginx-vhosts/templates/502-error.html
@@ -43,7 +43,6 @@
             retry_current *= 2;
           }
 
-          console.dir('Trying', window.location.href, 'in', retry_current, 'seconds...');
           setTimeout(check_gateway, retry_current * 1000);
         }
       };
--- a/plugins/nginx-vhosts/templates/502-error.html
+++ b/plugins/nginx-vhosts/templates/502-error.html
@@ -34,8 +34,6 @@
 
       request.onload = function () {
 
-        console.dir(window.location.href, 'returned', request.status);
-
         if (request.status >= 200 && request.status < 400) {
           window.location.reload(true);
         } else {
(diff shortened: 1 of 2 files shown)


plugins/nginx-vhosts/templates/502-error.html
```diff
@@ -0,0 +1,64 @@
...
+    var retry_limit = 4096; // Maximum backoff time, ~68 minutes
+
+    check_gateway = function () {
+
+      var request = new XMLHttpRequest();
+      request.open('HEAD', window.location.href, true);
+
+      request.onload = function () {
+
+        console.dir(window.location.href, 'returned', request.status);
```
josegonzalez: Can you remove the console log calls?

alexymik: Done

alexymik:
The page title could be renamed to something else and the return status could be modified to check for only 502. I had it in the 500-error.html originally

plugins/nginx-vhosts/templates/502-error.html
josegonzalez: Missed one

alexymik: Oops, sorry

plugins/nginx-vhosts/templates/502-error.html
```diff
@@ -0,0 +1,61 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>We're sorry, but something went wrong (500)</title>
```
josegonzalez: Should say 502 here (just to distinguish it since we can).

PR #3609, opened 2019-07-08
Start long running containers with --init with tests

jayjun:
See #3422 for earlier discussion.

Tests added as requested. The tests deploy a Go web app that spawns orphaned zombie processes then assert they are reaped.

Three types of apps are tested,
- Buildpack
- Dockerfile
- Dockerfile with Tini


josegonzalez:
Can you repush _just_ the tests, wait till they fail - tests should fail for all three types of apps you mentioned - _then_ push the fix? I want to be sure this actually fixes a problem that we have

jayjun:
@josegonzalez Rebased and pushed with just tests.

```
not ok 28 (zombies) buildpack in 52sec
ok 29 (zombies) dockerfile no tini # skip previous test failed! skipping remaining tests...
ok 30 (zombies) dockerfile with tini # skip previous test failed! skipping remaining tests...
```

However, Dockerfile tests are skipped. Should I split the tests into separate Bats files?

josegonzalez:
Yeah that might be best. Maybe put them in:

- `50_init_1.bats`
- `50_init_2.bats`
- `50_init_3.bats`

jayjun:
@josegonzalez Done. `50_init_3.bats` still passes because it tests Dockerfile with Tini, which _should_ reap zombies before the patch.

The point of the test is to make sure everything still works after the patch, namely running Tini with Tini.

josegonzalez:
Awesome, can you add the next commit now to add `--init`? Please don't rebase out that existing commit, I want it to be a record of this actually fixing something, or I'm liable to revert this at first sign of trouble in the future 😂

plugins/scheduler-docker-local/scheduler-deploy
```diff
@@ -98,16 +98,16 @@ scheduler-docker-local-scheduler-deploy() {
...
           # shellcheck disable=SC2086
-          cid=$("$DOCKER_BIN" run "${DOCKER_RUN_LABEL_ARGS[@]}" $DOKKU_GLOBAL_RUN_ARGS -d $DOKKU_DOCKER_PORT_ARGS -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
+          cid=$("$DOCKER_BIN" run "${DOCKER_RUN_LABEL_ARGS[@]}" $DOKKU_GLOBAL_RUN_ARGS --init -d $DOKKU_DOCKER_PORT_ARGS -e PORT=$DOKKU_PORT "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
         fi
       else
         START_CMD=$(fn-scheduler-docker-local-extract-start-cmd "$APP" "$PROC_TYPE" "$START_CMD" "$DOKKU_HEROKUISH")
 
         # shellcheck disable=SC2086
-        cid=$("$DOCKER_BIN" run "${DOCKER_RUN_LABEL_ARGS[@]}" $DOKKU_GLOBAL_RUN_ARGS -d "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
+        cid=$("$DOCKER_BIN" run "${DOCKER_RUN_LABEL_ARGS[@]}" $DOKKU_GLOBAL_RUN_ARGS --init -d "${ARG_ARRAY[@]}" $IMAGE $START_CMD)
```
josegonzalez: Could we instead add this once to `DOKKU_GLOBAL_RUN_ARGS` in the `dokku` binary?

jayjun: The intent is to target long running containers only, because zombies are reaped after a container stops.

Adding that to `$DOKKU_GLOBAL_RUN_ARGS` would mean using Tini everywhere, from building images to renaming apps.

josegonzalez:
@jayjun can you take a look at the init tests? They are extremely flaky, and usually cause CI to fail.

PR #3610, opened 2019-07-09
Correct Bats path in single tests

jayjun:
**Fixed paths to Bats script**
Bats-core moved scripts to `libexec/bats-core/` (see https://github.com/bats-core/bats-core/commit/e8a2774534c070990b4b13818ac8ddacae6009c5).

**Set `BATS_TEST_PATTERN` for `bats-preprocess`**
Regex pattern for tests moved out of `bats-preprocess` into `bats` (see https://github.com/bats-core/bats-core/commit/f5acd286129e3b68945a33da7d74ece82564ac57).

Now the pattern must be exported as `BATS_TEST_PATTERN` before executing `bat-preprocess`, so that line is grepped and executed.

`eval` is used because otherwise `"` and `\$` are read literally.

### DIFF
--- a/tests/bats-exec-test-single
+++ b/tests/bats-exec-test-single
@@ -46,7 +46,8 @@ main() {
   TMP_OUTPUT=$(mktemp "/tmp/${FUNCNAME[0]}.XXXX")
   trap 'rm -rf "$TMP_OUTPUT" >/dev/null' RETURN INT TERM EXIT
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
I use a bats fork to implement the junit support so this might not work...

josegonzalez:
Oh heck it does!

PR #3613, opened 2019-07-13
Remove all uses of proxy_set_header Connection "upgrade"

Lyelt:
Remove all uses of proxy_set_header Connection "upgrade"

Setting this header to "upgrade" was fixed in the nginx.conf.sigil file in [this commit](https://github.com/dokku/dokku/commit/4e4e256a2774621223efa168809848adc826c950). It should be fixed in the remaining docs/tests where it was still being set.


josegonzalez:
Thanks for the pull request!

Issue #3606, opened 2019-07-05
config:set and config:unset do not verify correct number of args


alexymik:
## Description of problem

It is possible to run `dokku config:set` and `config:unset` without specifying any environment variables, which could result in a needless container reboot.

### How to reproduce

Run `dokku config:set app` or `dokku config:unset app` without specifying a key/value or key, respectively.

#### Actual Results

The application will be restarted without any changes to environment variables.

#### Expected Results

The command should be rejected due to insufficient args.

### Additional information

From documentation:

```
config:set [--encoded] [--no-restart] (<app>|--global) KEY1=VALUE1 [KEY2=VALUE2 ...]  Set one or more config vars
config:unset [--no-restart] (<app>|--global) KEY1 [KEY2 ...]                          Unset one or more config vars
```

josegonzalez:
@alexquick Is this something you have time to work on?

alexquick:
Sure thing, I'll take care of it in the next few days.

PR #3614, opened 2019-07-14
[config] validate args for config:set and config:unset

alexquick:
Fixes dokku/dokku#3606

```
$ dokku config:unset envtest
FAILED: At least one key must be given
$ dokku config:set envtest
FAILED: At least one env pair must be given
```

Related: at the moment it's a warning to `config:unset` a key that doesn't exist but the app is still restarted. Likewise, setting a key to the same value will restart the app. Heroku seems to restart the app unconditionally in those cases like dokku does, so that is all OK right?

### DIFF
--- a/plugins/config/subcommands.go
+++ b/plugins/config/subcommands.go
@@ -62,6 +62,9 @@ func CommandClear(args []string, global bool, noRestart bool) {
 //CommandUnset implements config:unset
 func CommandUnset(args []string, global bool, noRestart bool) {
 	appName, keys := getCommonArgs(global, args)
+	if len(keys) == 0 {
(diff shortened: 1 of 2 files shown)


josegonzalez:
I'm of the opinion that if the configuration of the app doesn't change, we shouldn't also trigger a deploy of the app. My reasoning is that there is no actual state change to "reload", so in effect the command would then be "idempotent" in a sense.

Happy to have other opinions against that. Thoughts from you or @michaelshobbs ?

josegonzalez:
Merging. Thinking about this more, consistency with what Heroku does is important, and if we require removing the extra restart in the future, we can do that at a minor release.

PR #3623, opened 2019-07-29
Add web installer screenshot to docs

alexymik:
It might be a bit unclear that there is a web installer component on compatible systems. Adding a screenshot might make it more clear.


josegonzalez:
I'd actually like to get rid of the web installer in favor of moving that to the CLI. The reason is that many folks end up _not_ following the instructions - regardless of what we tell them - and then leave the installer up for any enterprising bitcoin miner to hack.

josegonzalez:
I'm going to close this. The install docs are already a bit long, but if people can't be bothered to read them, then adding a screenshot won't help. We're also going to remove this completely at some point before the 1.0, so adding the screenshot to remove it later seems unnecessary to me.

Regardless, thanks for the PR :)

PR #3650, opened 2019-08-15
Clarify --cpu mapping for docker-local scheduler

vincelwt:
[ci skip]

Clarifies a bit the default mapping of `resource:limit` for `cpu` to the underlying docker-local scheduler `--cpus` flag. 
Should make it clearer what the examples actually do when using `resource:limit --cpu 100`.


josegonzalez:
I opted to point to the docker-local scheduler docs, and also have _those_ docs point off to the official Docker documentation. If the upstream docs don't cut it, I think its best that folks explain _there_ what the docker flags mean, rather than us coming up potentially bad examples.

PR #3655, opened 2019-08-20
Fix filename of sshcommand tarball

znz:


josegonzalez:
Wow how did I not catch this...

PR #3659, opened 2019-08-21
nginx-vhost: add support for grpc/grpcs port forward

### LINKED ISSUE #3658: Add support for grpc schema in port forward
Currently, two `http` and `https` schema is supported using Nginx vhost (based on the template) Nginx supports the grpc backend, add support for grpc backend in the Nginx vhost plugin, so adding grpc port would be like this : 
```
dokku proxy:ports-add <app> grpc:XXXX:YYYY 
```
Also supporting secure grpc (grpcs?) is possible.

fzerorubigd:
fixes #3658 

I am not sure how I can add test for this change, but I test it in my installation, I was able to connect to the gRPC server (using go server and client) in both secure (with lets-encrypt certs) and insecure (with the insecure option in client). 

The support for grpc is added in nginx `1.13.10` (here: https://www.nginx.com/blog/nginx-1-13-10-grpc/)  also it checks if the current nginx supports `http2` or not (basically the current implementation check only for version, so if the grpc is enabled, http2 is enabled too, but I prefer to pass two different variables here, maybe switch to some other proper feature testing method?)


josegonzalez:
You'll need to add a grpc-enabled app in `tests/apps`, and then add the requisite test deploy in either of the `nginx-vhosts.bats` files. The test should make a grpc call - maybe try with curl? - to the service and get the appropriate response back.

fzerorubigd:
Now I understand, but there is no way to test the gRPC using curl only. I can use a simple go client, but the question is, is it possible to build a Go app during the test? Not in Dokku way, just building one in the bats file, use it to call the grpc endpoint. also, I can use Python, if installing Go is not an option, but then we need some specific python module.

josegonzalez:
A simple go client is fine, build/run it in a dockerfile.

fzerorubigd:
@josegonzalez added some test, and a note in docs, not sure if it's enough.

plugins/nginx-vhosts/functions
```diff
@@ -259,6 +259,29 @@ is_http2_enabled() {
...
+      HAS_SUPPORT=true
+    elif [[ "$MINOR_VERSION" -ge "14" ]]; then
+      HAS_SUPPORT=true
+    fi
+  fi
+
+  echo $HAS_SUPPORT
+}
+
+
```
josegonzalez: Remove this extra newline please.

tests/unit/40_nginx-vhosts_3.bats
```diff
@@ -0,0 +1,34 @@
...
+teardown() {
+  destroy_app 0 $TEST_APP
+  [[ -f "$DOKKU_ROOT/VHOST.bak" ]] && mv "$DOKKU_ROOT/VHOST.bak" "$DOKKU_ROOT/VHOST" && chown dokku:dokku "$DOKKU_ROOT/VHOST"
+  [[ -f "$DOKKU_ROOT/HOSTNAME.bak" ]] && mv "$DOKKU_ROOT/HOSTNAME.bak" "$DOKKU_ROOT/HOSTNAME" && chown dokku:dokku "$DOKKU_ROOT/HOSTNAME"
+  global_teardown
+}
+
+@test "(nginx-vhosts) grpc endpoint" {
+  deploy_app gogrpc
+  dokku proxy:ports-add "$TEST_APP" "grpc:80:50051"
```
josegonzalez: Can you add a test where we use `8080` instead of port `80` for the host as well?

fzerorubigd:
@josegonzalez Done.

fzerorubigd:
Thank you!

zph:
@josegonzalez @fzerorubigd Thank you both for working on this. I'm interested to try it out :). Is the best way to do it pre-release to install dokku from scratch on a new VM using:

```
# Use most recent bootstrap
wget https://raw.githubusercontent.com/dokku/dokku/v0.18.3/bootstrap.sh
sudo DOKKU_BRANCH=master bash bootstrap.sh
```

Or is there a better way to attempt this? Thanks you!

fzerorubigd:
There is a Vagrantfile in the project root, I never used it tho. Makefile is my choice on installing and testing on a local machine.

zph:
@fzerorubigd I'll give those a whirl. Thank you 👍

zph:
> There is a Vagrantfile in the project root, I never used it tho. Makefile is my choice on installing and testing on a local machine.

@fzerorubigd I've been using the GRPC mechanism you setup with this PR and 👍 it's going well! Thank you.

PR #3697, opened 2019-09-24
Migrate builds plugin to go and track per-build records

josegonzalez:
- Migrates the `builds` plugin to Go and starts tracking every deploy as a structured record on disk under `data/builds/<app>/<build-id>.{json,log}`, keyed on a stable `DOKKU_BUILD_ID`.
- New commands `builds:list`, `builds:info`, `builds:cancel`, `builds:output`, `builds:prune`, `builds:report`, and `builds:set retention` let operators inspect running and historical builds without grepping journalctl.

Related: https://github.com/heroku/heroku-builds


josegonzalez:
I might not actually add the builds:list or builds:info commands. Still thinking about this PR.

PR #3703, opened 2019-10-03
Update DOKKU_SCALE from Procfile during ps:scale

scjody:
Part of https://github.com/plotly/streambed/issues/13523

This is intended to make the minimal change needed to fix the problem we're seeing. I've logged internal issues for:
* Consider how to handle processes removed from the `Procfile` that are already in `DOKKU_SCALE`: https://github.com/plotly/streambed/issues/13593
* `ps:scale` cleanups (avoid creating `DOKKU_SCALE` during the "can-scale" check, and consider some refactoring in `generate_scale_file`): https://github.com/plotly/streambed/issues/13594

I'm happy to log these in this repo as well if you think that's useful.

### DIFF
--- a/plugins/ps/functions
+++ b/plugins/ps/functions
@@ -194,7 +194,17 @@ ps_scale() {
   local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
   shift 1
 
-  [[ ! -e $DOKKU_SCALE_FILE ]] && generate_scale_file "$APP" "$IMAGE_TAG" >/dev/null 2>&1
+  # shellcheck disable=SC2015
+  [[ ! -e $DOKKU_PROCFILE ]] && extract_procfile "$APP" "$IMAGE_TAG" >/dev/null 2>&1 || true
+  if [[ ! -e $DOKKU_SCALE_FILE ]] ; then
+    generate_scale_file "$APP" "$IMAGE_TAG" >/dev/null 2>&1
+  elif [[ -e $DOKKU_PROCFILE ]] ; then
+    for PROC_NAME in $(procfile-util list --procfile "$DOKKU_PROCFILE"); do
+      if ! grep -qE "^${PROC_NAME}=" "$DOKKU_SCALE_FILE"; then
+        echo "$PROC_NAME=0" >> "$DOKKU_SCALE_FILE"
+      fi
+    done
+  fi
   if [[ -z "$@" ]]; then
     dokku_log_info1_quiet "Scaling for $APP"
     dokku_col_log_msg_quiet "proctype" "qty"
--- a/tests/unit/10_ps-general.bats
+++ b/tests/unit/10_ps-general.bats
@@ -41,6 +41,31 @@ EOF
   assert_output "node worker.js"
 }
 
+@test "(ps:scale) update DOKKU_SCALE from Procfile" {
+  local TMP=$(mktemp -d "/tmp/dokku.me.XXXXX")
+  trap 'popd &>/dev/null || true; rm -rf "$TMP"' INT TERM
+
+  CUSTOM_TMP="$TMP" deploy_app nodejs-express
+
+  run /bin/bash -c "dokku --quiet ps:scale $TEST_APP"
+  output=$(echo "$output" | tr -s " ")
+  echo "output: ($output)"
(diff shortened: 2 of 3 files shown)


plugins/ps/functions
```diff
@@ -194,7 +194,16 @@ ps_scale() {
   local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
   shift 1
 
-  [[ ! -e $DOKKU_SCALE_FILE ]] && generate_scale_file "$APP" "$IMAGE_TAG" >/dev/null 2>&1
+  [[ ! -e $DOKKU_PROCFILE ]] && extract_procfile "$APP" "$IMAGE_TAG"
+  if [[ ! -e $DOKKU_SCALE_FILE ]] ; then
+    generate_scale_file "$APP" "$IMAGE_TAG"
```
josegonzalez: The output here should be redirected as before, right?

plugins/ps/functions
```diff
@@ -194,7 +194,16 @@ ps_scale() {
...
+  else
+    for PROC_NAME in $(procfile-util list --procfile "$DOKKU_PROCFILE"); do
```
josegonzalez: We should check if `DOKKU_PROCFILE` exists. If an app is not deployed but a user calls `ps:scale`, then no Procfile is extracted, and this will probably fail.

plugins/ps/functions
```diff
@@ -194,7 +194,16 @@ ps_scale() {
   local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
   shift 1
 
-  [[ ! -e $DOKKU_SCALE_FILE ]] && generate_scale_file "$APP" "$IMAGE_TAG" >/dev/null 2>&1
+  [[ ! -e $DOKKU_PROCFILE ]] && extract_procfile "$APP" "$IMAGE_TAG"
```
josegonzalez: Might want to append `>/dev/null 2>&1 || true` to the end of this so we don't get any errors when the app isn't deployed yet and doesn't have a built image.

josegonzalez:
This is a bit simpler than my initial stab would have been - rebuilding the file during `core-post-deploy`. Outside of my comments, should work unless the image was removed (which should be fairly rare).

plugins/ps/functions
scjody: 👍

The only case I can see where this will fail is if there are errors in the `Procfile`, but we don't want to fail in those cases either.

plugins/ps/functions
scjody: Yes, fixed.

plugins/ps/functions
scjody: 👍

scjody:
@josegonzalez This is ready for re-review. I expect tests to pass (a previous run with just `10_ps-general.bats` succeeded).

I logged two internal issues for future work. I'm happy to log these in this repo as well if you think that's useful.

josegonzalez:
Would be good to log here in Dokku - there is a possibility someone else out there grabs the issue :)

plugins/ps/functions
```diff
@@ -194,7 +194,17 @@ ps_scale() {
   local DOKKU_PROCFILE="$DOKKU_ROOT/$APP/DOKKU_PROCFILE"
   shift 1
 
-  [[ ! -e $DOKKU_SCALE_FILE ]] && generate_scale_file "$APP" "$IMAGE_TAG" >/dev/null 2>&1
+  # shellcheck disable=SC2015
+  [[ ! -e $DOKKU_PROCFILE ]] && extract_procfile "$APP" "$IMAGE_TAG" >/dev/null 2>&1 || true
```
josegonzalez: Can we change this to:

```shell
if [[ ! -e $DOKKU_PROCFILE ]]; then
  extract_procfile "$APP" "$IMAGE_TAG" >/dev/null 2>&1 || true
fi
```

I think I saw a lint error related to this last night, hence the request.

scjody: I added `# shellcheck disable=SC2015` to fix the lint error. According to the ShellCheck wiki this is the correct thing to do in this case. I'm happy to change the code to an `if` if you still feel like it's useful though.

josegonzalez: Eh this is fine for now then. At some point I'll rewrite this plugin in Golang, at which point all the shellcheck issues go away :)

PR #3707, opened 2019-10-08
Isolated ps:scale

gyfis:
This is a draft / semi-pseudocode PR for implementing a new feature: **isolated ps:scale**. I'll try to improve the PR more, but wanted to put it out to catch early feature design issues.

Use case: When you call **dokku ps:scale [APP] ...**, you may not want to restart every dyno; instead, you may want to only stop and start the dyno(s) mentioned in the ps:scale command. This PR tries to implement this feature, via a config `DOKKU_ISOLATED_SCALE`.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -561,6 +561,7 @@ release_and_deploy() {
 
   local APP="$1"
   local IMAGE_TAG="$2"
+  local CUSTOM_SCALE="$3"
   local IMAGE=$(get_app_image_name "$APP" "$IMAGE_TAG")
   local DOKKU_DOCKERFILE_PORTS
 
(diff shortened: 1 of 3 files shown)


josegonzalez:
Why not just always isolate scaling?

gyfis:
@josegonzales I didn’t want to break backward compatibility (e.g. if someone’s using it for app restarts), but it makes sense to me to make ps:scale always isolated.

josegonzalez:
If they are using it for app restarts then they should switch to `ps:restart`, which is more appropriate. I'm less concerned about breaking that use case.

josegonzalez:
Closing this for now as I'll be rewriting the ps plugin in golang in an upcoming PR, which will make these sorts of changes more feasible in the future.

gyfis:
@josegonzalez That sounds wonderful! Good luck :) Thanks for closing this!

Issue #3598, opened 2019-06-23
Installing Vagrant's dokku-windows only shows nginx screen


mmahandev:
## Description of problem

Doing a fresh install of dokku-windows on vagrant shows a default "Welcome to nginx!" on [dokku.me](http://dokku.me/):

![image](https://user-images.githubusercontent.com/20646367/59974809-eeb40c00-9575-11e9-97cd-149da0a91504.png)

### Steps to Reproduce

Done by following http://dokku.viewdocs.io/dokku/getting-started/install/vagrant/

1. git clone https://github.com/dokku/dokku.git dokku-test
2. cd dokku-test
3. vagrant up dokku-windows
4. Setup SSH Config in ~/.ssh/config (Did this on the host machine)
5. Access [dokku.me](http://dokku.me/)

#### Actual Results

Shows an nginx page

#### Expected Results

Should show the dokku page

## Environment Information

Windows 10 64-bit. Vagrant 2.2.4

### `dokku report APP_NAME` output

<details>
<summary>I don't know what APP_NAME should be so I just ran `dokku report` after ssh'ing in:</summary>
-----> uname: Linux dokku 4.15.0-51-generic #55-Ubuntu SMP Wed May 15 14:27:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
-----> memory:
                     total        used        free      shared  buff/cache   available
       Mem:            985         128         194           3         662         700
       Swap:           979          23         956
-----> docker version:
       Client:
        Version:           18.09.6
        API version:       1.39
        Go version:        go1.10.8
        Git commit:        481bc77
        Built:             Sat May  4 02:35:57 2019
        OS/Arch:           linux/amd64
        Experimental:      false

       Server: Docker Engine - Community
        Engine:
         Version:          18.09.6
         API version:      1.39 (minimum version 1.12)
         Go version:       go1.10.8
         Git commit:       481bc77
         Built:            Sat May  4 01:59:36 2019
         OS/Arch:          linux/amd64
         Experimental:     false
-----> docker daemon info:
WARNING: No swap limit support
       Containers: 0
        Running: 0
        Paused: 0
        Stopped: 0
       Images: 2
       Server Version: 18.09.6
       Storage Driver: overlay2
        Backing Filesystem: extfs
        Supports d_type: true
        Native Overlay Diff: true
       Logging Driver: json-file
       Cgroup Driver: cgroupfs
       Plugins:
        Volume: local
        Network: bridge host macvlan null overlay
[...]

PR #3757, opened 2019-10-23
Add provision line to Windows Vagrant box for dokku-installer

pickettd:
Add provision line to Windows Vagrant box definition so dokku-installer service is running/available by default in dokku-windows box

This is to fix issue #3598

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -72,6 +72,7 @@ Vagrant::configure("2") do |config|
     vm.vm.network :private_network, ip: DOKKU_IP
     vm.vm.provision :shell, :inline => "export DEBIAN_FRONTEND=noninteractive && apt-get update >/dev/null && apt-get -qq -y install git dos2unix >/dev/null"
     vm.vm.provision :shell, :inline => "cd /vagrant/ && export DOKKU_BRANCH=`git symbolic-ref -q --short HEAD 2>/dev/null` && export DOKKU_TAG=`git describe --tags --exact-match 2>/dev/null` && cd /root/ && cp /vagrant/bootstrap.sh ./ && dos2unix bootstrap.sh && bash bootstrap.sh"
+    vm.vm.provision :shell, :inline => "cd /root/dokku && make dokku-installer"
   end
 
   config.vm.define "dokku-deb", autostart: false do |vm|


josegonzalez:
The bootstrap script should already properly handle this as the dokku-installer gets installed via the package. Is that not the case?

pickettd:
Not sure exactly. On my dokku-windows vagrant box when I was first trying to debug the issue I was able to run `ps auxf | grep dokku-installer` and see a listing for dokku-installer. However when I navigated to http://dokku.me it just showed the blank nginx welcome page (as @mmahandev mentioned in the attached issue).

After I added the provisioner line from this PR I tried `vagrant destroy dokku-windows` and `vagrant up dokku-windows` from an elevated command prompt and everything appeared to work normally. However at that point yesterday I just halted the VM (`vagrant halt dokku-windows`) without completing the web-ui setup. Now today I brought the machine back up and got an nginx 502 bad gateway when trying to do web-ui setup.

josegonzalez:
The web ui is only available once, and removes itself after shutdown. Thats by design, as we don't want folks to be able to reconfigure an install just by stumbling to the right url. If the vm is halted, then the service running the installer is shut down, and therefore it'll remove itself.

pickettd:
Ok. That makes sense that the web ui removes itself during shutdown.

I just tested again on a fresh VM to make sure I can definitely repo issue #3598 and verified that I still see @mmahandev's same 'Welcome to nginx' by default in the dokku-windows box. And if I add the line from this PR to my Vagrantfile then everything works as described in the documentation. So I guess that means in the dokku-windows box that the bootstrap script is not properly installing dokku-installer.

nemanjapetrovic:
Hi @josegonzalez 

I can also confirm the issue https://github.com/dokku/dokku/issues/3598 on my Windows 10 machine. With this fix everything seems to work just as it should.

josegonzalez:
Just commenting to say that I haven't merged this because I think this is the wrong fix. The bootstrap script should already setup everything for this, and if it doesn't, then that means there is either a bug in the script or an error related to being a windows vm. Without access to a windows machine, it's a bit hard to test this.

josegonzalez:
Going to merge this in. I realize I've been sitting a patch for far too long that others are happy with, and since this doesnt break the target experience, I think this is fine.

PR #3739, opened 2019-10-19
dokku run do not supports interactive mode.

safeforge:
After [!3687](https://github.com/dokku/dokku/pull/3687) I discovered is not possible anymore to use command `dokku run` in the interactive mode. e.g. using `bash`

The error seems to be related with the docker command. The command `docker run` accepts the flags `--interactive` which permits to keep STDIN/STDOUT opened. Instead `--interactive` param in the `docker create` command seems to have a different behaviour, indeed `docker start` accepts `--interactive` param as well.

The PR appends the param `--interactive` to `docker start` command.

```bash
vagrant@dokku:~$ dokku --version
dokku version 0.19.2
vagrant@dokku:~$ dokku apps:create test-app
-----> Creating test-app... done
vagrant@dokku:~$ docker pull alpine:latest
latest: Pulling from library/alpine
9d48c3bd43c5: Pull complete
Digest: sha256:72c42ed48c3a2db31b7dafe17d275b634664a708d901ec9fd57b1529280f01fb
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest
vagrant@dokku:~$ docker tag alpine:latest dokku/test-app:latest
vagrant@dokku:~$ dokku run test-app sh
/ # ls -la
```

```
vagrant@dokku:~$ dokku --version
dokku version 0.18.5
vagrant@dokku:~$ dokku apps:create test-app
-----> Creating test-app... done
vagrant@dokku:~$ docker pull alpine:latest
latest: Pulling from library/alpine
9d48c3bd43c5: Pull complete
Digest: sha256:72c42ed48c3a2db31b7dafe17d275b634664a708d901ec9fd57b1529280f01fb
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest
vagrant@dokku:~$ docker tag alpine:latest dokku/test-app:latest
vagrant@dokku:~$ dokku run test-app sh
/ # ls -la
total 64
drwxr-xr-x    1 root     root          4096 Oct 19 09:53 .
drwxr-xr-x    1 root     root          4096 Oct 19 09:53 ..
-rwxr-xr-x    1 root     root             0 Oct 19 09:53 .dockerenv
drwxr-xr-x    2 root     root          4096 Aug 20 10:30 bin
drwxr-xr-x    5 root     root           360 Oct 19 09:54 dev
drwxr-xr-x    1 root     root          4096 Oct 19 09:53 etc
drwxr-xr-x    2 root     root          4096 Aug 20 10:30 home
[...]

PR #3760, opened 2019-10-26
Fix lost ability to run an interactive shell

benjamin-thomas:
Bug introduced between 0.18.5 and 0.19.5

A simple `dokku run app /bin/bash` would hang otherwise.


benjamin-thomas:
I wasn't sure how to go about testing this.

I tested the changes on my deployed dokku instance directly:

    /var/lib/dokku/plugins/enabled/scheduler-docker-local/scheduler-run

Using `dokku trace:on` highlighted what seems to be the problem.

Please note that I'm not familiar with the `docker container start` sub-command.

benjamin-thomas:
Oops, closing as duplicate of: https://github.com/dokku/dokku/pull/3739

PR #3761, opened 2019-10-28
Fix type-errors in dokku-installer.py

pithyless:
Ran the installer on a clean Ubuntu 18.04.3 LTS node and got a 502 Bad Gateway on the initial setup page.

After debugging, identified there were two problems with the `dokku-installer.py` script (Python 3.6.8).

```
TypeError: cannot use a string pattern on a bytes-like object

TypeError: '>' not supported between instances of 'str' and 'int'
```

The two changes to the script fix these two type-errors (if I understood the intent of the code correctly).


jccr:
I'm just setting up a new dokku installation on a clean Ubuntu 18.04.3 LTS droplet and I'm experiencing this same issue.

PR #3784, opened 2019-11-23
Ensure checks attempts are tracked per-check instead of globally

josegonzalez:
This PR minimizes the duplicate output regarding checks customization, and also fixes a bug where attempts were kept track of globally instead of per-check.

@michaelshobbs do you have time to review this? I would appreciate it if so.

### DIFF
--- a/plugins/scheduler-docker-local/check-deploy
+++ b/plugins/scheduler-docker-local/check-deploy
@@ -147,13 +147,40 @@ scheduler-docker-local-check-deploy() {
     fi
   done
 
-  local ATTEMPT=0
+  exec <"$CHECKS_FILENAME"
+  local CHECK_URL
+  local EXPECTED
+  local FAILEDCHECKS=0
+  while read -r CHECK_URL EXPECTED; do
+    # Ignore empty lines and lines starting with #
+    # shellcheck disable=SC1001
+    [[ -z "$CHECK_URL" || "$CHECK_URL" =~ ^\# ]] && continue
+    # Ignore if it's not a URL in a supported format
+    # shellcheck disable=SC1001
+    ! [[ "$CHECK_URL" =~ ^(http(s)?:)?\/.* ]] && continue
 
-  until [[ $SUCCESS == 1 || $ATTEMPT -ge $ATTEMPTS ]]; do
-    local FAILEDCHECKS=0
-    local ATTEMPT=$((ATTEMPT + 1))
-    dokku_log_verbose "Attempt $ATTEMPT/$ATTEMPTS. Waiting for $WAIT seconds ..."
-    sleep "$WAIT"
+    if [[ "$CHECK_URL" =~ ^https?: ]]; then
+      local URL_PROTOCOL=${CHECK_URL%:*}
+      local CHECK_URL=${CHECK_URL#*:}
+    else
+      local URL_PROTOCOL="http"
+    fi
+
+    if [[ "$CHECK_URL" =~ ^//.+ ]]; then
+      # To test a URL with specific host name, we still make request to localhost,
+      # but we set Host header to $SEND_HOST.
+      #
+      # The pattern is
+      #   //SEND_HOST/PATHNAME
+      local UNPREFIXED=${CHECK_URL#//}
+      local URL_HOSTNAME=${UNPREFIXED%%/*}
+      local URL_PATHNAME=${UNPREFIXED#$URL_HOSTNAME}
+
+      local HEADERS="-H Host:$URL_HOSTNAME"
+    else
+      local URL_HOSTNAME=localhost
+      local URL_PATHNAME=$CHECK_URL
+    fi
 
     # -q           Do not use .curlrc (must come first)
     # --compressed Test compression handled correctly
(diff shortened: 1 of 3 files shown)


plugins/scheduler-docker-local/check-deploy
```diff
@@ -99,20 +98,13 @@ scheduler-docker-local-check-deploy() {
...
-    else
-      dokku_log_verbose "Non web container detected: Running default container check..."
-    fi
-
     rm -rf "$CHECK_DEPLOY_TMP_WORK_DIR" &>/dev/null || true
 
     # simple default check to see if the container stuck around
     local DOKKU_DEFAULT_CHECKS_WAIT="${DOKKU_DEFAULT_CHECKS_WAIT:-10}"
-    dokku_log_info1 "Waiting for $DOKKU_DEFAULT_CHECKS_WAIT seconds ..."
+    dokku_log_verbose "Waiting for $DOKKU_DEFAULT_CHECKS_WAIT seconds ..."
```
michaelshobbs: i think this particular line should stay info1 in order to communicate to the user that we're intentionally not proceeding yet

plugins/scheduler-docker-local/check-deploy
```diff
@@ -176,48 +195,20 @@ scheduler-docker-local-check-deploy() {
...
-      # shellcheck disable=SC1001
-      ! [[ "$CHECK_URL" =~ ^(http(s)?:)?\/.* ]] && continue
-
-      if [[ "$CHECK_URL" =~ ^https?: ]]; then
-        local URL_PROTOCOL=${CHECK_URL%:*}
-        local CHECK_URL=${CHECK_URL#*:}
-      else
-        local URL_PROTOCOL="http"
-      fi
+    # This URL will show up in the messages
```
michaelshobbs: perhaps this reads more clearly
```
LOG_URL is used in verbose output below
```

plugins/scheduler-docker-local/check-deploy
```diff
@@ -176,48 +195,20 @@ scheduler-docker-local-check-deploy() {
...
+    local LOG_URL="$URL_PROTOCOL://$URL_HOSTNAME$URL_PATHNAME"
+    # And how we formulate the CURL request
```
michaelshobbs:
```
CURL_ARGS includes the behinds the scenes options for hitting the container service directly
```

plugins/scheduler-docker-local/check-deploy
```diff
@@ -176,48 +195,20 @@ scheduler-docker-local-check-deploy() {
...
+    dokku_log_verbose "$LOG_URL => \"$EXPECTED\""
+    [[ $DOKKU_TRACE ]] && dokku_log_verbose "$ curl $CURL_ARGS"
 
-      dokku_log_verbose "CHECKS expected result:"
-      dokku_log_verbose "$LOG_URL => \"$EXPECTED\""
-      [[ $DOKKU_TRACE ]] && dokku_log_verbose "$ curl $CURL_ARGS"
+    local ATTEMPT=0
+    until [[ $SUCCESS == 1 || $ATTEMPT -ge $ATTEMPTS ]]; do
+      local ATTEMPT=$((ATTEMPT + 1))
+      dokku_log_verbose "Attempt $ATTEMPT/$ATTEMPTS. Waiting for $WAIT seconds ..."
```
michaelshobbs: again i think it's important to communicate when we are intentionally not proceeding

plugins/scheduler-docker-local/scheduler-deploy
```diff
@@ -130,7 +130,7 @@ scheduler-docker-local-scheduler-deploy() {
       # run checks first, then post-deploy hooks, which switches proxy traffic
       trap 'kill_new $cid $PROC_TYPE $CONTAINER_INDEX' INT TERM EXIT
       if [[ "$(is_app_proctype_checks_disabled "$APP" "$PROC_TYPE")" == "false" ]]; then
-        dokku_log_info1 "Attempting pre-flight checks"
+        dokku_log_info1 "Attempting pre-flight checks ($PROC_TYPE.$CONTAINER_INDEX)"
```
josegonzalez: The reason I moved from `log1` to `verbose` was because of the above log line. I wanted to signify that the sections below were all related, versus having them be completely separate.

josegonzalez: @michaelshobbs ^

michaelshobbs: Taking a step back it's not clear even why these logging changes are in a bug fix PR. That minor point aside, it's still not clear to me what problem hiding those 2 specific log messages behind the verbose flag actually solves. Clarify that would probably help.

In absence of further clarity, however, those pauses above default to 10 seconds and are user configurable to be as high as one may desire. As part of providing a solid user experience we should be clear about any time we've paused processing of a command. Communicating what a program is doing in the background is a key principle of good UX. If there is, however, some question of the value of these log messages, perhaps we can modify them to provide more value?

josegonzalez: You're right, I can definitely separate out the two changes.

My original intention here was to reduce the amount of duplicative output. If a user is scaling a certain process up, then they'll get the same message about the CHECKS file over and over. Rather than do that, I wanted to only alert once so as to not to distract from the work being performed.

As I was refactoring, I also wanted to reduce the number of messages around attempts - or at least make them more clear. I noticed that we had a bug in the order in which attempts were performed, hence why that change tagged along in this PR.

Does that make sense?

michaelshobbs: Got it. I wasn’t putting together the execution flow. Thank you for clarifying

michaelshobbs:
Just a couple of suggestions on comment language. Otherwise looks good to me.

josegonzalez:
Do you want me to separate out the changes?

michaelshobbs:
I don’t think that’s necessary now. I just wasn’t clear on why they were grouped together.

Issue #3821, opened 2020-01-18
apps:rename doesn't copy buildpacks and app doesn't deploy (after adding manually)


turicas:
## Description of problem

I had an app running and since I've renamed it, it doesn't work anymore. Checking the logs I found some binaries (which were installed on the buildpacks) were missing and then the `dokku buildpacks <new-name>` listed no buildpacks. After adding the needed buildpacks I can't make the app run.

### Steps to Reproduce

1. Create an app
2. Deploy the app
3. Run `dokku apps:rename <old-name> <new-name>`
4. Run `dokku ps:restart <new-name>`

#### Actual Results

- Buildpacks won't be added automatically to the new app
- App won't deploy (after some attempts of running `ps:resrtart it finally worked)

#### Expected Results

- New app will run as the old one (it's just a rename, afterall)

### `dokku report APP_NAME` output

```
-----> uname: Linux fazzer 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u1 (2019-09-20) x86_64 GNU/Linux
-----> memory: 
                     total        used        free      shared  buff/cache   available
       Mem:           7877        1311         118         164        6447        6113
       Swap:         15998         611       15387
-----> docker version: 
       Client: Docker Engine - Community
        Version:           19.03.5
        API version:       1.40
        Go version:        go1.12.12
        Git commit:        633a0ea838
        Built:             Wed Nov 13 07:25:38 2019
        OS/Arch:           linux/amd64
        Experimental:      false
       
       Server: Docker Engine - Community
        Engine:
         Version:          19.03.5
         API version:      1.40 (minimum version 1.12)
         Go version:       go1.12.12
         Git commit:       633a0ea838
         Built:            Wed Nov 13 07:24:09 2019
         OS/Arch:          linux/amd64
         Experimental:     false
        containerd:
         Version:          1.2.10
         GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
        runc:
         Version:          1.0.0-rc8+dev
         GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
        docker-init:
         Version:          0.18.0
         GitCommit:        fec3683
-----> docker daemon info: 
       Client:
        Debug Mode: true
       
       Server:
        Containers: 23
         Running: 4
         Paused: 0
         Stopped: 19
        Images: 48
        Server Version: 19.03.5
        Storage Driver: overlay2
         Backing Filesystem: extfs
[...]

PR #3835, opened 2020-01-30
Added piece of additional information

mshafiq235:
When renaming the app, buildpacks are not preserved. Added this into documentation so that the person who is renaming app should know why the app might be failing to deploy.


josegonzalez:
We should just fix the bug?

josegonzalez:
Closing this in favor of fixing the bug as mentioned here: https://github.com/dokku/dokku/issues/3821#issuecomment-581343142

PR #3838, opened 2020-01-31
Expand docs surrounding access control

ltalirz:
[ci skip]

The deployment documentation on scoping user access pointed only to the
development documentation for implementing plugins to handle user
scoping.
Added link to the dokku-acl plugin, which already handles user scoping
out of the box.

### DIFF
--- a/docs/deployment/user-management.md
+++ b/docs/deployment/user-management.md
@@ -72,7 +72,9 @@ dokku ssh-keys:remove KEY_NAME
 
 ## Scoping commands to specific users
 
-Keys are given unique names, which can be used in conjunction with the [user-auth](/docs/development/plugin-triggers.md#user-auth) plugin trigger to handle command authorization. Please see the documentation on that trigger for more information.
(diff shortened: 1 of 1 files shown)


josegonzalez:
I don't want to do this. It's not an official plugin, so it should only be pointed out on our plugin index page.

ltalirz:
Up to you, of course.
It's just that this place in the documentation is where I ended up looking for how to do user management, and the link provided was not very helpful - that's why I needed to ask in the slack channel.

ltalirz:
One could suggest to have a look at the plugin page

josegonzalez:
Might be better to point to the plugin page. Do we have other auth plugins around?

ltalirz:
There's http auth, but in terms of management for docker users I don't see another one in the list

ltalirz:
I've now removed the explicit mention of dokku-acl and replaced it by a link to the list of contributed plugins.
I find it a bit less helpful than mentioning dokku-acl directly, but still better than before.

PR #3865, opened 2020-02-18
Cleanup injected docker labels

josegonzalez:


josegonzalez:
Tests fail due to an unrelated issue with grepping defunct processes inside of containers.

PR #3888, opened 2020-03-07
Mark 9+ for Debian version

decentral1se:
Debian 10 works and not seeing the "+" might scare off some users who think there is missing support.


tyt2y3:
Its even better if marked explicity supporting 'Debian 9/10' in my opinion. I was scared for a moment too. Anyways thanks!

josegonzalez:
@tyt2y3 pull requests welcome.

PR #3893, opened 2020-03-11
Rewrite apps plugin in golang

josegonzalez:


josegonzalez:
Merging with the failing test as that is unrelated and is a flakey test anyhow.

PR #3914, opened 2020-04-01
Add ability to change or disable the access/error log path

### LINKED ISSUE #3911: Feature: Add option to disable webserver logs
## Description of the feature

As an app maintainer I would like to disable/enable webserver logs with one setting to 
- increase performance, maybe even for just a short period of time (eg. high traffic app, writing logs which are not needed increases execution time with no benefit)
- support privacy policies / GDPR (eg. dont collect any personal data - access logs however contain an IP address and in case of GET requests maybe other personal data)

### How to solve

Add an option (like the [bind-address-ipv4](http://dokku.viewdocs.io/dokku/configuration/nginx/#binding-to-specific-addresses) or an environment variable) to toggle `access_log` and `error_log` in [nginx.conf](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/templates/nginx.conf.sigil) from file-path to `/dev/null/`/`off`.

josegonzalez:
Closes #3911

### DIFF
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -89,7 +89,7 @@ The path to where log files are stored can be changed by calling the `nginx:set`
 - `access-log-path` (type: string, default: `${NGINX_LOG_ROOT}/${APP}-access.log`): Log path for nginx access logs
 - `error-log-path` (type: string, default: `${NGINX_LOG_ROOT}/${APP}-error.log`): Log path for nginx error logs
 
-The defaults should not be changed without verifying that the paths will be writeable by nginx. However, this setting is useful for enabling or disabling logging by setting the values to `off`
+The defaults should not be changed without verifying that the paths will be writeable by nginx. However, this setting is useful for enabling or disabling logging by setting the values to `off`.
 
 ```shell
 dokku nginx:set node-js-app access-log-path off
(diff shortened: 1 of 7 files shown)


docs/configuration/nginx.md
```diff
@@ -82,6 +82,29 @@ You may also follow the logs by specifying the `-t` flag.
 dokku nginx:error-logs node-js-app -t
 ```
 
+## Changing log path
+
+The path to where log files are stored can be changed by calling the `nginx:set` command with the following options:
+
+- `access-log-path` (type: string, default: `${NGINX_LOG_ROOT}/${APP}-error.log`): Log path for nginx access logs
```
decentral1se, line 89:
```suggestion
- `access-log-path` (type: string, default: `${NGINX_LOG_ROOT}/${APP}-access.log`): Log path for nginx access logs
```

docs/configuration/nginx.md
```diff
@@ -82,6 +82,29 @@ You may also follow the logs by specifying the `-t` flag.
...
+- `error-log-path` (type: string, default: `${NGINX_LOG_ROOT}/${APP}-error.log`): Log path for nginx error logs
+
+The defaults should not be changed without verifying that the paths will be writeable by nginx. However, this setting is useful for enabling or disabling logging by setting the values to `off`
```
decentral1se, line 92:
```suggestion
The defaults should not be changed without verifying that the paths will be writeable by nginx. However, this setting is useful for enabling or disabling logging by setting the values to `off`.
```

PR #3958, opened 2020-05-01
feat: allow remotes named other than dokku for CLI

LukasSkywalker:
This change allows CLI users to use remotes named other than `dokku`. Users can prefix the dokku commands with the remote name to run a command on a specified remote.

This allows users to run commands on multiple environments or servers from the same git repo.

Example:

Given that the user has the following git remotes:
```bash
origin git@github.com:dokku/dokku.git
test dokku@test.example.com:myapp
production dokku@example.com:myapp
```
It is possible to run the commands on the specified remote like this:

```bash
dokku test ps:rebuild
dokku production logs -t
```

The change is backwards compatible with the existing CLI.

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -61,10 +61,29 @@ fn-dokku-host() {
   echo "$DOKKU_HOST"
 }
 
+fn-dokku-remotes() {
+  git remote -v 2>/dev/null | grep -Ei "dokku@" | cut -f1 | uniq
+}
+
+fn-alternative-dokku-remote() {
+  declare remote_name="$1"
+
+  dokku_remotes=$(fn-dokku-remotes)
+  match=$(echo "$dokku_remotes" | grep -Ei "^${remote_name}$")
+
+  echo "$match"
+}
+
 main() {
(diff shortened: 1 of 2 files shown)


josegonzalez:
This is already possible via the `--remote` flag, as implemented [here](https://github.com/dokku/dokku/blob/9827dd65bb9c09fcb693143c3726509737a49785/contrib/dokku_client.sh#L99-L100). I'd rather not re-implement the wheel without a good reason.

josegonzalez:
@LukasSkywalker any reason why the pre-existing flag doesn't work for your use case?

LukasSkywalker:
No, I just missed the flag. The method implemented by the PR would require less typing though. But I agree that it's the same feature duplicated.

josegonzalez:
While it is more typing, I think not duplicating the existing flag is the better move. It also ensures that we don't accidentally trample over a command name.

I would create aliases for the commands or potentially shell wrappers you distribute in your codebase's bin directory.

LukasSkywalker:
Yup, that‘s fair. Thanks for creating Dokku :)

PR #3967, opened 2020-05-06
Add Ubuntu 20.04 support

### LINKED ISSUE #3961: Ubuntu 20.04 LTS support
Ubuntu 20.04 Focal Fossa is out and is the new LTS release, supported until 2025. Having dokku support it would be neat, as most cloud providers will start providing the image very soon.

Thanks

josegonzalez:
Will still need package releases, but everything should be ready for 0.21.0.

Closes #3961

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -3,7 +3,7 @@ set -eo pipefail
 [[ $TRACE ]] && set -x
 
 # A script to bootstrap dokku.
-# It expects to be run on Ubuntu 16.04/18.04, or CentOS 7 via 'sudo'
+# It expects to be run on Ubuntu 16.04/18.04/20.04, or CentOS 7 via 'sudo'
 # If installing a tag higher than 0.3.13, it may install dokku via a package (so long as the package is higher than 0.3.13)
 # It checks out the dokku source code from Github into ~/dokku and then runs 'make install' from dokku source.
 
@@ -12,7 +12,7 @@ set -eo pipefail
 # That's good because it prevents our output overlapping with wget's.
 # It also means that we can't run a partially downloaded script.
 
-SUPPORTED_VERSIONS="Debian [9, 10], CentOS [7], Ubuntu [16.04, 18.04]"
+SUPPORTED_VERSIONS="Debian [9, 10], CentOS [7], Ubuntu [16.04, 18.04, 20.04]"
 
 log-fail() {
   declare desc="log fail formatter"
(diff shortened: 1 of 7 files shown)


boxed:
I was bitten by this yesterday.

Might I suggest that the install script is changed to warn instead of erroring out? Looking at this change, there isn't actually any code change required, so it would be nice that when a new ubuntu version comes out (which probably won't require changes), the install script will then say "dokku isn't tested on this version of ubuntu, do you wish to continue anyway?" for example.

boxed:
How can I test this? I tried just changing bootstrap.sh to what this PR suggest, but I get an error message because of this code:

```
  if ! in-array "$DOKKU_DISTRO_VERSION" "16.04" "18.04" "9" "10"; then
    log-fail "Unsupported Linux distribution. Only the following versions are supported: $SUPPORTED_VERSIONS"
  fi
```
I tried commenting that out and then I get the error

```
E: The repository 'https://download.docker.com/linux/ubuntu focal Release' does not have a Release file.
```

Which seems weird becuase `https://download.docker.com/linux/ubuntu focal Release` doesn't look like a valid URL.

dstaley:
@josegonzalez While we're waiting for the v0.21.0 release, what's the suggested way to get up and running with dokku on Ubuntu 20.04? Should we compile from source, or would installing the Bionic packages be preferred?

josegonzalez:
Bionic package works fine, so manually download that package for now.

I publish all the same packages for each OS, nothing special about any of them.

Actually I'll just copy the 0.20.4 package to Focal now.

josegonzalez:
Okay 0.20.4 has been published. I can merge this now.

dstaley:
So would installation be the following?

```bash
wget https://raw.githubusercontent.com/dokku/dokku/v0.20.4/bootstrap.sh
sudo DOKKU_TAG=v0.20.4 bash bootstrap.sh
```

josegonzalez:
Yeah that should work now.

dstaley:
Gotcha! I'll report back if there are any issues!

dstaley:
Just for anyone who stumbles on this: as of writing (May 15 2020) Docker has not published a version officially supporting Ubuntu 20.04. In order to install it, you'll need to run the following:

```
wget https://raw.githubusercontent.com/dokku/dokku/master/bootstrap.sh
sudo DOKKU_TAG=v0.20.4 bash bootstrap.sh
```

The second command _should_ fail with `E: Package 'docker-ce' has no installation candidate`. If that's the case, run the following:

```
sudo nano /etc/apt/sources.list.d/docker.list
```

Replace `focal` with `eoan`. Then run:

```
apt-get update
apt-get install -y --no-install-recommends docker-ce
```

After that you should be able to fully install Dokku with:

```
sudo DOKKU_TAG=v0.20.4 bash bootstrap.sh
```

PR #4007, opened 2020-05-27
Fix documentation about custom nginx config when using Dockerfile deployment

turicas:
I needed a custom nginx config file for an app and was deploying using `Dockerfile`, but it was not clear in documentation that I need also to set a `WORKDIR` in this case. After [investigating dokku's source code](https://github.com/dokku/dokku/blob/master/plugins/common/functions#L409) I find out it's not needed to set `WORKDIR` when deploying an app using `Dockerfile`, but it's needed when we have a custom nginx config since this feature uses `copy_from_image` function (actually, all features using this function require setting `WORKDIR`).

[ci skip]

### DIFF
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -155,8 +155,9 @@ Dokku uses a templating library by the name of [sigil](https://github.com/glider
 
 - Copy the following example template to a file named `nginx.conf.sigil` and either:
   - If using a buildpack application, you __must__ check it into the root of your app repo.
-  - If using Dockerfile deployment, set `WORKDIR` and `ADD` the file to this
-    directory (example: `WORKDIR /app` and `ADD nginx.conf.sigil /app`).
+  - If using a dockerfile or docker image for deploys, either:
+    - If WORKDIR is specified, add the file to the `WORKDIR` specified in the last Dockerfile stage  (example: `WORKDIR /app` and `ADD nginx.conf.sigil /app`).
+    - If no WORKDIR is specified, add the file to the root (`/`) of the docker image  (example:ADD nginx.conf.sigil /`).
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
This isn't true. If you don't specify a `WORKDIR`, the container path will be relative to the root of the container, as documented [here](https://docs.docker.com/engine/reference/commandline/cp/).

I've clarified your PR further to describe what should/should not happen, which hopefully makes up for my lack of response here (wanted to investigate further before merging and only just had time).

Thanks for the PR!

turicas:
Thanks for the clarification! :)

docs/configuration/nginx.md
```diff
@@ -155,8 +155,9 @@ Dokku uses a templating library by the name of [sigil](https://github.com/glider
 
 - Copy the following example template to a file named `nginx.conf.sigil` and either:
   - If using a buildpack application, you __must__ check it into the root of your app repo.
-  - If using Dockerfile deployment, set `WORKDIR` and `ADD` the file to this
-    directory (example: `WORKDIR /app` and `ADD nginx.conf.sigil /app`).
+  - If using a dockerfile or docker image for deploys, either:
+    - If WORKDIR is specified, add the file to the `WORKDIR` specified in the last Dockerfile stage  (example: `WORKDIR /app` and `ADD nginx.conf.sigil /app`).
+    - If no WORKDIR is specified, add the file to the root (`/`) of the docker image  (example:ADD nginx.conf.sigil /`).
```
turicas, line 160:
```suggestion
    - If no WORKDIR is specified, add the file to the root (`/`) of the docker image  (example: `ADD nginx.conf.sigil /`).
```

josegonzalez: Mind making a new pr for this change? I've already merged this PR...

turicas: https://github.com/dokku/dokku/pull/4100

PR #4041, opened 2020-07-01
feat: Add download option to the certs plugin

rvanlaar:
The 'server.key' and 'server.crt' can downloaded.
The command will print the certificates to stdin.

Is this something dokku is interested in to get mainlined?

This helped my use case where I spin up a new dokku when I'm testing and developing my app
and remove it when I'm done. This made me hit a domain request limit on let's encrypt.

With the changes in this PR I can download the certificates, so that I can upload them
with the certs plugin when I spin up a new dokku vm.

### DIFF
--- a/docs/configuration/ssl.md
+++ b/docs/configuration/ssl.md
@@ -9,6 +9,7 @@ certs:add <app> CRT KEY                  # Add an ssl endpoint to an app. Can al
 certs:generate <app> DOMAIN              # Generate a key and certificate signing request (and self-signed certificate)
 certs:remove <app>                       # Remove an SSL Endpoint from an app.
 certs:report [<app>] [<flag>]            # Displays an ssl report for one or more apps
+certs:show <app> <crt|key>               # Show the server.crt or server.key on stdout
 certs:update <app> CRT KEY               # Update an SSL Endpoint on an app. Can also import from a tarball on stdin
 ```
 
@@ -55,6 +56,16 @@ If you decide to obtain a CA signed certificate, you can import that certificate
 ### Certificate removal
 
 The `certs:remove` command only works on app-specific certificates. It will `rm` the app-specific tls directory, rebuild the nginx configuration, and reload nginx.
+
+### Showing the certificate
+
+The `certs:show` command can be used to show your configured certs for an app. The show command can be used for example to export Let's Encrypt certificates
+after they've been generated. You can export it as follows:
+
+```shell
+dokku certs:show node-js-app crt > server.crt
+dokku certs:show node-js-app key > server.key
+```
 
 ### Displaying certificate reports for an app
 
--- a/plugins/certs/help-functions
+++ b/plugins/certs/help-functions
@@ -34,6 +34,7 @@ fn-help-content() {
     certs:remove <app>, Remove an SSL Endpoint from an app
     certs:report [<app>] [<flag>], Displays an ssl report for one or more apps
     certs:rollback <app>, [NOT IMPLEMENTED] Rollback an SSL Endpoint for an app
+    certs:show <app> <crt|key>, Show the server.crt or server.key on stdout
     certs:update <app> CRT KEY, Update an SSL Endpoint on an app. Can also import from a tarball on stdin
 help_content
 }
--- a/plugins/certs/subcommands/show
+++ b/plugins/certs/subcommands/show
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -eo pipefail
+[[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/certs/functions"
+
+cmd-certs-show() {
+  declare desc="Show the server.crt or server.key on stdout"
+  declare cmd="certs:show"
+  [[ "$1" == "$cmd" ]] && shift 1
+  declare APP="$1" KEY_TYPE="$2"
+
+  [[ -z "$APP" ]] && dokku_log_fail "Please specify an app to run the command on"
+  verify_app_name "$APP"
+  local APP_SSL_PATH="$DOKKU_ROOT/$APP/tls"
+
+  if [[ "$KEY_TYPE" != "key" ]] && [[ "$KEY_TYPE" != "crt" ]]; then
+    dokku_log_fail "specify either 'key' or 'crt'"
+  fi
+
+  if ! is_ssl_enabled "$APP"; then
+    dokku_log_fail "$APP doesn't have an SSL endpoint defined"
+  fi
+
+  cat "$APP_SSL_PATH/server.$KEY_TYPE"
+}
+
+cmd-certs-show "$@"
--- a/tests/unit/certs.bats
+++ b/tests/unit/certs.bats
@@ -75,3 +75,29 @@ teardown() {
   echo "status: $status"
   assert_success
 }
+
+@test "(certs) certs:show" {
+  run /bin/bash -c "dokku certs:show fake-app-name 2>&1"
+  echo "output: $output"
+  echo "status: $status"
+  assert_output_contains "App fake-app-name does not exist"
+  assert_failure
+
+  run /bin/bash -c "dokku certs:show $TEST_APP fake-var-name"
+  echo "output: $output"
+  echo "status: $status"
+  assert_output_contains "specify either 'key' or 'crt'"
+  assert_failure
+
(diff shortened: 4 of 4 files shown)


plugins/certs/help-functions
```diff
@@ -29,6 +29,7 @@ fn-help-content() {
   cat <<help_content
     certs:add <app> CRT KEY, Add an ssl endpoint to an app. Can also import from a tarball on stdin
     certs:chain CRT [CRT ...], [NOT IMPLEMENTED] Print the ordered and complete chain for the given certificate
+    certs:download <app> key|crt, Print the server.key or server.crt on stdout
```
josegonzalez, line 32: Should be `[key|crt]`.

plugins/certs/subcommands/download
```diff
@@ -0,0 +1,28 @@
...
+  declare desc="exports SSL cert or key via STDOUT"
+  declare cmd="$1"
+  [[ "$1" == "$cmd" ]] && shift 1
+  declare APP="$1" KEY_TYPE="$2"
+
+  [[ -z "$APP" ]] && dokku_log_fail "Please specify an app to run the command on"
+  verify_app_name "$APP"
+  local APP_SSL_PATH="$DOKKU_ROOT/$APP/tls"
+
+  if [[ -f "$APP_SSL_PATH/server.key" ]] && [[ -f "$APP_SSL_PATH/server.crt" ]]; then
```
josegonzalez, line 17: Can we flatten this `if/else/fi` block? So exit early if the app doesn't have ssl.

Also, is there a helper function for checking for ssl support for an app?

plugins/certs/subcommands/download
```diff
@@ -0,0 +1,28 @@
...
+    if [[ "$KEY_TYPE" == "key" ]]  || [[ "$KEY_TYPE" == "crt" ]]; then
```
josegonzalez, line 18: Probably worth validating this above before even checking for ssl endpoint.

josegonzalez:
Also, should we call this `certs:show` instead? Download implies you will... well, download the file :D

plugins/certs/subcommands/download
rvanlaar: Updated.

plugins/certs/subcommands/download
rvanlaar: Done

plugins/certs/subcommands/print
```diff
@@ -0,0 +1,28 @@
...
+  fi
+
+  if ! is_ssl_enabled "$APP"; then
+    dokku_log_fail "$APP doesn't have an SSL endpoint defined"
+  fi
+
+  cat "$APP_SSL_PATH/server.$KEY_TYPE"
+}
+
+cmd-certs-download "$@"
```
josegonzalez, line 28: Can you add a newline at the end of the file?

plugins/certs/subcommands/print
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -eo pipefail
+[[ $DOKKU_TRACE ]] && set -x
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_AVAILABLE_PATH/certs/functions"
+
+cmd-certs-download() {
+  declare desc="exports SSL cert or key via STDOUT"
+  declare cmd="$1"
```
josegonzalez, line 9: I think this should be `certs:print`, right?

plugins/certs/subcommands/print
josegonzalez: Also, should rename the function to `cmd-certs-print` if you named the command that. I still think `show` is a better command, and is more inline with what we've been naming things elsewhere (config:show, nginx:show-config...).

rvanlaar: It's awesome how quick you are with feedback. I've changed it to show.

plugins/certs/subcommands/print
rvanlaar: It is now 'show SSL cert....' Is that what you meant?

plugins/certs/help-functions
rvanlaar: My understanding is that a `[]` block meant optional in documentation. I've changed it though.

plugins/certs/subcommands/print
josegonzalez: I just happen to be awake :P

plugins/certs/subcommands/print
josegonzalez: Well the cmd="$1" should be "certs:show".

Also I guess the description is out of date.

plugins/certs/help-functions
```diff
@@ -29,6 +29,7 @@ fn-help-content() {
   cat <<help_content
     certs:add <app> CRT KEY, Add an ssl endpoint to an app. Can also import from a tarball on stdin
     certs:chain CRT [CRT ...], [NOT IMPLEMENTED] Print the ordered and complete chain for the given certificate
+    certs:show <app> [key|crt], Show the server.key or server.crt on stdout
```
josegonzalez, line 32: Oh right, yeah this should be `<key|crt>`. My bad.

josegonzalez:
I should start drinking coffee...

plugins/certs/help-functions
rvanlaar: Fixed.

plugins/certs/subcommands/print
rvanlaar: The cmd="$1" was taken verbatim from: https://github.com/dokku/dokku/blob/master/plugins/certs/subcommands/add#L33

rvanlaar: fixed

josegonzalez: That one is weird due to symlinks, this one is better: https://github.com/dokku/dokku/blob/master/plugins/certs/subcommands/remove#L8

plugins/certs/help-functions
```diff
@@ -34,6 +34,7 @@ fn-help-content() {
     certs:remove <app>, Remove an SSL Endpoint from an app
     certs:report [<app>] [<flag>], Displays an ssl report for one or more apps
     certs:rollback <app>, [NOT IMPLEMENTED] Rollback an SSL Endpoint for an app
+    certs:show <app> <crt|key>, Show the server.crt or server.key on stdout
```
josegonzalez, line 37: This description should be the same as the following two:

- https://github.com/dokku/dokku/pull/4041/files#diff-3cfb1097cf8847f24c036fe8f8128a8eR12
- https://github.com/dokku/dokku/pull/4041/files#diff-5d6d6d7dafc57ca6c4c764ec0861c169R8

docs/configuration/ssl.md
```diff
@@ -56,6 +57,16 @@ If you decide to obtain a CA signed certificate, you can import that certificate
 
 The `certs:remove` command only works on app-specific certificates. It will `rm` the app-specific tls directory, rebuild the nginx configuration, and reload nginx.
 
+### Certificate export
```
josegonzalez, line 60: "Showing the certificate" would be a better title :)

josegonzalez: In addition, should we add a warning about this function as it exposes the internal certificate to anyone with access to the command?

rvanlaar: I would be more worried about database access and that I can't pin users to apps. Not that they can download the SSL cert. 

A rogue user that wants the private key could always upload a custom nginx sigil
that exposes the certificate directory.

plugins/certs/help-functions
rvanlaar: Done

docs/configuration/ssl.md
josegonzalez: Seems legit.

josegonzalez:
Last thing, would it be possible for you to add a test? That way when I'm porting this plugin to golang, I can verify that I didn't break any intended functionality.

rvanlaar:
> Last thing, would it be possible for you to add a test? That way when I'm porting this plugin to golang, I can verify that I didn't break any intended functionality.

Is this the file that needs a few checks added:
https://github.com/dokku/dokku/blob/master/tests/unit/certs.bat

With these:
- show nonsense: check error message
- show crt: check if cert is in output
- show key: check if key is in output

I didn't find:
1) tests that check how the program handles output when there's no $APP specified and
2) tests for certs in the tests/ci directory
Is that correct?

josegonzalez:
- Example test for invalid app name: https://github.com/dokku/dokku/blob/master/tests/unit/report.bats#L36
- Test that uses cert from test directory: https://github.com/dokku/dokku/blob/master/tests/unit/certs.bats#L44

rvanlaar:
Those were great examples. 

I've added tests for showing:
- show crt
- show key
- show with nonexisting app
- show where key/crt is misspelled.

josegonzalez:
Merging - test failures were from something else.

PR #4043, opened 2020-07-01
Allow controlling nginx proxy-read-timeout

### LINKED ISSUE #4042: Set `proxy_read_timeout` with nginx:set
## Description of problem

I have this websocket server which only purpose is to talk with another server. It should be a persistent connection but `proxy_read_timeout` resets it every 60 seconds (as it should by this docs: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout). It would be nice to be able to define the timeout using nginx:set.

josegonzalez:
Closes #4042

### DIFF
--- a/docs/configuration/nginx.md
+++ b/docs/configuration/nginx.md
@@ -106,6 +106,26 @@ dokku nginx:set node-js-app error-log-path
 
 In all cases, the nginx config must be regenerated after setting the above values.
 
+## Specifying a read timeout
+
+> New as of 0.21.0
+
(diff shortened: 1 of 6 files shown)


Yihao-G:
Looks like this will only set `proxy-read-timeout` for HTTP, not HTTPS schemes?

josegonzalez:
Damn you're right. It's late here, mind filing an issue so I can take a look and fix this for the next patch release? You are also free to make the pull request on your own, and I'll review/merge/release tomorrow :)

PR #4045, opened 2020-07-02
Filter gpus instead of nvidia-gpus from resource arguments

josegonzalez:
We transform the argument here: https://github.com/dokku/dokku/blob/master/plugins/resource/triggers.go#L58


josegonzalez:
Merging - tests are all flakey due to OOMkills, so this is good to go :)

PR #4046, opened 2020-07-03
Make `make test` pass on linting

rvanlaar:
Fixes linting problems that blocked `make test` inside the vagrant vm.


josegonzalez:
Seems legit! Thanks for the PR!

PR #4057, opened 2020-07-13
Install sudo when installing from source

hugopeixoto:
When installing from source on a brand new debian system, I got an error running `help2man`. It was calling `dokku --help`, which needed `sudo` to change to the `dokku` user. `sudo` was not installed by default, so I added it to the dependencies list in `bootstrap.sh`.


hugopeixoto:
This is the test that failed:

> not ok 4 (ps:scale) dockerfile with procfile in 51sec

I don't think this is related to the code I changed. Could this be a flaky test?

josegonzalez:
Flaky test - blech - but not worth waiting on a merge.

PR #4063, opened 2020-07-14
Clarify why we stop/rebuild apps during upgrade

josegonzalez:


docs/getting-started/upgrading.md
```diff
@@ -29,11 +28,18 @@ Before upgrading, check the migration guides to get comfortable with new feature
...
+If you'll be updating docker or the herokuish package simultaneously, it's recommended
+that you stop all applications before upgrading and rebuild afterwards. This is not
+required if the upgrade only impacts the `dokku` package.
+
+Why do we recommend stopping all apps?
+
+- `docker`: Containers may be randomly reset during the upgrade process, resulting in
+  requests being sent to the wrong containers. Acknowledging and scheduling downtime
+  thus becomes much more important.
+- `herokuish`: While not required, it may be useful to take advantage of the
```
chemicalkosek, line 43: - `herokuish`: While not required, it may be useful to take advantage of the

Missing text here?

chemicalkosek: I'm guessing from the next paragraphs that it was supposed to be 'of any new buildpacks that were released'

josegonzalez: Fixed, thanks for the callout.

PR #4072, opened 2020-07-17
Fix deprecation warning in config_all

Schlepptop:


josegonzalez:
Can you comment on what the warning was?

Schlepptop:
This one: https://github.com/dokku/dokku/blob/bcfa53005629f913e4e9a25fe5c91d61408bc927/plugins/config/src/commands/commands.go#L41

PR #4078, opened 2020-07-18
docs: add more details to `nginx-dokku-template-source` trigger

gurpreetatwal:
Feel free  to make any edits as you see fit! :D

### DIFF
--- a/docs/development/plugin-triggers.md
+++ b/docs/development/plugin-triggers.md
@@ -856,9 +856,9 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 ### `nginx-dokku-template-source`
 
-- Description: Customize the source location of the `dokku.conf` nginx configuration file
+- Description: Return the path to a `sigil` template that should be used to generate the `dokku.conf` nginx configuration file.
 - Invoked by: `nginx-vhosts#install`
-- Arguments:
+- Arguments: None, however the `sigil` template can make use of the following variables: `$.DOKKU_ROOT $.NGINX_ROOT`
 - Example:
 
 ```shell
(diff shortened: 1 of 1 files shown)


docs/development/plugin-triggers.md
```diff
@@ -856,19 +856,21 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
 
 ### `nginx-dokku-template-source`
 
-- Description: Customize the source location of the `dokku.conf` nginx configuration file
+- Description: Return the path to a `sigil` template that should be used to generate the `dokku.conf` nginx configuration file.
 - Invoked by: `nginx-vhosts#install`
-- Arguments:
+- Arguments: None, however the `sigil` template can make use of the following variables: `$.DOKKU_ROOT $.NGINX_ROOT`
```
josegonzalez, line 861: That’s not the full list, might be worth linking to the nginx customization docs instead.

docs/development/plugin-triggers.md
```diff
@@ -856,19 +856,21 @@ set -eo pipefail; [[ $DOKKU_TRACE ]] && set -x
...
 ```shell
 #!/usr/bin/env bash
 
-# TODO
+echo "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/dokku.conf.sigil"
 ```
 
+The default template is viewable here: [plugins/nginx-vhosts/templates/dokku.conf.sigil](https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/templates/dokku.conf.sigil)
```
josegonzalez, line 872: Same here, link to the docs rather than the master file, as the docs will point out more and reference the correct version for your dokku release.

docs/development/plugin-triggers.md
gurpreetatwal: Hmm, I think I might be missing something. 

- The ["Customizing the nginx Configuration"](http://dokku.viewdocs.io/dokku/configuration/nginx/#customizing-the-nginx-configuration) section of the docs only talks about customizing the application specific, not the global default.
- The call to `sigil` in the plugin only passes the two variables that I mentioned: 
https://github.com/dokku/dokku/blob/7c0eb20de8555a9f0065d0b70810b9710b276fdd/plugins/nginx-vhosts/install#L65-L71

docs/development/plugin-triggers.md
gurpreetatwal: Similar confusion here, the docs currently only have a reference to `nginx.conf.sigil`, not `dokku.conf.sigil`.

gurpreetatwal:
Also complete random sidenote, thanks for the super quick review haha

PR #4079, opened 2020-07-19
Remove web installer

hugopeixoto:
Remove web installer code and documentation.

@josegonzalez : Should I remove the web installer in the same PR as the docs removal, or create a separate one? I'll split them into two different commits anyway, just in case.


josegonzalez:
We should remove it in the same PR, but we should _also_ implement the alternative in the original issue, otherwise this just removes functionality without a good way for folks to continue on and have a good experience :( .

hugopeixoto:
I'll continue the discussion on the issue then, because I didn't fully understand what's needed

josegonzalez:
I'm going to close this for now since the PR has become slightly stale. Lets continue the convo in the original issue and restart this when you're ready :)

PR #4080, opened 2020-07-19
Stop using VHOST when listing app domains and urls 

hugopeixoto:
As mentioned here in https://github.com/dokku/dokku/issues/1558#issuecomment-659149054, I think that `$DOKKU_ROOT/VHOST` should only be used when building `$DOKKU_ROOT/$APP/VHOST`, as a prefix.

This PR removes the two direct usages of `$DOKKU_ROOT/VHOST`: app domains and app urls.


josegonzalez:
Finally got around to merging this, thanks for the pull request!

PR #4084, opened 2020-07-24
When `cert:add` remove previous cert before copying the new cert

AubreyHewes:
This fixes when a user (or plugin) has replaced the app cert with a symlink to a global cert; else the symlinked file would actually be overwritten. This mainly is in the case of a wildcard cert for apps without own domains.

e.g. 
I have apps without an own domain `app-one.app.dokku` and do not want to LE them individually, but use a wildcard cert for my app domain `*.app.dokku`.

See also for more background information:
 * https://github.com/josegonzalez/dokku-global-cert/issues/3 (issue)
 * https://github.com/josegonzalez/dokku-global-cert/pull/4 (pr)


plugins/certs/subcommands/add
```diff
@@ -71,6 +71,7 @@ cmd-certs-set() {
   fi
 
   mkdir -p "$APP_SSL_PATH"
+  rm "$APP_SSL_PATH/server.crt" "$APP_SSL_PATH/server.key" 2>/dev/null
```
ltalirz, line 74: Maybe a comment explaining that this is necessary for correctly handling symlinks would be useful here?
Also, instead of  `2>/dev/null`, what about `rm -f`?

AubreyHewes: @ltalirz I would assume the why would be due to the commit/change.. else the whole code base would be full of comments as to the why something is done?

I will change the rm command as proposed.. no idea why I did it like that, must have been late :sheep:

AubreyHewes: Let me know if you still want the comment.

ltalirz: Thanks @AubreyHewes , I just thought the reason for this `rm` is probably not obvious from looking at the code (which means someone might might delete it in the future).
Anyhow, when or not to put comments is a matter of taste, so just leave it as-is - @josegonzalez will anyhow give another review.

AubreyHewes: @ltalirz I understand and agree with the reasoning.. and I know that it could alleviate possible future problems (i.e. someone removing it as they think it is unnecessary as you describe).. I actually am for a little comment on the why. Though due to the very little amounts of explanatory comments in the code I was reluctant.. as it seems not to be the way in this codebase.. A change with the comment was already prepped ;-)

josegonzalez: I normally try making the commit message verbose to explain why I did a thing. A comment in code is more likely to diverge over time from the code underneath, whereas you can write whole narratives within commit messages.

AubreyHewes:
@ltalirz @josegonzalez Probably a stupid question... do I need to force commit the extra PR commit on to the original PR commit? or will you squash it when possibly merging it?... the extra commit message is not really helpful. Let me know if I need to do something.

josegonzalez:
I don't care about extra commits. It is nice to have, but not required, as long as the commit message explaining whats going on is easily searchable.

If you had a half-dozen or more commits, I might ask for a squash, but at this point it isn't super necessary unless you want to :)

AubreyHewes:
@josegonzalez I force pushed it all together as I also believe in a commit that explains what is happening and the extra commit diluted the why (i.e. you could not see directly why in the git log.. as the commits were on the same line.. now it is all together :people_holding_hands: :wink:

AubreyHewes:
@josegonzalez Should I do something else to get the checks working? And the patch possibly merged? I think they failed as I force pushed the change :sheep:

josegonzalez:
Mind rebasing the PR onto master and then pushing that up? I can check the PR after that.

Sorry, busy few weeks at work have taken me away from Dokku :(

AubreyHewes:
@josegonzalez yeah sure. Its ok, no problem, we all have other things to do ;-)

josegonzalez:
Seems this rebase isn't actually your changes on master, but master commits somehow applied onto your pr?

AubreyHewes:
@josegonzalez Hm yeah, wow, sorry about that. Bit busy at the now, will try to fix it asap.

ltalirz:
@AubreyHewes @josegonzalez since I'm also interested in getting the symlink feature in `dokku_global_cert` released, I've taken the liberty to cherry-pick Aubrey's commit and open a new PR: https://github.com/dokku/dokku/pull/4156

josegonzalez:
Closing in favor fo #4156. Thanks @AubreyHewes for the initial effort! :)

PR #4090, opened 2020-07-30
Official plugins no longer in beta starting with 0.21.2 release

ankane:
Removes `(beta)` from plugin names to prevent confusion
 
[ci skip]


josegonzalez:
Some haven't been in beta for a long time :)

ankane:
Thanks for all the work you do on Dokku @josegonzalez! It's really great.

PR #4129, opened 2020-09-10
Prohibit non-dns names for apps and process types

### LINKED ISSUE #4102: Consider prohibiting '_' in app names
It leads to problems with virtual hosting domains and the letsencrypt plugin, because letsencrypt does not issue certificates for domain names with '_' in domain names.

It has hit me twice. I did not learn the first time.

### LINKED ISSUE #4114: Only allow process names with characters valid for DNS records
Process names are used as part of internal networks when enabled in the `network` plugin, and schedulers may use the process names as part of DNS records for the services. We should be more strict on validation of process types as a result. This will allow us to better support DNS integration with other tools by checking values up-front.

This should be added to a `post-extract` hook in the `ps` plugin. This may be best to include in the `go-procfile-util` regex.

Refs #4102

josegonzalez:
This is necessary for ensuring ssl certificates can be auto-retrieved for apps, and also easing integration into schedulers that use names and process types as part of DNS records.

As well, this fixes an issue where we may potentially have invalid DNS entries when adding apps to custom networks.

Closes #4102
Closes #4114

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -185,13 +185,13 @@ is_valid_app_name() {
   local APP="$1"
   [[ -z "$APP" ]] && dokku_log_fail "APP must not be null"
   if [[ "$APP" =~ ^[a-z].* ]] || [[ "$APP" =~ ^[0-9].* ]]; then
-    if [[ ! $APP =~ [A-Z] ]] && [[ ! $APP =~ [:] ]]; then
+    if [[ ! $APP =~ [A-Z] ]] && [[ ! $APP =~ [:] ]] && [[ ! $APP =~ [_] ]]; then
       return 0
     fi
   fi
 
   [[ -d "$DOKKU_ROOT/$APP" ]] && rm -rf "${DOKKU_ROOT:?}/$APP"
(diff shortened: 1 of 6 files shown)


plugins/common/functions
```diff
@@ -185,13 +185,13 @@ is_valid_app_name() {
   local APP="$1"
   [[ -z "$APP" ]] && dokku_log_fail "APP must not be null"
   if [[ "$APP" =~ ^[a-z].* ]] || [[ "$APP" =~ ^[0-9].* ]]; then
-    if [[ ! $APP =~ [A-Z] ]] && [[ ! $APP =~ [:] ]]; then
+    if [[ ! $APP =~ [A-Z] ]] && [[ ! $APP =~ [:] ]] && [[ ! $APP =~ [_] ]]; then
```
Abbondanzo, line 188: fyi, this added check forces 193, which applies to existing apps during upgrade. Just had to restore my apps :(

josegonzalez: Damn sorry about that. I'll fix now in 0.22.0.

PR #4138, opened 2020-09-17
Allow controlling nginx proxy-buffer-size, proxy-buffering, proxy-buffers, proxy-busy-buffers-size

Yihao-G:
Very similar to #4043 but for `proxy-buffer-size`, `proxy-buffering`, `proxy-buffers` and `proxy-busy-buffers-size`.

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -40,7 +40,7 @@ fn-nginx-proxy-buffers() {
   declare desc="get the configured proxy buffers"
   declare APP="$1"
 
-  fn-plugin-property-get-default "nginx" "$APP" "nginx-proxy-buffers" ""
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffers" ""
 }
 
 fn-nginx-proxy-busy-buffers-size() {
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
   fn-plugin-property-get-default "nginx" "$APP" "access-log-path" "${NGINX_LOG_ROOT}/${APP}-access.log"
 }
 
+fn-nginx-proxy-buffer-size() {
+  declare desc="get the configured proxy read timeout"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffer-size" "$(getconf PAGESIZE)"
+}
+
+fn-nginx-proxy-buffering() {
+  declare desc="get the configured proxy read timeout"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffering" "on"
+}
+
+fn-nginx-proxy-buffers() {
+  declare desc="get the configured proxy read timeout"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "nginx-proxy-buffers" "8 $(getconf PAGESIZE)"
+}
+
+fn-nginx-proxy-busy-buffers-size() {
+  declare desc="get the configured proxy read timeout"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-busy-buffers-size" "$(($(getconf PAGESIZE) * 2))"
+}
+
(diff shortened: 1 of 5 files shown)


josegonzalez:
I’m mobile now, but are these the values used by kubernetes?

Yihao-G:
Not sure whether these values are used by Kubernetes. 

Sorry, I probably should post the motivation in the PR description. My use case was because one of my apps sometimes sends a huge header, and Nginx cannot handle it, responding with `502` with the error of `upstream sent too big header while reading response header from upstream`. 

A [solution](https://stackoverflow.com/a/27551259/9578295) is to increase the Nginx buffer size like so:
```
proxy_buffer_size 128k;
proxy_buffers 4 256k;
proxy_busy_buffers_size 256k;
```
Or, disable proxy buffering by:
```
proxy_buffering off;
```

Although we can supply a custom Nginx template to overcome this, I thought it would be better if Dokku provides such options so that users can directly manipulate these values without playing around the templates.

plugins/nginx-vhosts/functions
```diff
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
   fn-plugin-property-get-default "nginx" "$APP" "access-log-path" "${NGINX_LOG_ROOT}/${APP}-access.log"
 }
 
+fn-nginx-proxy-buffer-size() {
+  declare desc="get the configured proxy buffer size"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffer-size" "$(getconf PAGESIZE)"
```
josegonzalez, line 29: Where is `getconf` from?

plugins/nginx-vhosts/functions
```diff
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
...
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffering" "on"
+}
+
+fn-nginx-proxy-buffers() {
+  declare desc="get the configured proxy buffers"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "nginx-proxy-buffers" "8 $(getconf PAGESIZE)"
```
josegonzalez, line 43: Should the `8` always be there?

plugins/nginx-vhosts/functions
```diff
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
...
+}
+
+fn-nginx-proxy-busy-buffers-size() {
+  declare desc="get the configured proxy busy buffers size"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-busy-buffers-size" "$(($(getconf PAGESIZE) * 2))"
```
josegonzalez, line 50: Multiplication here won't work since the value might contain a `k`?

plugins/nginx-vhosts/functions
Yihao-G: The default value of [`proxy_buffers`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers) is `8 4k|8k`, where the first value indicates the number of buffers and the second indicates the size of each buffer (defaults to one memory page size).

plugins/nginx-vhosts/functions
Yihao-G: It's from `$PATH`. Probably I should have a function like `fn-nginx-vhosts-nginx-location` to get `getconf` location instead of directly invoking it?

josegonzalez: Is that from a specific OS package? If so, we'll want to have that as a dependency.

plugins/nginx-vhosts/functions
josegonzalez: Should the `8` be computed?

plugins/nginx-vhosts/functions
Yihao-G: Hmmm feeling like introducing a dependency is not a good idea although `getconf` should be available on most distros. How about we just hide the property key-value pair in `nginx.conf` if the value is not set? Let Nginx use its default value.

Yihao-G:
Well... The PR is a bit messy now (just realise that `CONTRIBUTING.md` says preferring single commit PR)

I'll close this one and open a new one later :)

PR #4139, opened 2020-09-19
Allow controlling Nginx proxy-buffer-size, proxy-buffering, proxy-buffers, proxy-busy-buffers-size

Yihao-G:
# Purpose
This PR enables user to set [`proxy_buffer_size`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size), [`proxy_buffering`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering), [`proxy_buffers`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers) and [`proxy_busy_buffers_size`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_busy_buffers_size) Nginx properties through `dokku nginx:set`.

# Background
A redone of #4138.

The approach is back to the first commit in that PR, which didn't hide these property values if they are not set. `PAGESIZE` value is from Golang's [`os.Getpagesize()`](https://golang.org/pkg/os/#Getpagesize) so no extra dependencies introduced and guarantee to get an integer.

# Motivation
From https://github.com/dokku/dokku/pull/4138#issuecomment-694302239:
> My use case was because one of my apps sometimes sends a huge header, and Nginx cannot handle it, responding with `502` with the error of `upstream sent too big header while reading response header from upstream`. 
>
> A [solution](https://stackoverflow.com/a/27551259/9578295) is to increase the Nginx buffer size like so:
> ```
> proxy_buffer_size 128k;
> proxy_buffers 4 256k;
> proxy_busy_buffers_size 256k;
> ```
> Or, disable proxy buffering by:
> ```
> proxy_buffering off;
> ```
> 
> Although we can supply a custom Nginx template to overcome this, I thought it would be better if Dokku provides such options so that users can directly manipulate these values without playing around the templates.

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
   fn-plugin-property-get-default "nginx" "$APP" "access-log-path" "${NGINX_LOG_ROOT}/${APP}-access.log"
 }
 
+fn-nginx-proxy-buffer-size() {
+  declare desc="get the configured proxy buffer size"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffer-size" "$(fn-get-pagesize)"
+}
+
+fn-nginx-proxy-buffering() {
+  declare desc="get the configured proxy buffering"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffering" "on"
+}
+
+fn-nginx-proxy-buffers() {
+  declare desc="get the configured proxy buffers"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffers" "8 $(fn-get-pagesize)"
+}
+
+fn-nginx-proxy-busy-buffers-size() {
+  declare desc="get the configured proxy busy buffers size"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-busy-buffers-size" "$(($(fn-get-pagesize) * 2))"
+}
+
 fn-nginx-proxy-read-timeout() {
   declare desc="get the configured proxy read timeout"
   declare APP="$1"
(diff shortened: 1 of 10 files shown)


plugins/nginx-vhosts/functions
```diff
@@ -22,6 +22,34 @@ fn-nginx-access-log-path() {
...
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffering" "on"
+}
+
+fn-nginx-proxy-buffers() {
+  declare desc="get the configured proxy buffers"
+  declare APP="$1"
+
+  fn-plugin-property-get-default "nginx" "$APP" "proxy-buffers" "8 $(fn-get-pagesize)"
```
Yihao-G, line 43: Continued from https://github.com/dokku/dokku/pull/4138#discussion_r490412785:
> Should the `8` be computed?

Can you please explain a bit more why `8` should be computed? The default buffer number should be static (fixed) according to [Nginx's docs](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers).

josegonzalez: Hmmm fair, didn't read that this was `number`. I guess as a default, its fine.

tests/unit/nginx-vhosts_8.bats
```diff
@@ -173,6 +173,142 @@ teardown() {
   assert_output_contains "45s;" 0
 }
 
+@test "(nginx-vhosts) nginx:set proxy-buffer-size" {
+  deploy_app
+
+  run /bin/bash -c "dokku nginx:set $TEST_APP proxy-buffer-size 2k"
```
josegonzalez, line 179: We should test a different value here, right?

plugins/nginx-vhosts/src/pagesize/pagesize.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+func main() {
+	fmt.Println(os.Getpagesize())
```
josegonzalez, line 9: Will this value break tests if we give the server more resources? Just curious, haven't run it locally yet.

PR #4167, opened 2020-10-27
The default branch for ruby-getting-started is 'main', not 'master

nateww:
- Ensure the user is using the master branch to match later instructions to do a git push master.

[ci skip]

### DIFF
--- a/docs/deployment/application-deployment.md
+++ b/docs/deployment/application-deployment.md
@@ -9,7 +9,7 @@ Once you have configured Dokku with at least one user, you can deploy applicatio
 ```shell
 # from your local machine
 # SSH access to github must be enabled on this host
-git clone https://github.com/heroku/ruby-getting-started
+git clone https://github.com/heroku/ruby-getting-started --branch master
 ```
 
 ### Create the app


hugopeixoto:
Hm, I did this PR a couple of time ago: https://github.com/dokku/dokku/pull/4136

I think this will make conflicting instructions.

josegonzalez:
Dangit.

hugopeixoto:
`ruby-getting-started` has a default branch `main`. Unfortunately dokku only deploys `master` by default, so we need to either clone ruby-getting-started with `master` (this PR), or push with branch renaming (#4136). But I think we can't have both.

Since github now defaults to `main`, and there are a lot of people moving, having the explicit branch renaming will be helpful for them. It might be confusing for folks still using `master`, though. Maybe we should add a paragraph explaining this?

Ideally dokku would detect either branch, but it doesn't seem a simple change. Maybe this could be done by using the `git:receive-branch` hook: if it's `main`, and there's no `master` yet, set `GIT_DEPLOY_BRANCH=main`. This would simplify the docs as well. What do you think?

josegonzalez:
Could we just clone a sample app that uses `master` as it's branch? Or modify the instructions to show how to change the deploy branch?

juliosantos:
> Ideally dokku would detect either branch

This is the better long term fix. Some number of other "sample apps" are likely to follow suit. Having this flexibility helps avoid further inevitable documentation mismatches

hugopeixoto:
I think that supporting both `master` and `main` is going to be important in the long term, as more people migrate. Ideally neither `master` nor `main` users have to do extra setup, to minimize the number of things that can go wrong, and minimize support requests.

My previous PR was a temporary solution to guarantee that the instructions would work, but it favors `main` users and doesn't explain what's going on to `master` users. If you prefer to have it the other way round, I think that's fine, since probably most folks are still using `master` (there was a support request on slack just now, and I expect there will be more). But supporting both seamlessly is probably the right thing to do in the long term.

josegonzalez:
If someone wants to make a PR (with tests!) that supports both for the initial push and correctly handles subsequent pushes without waffling, that would be excellent. Otherwise, I am unlikely to do it, as it seems like it would complicate the core when a documentation change is simple enough.

juliosantos:
> it seems like it would complicate the core when a documentation change is simple enough

I appreciate your contribution to the project, and agree that the principle behind this trade-off is generally a good heuristic.

I also believe [we have some evidence](https://trends.google.com/trends/explore?date=today%205-y&q=github%20main) of an upwards trend in default branch renaming. This could lead to a number of frustrated novices and new issues opened. That is why I believe that Dokku, with its focus on developer experience, should consider default branch name independence a core feature.

I am not familiar with Dokku's core nor with how hard this would be to implement (https://github.com/dokku/dokku/pull/4167#issuecomment-717607136 seems harmless, but I've built enough software to never assume change is easy). Just pointing out that the git protocol lost a de facto constant, that adapting to this fact will contribute to a more pleasant DX, and that this may be a good reason to consider complicating the core.

josegonzalez:
The new functionality - contributed by @znz - to use the first-pushed branch as the push branch will be released in 0.22.1 :)

Issue #4166, opened 2020-10-24
Vagrant Installation with Shell Provisioning Fails


ronen25:
## Description of problem
I followed the [Vagrant installation guide](http://dokku.viewdocs.io/dokku/getting-started/install/vagrant/), but when I tried to `vagrant up` the environment, I got the following error:
```==> empty: VM not created. Moving on...
==> dokku: Running provisioner: shell...
    dokku: Running: inline script
    dokku: /tmp/vagrant-shell: line 1: make: command not found
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

Any attempts to provision the VM manually after that (using `vagrant provision`) gave me the same error.

**SSHing to the VM works**, but it does not have **any** dokku binaries or tools.
After connecting via `vagrant ssh dokku` and then running:
```
sudo apt install build-essential
```
The problem was fixed and I was able to provision the VM properly and carry on with the installation.

### How reproducible
Every time, just start with a clean dokku clone on Ubuntu 20.04 LTS.

### Steps to Reproduce

1. Follow the Vagrant [Vagrant installation guide](http://dokku.viewdocs.io/dokku/getting-started/install/vagrant/) right up until step 4 when you `vagrant up` the VM.
2. You will encounter the error about the missing `make` utility
3. Run:
```vagrant ssh dokku```
4. Notice that there are no dokku unit files or binaries, because it hasn't installed properly.

#### Actual Results
Provisioning error about a missing `make utillity`

#### Expected Results
Expected the `make` utility to be already installed, and the machine provisioned correctly.

## Environment Information
Ubuntu 20.04 LTS, Kernel 5.4.0-52-generic (AMD64)
Dokku last cloned with `57b317b0a4d6b409baed2c85ddb69ab3fda0448d` as `HEAD`
dokku version v0.21.4-91-g57b317b0

### `dokku report APP_NAME` output
N/A, VM loaded without any dokku tools.

### How (deb/make/rpm) and where (AWS, VirtualBox, physical, etc.) was Dokku installed?:
Virtualbox 6.1, via Vagrant 2.2.10

### Additional information
N/A

PR #4173, opened 2020-11-02
Add build-essential to Vagrant provision

karimsan:
When I follow the documentation for [Vagrant installation](http://dokku.viewdocs.io/dokku/getting-started/install/vagrant/), after running `vagrant up` during the provision step I get the error: 
```
dokku: Running provisioner: shell...
dokku: Running: inline script
dokku: /tmp/vagrant-shell: line 1: make: command not found
```

If I install `build-essential` package on the target VM, everything works flawlessly. 

I found an existing issue with a detailed report https://github.com/dokku/dokku/issues/4166 and this PR aims to fix it.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -53,7 +53,7 @@ Vagrant::configure("2") do |config|
       vb.customize ["modifyvm", :id, "--cableconnected1", "on"]
     end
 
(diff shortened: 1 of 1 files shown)


badsyntax:
Just confirming that adding `build-essential` fixes things for me too.

josegonzalez:
I wonder if the virtualbox vagrant box we use doesn't have this anymore. Ah well.

Issue #4151, opened 2020-10-12
Provide official GitHub action to deploy from GitHub releases


jezdez:
## Description of problem

I noticed that there are quite a few [GitHub actions for Dokku on the GitHub marketplace](https://github.com/marketplace?query=dokku) which indicates to me that there is a need in the community. I would suggest to offer an official GitHub action for security and documentation reasons.

josegonzalez:
This makes sense. What features should the push action have? Is there a community one we can just adopt/make official (with the consent of the maintainer of course)?

jezdez:
I'm not completely sure, there are a few differences in their approaches, but I would say whatever you chose it should be using native workflow arguments instead relying on environment variable to pass configuration parameters. I would also lean towards providing a high-level approach instead of exposing all kinds of SSH options so that regular users don't accidentally misconfigure it. E.g. using SSH passwords for deployment should be not possible.

julienma:
I too would prefer an "official" action.

I've been using https://github.com/obrassard/action-dokku-push with success. There are multiple plugins, but this one seems the most straightforward to me (only required options, uses a standard git repo URL in ssh format—the same that I use to create my git remotes—instead of splitting this in multiple params), have the ability to cancel previous runs.
Could be a good starting point?

jsuchal:
We use this one https://github.com/marketplace/actions/dokku-deploy-action and seems to do its job good.

josegonzalez:
@obrassard would you be fine with me making your action the official one? I have a few (small) changes to make to it, but we'd republish it under the Dokku org and I would give you commit bit.

If not, I'll write one from scratch, but I'd rather save myself some time and effort in the long-run :)

obrassard:
Hi @josegonzalez I am fine with that, you can go ahead! 👍

josegonzalez:
@obrassard mind adding me as a repo-admin on your repo? I'll grant you commit bit and everything once the repo has been moved.

josegonzalez:
@obrassard what is the Dockerfile for `obrassard/alpine-git-ssh` ? That is currently in use by that github action.

obrassard:
Here it is : https://github.com/obrassard/dockerfiles/blob/master/alpine-git-ssh/Dockerfile

josegonzalez:
I went through @obrassard's github action as well as others and did a few things:

- created a new action based on @obrassard's work. I also kept your copyright, though added myself due to the extra work
- general file and variable refactor
[...]

PR #4179, opened 2020-11-07
Add deploy with GitHub Actions doc

badsyntax:
I noticed that building docker images can negatively affect my runtime apps performance, so I setup builds in CI using [these docs as a guide](http://dokku.viewdocs.io/dokku/deployment/methods/images/#deploying-an-image-from-ci).

I think adding docs specific to GitHub Actions might be useful for others.

### DIFF
--- a/docs/community/tutorials/deploying-with-github-actions.md
+++ b/docs/community/tutorials/deploying-with-github-actions.md
@@ -1,0 +1,61 @@
+# Deploying with GitHub Actions
+
+GitHub actions can be used to build the app image and deploy it.
+
+## Prerequisites
+
+You'll need a SSH keypair created (without a passphrase) to allow the GitHub Action runner to SSH to the dokku server.
+
+On your dokku server:
+
+```bash
+ssh-keygen -N "" -f /root/.ssh/githubactions
+```
+
+Add the `githubactions` public key to authorized_keys:
+
+```bash
+cat /root/.ssh/githubactions.pub >> /root/.ssh/authorized_keys
+```
+
+## Setting up the GitHub Action workflow
+
+### Secrets
+
+Create the following secrets in your GitHub repo:
+
+- `SSH_PRIVATE_KEY` (the contents of `/root/.ssh/githubactions`)
+- `DOKKU_HOST` (the hostname of your dokku server)
+- `DOKKU_APP_NAME` (the name of the dokku app you want to deploy)
+
+### Workflow
+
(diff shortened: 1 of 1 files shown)


josegonzalez:
While I understand the use case, I think we're conflating github actions with loading a remote image - something already documented [here](http://dokku.viewdocs.io/dokku/deployment/methods/images/#deploying-an-image-from-ci). Any reason why that doc doesn't suffice (other than not filling in the other Github Actions bits)?

I think the upcoming work in #4151 neatly handles the general Github Actions use case. Going to close this for now.

badsyntax:
> I think we're conflating github actions with loading a remote image - something already documented here

I don't follow. The purpose of this PR was to provide an example of how to automate (with CI) what's mentioned in the docs. Now I could be missing something but I don't see how this use case is covered by https://github.com/dokku/github-action.

> Any reason why that doc doesn't suffice (other than not filling in the other Github Actions bits)?

I could ask the same about about the reason for building https://github.com/dokku/github-action - it's about making life a little easier for dokku users.

I personally feel that building docker images on a runtime server is a bad idea. My first experience of using dokku was experiencing degraded runtime performance due to this approach. 

Would you consider adding this use case (building an image in CI and loading it with dokku remotely) to https://github.com/dokku/github-action?

josegonzalez:
A few things:

- The problem with your doc addition is that it is a specific, already documented workflow tailored to a specific CI environment. Folks will want that same example on other CI systems, and I don't wish to do that.
-  Yes, we have a Gitlab-specific example that is contrary to my first point, but that will actually be removed in favor of #3433.
- Part of the work for all the CI docs is to unify the workflows so there is one single image - https://github.com/dokku/ci-docker-image - that handles the general integration for all CI systems. Adding support there is the correct place, and then we'd flow that through everywhere else.
- Your workflow actually doesn't make things simpler.
  - It only deploys on release creation, whereas a more common workflow is on push to the `master`/`main` branch.
  - It assumes the user will enable root access for deploys, which is the opposite of what we want to enable (all actions should flow through the dokku user).
- Yes, I could have merged this as is, but then I'd have to do the extra work to delete it later and break links.
- The workflow uses docker commands directly. This has never been great, and while I've never had a use case for it, we should probably have something in the core for it. I filed #4219 to enable that work.

josegonzalez:
0.24.0 will have the `git:from-image` command, which will support this sort of workflow by embedding the image pulling into the `Dockerfile` being built :)

PR #4186, opened 2020-11-10
Add dokku-discourse to community plugins list

badsyntax:
This change adds the dokku-discourse plugin to the plugins list.

I'm not really sure what compatibility version to use, so i chose the one i'm using. Are there some guidelines I can follow to understand what version my plugin is compatible with?

[ci skip]


josegonzalez:
Minimum version is whatever version you're willing to test with :)

PR #4197, opened 2020-11-22
Retire intermediate containers after use

josegonzalez:
Rather than waiting for the next deploy, take advantage of the retire system to retire these containers immediately.

Note that since the retire process happens out of band, the containers may stick around for up to 5 minutes, after which point they will be removed.

Customization of the wait time to retire - currently 60 seconds - is up for future debate. The containers ultimately haven't been useful for debugging, so keeping them around for longer won't help in most cases, and folks can disable the dokku-retire service (or cron) if that is desired.


josegonzalez:
Failures are due to flaky builds - ran those locally, they worked - so merging.

PR #4205, opened 2020-11-29
Use https for links to documentation

gamwe6:


josegonzalez:
Neither of these urls currently support https - I don't yet have DNS access for viewdocs - so I'll close this for now. Sorry.

PR #4208, opened 2020-12-01
feat: add ability to change the access log format

ankane:
Hi @josegonzalez, thanks for all the amazing work on Dokku!

This PR adds the ability to set a custom access log format for NGINX, which can be useful for things like [anonymizing IP addresses](https://chriswiegman.com/2019/09/anonymizing-nginx-logs/).

```sh
dokku nginx:set node-js-app access-log-format custom
```

There's not a way to customize the error log format (as far as I know), and it plays nicely with `access-log-path off` :tada: Also, I wasn't sure what the next release would be for the docs.


ankane:
Thanks @josegonzalez!

Issue #4251, opened 2020-12-17
Supplying certificate paths fails if given symlink


magikid:
## Description of problem

When the certificate or key file are symlinks and given to the `certs:add` command, the command will fail with the error "CRT file specified not found, please check file paths".  Even though the file is a symlink, it is still readable and should be just fine to use.  

### How reproducible

Create a symlink to your already existing certificate.  Supply that path to the `certs:add` command.

### Steps to Reproduce

1. `ln -s /var/www/real_cert.pem /var/www/current_cert.pem`
2. `ln -s /var/www/real_key.pem /var/www/current_key.pem`
3. `sudo dokku certs:add app-name /var/www/current_cert.pem /var/www/current_key.pem`

#### Actual Results

```
 !     CRT file specified not found, please check file paths
```

#### Expected Results

Certificate and key accepted as valid.

### `dokku report nodejs-app` output

```
-----> uname: Linux zoidberg 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
-----> memory: 
                     total        used        free      shared  buff/cache   available
       Mem:            981         333         165           1         482         495
       Swap:             0           0           0
-----> docker version: 
       Client: Docker Engine - Community
        Version:           20.10.1
        API version:       1.41
        Go version:        go1.13.15
        Git commit:        831ebea
        Built:             Tue Dec 15 04:34:58 2020
        OS/Arch:           linux/amd64
        Context:           default
        Experimental:      true
       
       Server: Docker Engine - Community
        Engine:
         Version:          20.10.1
         API version:      1.41 (minimum version 1.12)
         Go version:       go1.13.15
         Git commit:       f001486
         Built:            Tue Dec 15 04:32:52 2020
         OS/Arch:          linux/amd64
         Experimental:     false
        containerd:
         Version:          1.4.3
         GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b
        runc:
         Version:          1.0.0-rc92
         GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff
        docker-init:
         Version:          0.19.0
         GitCommit:        de40ad0
-----> docker daemon info: 
       Client:
        Context:    default
        Debug Mode: true
        Plugins:
         app: Docker App (Docker Inc., v0.9.1-beta3)
         buildx: Build with BuildKit (Docker Inc., v0.5.0-docker)
       
[...]

PR #4252, opened 2020-12-17
Allow symbolic links for certificate and key files

magikid:
This corrects an issue where if a symbolic link was given to the
`certs:add` command, an error would be thrown even though the file was
valid.

Issue: #4251


josegonzalez:
Does this add the actual file underneath or a symlink to the original cert? I'm torn on which is the "correct" behavior.

magikid:
I've been fiddling around with this all afternoon and the PR as it currently is written only adds the symlink and not the file underneath.  Since it copies the file into the $APP_DIR/tls folder and set nginx to read from that, I don't see that it makes much difference either way.

PR #4261, opened 2020-12-19
Add remove by fingerprint and json format output to ssh-keys plugin

josegonzalez:


josegonzalez:
Weird, circleci is not reporting tests as running 😅 

https://app.circleci.com/pipelines/github/dokku/dokku/667/workflows/7c4fda94-7dea-4b9b-a823-f692601a9ad6

Issue #4267, opened 2020-12-21
Add ability to rename apps with old name restrictions to new name restrictions


sinys15:
I have an old one project with the name updater_local. And when I want to do something with this project I got a message "App name must begin with lowercase alphanumeric character, and cannot include uppercase characters, colons, or underscores". And now I can't do anything with this project. I can't get settings and I can't rename it:
$ dokku apps:rename updater_local updater-local
 !     App name must begin with lowercase alphanumeric character, and cannot include uppercase characters, colons, or underscores

Could you please fix it? I think it will be fair to allow renaming old projects to a new name with the new restrictions.

afuno:
This is a problem in the new version (0.22.x). When updating Dokku, for some reason, protection didn't work (or it does not exist at all). But after updating Dokku almost completely stops working and displays this error text.

josegonzalez:
There was a migration guide entry for this [here](http://dokku.viewdocs.io/dokku/appendices/0.22.0-migration-guide/), and that should have been performed prior to an upgrade.

That said, I can certainly add the ability to rename apps even if the old app is incorrectly named. I'll add that as a feature to Dokku 0.22.3.

afuno:
> There was a migration guide entry for this here, and that should have been performed prior to an upgrade.

It's good. Well it would be nice to prohibit updating until the problems are fixed.

> That said, I can certainly add the ability to rename apps even if the old app is incorrectly named. I'll add that as a feature to Dokku 0.22.3.

Thank you. This should help other people get rid of their headaches. Now the situation in which there is Tom and his ilk really looks bad. It is extremely difficult to return from it.

josegonzalez:
Users seeing this issue can downgrade dokku to the previous release via yum or apt. Here is the command for apt:

```shell
apt install dokku=0.21.4
```

josegonzalez:
The linked PR (#4270) fixes this :)

PR #4270, opened 2020-12-22
Allow renaming old applications to new format

josegonzalez:
0.22.0 changed allowed values for app names, making it more difficult for users to fix their app names - you need to either downgrade+rename or recreate the app. The `apps:rename`  command can now rename these old applications to the new standard.

Also drop some app name verification from internal commands. Duplicate verification doesn't help much, and though it isn't excessively slow, it does prevent us from running commands against old apps that exist.

Refs #4267

### DIFF
--- a/plugins/apps/functions.go
+++ b/plugins/apps/functions.go
@@ -83,6 +83,17 @@ func destroyApp(appName string) error {
 
 	forceCleanup := true
 	common.DockerCleanup(appName, forceCleanup)
+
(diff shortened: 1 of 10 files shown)


josegonzalez:
I manually verified this worked by:

- installing 0.21.4
- creating and deploying an app named php_vendor
- upgrading to this patch release (via make copyfiles)
- running `dokku ps:rename php--vendor`

Seems legit.

PR #4300, opened 2021-01-06
allow disabling hsts globally and explicitly enable per app

AubreyHewes:
The HSTS Header feature introduced in https://github.com/dokku/dokku/pull/3843 breaks current apps that already do their own HSTS header. If an app already emits a HSTS header; then two headers are added; breaking the spec / also may be contradictory values.

An example of this bad default behaviour and contradictory values is when an app explicitly does not include sub domains; this feature does (by default). This entails that possible subdomains not hosted in dokku; maybe private and not https also require a tls cert; the browser will not open them.

Instead of checking every app on multiple dokku servers (and their source code) I would like to disable the feature globally and opt-in on a per app basis, vis a vis other dokku settings.

This PR adds a global config option `DOKKU_GLOBAL_DISABLE_HSTS` to disable the feature (opt-out) and bring the control back to the developer (opt-in) on a per app basis.

Setting `DOKKU_GLOBAL_DISABLE_HSTS` to `true` implies that you have to opt-in to the functionality on a per app basis.
> After setting this you will have to restart apps to remove the dokku-HSTS Header

```shell
dokku config:set -global DOKKU_GLOBAL_DISABLE_HSTS=true
```

Then opt-in per app

```shell
dokku nginx:set node-js-app hsts true
```

I also added a description to the docs, let me know if anything needs to be changed.


josegonzalez:
I don't want to add yet another global environment variable. Can we use normal properties for this?

AubreyHewes:
@josegonzalez yeah sure.. 

I tried `dokku nginx:set hsts false` but that fails (as no app name; the error is wrong; says incorrect prop).. that's why I used a global config var

Currently you can not do `dokku nginx-proxy:set hsts false` or `dokku nginx-proxy:set -global hsts false`

Can you explain how it can be done? Thanks!

josegonzalez:
Well I mean we should fix whatever is blocking us from allowing `--global` to work for this property, and then use that. The golang implementation of `:set` supports global properties (as long as they are whitelisted) so we should probably do something similar here.

AubreyHewes:
So the following should work (but doesn't at the moment) ??

    dokku nginx:set --global hsts false

Is there a reason it is not `-global` like the config command?

how would you then retrieve the global property in the internal-functions?

josegonzalez:
It doesn't work at the moment because it wasn't necesssary. I'm saying you should make the changes that would allow it to work for this particular property.

The config command allows `-global` both because of how the pflag golang module works (it allows single dashes, like the standard lib flag module does) and because its internally interpreted as a boolean (global config or not global) and stored in a completely different place. App env vars are not handled by the property system. Of note, I don't even document or test the use of `-global` (single dash), just `--global`, so if that gets removed at some point, I would not consider it a BC break.

Retrieval of the property would be in the same way we retrieve a normal property, except with `--global` instead of the app name?

AubreyHewes:
hm.. yeah ok...

    dokku nginx:set -global hsts false

actually works, but sets as app name -global in the props dir (`/var/lib/dokku/config/nginx/-global` or with `--global` at `/var/lib/dokku/config/nginx/-global`)

so then reading it is just using -global as the app name...

I can do this ;-)

josegonzalez:
No, we should not support `-global`. Please don't use single dash going forward. I'm very likely to break that usage once I tighten up property setting for bash plugins.

AubreyHewes:
aight.. but that is pure docs

josegonzalez:
Internally, the property would be stored with the name you use, so if you tried supporting both --global and -global, those would technically be two different properties.

AubreyHewes:
@josegonzalez I updated the PR, see the latest commit

Docs description;

You can opt-out of the HSTS Header being set on all apps. 

```shell
dokku nginx:set --global hsts false
```

Setting this implies that you have to opt-in to the functionality on a per app basis.
> After setting this you will have to restart apps to remove the dokku-HSTS Header

Then opt-in per app

```shell
dokku nginx:set node-js-app hsts true

File: docs/configuration/nginx.md [hunk 1]
```diff
@@ -61,6 +61,24 @@ The following options are also available via the `nginx:set` command:
 
 Beware that if you enable the header and a subsequent deploy of your application results in an HTTP deploy (for whatever reason), the way the header works means that a browser will not attempt to request the HTTP version of your site if the HTTPS version fails until the max-age is reached.
 
+#### Disable HSTS Header
+
+You can opt-out of the HSTS Header being set on all apps.
```
josegonzalez, line 66: Rewrite this as:

```
HSTS Header can be disabled for all apps by setting the `hsts` property to false after passing the `--global` flag to `nginx:set`. Changing this value globally or on a per-app basis will require rebuilding the nginx config via the `nginx:build-config` command.

File: docs/configuration/nginx.md [hunk 2]
```diff
@@ -61,6 +61,24 @@ The following options are also available via the `nginx:set` command:
...
+#### Disable HSTS Header
+
+You can opt-out of the HSTS Header being set on all apps. 
+
+```shell
+dokku nginx:set --global hsts false
+```
+
+Setting this implies that you have to opt-in to the functionality on a per app basis.
+> After setting this you will have to restart apps to remove the dokku-HSTS Header
```
josegonzalez, line 73: Lets remove these two lines.

docs/configuration/nginx.md
```diff
@@ -61,6 +61,24 @@ The following options are also available via the `nginx:set` command:
...
+
+Then opt-in per app
```
josegonzalez, line 75:
```
Once the HSTS setting is disabled globally, it can be re-enabled on a per-app basis by setting the `hsts` property as normal.
```

josegonzalez (changes requested):
The global property should also be added to the nginx:report output.

josegonzalez:
We should also have a test for this. Checkout `nginx-vhosts_6.bats` (`(nginx-vhosts) nginx:set hsts` is the test) for an example. I would make a separate test in that file that checks the permutations.

plugins/nginx-vhosts/command-functions
```diff
@@ -44,7 +44,7 @@ cmd-nginx-report-single() {
     "--nginx-bind-address-ipv6: $(fn-plugin-property-get-default "nginx" "$APP" "bind-address-ipv6" "::")"
     "--nginx-disable-custom-config: $(fn-plugin-property-get-default "nginx" "$APP" "disable-custom-config" "false")"
     "--nginx-error-log-path: $(fn-nginx-error-log-path "$APP")"
-    "--nginx-hsts: $(fn-plugin-property-get-default "nginx" "$APP" "hsts" "true")"
+    "--nginx-hsts: $(fn-plugin-property-get-default "nginx" "$APP" "hsts" "$(fn-plugin-property-get-default "nginx" "--global" "hsts" "true")")"
```
josegonzalez, line 47: No, we should have separate report outputs for global vs app specific. So `--nginx-hsts` and `--nginx-global-hsts`. No need to change the default for the app-specific one.

josegonzalez (changes requested):

plugins/nginx-vhosts/command-functions
AubreyHewes: hm.. yeah ok.. but the app specific default is dependent on the global

AubreyHewes: so if the global is false then the app default needs to be false

AubreyHewes:
````shell
    "--nginx-global-hsts: $(fn-plugin-property-get-default "nginx" "--global" "hsts" "true")"
    "--nginx-hsts: $(fn-plugin-property-get-default "nginx" "$APP" "hsts" "$(fn-plugin-property-get-default "nginx" "--global" "hsts" "true")")"
````

Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:47]: else it is confusing.... global is false but app is true.. but as app has no own setting it is actually false.. so should display false

Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:47]: Right but this is the report. Downstream tools can interpret the two values to figure out what to do.

If anything, we should change how the value gets pulled internally - maybe a helper function to fetch the global value with fallback to the app one? - and then report the values with empty default for app-specific and `true` default for global.

**Comment by AubreyHewes** [comment]
@josegonzalez is there a docker file for running the tests in an isolated env.. it is f-ing up my local dokku dev env :sheep:

**Comment by josegonzalez** [comment]
Nah not really. Thats a good idea though.

I would just clone the repo again to another path and setup a new vagrant vm for development.

**Comment by AubreyHewes** [comment]
hm yeah I don't use vagrant, docker is enough ;-)

May look into adding a Dockerfile for running the test locally another time.. for now will just see where it fails in circle

File: plugins/nginx-vhosts/command-functions (hunk 4 above)
Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:48]: yeah but if the default is false and the app has never explicitly set a value it should be false.. a tool would not know this

i.e. if an app has never enabled hsts it's default value is true, but the global value is false. So the app should display false. If an app actually sets it as true then it is true..?

Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:48]: Right but again, this is the report. Lets say someone wants to force set it to true and has it globally set to true. The report wouldn't give the actual value of the property, but a computed one.

You'll see something similar with the resource properties that can have defaults set for all process types within an app. We show the values separately, not in computed fashion.

Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:48]: hm ok, still as a user/dev I find that confusing.. the report says that the global property is disabled, but the app property is enabled. Even though (as the property has never been set) the app is actually disabled...

We seem to be speaking in circles ;-)

Out of the box dokku the app hsts default property is true
Setting hsts property globally to false gives you a report that says that globally it is false but the app is true (as it is the default)
This makes it confusing, as my report tells me that it is disabled but enabled for this app; which is not true

Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:48]: What I'm saying is that we remove the app default, so `true` => "" (empty string). A user would need to parse that in their head.

The default global value would be true.

Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:48]: Maybe I do not understand what you mean with report... I assume the current active config/props... something that displays what I want to see to understand how/what is active..

Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:48]: If the app is forced to true then it has an entry? so would not go to the default; which is the global check

Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:48]: So this particular diff comment is on a piece of code that gets run when someone calls `nginx:report` (or `report $APP`).

We want to expose the properties set by the user so that either automation can set the values as expected, or that they can report issues and I can debug what the actual state of their system is without asking more questions.

In that case, we should report the values _as set by the user_ unless there are system defaults. If we add a global default, then the app-specific version has no actual defaults, and should be an empty string.

Outside of this usage, we should probably have a single function - `fn-nginx-vhosts-hsts-enabled` maybe? - that fetches the app-specific value (with a default of empty string). If the fetched value is not empty, then it is returned. Otherwise, we fetch the global property value (with a default of `true`) and return that. This function would be used in every place outside of the report function to fetch whether or not hsts is enabled.

Comment (AubreyHewes) [reply plugins/nginx-vhosts/command-functions:48]: so if global is set and app has never been set it should be empty for the app in the report?

could you otherwise post an example of how you would like the report to look like in the following situations:

 * out of the box app hsts
 * out of the box app hsts, with global hsts disabled
 * out of the box app hsts, with global hsts disabled, app hsts enabled
 * out of the box app hsts, with global hsts disabled, app hsts disabled

I think those are the cases

The report itself is also up to discussion as if I have disabled hsts on an app why does the report display hsts things that are not useful.. ; if disabled do not display parameters; disabled is disabled

Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:48]:
```
    "--nginx-global-hsts: $(fn-plugin-property-get-default "nginx" "--global" "hsts" "true")"
    "--nginx-hsts: $(fn-plugin-property-get-default "nginx" "$APP" "hsts" "")"
```

Those are the two lines that should be in the report function diff.

**Comment by AubreyHewes** [comment]
wow >=30mins for a testsuite

File: plugins/nginx-vhosts/command-functions (hunk 4 above)
Comment (josegonzalez) [reply plugins/nginx-vhosts/command-functions:48]: The app-specific report value for nginx-hsts would have an empty default.

A similar condition occurs for the `deploy-branch` in the git plugin (and actually the output there should be default empty string, so thats wrong and will be fixed in a minor release).

The resource plugin is also similar in that you can set a default value - `_default_` - and that is shown separately from the explicitly set value.

File: plugins/nginx-vhosts/internal-functions [hunk 5]
```diff
@@ -22,6 +22,13 @@ fn-nginx-vhosts-manage-hsts() {
   local NGINX_HSTS_CONF="$DOKKU_ROOT/$APP/nginx.conf.d/hsts.conf"
   local HSTS_TEMPLATE="$PLUGIN_AVAILABLE_PATH/nginx-vhosts/templates/hsts.conf.sigil"
 
+  # if hsts is disabled globally then only apply it if an app has it set explicitly
+  if [[ "$(fn-plugin-property-get-default "nginx" "--global" "hsts" "true")" = "false" ]] && \
+     [[ "$(fn-plugin-property-get-default "nginx" "$APP" "hsts" "false")" = "false" ]]; then
+    rm -rf "$NGINX_HSTS_CONF"
+    return;
+  fi
+
```
josegonzalez, line 31: This addition should become a new function:

```
fn-nginx-vhosts-hsts-enabled() {
  declare APP="$1"

  if [[ "$(fn-plugin-property-get-default "nginx" "$APP" "hsts" "")" == "true" ]]; then
    echo "true"
    return
  fi

  fn-plugin-property-get-default "nginx" "--global" "hsts" "true"
}
```

And then the `local HSTS` line near the beginning of `fn-nginx-vhosts-manage-hsts` should be:

```
local HSTS="$(fn-nginx-vhosts-hsts-enabled "$APP")"
```

plugins/nginx-vhosts/command-functions
AubreyHewes: ah but you are missing the point the app hsts default is true

.. so even without this change the app hsts default displayed value must be true

current:
`"--nginx-hsts: $(fn-plugin-property-get-default "nginx" "$APP" "hsts" "true")"`

AubreyHewes: else it has no meaning;

if hsts global false the app must display a default (if not set) that is correct and understandable; i.e. false or true

josegonzalez:
Just commented with what the diff to support this should be. Hope that makes it a bit more clear.

Also yeah, tests take about an hour. Its somewhat frustrating but like... half the tests require a deployed app, and deploying takes a minute or so, so a bunch of tests take 1-2 minutes because we create new apps for every test (blank slate).

plugins/nginx-vhosts/command-functions
josegonzalez: I understand your point. From your vantage point, it is nice to know the computed value vs the actual. For me - and for automation purposes - the computed value is not interesting.

If you want, adding a new flag - `--nginx-hsts-is-enabled` or something - would be fine. Not all properties can be set - there's definitely a bunch of precedent for that - and this allows us to have the computed value show up in addition to setable ones.

AubreyHewes: anyway semantics.

If you want what you are describing that is fine.

As a user I would like a report that actually describes the state; a readable report.. not something that I would have to dig in to the config/props to find out what the state actually is.

plugins/nginx-vhosts/internal-functions
AubreyHewes: this would break bc.. current default for apps that have never set the property is true

the PR was to allow an option to opt out of the global hsts that is applied

moving it to a reusable function is great, but does not address the issue

plugins/nginx-vhosts/command-functions
josegonzalez: But thats the thing, this PR is effectively saying that the app specific value has no default and falls back to a global value - which is already a BC-break. The state of the system is actually changed by this PR, as the default interpolated value is no longer actually correct.

Using a computed value - where we fallback to the global - as the interpolated value would be wrong from an automation sense:

- Globally `true`, app-specific has no user-set value: If I skip setting the app-specific value to `true` in my automation because it shows as `true`, and my next step is to set it globally to `false`, then the system is actually out of sync. Needing to remember order of operations is bad

For a report sense, I need to know if the user has the property set in one way or another so I can effectively debug what they have setup.

If we want to have something that can be inferred by the user, then adding a new computed report value is the way to go.

One thing that might make this more clear is if we had different prefixes for setable properties, so `--nginx-prop-hsts` would be the exact property setting, while `--nginx-hsts-is-enabled` would be the user-friendly report value.

plugins/nginx-vhosts/internal-functions
josegonzalez: If the property has never been set, the global value is defaulted to true, so there is no change in the internal workings of Dokku. The only bc-break would be in the report output.

AubreyHewes:
> Just commented with what the diff to support this should be. Hope that makes it a bit more clear.
> 
> Also yeah, tests take about an hour. Its somewhat frustrating but like... half the tests require a deployed app, and deploying takes a minute or so, so a bunch of tests take 1-2 minutes because we create new apps for every test (blank slate).

Yeah, I noticed.. I am thinking most of the non deployment tests could actually be tested without a deployment, pure unit testing (does my code work)

plugins/nginx-vhosts/internal-functions
josegonzalez: If a user wants to opt out of the system's default of enabling hsts, they can set the global value to false. The apps won't have an implicit value of `true` set anymore with the diff I am suggesting, so the system will disable hsts for all apps.

If a user _has_ set a value, then that value would use that value.

I've also updated the diff to reflect that since I'm a noob and realized it would only respect the user's app setting if the app setting was true 😅

josegonzalez:
Yeah I think the idea regarding tests is that they are mostly actually not unit tests but integration tests. For a project like Dokku, integration tests are super important because of the many interactions that can happen. Lots of interactions are only really testable during the deploy process because of implied system state.

Our bash code isn't super well structured for unit tests  - though we should certainly have them - and the idea behind the golang rewrite is that it would enable unit testing - though I've been lazy and avoiding writing those. Happy to have help with that :)

tests/unit/nginx-vhosts_6.bats
```diff
@@ -107,6 +107,44 @@ teardown() {
...
+  # apply on app
+  run /bin/bash -c "dokku nginx:set $TEST_APP hsts true"
+  # check it is now applied
+  run /bin/bash -c "dokku nginx:build-config $TEST_APP"
+  echo "output: $output"
+  echo "status: $status"
+  assert_success
+  assert_output_contains "Enabling HSTS" 1
+
+  # disable globally
```
josegonzalez, line 138: Should say `implicit enable globally`.

Should we also have an assert for implicitly enabling?

AubreyHewes: hm first need to get the test working... then will look at code comments... the non-ability to just run one test-case without waiting an hour is just enough to make me think about walking away.. I think the code works (works in my env; cat>./file)

AubreyHewes: this PR has nothing to do with implicitly enabling.. the dokku codebase implicitly enables; this pr just allows to disable it.. so if it is disabled it is disabled; if it is enabled you get the same message

josegonzalez:
Mind hopping on slack for a bit? I think we're talking past each other here and I'm probably pissing you off :( I'm happy to checkout your branch and run tests for you on my machine :)

AubreyHewes:
I changed it; 

 * extracted the determination logic to a possible reusable method
 * added the computed value to the report (default is nginx-app-hsts, nginx-hsts is the computed value.. )

> Aside, I wonder if we can add a test step that only tests plugins that were changed - so in this case we'd have a separate step after the go-tests but before test that would run the plugin-specific tests. Might even be a way to do this with circleci pipelines...

You could do a git change (bash) script to determine the tests paths that should be applied for the commit file changes... though the tests would need to be in relatable dirs... i.e. split from all in one files

josegonzalez:
I think changing what the existing report value reflects (the value of the `hsts` property to the "union" of the app and global `hsts` property value) is a bc-incompatible change. That said, I'm happy to do that for the next minor, which should be soon anyhow, as there are a few major features and such that I'd like to ship.

Regarding tests, we could probably do something like this:

```
https://dev.to/acro5piano/exit-circleci-jobs-if-changed-files-do-not-match-specific-pattern-mel
```

And have a separate step for each plugin _prior to_ the full tests.

AubreyHewes:
I have changed it;
````
  Nginx global hsts:             false                    
  Nginx applied hsts:            false                    
  Nginx hsts:                    true                 
````

Nginx hsts is now the same as before (bc).
Nginx applied hsts is the if it will be applied...

AubreyHewes:
> Regarding tests, we could probably do something like this

Yeah well this is a known problem on all CI... what to do and when... CI resource management

in this case changing an internal function should only imply testing that function (and where it is used).. not testing non related functions

josegonzalez:
@AubreyHewes happy to take over this PR + write tests if you need that :)

AubreyHewes:
hey @josegonzalez, it's alright, I've been really busy. I am going to look at it again either this evening or tomorrow evening. Thanks.

josegonzalez:
Taking over this PR to get it over the finish line for 0.23.0.

AubreyHewes:
@josegonzalez Cool, If you have any questions you know where to find me ;-)

AubreyHewes:
@josegonzalez Cool, you merged it... but I don't see the difference else from documentation and the build-config queried.. was it the plugin `dokku nginx:build-config` vs `dokku proxy:build-config` ?

josegonzalez:
You didnt implement any way of setting keys globally, so those were just failing. The failures werent caught because your tests didn't assert anything for those calls.

AubreyHewes:
@josegonzalez I have one last peeve on the HSTS default that was introduced and that is the includeSubDomains=true... this should never have been true, but without breaking bc (which it broke).. probably can never get rid of it now :-D

AubreyHewes:
@josegonzalez ah... I was setting as `--global`.. which set them as config on app `--global` I thought that was the way...

josegonzalez:
We can make the same change for subdomains. Make that an option for both app/global, default on in global, default empty value in app.

AubreyHewes:
Yeah probably should do that as well, I think...

possibly all hsts options should be individually possible.. Was the main reason for the PR that the HSTS messed up domains that had apps on dokku and apps in other places.. without tls.. so those failed.

AubreyHewes:
But anyway; Nice work, and thanks for fixing the tests and understanding what I was doing wrong!

PR #4311, opened 2021-01-10
Allow named volume mounts in regex

nisseknudsen:

- Allows the usage of named volumes in storage:mount by validating correctly in Regex
- Embeds [Docker rules on named volumes](https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/daemon/names/names.go#L6):
 - Min 2 characters long in name
 - Needs to follow schema `[a-zA-Z0-9][a-zA-Z0-9_.-]`

### DIFF
--- a/plugins/storage/functions
+++ b/plugins/storage/functions
@@ -7,7 +7,7 @@ source "$PLUGIN_AVAILABLE_PATH/docker-options/functions"
 verify_paths() {
   declare desc="verifies storage paths"
   local -r passed_path=$1
-  if [[ "$passed_path" = \* ]]; then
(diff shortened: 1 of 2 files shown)


josegonzalez:
@nisseknudsen mind adding a test case for this?

josegonzalez:
@nisseknudsen minor _bump_. I'm also happy to write the tests for you if you need that :)

nisseknudsen:
@josegonzalez : Totally - just got out of work for the week. I will check it rn - I hope I can directly understand how and where to write tests.

PR #4337, opened 2021-01-18
Add logrotation to container log files

### LINKED ISSUE #4023: Add command to manage docker container log truncation
This can be as simple as automatically creating a logrotate file:

```
https://blog.birkhoff.me/devops-truncate-docker-container-logs-periodically-to-free-up-server-disk-space/
```

And having this enabled/disabled via a global environment variable (DOKKU_LOGROTATE). If there is no value set, it is implied true. If it has a value of `false`, we delete the file. If it is set to `true`, we create the file. The value can be set via `logs:set`, and is output to `logs:report`.

On install of dokku, we'll set this value if it is not set.

josegonzalez:
Closes #4023

### DIFF
--- a/docs/deployment/logs.md
+++ b/docs/deployment/logs.md
@@ -59,7 +59,7 @@ dokku logs:failed --all
 
 ### Docker Log Retention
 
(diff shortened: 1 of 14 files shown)


josegonzalez:
I wonder if this is better suited to using injected docker-options...

PR #4340, opened 2021-01-18
Allow configuring x-forwarded-* proxy headers via nginx:set

### LINKED ISSUE #4339: Q: Resetting X-Forwarded-For header?
Hi,

I was wondering why the `X-Forwarded-For` header is being reset with the `$remote_addr` value in the [default Dokku nginx config](https://github.com/dokku/dokku/blob/e64b69248a4327bc41b3a3e1eb3736ccfb5b6bbb/plugins/nginx-vhosts/templates/nginx.conf.sigil#L100). This causes issues with fetching the client's IP address with an intermediate proxy server, like this example:

Client (1.1.1.1) --> Intermediate proxy (2.2.2.2) --> Dokku nginx --> Dokku app

In this case, the user's IP address in the Dokku app is reported as 2.2.2.2 (iso 1.1.1.1). This is because the Dokku nginx config overwrites any present `X-Forwarded-For` header with the current `$remote_addr` (2.2.2.2), rather than appending a second IP address.

An easy fix for this issue would be to replace $remote_addr with `$proxy_add_x_forwarded_for`.

P.s. Thank you for Dokku <3

bjornpost:
Fixes #4339

### DIFF
--- a/docs/configuration/ssl.md
+++ b/docs/configuration/ssl.md
@@ -143,6 +143,12 @@ dokku nginx:set node-js-app trust-x-forwarded-for true
 dokku nginx:set node-js-app trust-x-forwarded-for false
 ```
 
+Only use this option if:
+1. All requests are terminated at the load balancer, and forwarded to Nginx
+2. The load balancer is configured to send the `X-Forwarded-` headers (this may be off by default)
+
+If it's possible to make HTTP(S) requests directly to Nginx, bypassing the load balancer, or if the load balancer is not configured to set these headers, then it becomes possible for a client to set these headers to arbitrary values.
+
 ### SSL Port Exposure
 
 When your app is served from port `80` then the `/home/dokku/APP/nginx.conf` file will automatically be updated to instruct nginx to respond to ssl on port 443 as a new cert is added.  If your app uses a non-standard port (perhaps you have a dockerfile deploy exposing port `99999`) you may need to manually expose an ssl port via `dokku proxy:ports-add <APP> https:443:99999`.
--- a/docs/configuration/ssl.md
+++ b/docs/configuration/ssl.md
@@ -139,8 +139,9 @@ Your application has access to the HTTP headers `X-Forwarded-Proto`, `X-Forwarde
 If your server runs behind an HTTP(S) load balancer, then Nginx will see all requests as coming from the load balancer. If your load balancer sets the `X-Forwarded-` headers, you can tell Nginx to pass these headers from load balancer to your application via `nginx:set`:
 
 ```shell
-dokku nginx:set node-js-app trust-x-forwarded-for true
-dokku nginx:set node-js-app trust-x-forwarded-for false
+dokku nginx:set node-js-app x-forwarded-for-value "\$http_x_forwarded_for"
+dokku nginx:set node-js-app x-forwarded-port-value "\$http_x_forwarded_port"
+dokku nginx:set node-js-app x-forwarded-proto-value "\$http_x_forwarded_proto"
 ```
 
 Only use this option if:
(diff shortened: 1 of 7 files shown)


docs/configuration/ssl.md
```diff
@@ -136,44 +136,9 @@ Certain versions of nginx have bugs that prevent [HTTP/2](https://nginx.org/en/d
...
-
-  location    / {
-    proxy_pass  http://{{ .APP }};
-    proxy_http_version 1.1;
-    proxy_set_header Upgrade $http_upgrade;
-    proxy_set_header Connection $http_connection;
-    proxy_set_header Host $http_host;
-    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
-    proxy_set_header X-Forwarded-For $http_x_forwarded_for;
-    proxy_set_header X-Forwarded-Port $http_x_forwarded_port;
```
josegonzalez, line 157: Should we do anything about all the other properties here? Would it make sense to instead make these values toggleable via an `nginx:set` call?

josegonzalez:
Given your comment in the issue, seems like this shouldn't be default, and we should gate it behind an `nginx:set` flag.

docs/configuration/ssl.md
bjornpost: I don't think so. Only the value in `-For` is appended. `-Proto` and `-Port` are being replaced:

```bash
curl -H "X-Forwarded-Port: 1234" https://...

  ["HTTP_X_FORWARDED_PORT"]=>
  string(3) "443"
```

```bash
curl -H "X-Forwarded-Proto: Banana" https://...

  ["HTTP_X_FORWARDED_PROTO"]=>
  string(3) "https"
```

bjornpost:
I've refactored the PR to only set `$http_x_forwarded_for` when configured via `nginx:set`. I don't have a test env locally, so I haven't tested the code yet. Thanks for the quick response!

plugins/nginx-vhosts/functions
```diff
@@ -462,6 +462,8 @@ nginx_build_config() {
       local NGINX_BIND_ADDRESS_IP6="$(fn-plugin-property-get-default "nginx" "$APP" "bind-address-ipv6" "::")"
       [[ -z "$NGINX_BIND_ADDRESS_IP6" ]] && NGINX_BIND_ADDRESS_IP6="::"
 
+      local PROXY_TRUST_X_FORWARDED_FOR="$(fn-plugin-property-get-default "nginx" "$APP" "trust-x-forwarded-for" "false")"
```
josegonzalez, line 465: Can we add this to the nginx:report output?

tests/unit/test_helper.bash
```diff
@@ -406,7 +406,7 @@ server {
     proxy_set_header Connection \$http_connection;
     proxy_set_header Host \$http_host;
     proxy_set_header X-Forwarded-Proto \$scheme;
-    proxy_set_header X-Forwarded-For \$remote_addr;
+    proxy_set_header X-Forwarded-For \$proxy_add_x_forwarded_for;
```
josegonzalez, line 409: Should default back to `remote_addr`.

We also need a test of setting/unsetting this property to ensure it's set correctly/I dont break the option in a future refactor.

docs/configuration/ssl.md
josegonzalez: Yeah just thinking that removing this doc maybe isn't the best. Maybe we can add properties for each of these headers as well?

bjornpost: I'm not sure if I understand you fully. I will undo the removal of the paragraph "Only use this option if ..." and the paragraph after. Do you also want to undo the deletion of the custom template example? I did add an example of how to configure via `nginx:set`, which makes more sense I think.

Regarding making `-Proto` and `-Port` toggleable: I can do it, but I'm not sure if its a common use case. Also not sure between which values you should be able to toggle.

josegonzalez: If you're around on the gliderlabs slack, I'd be happy to jump on a quick call to explain what I mean. I'll be around for the next ~30 minutes.

bjornpost:
Updated the PR again based on your feedback!

josegonzalez (approved):

Haocen:
Sadly this is not working with `--global` flag.

PR #4412, opened 2021-02-17
Fix generation of crontab

markuspoerschke:
There was a missing placeholder in the template string.
That caused that `%!(EXTRA string=...` is added to the crontab file which causes commands to fail.

(I did not yet test this. If you can do it, fine, otherwise I will try to test tomorrow.)


josegonzalez:
I swear I fixed this...

PR #4450, opened 2021-03-01
Simplify docker image deploys via git:from-image

### LINKED ISSUE #4296: Implement git:from-image and deprecate tags plugin
## Description of problem

I'm using a `Dockerfile` to package my app. I can't follow the dokku/heroku workflow of `git push` and build on my server as my apps take forever to build and this kills my runtime server. 
So I'm building my docker images in CI and publishing to GitHub container registry, but before deploying the app i need to pull the docker image onto the server and tag it. 

It's somewhat awkward to achieve this as I can't use the `dokku` user to interact with `docker` remotely as it's using `sshcommand` which will forward all commands to `/usr/bin/dokku`. 

After reaching out to the dokku maintainers in slack, they suggest we do the following:

> I think the proper way would be to expose like… a git:from-image command
> So you build your image, push it to $LOCATION, and then call that command to deploy that image which would make a temporary repository (or get the existing one and clear out its contents), add a Dockerfile with a FROM $IMAGE, and trigger the deploy as normal 
> That way we avoid rebuild issues that currently happen with the tags plugin
> i think its worth implementing and deprecating the tags plugin

Notes:

- It would be useful to accommodate moving tags (like `latest`) which would require a `docker pull` before deploy.

josegonzalez:
The new command superscedes the previous tags plugin, and integrates docker image deployment with the general build process.

While `docker image load` is not supported, this otherwise completely handles all previous workflows supported by the `tags:deploy` command, while doing so in a much easier to use interface.

Closes #4296

### DIFF
--- a/.github/commands/bump-azure
+++ b/.github/commands/bump-azure
@@ -45,7 +45,7 @@ main() {
   echo "=====> Creating upstream pull request"
   ../$GH_FOLDER/bin/gh pr create --head dokku:dokku-$VERSION --repo Azure/azure-quickstart-templates --title 'Update dokku-vm dokku version to 0.23.7' --body ''
 
-  popd >/dev/null
+  popd &>/dev/null
 }
 
 main "$@"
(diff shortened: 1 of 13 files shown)


josegonzalez:
The build succeeded so I'm merging (the current build may fail because I force pushed a doc change).

Issue #4458, opened 2021-03-02
"Clone in progress" after git:sync fails during release


matijakolaric:
> If you're using Dokku - especially for commercial purposes - consider donating to project development via [OpenCollective](https://opencollective.com/dokku) or [Patreon](https://www.patreon.com/dokku). Funds go to general development, support, and infrastructure costs.
>
> If you'd like to sponsor specific functionality, see the project's [Sponsoring](https://github.com/dokku/.github/blob/master/SPONSORING.md) document.
>
> If you need help figuring out how to use a specific buildpack, are seeing an issue in during the buildpack process, or are having issues using multiple buildpacks, please see our [slack channels or github discussion board](https://dokku.com/docs/getting-started/where-to-get-help/#monitored-locations). Issues pertaining to buildpacks may be closed and locked.
>
> If you need support for a version of Dokku that is more than a year old, your issue may be closed without an answer. Please upgrade to a recent version.

## Description of problem

If git:sync fails during release, the next time it is run, there is an error "Clone in progress".

### Steps to Reproduce

1. Add an error-creating command to release: in Procfile
2. run git:sync --build
3. run git:sync --build  again

#### Actual Results

error "Clone in progress"
after deleting /var/lib/dokku/data/git/<app>/ folder, all is fine

#### Expected Results

git:sync to start over without a need to manually delete a folder.

Removed the rest, as this is not related to environment.

dokku version 0.24.0

josegonzalez:
Did the first `git:sync` fail, or did you ctrl+c it?

I think we can add a command similar to `apps:unlock` to allow users to remove the clone. Maybe `git:unlock-clone`?

matijakolaric:
It failed. (I dont excpect an app to handle CTRL-C nicely.)
It is a Django app, and I tested a bit, it happens when it fails on collect static (no folder), database migrations (wrong password) or commands in Procfile under release.

I must say that the error messageis a bit confusing, as there is nothing going on. I think "lock" should be somewhere in that message. "Clone in progress or failed, lock present, see git:unlock." I think git:unlock would be best in that case. It only took some 2 minutes to figure out what to delete, but then, not everyone may be as fast.

It would be nice if the error could be handled gracefully, but I have no clue if bash has try/catch.

Akirtovskis:
Facing the same issue and even if first build succeeds, second one is always failing with "Clone in progress" error. 

I am also fairly sure, that this didn't happen on the previous version.

josegonzalez:
[...]

PR #4465, opened 2021-03-07
Add git:unlock command

Akirtovskis:
Related issue : https://github.com/dokku/dokku/issues/4458 

To solve the cloning issue : 
- added git:unlock command
- called within git:sync so on every next build clone directory is removed and build continues 
- added error handling if no `tty` is present (?) 
- removed cloned directory check from `fn-git-clone` as we now have check for that in `git:sync`

Questions that I still have : 
- when trying to run `git:unlock app` separately from vm it is not working and I see `git:unlock` is not a dokku command. Is 
there something obviuos I missed in command creation flow? 
- is this good error message for `no tty` cases? 
**dokku_log_fail "Missing tty, try running ssh -t which will allocate new tty for you"**


josegonzalez:
The unlock shell script needs to be executable.

I would just error with a message telling the user to call git:unlock if no tty is present.

Akirtovskis:
@josegonzalez Thanks for feedbacks. Fixed both issues and successfully tested `git:unlock` functionality on local.

plugins/git/internal-functions
```diff
@@ -193,6 +193,20 @@ cmd-git-sync() {
 
   verify_app_name "$APP"
 
+  if [[ -d "$DOKKU_LIB_ROOT/data/git/$APP" ]]; then
+    if has_tty eq 0; then
+      cmd-git-unlock "$APP"
```
josegonzalez, line 198: We should _ask_ for confirmation before attempting to delete anything (thats the point of me asking you to check for a tty lol).

plugins/git/internal-functions
```diff
@@ -193,6 +193,20 @@ cmd-git-sync() {
...
+      if [[ -d "$DOKKU_LIB_ROOT/data/git/$APP" ]]; then
+        dokku_log_fail "Failed to delete existing clone folder"
+        exit 15
+      fi
+    else
+      dokku_log_fail "Missing tty, please run git:unlock command"
```
josegonzalez, line 204: Should just be "Run 'git:unlock' to remove the existing temporary clone"

plugins/git/internal-functions
```diff
@@ -193,6 +193,20 @@ cmd-git-sync() {
...
+    fi
+  else
+    dokku_log_info1 "No need to unlock app, carrying on with build"
```
josegonzalez, line 207: No need for this else block.

plugins/git/internal-functions
```diff
@@ -215,6 +229,26 @@ cmd-git-sync() {
...
+  if [[ -d "$APP_CLONE_ROOT" ]]; then
+    dokku_log_info1 "About to delete $APP_CLONE_ROOT"
+    chown dokku:dokku -R "$APP_CLONE_ROOT"
+    chmod 755 "$APP_CLONE_ROOT"
+    rm -rf "$APP_CLONE_ROOT"
+    dokku_log_info1 "Successfully unlocked $APP"
+  else
+    dokku_log_info1 "No clone folder exists app already unlocked"
+  fi
+
```
josegonzalez, line 249: Remove the extra newline please.

plugins/git/internal-functions
```diff
@@ -215,6 +229,26 @@ cmd-git-sync() {
...
+  local cmd="git:unlock"
+  [[ "$1" == "$cmd" ]] && shift 1
+  declare APP="$1"
+
+  local APP_CLONE_ROOT="$DOKKU_LIB_ROOT/data/git/$APP"
+
+  if [[ -d "$APP_CLONE_ROOT" ]]; then
+    dokku_log_info1 "About to delete $APP_CLONE_ROOT"
+    chown dokku:dokku -R "$APP_CLONE_ROOT"
+    chmod 755 "$APP_CLONE_ROOT"
```
josegonzalez, line 243: Do we need the chown/chmod?

josegonzalez (commented):
Can you also write a test? Should just mock creating the `clone` directory.

The test might fail on gitlab because they don't support tty, so no need to check for the inline unlock, just want to know that the regular `git:unlock` command does what we expect.

plugins/git/internal-functions
Akirtovskis: Ok, I will try to make this work and same for tests.

plugins/git/internal-functions
Akirtovskis: Double checking this now. I added commands because at some point was facing `permission-denied` errors.

Akirtovskis: We do not, removed

Akirtovskis:
@josegonzalez add the prompt and also --force flag for cases where it would make sense to bypass the prompt. 

Added tests also, but when I was running them they are failing with : 
```
(git) git:unlock [succes]
   (from function `global_setup' in file test_helper.bash, line 23,
    from function `setup' in test file git_6.bats, line 6)
     `global_setup' failed with status 127
   /Users/arturs/arturs/dev/dokku/tests/unit/test_helper.bash: line 23: free: command not found
   /Users/arturs/arturs/dev/dokku/tests/unit/test_helper.bash: line 216: dokku: command not found
   /Users/arturs/arturs/dev/dokku/tests/unit/test_helper.bash: line 37: dokku: command not found
```

Same is happening on other git test files as well and I am not sure how to fix that.

josegonzalez:
Tests cant be run on OSX, they need to run in a Linux VM (the included one is fine).

Akirtovskis:
@josegonzalez `git:unlock` command works as expected now and I even managed to make it cleaner by moving out the `fn-git-remove-clone-folder` to a separate function. 

But I would still appreciate some help with tests and fixing the lint issue.

josegonzalez:
Gonna take over the PR from here.

PR #4476, opened 2021-03-13
Update uninstalling.md

nunoh:
my linux fu is not great, but seems like there's a `/` missing

[ci skip]

### DIFF
--- a/docs/getting-started/uninstalling.md
+++ b/docs/getting-started/uninstalling.md
@@ -49,6 +49,6 @@ The following user/group must be deleted:
 
 The following directories must be deleted:
 
-- `~dokku`
+- `~/dokku`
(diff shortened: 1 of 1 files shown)


haliliceylan:
best PR ever xD

josegonzalez:
No the docs are actually correct:

```
root@dokku:~# echo ~dokku
/home/dokku
root@dokku:~# echo ~/dokku
/root/dokku
```

It's weird but if you don't have a `slash`, the suffix to `~` is treated as a username.

Thanks for the submission though, feel free to poke around for anything else that seems off :)

nunoh:
oh, that makes sense
just learned another thing today 
will do! thanks for such great project btw

Issue #4482, opened 2021-03-15
Support elementary OS


djyde:
I ran `bootstrap.sh` in elementary OS (https://elementary.io/) which is an OS based on Ubuntu 18.04. But the command failed with `Unsupported Linux distribution`

I read the source code and I figured out it is because the output of `/etc/os-release` on elementary OS is:

```bash
NAME="elementary OS"
VERSION="5.1.6 Hera"
ID=elementary
ID_LIKE=ubuntu
PRETTY_NAME="elementary OS 5.1.6 Hera"
LOGO=distributor-logo
VERSION_ID="5.1.6"
HOME_URL="https://elementary.io/"
SUPPORT_URL="https://elementary.io/support"
BUG_REPORT_URL="https://github.com/elementary/os/issues/new"
PRIVACY_POLICY_URL="https://elementary.io/privacy-policy"
VERSION_CODENAME=hera
UBUNTU_CODENAME=bionic
```

the `VERSION_ID` is pointed to the elementory OS version, instead of Ubuntu version, which is not in `SUPPORTED_VERSIONS`.

https://github.com/dokku/dokku/blob/4662ac2774edf8c66fa5668edb64b08abea50a7c/bootstrap.sh#L245

BTW, this is the output of `lsb_release -a -u` in elementary OS:

```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.4 LTS
Release:	18.04
Codename:	bionic
```

josegonzalez:
How does elementary packaging work? Is this something you'd be interested in working on?

djyde:
@josegonzalez Same as Ubuntu (apt-get). I tried hard code: 

```
if  [[ $DOKKU_DISTRO == "elementary" && $DOKKU_DISTRO_VERSION == "5.1.6"  ]]; then
	export DOKKU_DISTRO="ubuntu"
    export DOKKU_DISTRO_VERSION="18.04"
fi
```

and it works. But I'm not sure if it's a good solution.

josegonzalez:
Okay seems like we can maybe instead just release our deb packages for Elementary OS.

The alternative is running the docker image as is, but I don't see any harm in this :)

josegonzalez:
Going to close this. The recommended alternative is to run the Docker image.

lmdc45:
If anyone is doing the same with Elementary 6 odin, need to change bootstrap.sh with:
at line 265 in main function
```
# see https://github.com/dokku/dokku/issues/4482
[...]

PR #4483, opened 2021-03-15
Support elementary OS

djyde:
#4482


josegonzalez:
Does ElementaryOS just use Debian packages? If so, do our debs work? In such a case, we can modify [the release process](https://github.com/dokku/dokku/blob/master/contrib/release-dokku#L129) to push to our packagecloud repository under ElementaryOS (for Dokku and everything under the Dokku umbrella). That might be better than doing a hardcode here :)

josegonzalez:
Going to close this. I think the correct way forward for unofficially supported operating systems is to use the Docker image. It should support almost every workflow, with the advantage of being easy to upgrade/downgrade (just stop the existing container and start a new one).

PR #4484, opened 2021-03-15
Bumps vector image version

### LINKED ISSUE #4480: Upgrade Vector image for log shipping to 0.12.x
This will allow using aws auth parameters, and probably other fixes.

The fix is a matter of changing https://github.com/dokku/dokku/blob/4662ac2774edf8c66fa5668edb64b08abea50a7c/plugins/logs/logs.go#L27

lunswor:
Bumps version of Vector. Resolves #4480

### DIFF
--- a/plugins/logs/logs.go
+++ b/plugins/logs/logs.go
@@ -24,7 +24,7 @@ var (
 )
 
 // VectorImage contains the default vector image to run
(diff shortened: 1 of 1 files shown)


josegonzalez:
We should probably also make `encoding.codec` required, as per https://vector.dev/highlights/2021-02-16-0-12-upgrade-guide/ . Any chance you can work on that?

PR #4493, opened 2021-03-19
Ensure alternate-tags is properly injected for git:from-image deploys

Cellane:
This PR will (hopefully) add a test that indicates a failure in image labeling using `docker-image-labeler` during `git:from-image` deployment.


Cellane:
Thank you! 🙇‍♂️

josegonzalez:
Okay I figured it out. For `git:from-image`, we add custom labels via `--label` on the `docker build`, so the image actually _isn't_ the one you specify.

That said, we can add the `LABEL` for the docker-image-labeler, which I think is fine for our use case.

PR #4497, opened 2021-03-21
Migrate documentation to mkdocs

mom1:
Hi, I want to suggest a documentation approach using [mkdocs](https://www.mkdocs.org/).

- Added search for documentation
- Added syntax highlighting
- Fixed some bugs
- Footnotes now links
- Link to go to page editing
- New style for `info`, `warnings`, `New as xxx`

To work on the documentation locally:

`pip install -r docs/requirements.txt`
`mkdocs serve`
after that you can open `http://127.0.0.1:8000/`

To build the site you can use the command `mkdocs build`
After that, the site will be built in the `site` folder
This can be changed in the config

TODO:
  - I was unable to replicate the home page of the original site
  - I can add versioning, but it seems to me not necessary, because there are marks like `New as of 0.5.0`
  - Need more beautiful highlighting of terminal output
  - Need improve environment-variables table

Here are some examples:

Main page of documents:
image

Search:
image

Admonitions:
image

If you are interested, I can continue to work on it or pass it on to you.

[ci skip]


josegonzalez:
I gotta say that I'm not a huge fan for a few reasons:

- The url changes break browsing on github. Some people browse that way vs the website for whatever reason, and I'd like to support that.
- The new blockquote pullouts are weird. I realize markdown doesn't have a great way of doing these, but the method used isn't very common anywhere.
- Versioning support is actually quite useful, as the underlying docs do change as the behavior changes (and sometimes features disappear completely).
- Custom homepage is super desirable.

Overall I think if we didn't have anything, this would be a huge improvement. That said, outside of doc search (which is obviously beneficial) everything else here feels as though it could be added on to our existing docs.

I don't want to detract from your effort - since its overall quite good - but any chance you'd be interested instead of helping shore up the existing docs to better align with your goals?

josegonzalez:
Going to close this as there hasn't been any followup. Thanks for the initial exploration.

josegonzalez:
@mom1 just wanted to let you know that I revisited this effort in #5344. I took a lot of inspiration from your work, and I appreciate the time you put into this!

mom1:
@josegonzalez I am glad to hear it.  Thank you for a useful tool and now convenient documentation for it. 😂

PR #4509, opened 2021-03-22
Add ability to specify initial network

### LINKED ISSUE #4477: Add ability to set initial network via network:set
At the moment, we default to using the bridge network. We should instead allow an `initial-network` property that does the same thing as injecting `--network=$NETWORK`.

- ~Migrate existing `--network=$NETWORK` calls to the property~ lets let this be a conscious decision
- Inject that property during `docker-args-*`
- ~Disallow setting the flag via `docker-options`~ lets let this be a conscious decision

josegonzalez:
Todo:

- [x] Enable global and computed values for the other network settings
- [x] Add tests
- [x] Add a note to the migration guide for migrating to this property

Closes #4477


hiepxanh:
https://github.com/dokku/dokku/issues/4633
hello @josegonzalez, hope you have a good day, I just trying to recovery from backup, it look like missing the network data. is that  realate to this PR? 
I think if your PR completed. dokku is able to set initial-network after recovery from backup?

PR #4517, opened 2021-03-23
Update README with correct link to docs

illgitthat:
Changed https://dokku.com/docs/ to → https://dokku.com/docs/getting-started/installation/

The current link https://dokku.com/docs/ just seems to load dokku.com but with some broken image urls
Example: img tag links to `https://dokku.com/docs/assets/docker.png` instead of `https://dokku.com/assets/docker.png`


illgitthat:
Sorry, I saw to add [ci skip] only after I had committed and not sure how to revert this now.

josegonzalez:
The initial url should probably be just `https://dokku.com` instead of what it is. That said, landing directly in the docs might be a better idea...

PR #4531, opened 2021-03-28
Update config command to config:show

elamje:
Dokku now gives `config` deprecation message, and suggests to use `config:show`
[ci skip]


josegonzalez:
Thanks for the pull request!

PR #4548, opened 2021-04-11
Remove the remaining traces of dokku-update

halilim:
Re: #4527 [ci skip]
I guess [dokku-update](https://github.com/dokku/dokku-update) can be archived as well?


josegonzalez:
We actually meant to move the codebase to it's own repo and fix it up a bit. The only bit left is the release change.

Thanks for the PR though :)

PR #4563, opened 2021-04-14
Tough netrc file when setting git:auth entries

Akirtovskis:

Creating .netrc file before setting gith auth values

Removing .netrc file after removing gith auth values

### DIFF
--- a/plugins/git/internal-functions
+++ b/plugins/git/internal-functions
@@ -112,7 +112,6 @@ cmd-git-auth() {
   if [[ -z "$USERNAME" ]]; then
     dokku_log_info1 "Removing netrc auth entry for host $HOST"
     netrc unset "$HOST"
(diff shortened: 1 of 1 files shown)


plugins/git/internal-functions
```diff
@@ -112,10 +112,12 @@ cmd-git-auth() {
   if [[ -z "$USERNAME" ]]; then
     dokku_log_info1 "Removing netrc auth entry for host $HOST"
     netrc unset "$HOST"
+    rm "${DOKKU_ROOT}/.netrc"
```
josegonzalez, line 115: We probably don't need the remove?

Akirtovskis: Yes, I guess it is fine to leave the .netrc file once it has been created.

PR #4566, opened 2021-04-15
Normalize Dockerfile line terminators to handle issues in port extraction

jasiek:


plugins/builder-dockerfile/builder-build
```diff
@@ -20,6 +20,7 @@ trigger-builder-dockerfile-builder-build() {
   pushd "$SOURCECODE_WORK_DIR" &>/dev/null
 
   # extract first port from Dockerfile
+  dos2unix Dockerfile
```
josegonzalez, line 23: This will just write it to stdout, so I think you want something like:

```
    TMP_DOCKERFILE=$(mktemp "/tmp/dokku-${DOKKU_PID}-${FUNCNAME[0]}.XXXXXX")
    trap "rm -rf '$TMP_DOCKERFILE' >/dev/null 2>&1 || true" RETURN
    dos2unix -l <"Dockerfile" >"$TMP_FILE_COMMAND_OUTPUT"
    mv "$TMP_FILE_COMMAND_OUTPUT" Dockerfile


```

jasiek: It will change the file in-place:

```
➜  dokku git:(master) ✗ unix2dos LICENSE
unix2dos: converting file LICENSE to DOS format...
➜  dokku git:(master) ✗ file LICENSE
LICENSE: ASCII text, with very long lines, with CRLF line terminators
➜  dokku git:(master) ✗ dos2unix LICENSE
dos2unix: converting file LICENSE to Unix format...
➜  dokku git:(master) ✗ file LICENSE
LICENSE: ASCII text, with very long lines
```

josegonzalez: Is that the case on linux too? Just verifying.

josegonzalez:
I'm assuming you've tested this change locally and it works for your repo, correct?

plugins/builder-dockerfile/builder-build
jasiek: Yes, this is in the vagrant VM for this project. Maybe I ought to add an automated test for this?

josegonzalez: That would be great :)

josegonzalez: Honestly we should have a better way of extracting the exposed ports from a dockerfile by parsing the AST versus doing grep, but here we are :)

Issue #4380, opened 2021-02-04
Tags are deleted after build using dockerfile deploy


Kjwon15:
> If you're using Dokku - especially for commercial purposes - consider donating to project development via [OpenCollective](https://opencollective.com/dokku) or [Patreon](https://www.patreon.com/dokku). Funds go to general development, support, and infrastructure costs.
>
> If you'd like to sponsor specific functionality, see the project's [Sponsoring](https://github.com/dokku/.github/blob/master/SPONSORING.md) document.
>
> If you need help figuring out how to use a specific buildpack, are seeing an issue in during the buildpack process, or are having issues using multiple buildpacks, please see our [slack channels or github discussion board](http://dokku.viewdocs.io/dokku/getting-started/where-to-get-help/#monitored-locations). Issues pertaining to buildpacks may be closed and locked.
>
> If you need support for a version of Dokku that is more than a year old, your issue may be closed without an answer. Please upgrade to a recent version.

## Description of problem
Saved tags are deleted after new build using Dockerfile build
### Steps to Reproduce
using app named testapp

1. dokku tags:save testapp v1.0.0
2. dokku ps:rebuild testapp
3. dokku tags testapp

#### Actual Results

There is only `latest` tag

#### Expected Results
There are `v1.0.0` tag and `latest`

---
I am using dokku v0.23.0

Kjwon15:
Also after successful deploy, old containers are not removed

josegonzalez:
Can you post an image I can replicate this with? Old containers should def be removed, as we're using the normal deployment flow for these images...

josegonzalez:
Closing this. There is a test for this behavior [here](https://github.com/dokku/dokku/blob/757a82913749c64e77b83aaeae7038684d9d0fbf/tests/unit/tags.bats#L101) and I am unable to reproduce this on the latest Dokku.

PR #4596, opened 2021-05-09
Do not remove unused image to hold tags

tribela:
Related: #4380

Dokku was removing all unused images and that causes removed tagged images.
This PR will fix that problem.

### DIFF
--- a/plugins/common/functions
+++ b/plugins/common/functions
@@ -755,7 +755,7 @@ docker_cleanup() {
     "$DOCKER_BIN" image rm $("$DOCKER_BIN" image list --filter 'dangling=true' --filter "label=com.dokku.app-name=$APP" --quiet) &>/dev/null || true
 
     # delete unused images
-    "$DOCKER_BIN" image prune --all --filter "label=com.dokku.app-name=$APP" --force &>/dev/null || true
+    "$DOCKER_BIN" image prune --filter "label=com.dokku.app-name=$APP" --force &>/dev/null || true
   else
     # delete all non-running containers
     # shellcheck disable=SC2046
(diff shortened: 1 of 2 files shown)


josegonzalez:
```
% docker image prune -h
Flag shorthand -h has been deprecated, please use --help

Usage:  docker image prune [OPTIONS]

Remove unused images

Options:
  -a, --all             Remove all unused images, not just dangling ones
      --filter filter   Provide filter values (e.g. 'until=<timestamp>')
  -f, --force           Do not prompt for confirmation
```

Are you sure this is correct? The help output says its only unused images...

tribela:
That "unused" means "currently not running images".
tags should not be removed even they are not running right now

josegonzalez:
I don't think we should do this, as it will lead to folks having a ton of used up space and complaints that Dokku is leaving unused images around. Since the image is tagged with dokku's internal tag, I think we should assume Dokku is handling it's lifecycle.

Instead, it might be worth moving your workflow to the new `git:from-image` command?

tribela:
Oh I didn't know that git:tags was deprecated and `git:from-image` command.
I want to move to that `from-image` deploy Cause I build the docker image separately with CI.
Can you tell me how the `git:from-image` triggers to update(pull) to new image?

josegonzalez:
Docs here: https://dokku.com/docs/deployment/methods/git/#initializing-an-app-repository-from-a-docker-image

You will need to push the image to a registry and login as the dokku user.

tribela:
So if I want to deploy new image, execute `git:from-image` command after pushing into docker registry?

josegonzalez:
Yes.

PR #4622, opened 2021-05-27
Fix typo in nginx logging configuration example

allanitis:
Remove extra space in nginx logging config example.

This causes ingestion problems on AWS Cloudwatch and brings the attribute in line with the other attributes in the example. 


josegonzalez:
Thanks for the pull request!

PR #4701, opened 2021-07-27
Revamp dokku run command

josegonzalez:
- Deprecate `--detach`, `--rm` and `--rm-container` flags
- Add the `--no-tty` flag
- Add the `run:detached` and `run:list` commands
- Default to removing exited run containers
- Label cron containers differently


hgeorgilas:
@josegonzalez I think the documentation available here: https://dokku.com/docs/appendices/0.25.0-migration-guide/ conflicts with the actual behaviour of the `$ dokku run <app-name> COMMAND` command as it is described here: https://dokku.com/docs/processes/one-off-tasks/. 

In the migration guide is stated that: 
> The dokku run command now always removes the ephemeral container on exit. 

But the actual documentation describes that the container will not be automatically terminated in `detach` mode. I tried this with dokku `0.26.6`
> Running a process in detached mode will immediately return a CONTAINER_ID. It is up to the user to then further manage this container in whatever manner they see fit, as Dokku will not automatically terminate the container.

josegonzalez:
Ah yeah the docs are a bit wrong. We always remove the container now. I fixed the docs to reflect this.

hgeorgilas:
@josegonzalez Thanks for the swift response! Is there a way to keep the container around? I'm generating reports inside the container and want to move them outside after my task finishes, destroying the container after doing so does not allow me to export my report. Do you think ti would be good idea to introduce a `--keep` or equivalent argument that will allow for containers to stay around until the user explicitly removes them?

Thanks

josegonzalez:
Yeah that seems reasonable to me @hgeorgilas. If you want to file a ticket for that, I can take a look at implementing it in a future release (pull requests welcome of course!).

PR #4723, opened 2021-08-06
Clarify buildkit instructions

josegonzalez:
There is no need for a reboot

### DIFF
--- a/docs/deployment/builders/dockerfiles.md
+++ b/docs/deployment/builders/dockerfiles.md
@@ -145,7 +145,16 @@ RUN echo $NODE_ENV
 
 ### Building images with Docker Buildkit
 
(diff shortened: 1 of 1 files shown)


FinnWoelm:
Hi @josegonzalez,

amazing work on Dokku! ❤️

I don't think the new instructions work for enabling buildkit. I just tried both ways (`dokku config:set` both globally and for my app specifically and also via the `/etc/default/dokku` file).

The issue is that those env vars do not seem to be available/set when dokku runs the git `pre-receive` hook. So the build fails and the git push is rejected. This is by design, I believe?

image

I only got it working after modifying `/etc/environment` from `root` user and rebooting, as the instructions previously indicated. 

Should this commit be reverted? Or am I doing something wrong?

josegonzalez:
Potentially the config:set method might not work, but we source the `/etc/default/dokku` file [here](https://github.com/dokku/dokku/blob/master/dokku#L6-L10).

Maybe that should be:

```
echo "export DOCKER_BUILDKIT=1" | sudo tee -a /etc/default/dokku
echo "export BUILDKIT_PROGRESS=plain" | sudo tee -a /etc/default/dokku
```

Instead?

FinnWoelm:
That was it 👍 Amazing! 🙌

josegonzalez:
@FinnWoelm mind making a PR to our docs to clarify things?

FinnWoelm:
On it already :)

PR #4742, opened 2021-08-19
bootstrap: add support for Debian 11 (bullseye)

### LINKED ISSUE #4740: No install package for debian/bullseye
## Description of problem
https://packagecloud.io/dokku/dokku is missing debian/bullseye

### Steps to Reproduce

1. Change `/etc/apt/sources.list.d/dokku.list` from buster to bullseye
2. `apt update`

#### Actual Results

```
Err:9 https://packagecloud.io/dokku/dokku/debian bullseye Release                                  
  404  Not Found [IP: 54.193.9.65 443]
E: The repository 'https://packagecloud.io/dokku/dokku/debian bullseye Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

#### Expected Results

`apt update` completes successfully

### Results when running bootstrap.sh on bullseye

```
Preparing to install v0.25.1 from https://github.com/dokku/dokku.git...
--> Ensuring we have the proper dependencies
--> Note: Installing dokku for the first time will result in removal of
    files in the nginx 'sites-enabled' directory. Please manually
    restore any files that may be removed after the installation and
    web setup is complete.

    Installation will continue in 10 seconds.
Unsupported Linux distribution. Only the following versions are supported: Debian [9, 10], CentOS [7], Fedora (partial) [33, 34], Ubuntu [18.04, 20.04]
```

ashkulz:
fixes #4740

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -12,7 +12,7 @@ set -eo pipefail
 # That's good because it prevents our output overlapping with wget's.
 # It also means that we can't run a partially downloaded script.
 
-SUPPORTED_VERSIONS="Debian [9, 10], CentOS [7], Fedora (partial) [33, 34], Ubuntu [18.04, 20.04]"
+SUPPORTED_VERSIONS="Debian [9, 10, 11], CentOS [7], Fedora (partial) [33, 34], Ubuntu [18.04, 20.04]"
 
 log-fail() {
   declare desc="log fail formatter"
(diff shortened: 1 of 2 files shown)


bootstrap.sh
```diff
@@ -200,9 +200,9 @@ install-dokku-from-deb-package() {
       OS_ID="bionic"
     fi
   elif [[ "$DOKKU_DISTRO" == "debian" ]]; then
-    OS_IDS=("stretch" "buster")
+    OS_IDS=("stretch" "buster" "bullseye")
```
ashkulz, line 203: Note that this requires changes to be made to `packagecloud.io`, else should be held back till that is done.

josegonzalez: Mind also updating this: https://github.com/dokku/dokku/blob/ca680c3aebfd674c0cf52806122708ba3b8ef45d/contrib/release-dokku#L150

ashkulz: Done :+1:

ashkulz:
@josegonzalez anything more needed from my side?

PR #4743, opened 2021-08-19
Update LICENSE 2021

masguit42:


josegonzalez:
Is that necessary?

josegonzalez:
Looking at [this](https://opensource.stackexchange.com/a/6393), I don't think updating this line is necessary. Going to close for now unless someone gets antsy and pays for a lawyer that says we should update it (and to what!).

PR #4808, opened 2021-09-29
Added a comment about Dockerfile deployment

francipvb:
I found issues while trying to deploy an app from a dockerfile and I found that the expose instruction causes problems with the dokku proxy. I found the solution and I am contributing it.

### DIFF
--- a/docs/getting-started/troubleshooting.md
+++ b/docs/getting-started/troubleshooting.md
@@ -64,7 +64,24 @@ Save the file and try stopping nginx and starting it again:
 ## * Starting nginx nginx                                        [ OK ]
 ```
 
-If you deploy using a *Dockerfile*, this may result in issues at deployment time. The `EXPOSE` instruction may cause your app to be exposed in an incorrect port. Try removing the `EXPOSE` instruction from your dockerfile.
+---
+
+Using the `EXPOSE` directive in a Dockerfile may be another reason why you might see the default site. When the `EXPOSE` directive is in use and a proxy plugin is enabled (the default), the proxy plugin will listen to requests on the ports specified in the `EXPOSE` stanza.
+
+For example, if you have an `EXPOSE` directive like so:
+
+```Dockerfile
+EXPOSE 8000
+```
+
+The proxy port mapping will be `http:8000:8000`.
+
+To avoid this issue, either of the following can be done:
+
+- Remove `EXPOSE` directive: This will require respecting the `$PORT` environment variable (automatically set by Dokku). Once that change is deployed, the port mapping should be cleared via the `dokku proxy:ports-clear $APP` command (where `$APP` is your app name).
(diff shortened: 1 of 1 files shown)


josegonzalez:
When you say it is exposed on the incorrect port, what did you find it was exposed on?

francipvb:
Hello,

The default HTTP port is 80 and if you specify in the dockerfile that you expose the port 8888 you'll have to set up the proxy manually. That means the default nginx page is displayed.

However, if you don't add the `EXPOSE` instruction the proxy is set up correctly and the application is presented as expected.

josegonzalez:
Do you have an example Dockerfile app that doesn't work as expected and the change that you performed to make it work as expected?

francipvb:
Hello,

See this dockerfile:

```
FROM python:3.9 AS base

ARG DJANGO_SETTINGS_MODULE

WORKDIR /app

RUN pip3 --no-cache-dir install poetry \
    && poetry config virtualenvs.create false

COPY pyproject.toml poetry.lock ./
RUN poetry install --no-dev --no-root

COPY . ./

COPY ./docker/entrypoint.sh /
RUN chmod +x /entrypoint.sh

ENV DJANGO_SETTINGS_MODULE="vectorvgt.settings.runtime"

RUN python manage.py collectstatic --no-input
RUN chmod +x ./docker/post_deploy.sh

EXPOSE 8000
ENTRYPOINT [ "/entrypoint.sh" ]
CMD daphne -b 0.0.0.0 vectorvgt.asgi:application
```

And here is the procfile:

```
web: daphne -b 0.0.0.0 -p $PORT vectorvgt.asgi:application
release: python manage.py migrate
worker: celery --application vectorvgt.celery:app worker --events
beat: celery --application vectorvgt.celery:app beat --events
```

This app won't be presented in the default HTTP port. Insthead the default nginx page will be presented.

To fix this issue, I just removed the `EXPOSE` instruction.

josegonzalez:
Right, when you set an `EXPOSE` directive and have the nginx proxy enabled, we use that port as the mapping. So nginx listens on port `8000` and routes that to the container. The port mapping is thus `http:8000:8000` (`scheme:host-port:container-port`).

That is documented [here](https://dokku.com/docs/networking/port-management/#usage) under port management (and that section is linked to from [here](https://dokku.com/docs/deployment/builders/dockerfiles/#exposed-ports).

Unfortunately, nginx does fallback routing, so if there is already a site set (which is usually something like `/etc/nginx/sites-enabled/00-default`), then it'll route to that if there is no matching vhost:port server block for the request.

I think your doc link might need a bit more explanation as to why not using an EXPOSE might help. I'll update the PR.

francipvb:
I see.

What happen if you push changes to the app?

josegonzalez:
I think the doc update I wrote clarifies things. Can you check the diff?

francipvb:
Thank you, seems better than mine.

josegonzalez:
Awesome, I'll merge it then.

PR #4950, opened 2021-11-25
Use exported environment vars during pack build

emdienn:
Hey all, in using the builder-pack packer, I came across the scenario where my apps' config variables weren't being drawn into the build process like the documentation intimated it would/should.

Digging into it, I discovered that in the builder-pack build script, the `config_export` is called and stored in a file, but then that file is never touched again. 

Figuring this was an oversight, I made a fix and would like to share it back into the codebase.

### Why did I use `sed` on the "pretty" format instead of just use the "envfile" format?

Great question.

The tl:dr; is that the "envfile" format provided by `config` is actually incorrect: it surrounds the value of the variable in single quotes, ie.
```
APPLICATION_ID='49b2195e-da01-5f1c-b1c8-61455a695d93'
```
causing the value to _include_ the quotes, because environment variables _should*_ be specified as
```
APPLICATION_ID=49b2195e-da01-5f1c-b1c8-61455a695d93
```

After digging through the `config` subcommand codebase, I determined that this is being done by a dependency package (namely `github.com/joho/godotenv`) - rather than start wrangling the format (which might have implications elsewhere) I opted for a more localised solution.

_*I couldn't find an authoritative source for the exact semantics of a dotenv file, but every other platform and system I have ever seen or used does not wrap the value in quotes. I'm happy to be wrong, but `pack` requires no-quote syntax regardless._

---

Anyway. Love dokku, hope you find this PR helpful.


plugins/builder-pack/builder-build
```diff
@@ -31,11 +31,12 @@ trigger-builder-pack-builder-build() {
   local TMP_FILE=$(mktemp "/tmp/dokku-${DOKKU_PID}-${FUNCNAME[0]}.XXXXXX")
   trap "rm -rf '$TMP_FILE' >/dev/null" RETURN INT TERM
 
-  config_export app "$APP" --format envfile --merged >"$TMP_FILE"
+  config_export app "$APP" --format pretty --merged >"$TMP_FILE"
```
josegonzalez, line 34: Could you change this to an eval like [here](https://github.com/dokku/dokku/blob/master/plugins/builder-herokuish/builder-build#L32) (probably with --merged of course to get globals, which makes me think we're missing merged for herokuish...) and then fetch the extra build args [like so](https://github.com/dokku/dokku/blob/0b3454412542c510c482eb6bbb42dc3e0c1ab579/plugins/config/docker-args-deploy#L15)? The `docker-args-keys` format is the same between `pack build` and `docker-build`. This should remove the need for the extra sed call and also properly handle newlines imo.

Well, at least I'm hoping `pack build` supports `--env KEY` the same way `docker build` does.

josegonzalez (changes requested):

plugins/builder-pack/builder-build
emdienn: Makes total sense to keep the approaches consistent. I've had a shot at it, and annoyingly, there's _just enough_ discrepancy between the docker args format and the pack args format that I don't think there's going to be a way to avoid string manipulation of some sort somewhere.

Using
```sh
ENV_ARGS="$(config_export app "$APP" --format docker-args-keys --merged)"
eval "$(config_export app "$APP" --merged)"
pack build "$IMAGE" --builder "$DOKKU_CNB_BUILDER" --path "$SOURCECODE_WORK_DIR" \
--default-process web "$ENV_ARGS"
```

The result is:
```
# assuming FOO=foo and BAR=bar in ENV
pack build my-image:version --builder my-builder --path /src --default-process web --env=FOO --env=BAR
```

Called like this, pack will build, but it ignores those `env` declarations because it expects a space, not an equals sign for its arguments :disappointed: 

I reworked my solution based on your suggestions, latest commit still does some string manipulation :slightly_frowning_face: but I released the dependency on `sed`, at least.

plugins/builder-pack/builder-build
```diff
@@ -28,15 +28,17 @@ trigger-builder-pack-builder-build() {
...
+  while IFS= read -r line; do
+    ENV_ARGS="$ENV_ARGS--env ${line%=*} "
+  done <<< "$ENV_ARGS_VALUES"
+
+  eval "$(config_export app "$APP" --merged)"
 
   plugn trigger pre-build-pack "$APP" "$SOURCECODE_WORK_DIR"
 
-  pack build "$IMAGE" --builder "$DOKKU_CNB_BUILDER" --path "$SOURCECODE_WORK_DIR" --default-process web --env-file "$TMP_FILE"
+  eval "pack build "$IMAGE" --builder "$DOKKU_CNB_BUILDER" --path "$SOURCECODE_WORK_DIR" --default-process web "$ENV_ARGS""
```
emdienn, line 41: I don't know why (bash isn't exactly my strong suit), but `eval "$(pack build...)"` _doesn't_ work

plugins/builder-pack/builder-build
josegonzalez: I refactored the MR a bit to use my idea with a new export format for pack.

plugins/builder-pack/builder-build
josegonzalez: My refactor drops the eval usage :)

PR #4975, opened 2021-12-14
add arm64 to the build

tcurdt:
The builds system is rather complex and I couldn't really test this - but I hope it should work like that?
Let me know.

### DIFF
--- a/contrib/build-dokku.Dockerfile
+++ b/contrib/build-dokku.Dockerfile
@@ -41,7 +41,7 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
 
 FROM builder as armhf
 
-COPY --from=amd64 /tmp /tmp
+COPY --from=armhf /tmp /tmp
 COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
 
 RUN rm -rf /tmp/build-dokku
@@ -61,7 +61,7 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
 
 FROM builder as arm64
 
-COPY --from=amd64 /tmp /tmp
+COPY --from=arm64 /tmp /tmp
 COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
 
 RUN rm -rf /tmp/build-dokku
--- a/contrib/build-dokku.Dockerfile
+++ b/contrib/build-dokku.Dockerfile
@@ -41,7 +41,7 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
 
 FROM builder as armhf
 
-COPY --from=armhf /tmp /tmp
+COPY --from=amd64 /tmp /tmp
 COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
 
 RUN rm -rf /tmp/build-dokku
@@ -61,7 +61,7 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
 
 FROM builder as arm64
 
-COPY --from=arm64 /tmp /tmp
+COPY --from=armhf /tmp /tmp
 COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
 
 RUN rm -rf /tmp/build-dokku
--- a/contrib/build-dokku.Dockerfile
+++ b/contrib/build-dokku.Dockerfile
@@ -24,6 +24,7 @@ COPY . ${WORKDIR}
 
 ENV GOPATH=/go
 
+
 FROM builder as amd64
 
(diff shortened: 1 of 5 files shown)


contrib/build-dokku.Dockerfile
```diff
@@ -24,6 +24,7 @@ COPY . ${WORKDIR}
 
 ENV GOPATH=/go
 
+
```
josegonzalez, line 27: Mind removing all these extra newlines you added?

contrib/build-dokku.Dockerfile
```diff
@@ -59,4 +61,26 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
   GOARCH=arm make version copyfiles \
   && DOKKU_ARCHITECTURE=armhf GOARCH=arm make deb-dokku
 
+
+FROM builder as arm64
+
+COPY --from=amd64 /tmp /tmp
```
josegonzalez, line 67: Probably should copy from `armhf` instead, as that will include the amd64 files.

contrib/build-dokku.Dockerfile
```diff
@@ -59,4 +61,26 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
...
+COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
```
josegonzalez, line 68: This can stay as amd64.

contrib/release-dokku
```diff
@@ -415,6 +417,8 @@ main() {
     git checkout -- plugins/*/plugin.toml
   fi
 
+  # what about this?
+  # echo "build/dokku_${NEXT_VERSION}_arm64.deb" >build/deb-filename
```
josegonzalez, line 421: Drop these two lines.

josegonzalez (changes requested):

contrib/build-dokku.Dockerfile
tcurdt: Sure can - I added them to make the file a little bit more readable.

contrib/build-dokku.Dockerfile
tcurdt: It's all the same, no?

tcurdt: Ups.

contrib/release-dokku
tcurdt: What's the idea behind those other two filename lines?

contrib/build-dokku.Dockerfile
tcurdt: `arm64` you mean.

contrib/release-dokku
josegonzalez: They are used by tests to extract the correct file. The tests run in amd64.

contrib/build-dokku.Dockerfile
tcurdt: Hm. Shouldn't it be
```
FROM builder as armhf

COPY --from=armhf /tmp /tmp
COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
...
FROM builder as arm64

COPY --from=arm64 /tmp /tmp
COPY --from=amd64 /usr/local/share/man/man1/dokku.1 /usr/local/share/man/man1/dokku.1-generated
```

contrib/build-dokku.Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN PLUGIN_MAKE_TARGET=${PLUGIN_MAKE_TARGET} \
 
 FROM builder as armhf
 
-COPY --from=amd64 /tmp /tmp
+COPY --from=armhf /tmp /tmp
```
josegonzalez, line 44: Please revert this. You're _in_ the `armhf` build stage. We're trying to grab the `/tmp` folder _from_ the `amd64` stage.

Maybe useful reading:

- https://www.docker.com/blog/advanced-dockerfiles-faster-builds-and-smaller-images-using-buildkit-and-multistage-builds/
- https://levelup.gitconnected.com/docker-multi-stage-builds-and-copy-from-other-images-3bf1a2b095e0

contrib/build-dokku.Dockerfile
tcurdt: OK, now I get the flow. This is incrementally adding and so chaining the folders. That was not obvious.

contrib/build-dokku.Dockerfile
tcurdt: Already did. Thanks.

tcurdt: It wasn't clear why I would want to grab a `amd64` folder in an `armhf` stage.
That would make a useful comment in the dockerfile ;)

josegonzalez (approved):

josegonzalez:
Since this still needs other dependencies, will be hard to test this without going through everything else. I don't have much time this week to do dokku-related work (maybe this weekend) so if you wanted to start on the other dependencies, that would be great. I can definitely give reviews too.

josegonzalez:
In _theory_ things will work with arm64 once this is merged (and we'll even have arm/arm64 docker images!) but we'll need to test it then.

PR #4979, opened 2021-12-23
Updated logs.md about log-rotation

yourivdlans:
While adding Vector with a file sink I was looking at the documentation where it stated no log-rotation is set up for `/var/log/dokku/apps`.

But looking at `/etc/logrotate.d/dokku` the documentation might be out-dated.

### DIFF
--- a/docs/deployment/logs.md
+++ b/docs/deployment/logs.md
@@ -115,7 +115,7 @@ The `vector` container will be started with the following volume mounts:
 - `/var/run/docker.sock:/var/run/docker.sock`
 - `/var/log/dokku/apps:/var/log/dokku/apps`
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
These are the contents for me:

```
/var/log/dokku/*.log {
        hourly
        rotate 7
        missingok
        notifempty
        su syslog dokku
        compress
        delaycompress
        postrotate
                reload rsyslog >/dev/null 2>&1 || true
        endscript
        create 664 syslog dokku
}
```

Not sure how that would match `/var/log/dokku/apps/somefile.log`?

yourivdlans:
Sorry I must've misread. I'll go setup logrotation for `/var/log/dokku/apps/*.log` manually.

Issue #4987, opened 2021-12-31
Error destroying an app if nginx is disabled.


fomojola:
## Description of problem

If the nginx service on the server running Dokku is disabled (as it is for me since I'm using an alternate proxy), attempting to destroy an application fails because the nginx-vhosts plugin (specifically https://github.com/dokku/dokku/blob/master/plugins/nginx-vhosts/functions#L131) always runs, even if the app doesn't use the nginx proxy.

### How reproducible

Always

### Steps to Reproduce

Done on an Ubuntu 20.04 LTS instance.

1. Install Dokku
2. Create an app: `dokku apps:create updates.p`
3. Disable nginx using `systemctl disable nginx`
4. Attempt to destroy the app: `dokku apps:destroy updates.p`

#### Actual Results
root@dev:/home/dokku/updates.p# dokku apps:destroy updates.p
 !     WARNING: Potentially Destructive Action
 !     This command will destroy app updates.p.
 !     To proceed, type "updates.p"
> updates.p
-----> Destroying updates.p (including all add-ons)
nginx.service is not active, cannot reload.
!     exit status 1

#### Expected Results
root@dev:/home/dokku/updates.p# dokku apps:destroy updates.p
 !     WARNING: Potentially Destructive Action
 !     This command will destroy app updates.p.
 !     To proceed, type "updates.p"
> updates.p
-----> Destroying updates.p (including all add-ons)
-----> Cleaning up...
-----> Retiring old containers and images

## Environment Information

App was destroyed before I got to dump this, but it always works.

### Additional information

The fix for this is to check if the app is using the nginx proxy and if not don't bother doing the reload. The following block fixes this for me:

    local PROXY_ENABLED="$(plugn trigger proxy-is-enabled "$APP")" 
    local PROXY_TYPE="$(plugn trigger proxy-type "$APP")"      
    if [[ "$PROXY_ENABLED" == "true" ]]; then                  
      if [[ "$PROXY_TYPE" == "nginx" ]]; then                
        fn-nginx-vhosts-nginx-init-cmd "reload"            
      fi                                                     
    fi

- Output of failing Dokku commands after running `dokku trace:on`

+ declare 'desc=destroy the properties for an app'
+ declare PLUGIN=git APP=updates.p
+ /var/lib/dokku/core-plugins/available/common/prop destroy git updates.p
+ source /var/lib/dokku/core-plugins/available/common/property-functions
++ set -eo pipefail
++ [[ -n 1 ]]
++ set -x
[...]

PR #4988, opened 2021-12-31
Conditionally restart NGINX

fomojola:
Before restarting nginx, check if the current app is using a proxy and if the selected proxy is in fact NGINX. Only then should we bother invoking the nginx reload.

Fixes https://github.com/dokku/dokku/issues/4987

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -128,7 +128,13 @@ validate_nginx() {
 
 restart_nginx() {
   declare desc="restart nginx for given distros"
-  fn-nginx-vhosts-nginx-init-cmd "reload"
+  local PROXY_ENABLED="$(plugn trigger proxy-is-enabled "$APP")" 
+  local PROXY_TYPE="$(plugn trigger proxy-type "$APP")"      
+  if [[ "$PROXY_ENABLED" == "true" ]]; then                  
+    if [[ "$PROXY_TYPE" == "nginx" ]]; then                
+      fn-nginx-vhosts-nginx-init-cmd "reload"            
+    fi                                                     
+  fi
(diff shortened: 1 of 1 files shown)


plugins/nginx-vhosts/functions
```diff
@@ -128,7 +128,13 @@ validate_nginx() {
 
 restart_nginx() {
   declare desc="restart nginx for given distros"
-  fn-nginx-vhosts-nginx-init-cmd "reload"
+  local PROXY_ENABLED="$(plugn trigger proxy-is-enabled "$APP")" 
+  local PROXY_TYPE="$(plugn trigger proxy-type "$APP")"      
+  if [[ "$PROXY_ENABLED" == "true" ]]; then                  
+    if [[ "$PROXY_TYPE" == "nginx" ]]; then
```
josegonzalez, line 134: Mind changing this to:

```
if [[ "$PROXY_ENABLED" == "true" ]] && [[ "$PROXY_TYPE" == "nginx" ]]; then
```

Also, there is a formatting issue, should be easy to check CI output to see whats up :)

josegonzalez (changes requested):

josegonzalez:
Thanks for the pull request, it's much appreciated!

plugins/nginx-vhosts/functions
fomojola: And done (I think)! I don't spend enough time in shell land: I wasn't confident I could stick the double.

josegonzalez: No worries, you'll get more comfortable as you make more pull requests :)

fomojola:
Had to install shfmt: should be good to go now.

PR #4999, opened 2022-01-13
Purge cache for all apps when calling buildpacks:set-property with --global flag

elhu:
Fixes dokku#4990


josegonzalez:
Any chance you can add a test for this?

elhu:
For sure! Any pointers on where to look?

josegonzalez:
I would create a test that runs the command and checks for zero exit code.

elhu:
Got it, I'll see if I can add that today or tomorrow.

elhu:
Alright, I have a test, let's see if that works!

PR #5043, opened 2022-02-19
Upgrade default Vagrantfile box to bento/ubuntu-20.04

holamendi:
20.04 is the default Ubuntu LTS version on most cloud providers. Personally I've been using this image for months without any issues.


josegonzalez:
I'm sticking to 20.04 for now. All of our testing otherwise uses 20.04 as the base, and will do so for another year until the end of standard support. Thanks for the MR though!

PR #5180, opened 2022-05-15
fixes rst-versions height

Bilal-io:


Bilal-io:
Sorry about that, the calculated height should have only been applied to the container when expanded.

PR #5196, opened 2022-05-22
Disable herokuish usage on both armhf and arm64 platforms

josegonzalez:
We will not provide support for herokuish on either as herokuish is meant to be Heroku-compatible, and the Heroku stack does not support arm. Use 'pack' or Dockerfiles for app building instead.

### DIFF
--- a/plugins/00_dokku-standard/subcommands/report
+++ b/plugins/00_dokku-standard/subcommands/report
@@ -8,6 +8,7 @@ cmd-report() {
   declare cmd="report"
   [[ "$1" == "$cmd" ]] && shift 1
   declare APP="$1"
+  local ARCHITECTURE="$(dpkg --print-architecture 2>/dev/null || true)"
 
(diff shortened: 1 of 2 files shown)


0wx:
Why I'm still get herokuish tho
```sh
-----> Set main to DOKKU_DEPLOY_BRANCH.
-----> Cleaning up...
-----> Building test from herokuish
remote: WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
remote:  !     Failure extracting app code
remote:  !     Removing invalid image tag dokku/test:latest
remote:  !     App build failed
To pi.local:test
 ! [remote rejected] main -> main (pre-receive hook declined)
error: failed to push some refs to 'pi.local:test'
```

```bash
root@pi:~# dpkg --print-architecture 2>/dev/null || true
arm64
root@pi:~# dokku version
dokku version 0.27.5
root@pi:~# 
```

PR #5197, opened 2022-05-25
fix: Vector fails to start container due to --watch-config flag

holamendi:
Vector doesn't accept an argument after the `--watch-config` flag anymore, so executing `dokku logs:vector-start` fails to start the container.

The output of `dokku logs:vector-logs`:

```
-----> Vector container logs
 !     USAGE:
 !         vector [OPTIONS] [SUBCOMMAND]
 !     For more information try --help
 !     error: Found argument '1' which wasn't expected, or isn't valid in this context
 ``` 

The output of `docker ps --no-trunc`:
```
CONTAINER ID                                                       IMAGE                           COMMAND                                                                              CREATED          STATUS                          PORTS      NAMES
f3bb3fae0464081c3c8c39cfb7f777ed920a8bd4f89392bc922d949e9e721886   timberio/vector:latest-debian   "/usr/bin/vector --config /etc/vector/vector.json --watch-config 1"                  35 minutes ago   Restarting (2) 30 seconds ago              vector
```

### DIFF
--- a/plugins/logs/functions.go
+++ b/plugins/logs/functions.go
@@ -77,7 +77,7 @@ func startVectorContainer(vectorImage string) error {
 		"--volume", common.MustGetEnv("DOKKU_LOGS_HOST_DIR") + ":/var/logs/dokku/apps",
 		"--volume", common.MustGetEnv("DOKKU_LOGS_HOST_DIR") + "/apps:/var/log/dokku/apps",
 		vectorImage,
-		"--config", "/etc/vector/vector.json", "--watch-config", "1"}, " "))
+		"--config", "/etc/vector/vector.json", "--watch-config"}, " "))
 	cmd.ShowOutput = false
 
 	if !cmd.Execute() {


holamendi:
I see the CI tests failed on master: https://github.com/dokku/dokku/runs/6595025431?check_suite_focus=true#step:5:41

Could it be due to a Vector version difference?

josegonzalez:
Yeah seems like maybe you were using a newer version manually? I'm going to revert this for now and we can merge this in later once I do the minor release (next release is a patch).

holamendi:
You're right. I used `dokku logs:vector-start --vector-image timberio/vector:latest-debian`. Sorry for the confusion.

Dokku's [Vector doc page shows that same command](https://dokku.com/docs/deployment/logs/#vector-logging-shipping) as an example for starting the container with a custom Image, which in this case broke the plugin.

josegonzalez:
Well we start it with an image that is tested and compatible. If the upstream image changes, it changes, can't do much about that.

holamendi:
Totally, sorry again for the false alarm.

josegonzalez:
This should be fixed in the next minor :)

PR #5206, opened 2022-06-01
Added dokku-litestream

AxelTheGerman:
I recently created a plugin for SQLite replication via Litestream to S3. Been using it myself in production for a couple of weeks and was able to successfully restore a backed up DB so I guess it's ready for people to try and be added to this list.

https://github.com/AxelTheGerman/dokku-litestream


AxelTheGerman:
Not sure if that one static analysis error is important or not

PR #5263, opened 2022-07-24
Update deprecated apt-key command

alexgleason:
Upon running the provided `apt-key` command, I get a deprecation warning:

```sh
root@pop-os:/home/alex# wget -nv -O - https://packagecloud.io/dokku/dokku/gpgkey | apt-key add -
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
2022-07-24 11:01:36 URL:https://d28dx6y1hfq314.cloudfront.net/505/623/gpg/dokku-dokku-FB2B6AA421CD193F.pub.gpg?t=1658678796_f906c69452735f2338f96cd23ea0bcf3ba2b60b1 [3937/3937] -> "-" [1]
OK
```

Furthermore in `apt update`:

```sh
root@pop-os:/home/alex# apt update
...
W: https://packagecloud.io/dokku/dokku/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

Checking the docs for `man apt-key`:

```sh
DEPRECATION
       Except for using apt-key del in maintainer scripts, the use of apt-key is deprecated. This section shows how to replace existing use of apt-key.

       If your existing use of apt-key add looks like this:

       wget -qO- https://myrepo.example/myrepo.asc | sudo apt-key add -

       Then you can directly replace this with (though note the recommendation below):

       wget -qO- https://myrepo.example/myrepo.asc | sudo tee /etc/apt/trusted.gpg.d/myrepo.asc

       Make sure to use the "asc" extension for ASCII armored keys and the "gpg" extension for the binary OpenPGP format (also known as "GPG key public ring"). The binary OpenPGP format
       works for all apt versions, while the ASCII armored format works for apt version >= 1.4.

       Recommended: Instead of placing keys into the /etc/apt/trusted.gpg.d directory, you can place them anywhere on your filesystem by using the Signed-By option in your sources.list
       and pointing to the filename of the key. See sources.list(5) for details. Since APT 2.4, /etc/apt/keyrings is provided as the recommended location for keys not managed by
       packages. When using a deb822-style sources.list, and with apt version >= 2.4, the Signed-By option can also be used to include the full ASCII armored keyring directly in the
       sources.list without an additional file.
```

I was able to remove the old key with:

```sh
apt-key del 288B3315
```

Then reinstall install it the new way:

```sh
wget -qO- https://packagecloud.io/dokku/dokku/gpgkey | sudo tee /etc/apt/trusted.gpg.d/dokku.asc
```

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -212,7 +212,7 @@ install-dokku-from-deb-package() {
   fi
 
   echo "--> Installing dokku"
-  wget -nv -O - https://packagecloud.io/dokku/dokku/gpgkey | apt-key add -
+  wget -qO- https://packagecloud.io/dokku/dokku/gpgkey | sudo tee /etc/apt/trusted.gpg.d/dokku.asc
   echo "deb https://packagecloud.io/dokku/dokku/$DOKKU_DISTRO/ $OS_ID main" | tee /etc/apt/sources.list.d/dokku.list
   apt-get update -qq >/dev/null
 
--- a/deb.mk
+++ b/deb.mk
@@ -19,7 +19,7 @@ install-from-deb:
 	wget -nv -O - https://get.docker.com/ | sh
 
 	@echo "--> Installing dokku"
-	wget -nv -O - https://packagecloud.io/dokku/dokku/gpgkey | apt-key add -
+	wget -qO- https://packagecloud.io/dokku/dokku/gpgkey | sudo tee /etc/apt/trusted.gpg.d/dokku.asc
 	@echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ $(shell lsb_release -cs 2>/dev/null || echo "bionic") main" | sudo tee /etc/apt/sources.list.d/dokku.list
 	sudo apt-get update -qq >/dev/null
 	sudo DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true apt-get -qq -y --no-install-recommends install dokku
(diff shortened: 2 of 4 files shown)


josegonzalez:
Does this work for all supported OS Versions? Specifically Ubuntu 18.04 (I don't care so much about Debian 9 as we're dropping support for it in the next major release).

josegonzalez:
@alexgleason does this work on ubuntu 18.04 and Debian 10?

alexgleason:
I'm not sure. :thinking: However, they plan to completely remove the `apt-key` command from the next version of Ubuntu, so it at least should be merged before then.

josegonzalez:
Mind doing some investigation? If it works, we can just merge this as is.

alexgleason:
Here's my giant log of what I did: https://www.toptal.com/developers/hastebin/tikurasujo.sql

Seems fine in `ubuntu:18.04` docker image.

Condensed/amended version for simplicity:

```
➜  ~ docker run -it ubuntu:18.04 bash
root@4c97acfb4ba4:/# apt update && apt install -y wget
root@4c97acfb4ba4:/# wget -qO- https://packagecloud.io/dokku/dokku/gpgkey | tee /etc/apt/trusted.gpg.d/dokku.asc
root@4c97acfb4ba4:/# echo "deb https://packagecloud.io/dokku/dokku/ubuntu/ bionic main" | tee
root@4c97acfb4ba4:/# apt update
root@4c97acfb4ba4:/# apt install -y dokku
root@4c97acfb4ba4:/# dokku apps:list
=====> My Apps
 !     You haven't deployed any applications yet
root@4c97acfb4ba4:/# 
```

PR #5273, opened 2022-07-30
Correct issue where more complex docker args were not properly parsed when scheduling containers

josegonzalez:
In cases where a docker arg uses backticks - which may happen with docker labels - scheduled containers would fail to support these due to performing a double-eval. This is particularly noticeable for traefik labels, where backticks and parenthesis are common. This fix properly handles docker args as an array instead of performing an extra eval when scheduling containers.


iot-resister (approved):
lgtm

PR #5293, opened 2022-08-10
Add an official traefik proxy implementation

josegonzalez:
This plugin uses a docker-compose based Traefik installation in conjunction with injected container labels to route requests. It only exposes the minimal necessary for routing traffic to docker containers. Users wishing to customize further labels may explore using the docker-options plugin to attach additional labels during the 'deploy' phase.

Requests are routed as soon as the container is running and passing [healthchecks](https://doc.traefik.io/traefik/routing/services/#health-check).


plugins/traefik-vhosts/docker-args-process-deploy
```diff
@@ -0,0 +1,121 @@
...
+
+    if [[ -n "$proxy_container_https_port" ]] || [[ -n "$proxy_container_https_port_candidate" ]]; then
+      if [[ -z "$proxy_container_https_port" ]]; then
+        dokku_log_warn "Warning: https:443 port mapping not found"
+        dokku_log_warn "Utilizing first https port mapping, http:$proxy_host_https_port_candidate:$proxy_container_https_port_candidate"
+        proxy_container_https_port="$proxy_container_https_port_candidate"
+      fi
+
+      output="$output --label traefik.http.services.$APP-$PROC_TYPE-https.loadbalancer.server.port=$proxy_container_https_port"
+      output="$output --label traefik.http.routers.$APP-$PROC_TYPE-https.entrypoints=https"
```
timadevelop, line 109: Shouldn't we specify a `certresolver` for each service?

Like
```bash
output="$output --label traefik.http.routers.$APP-$PROC_TYPE-https.tls.certresolver=leresolver"
```

Without specifying a `certresolver` it seems like Traefik won't even try to use TLS for this route...

plugins/traefik-vhosts/templates/compose.yml.sigil
```diff
@@ -0,0 +1,58 @@
...
+      - --certificatesresolvers.leresolver.acme.storage=/acme.json
+      - --certificatesresolvers.leresolver.acme.tlschallenge=true
+      {{ end }}
+
+    labels:
+      # Dashboard
+      - "traefik.http.routers.api.rule=Host(`{{ $.TRAEFIK_API_VHOST }}`)"
+      - "traefik.http.routers.api.service=api@internal"
+      - "traefik.http.routers.api.entrypoints={{ if $.TRAEFIK_LETSENCRYPT_EMAIL }}https{{ else }}http{{ end }}"
+      - "traefik.http.routers.middlewares=auth"
```
timadevelop, line 31: What version of Traefik was used to test this basic_auth?

Shouldn't this middleware be under `.api`? i.e.
```diff
-     - "traefik.http.routers.middlewares=auth"
+     - "traefik.http.routers.api.middlewares=auth"
```

josegonzalez: traefik:v2.8

PR #5308, opened 2022-08-17
Implement haproxy proxy plugin

josegonzalez:
This will be based on https://github.com/byjg/docker-easy-haproxy/

Todo:

- tests
- letsencrypt integration
- tcp proxying?


tmontovaneli:
👍 nice work!

douglasmelo (approved):

josegonzalez:
Okay this is pretty close. The big issue is that the upstream project uses the network alias vs container ip, and the former doesn't work under the default docker bridge network. I've filed https://github.com/byjg/docker-easy-haproxy/issues/33 and will try to get to it but won't block the next minor on including this if it doesn't land in time.

PR #5348, opened 2022-09-07
Add nixpacks builder support

### LINKED ISSUE #5338: Support Nixpacks as a builder
### Description of feature

Nixpacks by Railway (https://nixpacks.com) is a buildpack alternative that will work wonderfully with Dokku. Since Dokku already supports running docker images, much of the plumbing should already be in place to be able to add Nixpacks as a builder. 

I can research into doing it if this feature request is approved.

josegonzalez:
Todo

- [x] tests

Closes #5338


IgnisDa:
Would be great if the docs mentioned that the `nixpack` cli needs to be installed for it to work.

IgnisDa:
When is this expected to land?

josegonzalez:
Whenever I get time to work on it. I try not to land new features in patch releases.

josegonzalez:
If you want the feature to come out faster, you can [consider sponsoring it](https://github.com/dokku/.github/blob/master/SPONSORING.md).

IgnisDa:
@josegonzalez Can I continue working in this?

IgnisDa:
@josegonzalez Are there any updates on this?

josegonzalez:
Working on adding tests for this and a few other MRs since I want to get out the next version of Dokku and Dokku Pro out this month :)

josegonzalez:
Okay, there are a few blockers here:

- https://github.com/railwayapp/nixpacks/issues/651: We can probably get around this for the global build args by reparsing and setting them manually, so not a _true_ blocker, but still something I need to deal with.
- https://github.com/railwayapp/nixpacks/issues/658: This breaks predeployment tasks and probably just any Procfile usage to be honest. It's not quite clear whats going on, but this is a hard blocker.
- https://github.com/railwayapp/nixpacks/issues/660: This is a weird one because nixpacks is doing something I wouldn't expect it to do (the `release` process is usually run by the platform after a build, at least for all Procfile-based platforms I'm aware of). We can disable the `release` command (by removing it) but honestly I'd rather not if we can avoid it, since it removes Dokku functionality (though we have `app.json` support that folks can use otherwise).

josegonzalez:
I made an MR for the label parsing upstream. Going to try and make another for the entrypoint issue.

josegonzalez:
Made a PR for the array command issue [here](https://github.com/railwayapp/nixpacks/pull/906) as well. I think that should work without test changes but we'll see.

josegonzalez:
Okay I think we can sort of workaround the nixpacks issues:

- Have a separate Dockerfile that injects a "corrected" entrypoint for the app. Basically [this](https://github.com/josegonzalez/nixpacks/commit/03ebcd09ff3c8ce77d3312a293ff7843fd40d9ce).
- Parse the build args and set them manually as a sublist...

josegonzalez:
Alright, seems like the nixpacks cli now properly supports list-like arguments, so that can be crossed off the list. Hoping this works!

Issue #4067, opened 2020-07-16
dokku does not respect EXPOSEd port


ltalirz:
## Description of problem
Pushing a [git repository containing a `Dockerfile`](https://github.com/materialscloud-org/tools-seekpath/) that does `EXPOSE 80`.
This is recognized by dokku as evidenced by the logs during the push:

```
-----> Setting config vars
       DOKKU_DOCKERFILE_PORTS:  80
``` 
Reading the [dokku documentation](http://dokku.viewdocs.io/dokku/networking/port-management/#applications-using-expose), I expect that this will cause dokku to map the corresponding port on the remote.

However, when I then inspect the app on the dokku instance, the port mappings are not set up correctly - in particular, port 80 is still mapped to the default port 5000 inside the container:
```
dokku proxy:ports seekpath
-----> Port mappings for seekpath
-----> scheme             host port                 container port
http                      80                        5000
https                     443                       5000
```

### How reproducible
I've tried 

 * deleting the app via `dokku apps:destroy seekpath` and recreating it again
 * the new `dokku proxy:build-config seekpath` command

and keep getting the same result.
The only solution I found is to manually set the corresponding port mapping.

### Steps to Reproduce

1. `git clone https://github.com/materialscloud-org/tools-seekpath/`
2. `git remote add dokku@yourdokku:appname`
3. ssh into dokku instance and run `dokku proxy:ports appname`

#### Actual Results
```
dokku proxy:ports seekpath
-----> Port mappings for seekpath
-----> scheme             host port                 container port
http                      80                        5000
https                     443                       5000
```
#### Expected Results
```
dokku proxy:ports seekpath
-----> Port mappings for seekpath
-----> scheme             host port                 container port
http                      80                        80
https                     443                       80
```
Following the dokku documentation, I may need another `EXPOSE 443`, which would then result in mapping port 443:443.
In any case, the configuration above is the mapping that I would *like* to have.

### `dokku report APP_NAME` output

```
ubuntu@dokku:~$ dokku report seekpath
-----> uname: Linux dokku 4.15.0-22-generic #24-Ubuntu SMP Wed May 16 12:15:17 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
-----> memory:
                     total        used        free      shared  buff/cache   available
       Mem:          32167         495       23005           1        8666       31235
       Swap:             0           0           0
-----> docker version:
       Client: Docker Engine - Community
        Version:           19.03.12
        API version:       1.40
[...]

Issue #2903, opened 2017-09-15
Reset proxy ports, if no domain is set on first Dockerfile deploy


mimischi:
Dokku will clear all values in `DOKKU_PROXY_PORT_MAP` and set the variable to `http:9000:9000`, when deploying an app for the first time via Dockerfile (where `9000` is the exposed port, specified in the Dockerfile). Subsequent deploys are fine.
*Note*: global vhosts are disabled and no domain is specified for the app:

```user@host:~$ dokku domains:report test
=====> test domains information
       Domains app enabled: false                    
       Domains app vhosts:                           
       Domains global enabled: false                    
       Domains global vhosts
```

### `dokku report`
```
-----> uname: Linux poc3 4.4.0-93-generic #116-Ubuntu SMP Fri Aug 11 21:17:51 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
-----> memory: 
              total        used        free      shared  buff/cache   available
Mem:          64426        4763       55859         126        3803       58755
Swap:         51245           0       51245
-----> docker version: 
Client:
 Version:      17.07.0-ce
 API version:  1.31
 Go version:   go1.8.3
 Git commit:   8784753
 Built:        Tue Aug 29 17:42:53 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.07.0-ce
 API version:  1.31 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   8784753
 Built:        Tue Aug 29 17:41:43 2017
 OS/Arch:      linux/amd64
 Experimental: false
-----> docker daemon info: 
Containers: 52
 Running: 25
 Paused: 0
 Stopped: 27
Images: 157
Server Version: 17.07.0-ce
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins: 
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 3addd840653146c90a254301d6c3a663c7fd6429
runc version: 2d41c047c83e09a6d61d464906feb2a2f3c52aa4
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.4.0-93-generic
[...]

PR #5548, opened 2022-12-27
Simplify port handling and centralize all port management in new ports plugin

josegonzalez:
Port management is _all_ over the place in the Dokku codebase, and makes certain operations for users more annoying - particularly anything in regards to exposing ports for dockerfile/image-based deployments.

This change centralizes all that logic and removes the dependency on the weird port env vars.

Todo:

- [ ] Move `DOKKU_PROXY_PORT_MAP` to a property
- [ ] Remove the `DOKKU_PROXY_PORT` and `DOKKU_PROXY_SSL_PORT` usage
- [ ] Make the computed `DOKKU_PROXY_PORT_MAP` a new property that can be overridden as necessary #2903
- [ ] Drop DOKKU_DOCKERFILE_PORTS #4067


josegonzalez:
There is too much going on here. I think I need to redo this PR so that we:

- don't break the proxy:ports commands (that can be done after the ansible repo is updated)
- it separates the port command refactor from the port handling code (to make it easier to review/debug)

PR #5558, opened 2022-12-29
Remove references to ubuntu 18.04 ahead of EOL

josegonzalez:
While we will still build for 18.04, we will now stop testing against it. Instead, we will test against 20.04, the next EOL release after 18.04.

Once Ubuntu 18.04 is EOL, any references to it will also be removed.

Note that 18.04 support was already deprecated in Dokku 0.28.0, so this change is just acknowledging that deprecation.


josegonzalez:
The failures are because the ssl files are written with an old encryption format that isn't supported by 18.04. We'll need to regenerate those (though idk how they were generated in the first place lol).

Maybe this will help for the SAN: https://gist.github.com/KeithYeh/bb07cadd23645a6a62509b1ec8986bbc

josegonzalez:
@michaelshobbs any chance you have any idea how this stuff worked/can regenerate the tls files for me? :D

josegonzalez:
I regenerated the files via the following:

```
# checking the extracted certs
openssl x509 -text -in server.crt

# server_ssl (node-js-app.dokku.me)
openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout server.key -out server.crt -subj "/CN=dokku.me" -days 3650

# server_ssl_wildcard (*.dokku.me)
openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout server.key -out server.crt -subj "/CN=*.dokku.me" -days 3650

# server_ssl_san
openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout server.key -out server.crt -subj "/CN=*.dokku.me" -days 3650 -addext "subjectAltName = DNS:www.test.dokku.me, DNS:www.test.app.dokku.me"
```

PR #5563, opened 2023-01-02
Clarify documentation for default site under nginx

### LINKED ISSUE #5562: Documentation Issue
### Description of problem

I have a specific Dokku app which I use to catch all HTTP traffic to my server. This is what I would refer to as my "default app". On this same server, there are other deployed Dokku apps, each of which has one or more domain assigned to it. All HTTP requests which to not match any of the domains assigned to any of my running apps should end up hitting the "default app".

This is covered in the documentation here:

https://dokku.com/docs/networking/proxies/nginx/#default-site

It says:

>By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack.

However, this was not what actually happened. It looks like on a fresh install of Dokku on Ubuntu 20.04, nginx comes with a `default` site which is enabled under `/etc/nginx/sites-enabled/`, and which displays the "Welcome to Nginx" page. Once I removed the `default` symlink from that directory, and restarted nginx, everything worked as expected, with the default site (which is the lexicographically first site) handling all unknown hosts.

I would recommend mentioning this in this section of the docs. I am happy to make this change myself and submit a PR.

### Steps to reproduce

Install Dokku on Ubuntu 20.04.

Deploy two apps named:

- `0-app-1`
- `app-2`

Assign a domain to each app:

`dokku domains:add 0-app-1 app1.mydomain.com`
`dokku domains:add app-2 app2.mydomain.com`

Verify that traffic is being routed to the proper apps when you visit their respective domains.

Point the domain `app3.mydomain.com` to the same server, then visit that domain in a browser or send it a curl.

I would expect that the contents of `0-app-1` would be returned, since there is no app which as the domain `app3.mydomain.com`, but it returns the "Welcome to nginx" page, which is provided by the "default" site in `/etc/nginx/sites-enabled/`.

Remove the `default` symlink from `/etc/nginx/sites-enabled/`, and restart nginx with:

`service nginx restart`

Visit `app3.mydomain.com`, and receive the contents of `0-app-1`

### dokku report $APP_NAME

If this is necessary for this specific issue, please let me know, and I will provide it.

### Additional information

_No response_

_No response_

_No response_

kbuilds:
Added some wording to the Nginx docs about the default site which Nginx creates on install. This default site being enabled prevents Dokku from receiving traffic unless the HOST header matches one of the domains assigned to an app.

Fixes #5562

### DIFF
--- a/docs/networking/proxies/nginx.md
+++ b/docs/networking/proxies/nginx.md
@@ -397,7 +397,11 @@ These are provided as an alternative to the generic Nginx error page, are shared
 
 ### Default site
 
-By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack. If this is not the desired behavior, you may want to add the following configuration to the global nginx configuration.
+By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack.
+
+> Note that some versions of Nginx may create a default site (a static page which says "Welcome to Nginx") when installed. If this default site is enabled, it will not route any requests with an unknown HOST header to Dokku. If you want Dokku to receive all requests, disable this default site by removing the `/etc/nginx/sites-enabled/default` file, and then restart nginx.
+
+If Dokku handling all requests is not the desired behavior, you may want to add the following configuration to the global nginx configuration.
 
 Create the file at `/etc/nginx/conf.d/00-default-vhost.conf`:
 
--- a/docs/networking/proxies/nginx.md
+++ b/docs/networking/proxies/nginx.md
@@ -399,7 +399,13 @@ These are provided as an alternative to the generic Nginx error page, are shared
 
 By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack.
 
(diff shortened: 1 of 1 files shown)


docs/networking/proxies/nginx.md
```diff
@@ -397,7 +397,11 @@ These are provided as an alternative to the generic Nginx error page, are shared
 
 ### Default site
 
-By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack. If this is not the desired behavior, you may want to add the following configuration to the global nginx configuration.
+By default, Dokku will route any received request with an unknown HOST header value to the lexicographically first site in the nginx config stack.
+
+> Note that some versions of Nginx may create a default site when installed. This site is simply a static page which says "Welcome to Nginx", and if this default site is enabled, Nginx will not route any requests with an unknown HOST header to Dokku. If you want Dokku to receive all requests, disable this default site by removing the `/etc/nginx/sites-enabled/default` file, and then restart Nginx.
```
josegonzalez, line 402: Mind changing the prefix to `> Warning:` instead of `> Note`? It'll render like a warning :)

Also, maybe add the exact command users should run, like so:

```
> Warning: that some versions of Nginx may create a default site when installed. This site is simply a static page which says "Welcome to Nginx", and if this default site is enabled, Nginx will not route any requests with an unknown HOST header to Dokku. If you want Dokku to receive all requests, run the following commands:
>
> ```
> sudo rm -f /etc/nginx/sites-enabled/default
> dokku nginx:stop
> dokku nginx:start
> ```
```

This will make the work less ambiguous for users.

PR #5585, opened 2023-01-14
Improve instructions for disabling the default nginx site

NicolasLM:
The previous instructions for disabling the default site left things in an in-between state: it still returned an error page and did not disable the default HTTPS site unless the user figured out how to generate a valid SSL certificate.

The changes made in this commit assume that the user actively disabling the default site considers all requests to unknown vhosts as bogus. Therefore there is no point in responding and providing a valid SSL certificate to them.

As such, the recommended error code is now `444` to close the connection without responding and the HTTPS default site is disabled with a self-signed certificate.

Also as all distributions supported by Dokku use systemd, make use of it instead of the legacy `service` command which does not work anymore on some distributions like Arch.

### DIFF
--- a/docs/networking/proxies/nginx.md
+++ b/docs/networking/proxies/nginx.md
@@ -397,7 +397,7 @@ These are provided as an alternative to the generic Nginx error page, are shared
 
 ### Default site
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Technically the docker image doesn't use systemd, but I think thats fine.

Thanks for the docs update!

PR #5592, opened 2023-01-16
Remove support for DOKKU_SCALE file

josegonzalez:
It has been replaced by the 'formations' key within the 'app.json' file format.


ohld:
nooooooooooo

josegonzalez:
Switch to the formations key, its better.

PR #5737, opened 2023-03-13
Update instructions for setting letsencrypt global email

Samuelodan:
Since there's no instruction for how to use the `DOKKU_LETSENCRYPT_EMAIL` env variable, I'm not sure why it's included here. So, instead, I'm proposing to include the step for directly setting the global email for `dokku-letsencrypt`.

(It'd be nice to also get some explanation behind creating that env variable)


josegonzalez:
The environment variable referenced an old way of configuring that plugin. The plugin has since been refactored to use settings, hence why it is no longer available.

Samuelodan:
Ah, I see. Okay, thanks you.

Issue #5785, opened 2023-04-05
Detect id_ed25519.pub in git:public-key command


obrienmd:
At line:
https://github.com/dokku/dokku/blob/5af750b4695f20fa45af2bf2086d5c921a04b6fd/plugins/git/internal-functions#L260

It looks like dokku checks for the file `~/.ssh/id_rsa.pub` to determine if a public key is available to the dokku user. Upon failure, the system recommends the user run ` ssh-keygen -t ed25519 -C 'example@example.com'` to resolve the issue.

On each system I've tested with, this command by default saves the public key to `~/.ssh/id_ed25519.pub`, which makes sense.

There's probably some super complex way to test this dynamically, but IMHO just adding an OR condition for the ED25519 pub key file is probably OK.

If no major disagreement, I'll work up a pull request ASAP.

josegonzalez:
Seems like a good fix.

josegonzalez:
Closing as there is a pull request open.

PR #5786, opened 2023-04-05
Detect id_ed25519.pub in git:public-key command 

obrienmd:
Per issue:
https://github.com/dokku/dokku/issues/5785

### DIFF
--- a/plugins/git/internal-functions
+++ b/plugins/git/internal-functions
@@ -257,13 +257,13 @@ cmd-git-public-key() {
   local cmd="git:public-key"
   [[ "$1" == "$cmd" ]] && shift 1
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Probably should also cat the correct file?

obrienmd:
Ugh, this is what I get for trying to PR before rushing out the door. Sorry @josegonzalez, I'll try to do a decent job later tonight :)

obrienmd:
OK, now should work as expected. Tested on a couple of hosts, but not much diversity, all Ubuntu 22.04.1.

I think ultimately this would be better served by listing *.pub and pluralizing key, but I think this probably gets us close enough to correct to be worth it (if I didn't flub something obvious again).

PR #5862, opened 2023-05-19
Stop using the `latest` tag for the default `DOKKU_CNB_BUILDER`

### LINKED ISSUE #5861: DOKKU_CNB_BUILDER defaults to the `latest` tag of `heroku/buildpacks` which is about to be deleted
### Description of problem

The `DOKKU_CNB_BUILDER` config option defaults to `heroku/buildpacks` (ie: `heroku/buildpacks:latest`), and:

1. The `heroku/buildpacks:*` repo is deprecated in favour of the `heroku/builder:*` and `heroku/builder-classic:*` repos (see https://github.com/heroku/builder#heroku-builder-images)
2. (More urgently) The `latest` tag of `heroku/buildpacks` is no longer being updated, and will soon be deleted (in favour of using explicit version tags, such as `heroku/buildpacks:20`), since the latest tag is not something we document or recommend for stack images as it allows breaking changes with no warning. See: https://github.com/heroku/builder/pull/344

### Steps to reproduce

See:
https://github.com/dokku/dokku/blob/e72f5b1d90a12222db0c3e6813fa7fb945a8cb59/dokku#L45

### dokku report $APP_NAME

N/A

### Additional information

_No response_

_No response_

_No response_

edmorley:
Previously the default value for `DOKKU_CNB_BUILDER` was `heroku/buildpacks`, which for Docker is equivalent to `heroku/buildpacks:latest`.

The `latest` tag is about to be deleted (see heroku/builder#344), so this switches the default value to `heroku/buildpacks:20` - which is what the `latest` tag is currently as alias for.

Longer term, `DOKKU_CNB_BUILDER` should be switched to the non-deprecated `heroku/builder:*` images, however, that's potentially a breaking change, so is out of scope of this bug-fix PR.

For the list of available builders/tags, see:
https://github.com/heroku/builder#heroku-builder-images

Fixes #5861.

### DIFF
--- a/dokku
+++ b/dokku
@@ -42,7 +42,7 @@ DOKKU_DISTRO=$(
 export DOCKER_BIN=${DOCKER_BIN:="docker"}
 
 export DOKKU_IMAGE=${DOKKU_IMAGE:="gliderlabs/herokuish:latest-20"}
-export DOKKU_CNB_BUILDER=${DOKKU_CNB_BUILDER:="heroku/buildpacks"}
+export DOKKU_CNB_BUILDER=${DOKKU_CNB_BUILDER:="heroku/buildpacks:20"}
 export DOKKU_LIB_ROOT=${DOKKU_LIB_PATH:="/var/lib/dokku"}
 
 export PLUGIN_PATH=${PLUGIN_PATH:="$DOKKU_LIB_ROOT/plugins"}


josegonzalez:
I'll get this merged, but what is potentially breaking about switching to the `heroku/builder` image?

edmorley:
@josegonzalez Thank you for the merge. The breaking differences between the various images are:
- The `heroku/buildpacks:*` images contain a mixture of native CNBs and shimmed classic buildpacks. In contrast for the new images, we've split out the shimmed and native buildpacks into two separate builder namespaces: `heroku/builder:*` and `heroku/builder-classic:*`. For now (until we finish more CNBs), `heroku/buildpacks:*` supports a couple more languages/tools (PHP, Gradle) than the `heroku/builder:*` images. 
- The `heroku/builder:*` and `heroku/builder-classic:*` images only support Heroku-22+, meaning that switching to them requires upgrading from Heroku-20 to Heroku-22 too.
- The native CNBs for some languages have used this opportunity to drop old features (eg old versions of Rails in the case of the Ruby buildpack iirc), or haven't quite added support for all features yet (eg automatic Django collectstatic support is not yet finished for the Python CNB)

The feature parity and language support differences will reduce over the coming months as we finish/polish the native CNBs.

See also the image descriptions under the table here:
https://github.com/heroku/builder#heroku-builder-images

edmorley:
That is, it would be great if Dokku eventually switched the default to `heroku/builder:22`, however, if that was done right now, then for a small subset of people that would be a breaking change, and they would need to instead switch to `heroku/builder-classic:22` until the feature parity/language support of `heroku/builder:22` is improved. 

Switching the Dokku default now might be something you want to do anyway (depending on how many people rely on the defaults, your appetite for risk, and how many Dokku users you think might read the release notes clearly etc), but I wanted to at least keep this initial change simple, since it unblocks me deleting the (no longer updated, and IMO footgun-prone) `latest` tag :-)

josegonzalez:
I'll get a patch release out in a day or two, after which you can probably remove the image. If it breaks for folks, we have a [workaround](https://dokku.com/docs/deployment/builders/cloud-native-buildpacks/#customizing-the-buildpack-stack-builder) folks can use to switch to a supported builder.

I think moving to the `heroku/builder:22` might make sense in the next minor release.

josegonzalez:
Also, thank you for the explanation, that makes a ton of sense!

PR #5994, opened 2023-07-02
Update proxy-management.md

aochagavia:
Clarify that Dokku already supports Caddy, HAProxy, etc

Thanks for the great work on Dokku, by the way!


josegonzalez:
This will actually be removed in #5991 as part of an overhaul of the port/proxy docs, so I'm going to close this for now. Thanks for the pull request though!

PR #6007, opened 2023-07-07
Add application/wasm to nginx.conf.sigil gzip

Firfi:
Hi! `.wasm` files tend to be the largest out there and also the ones that benefit from gzip a lot. This PR adds them as candidates for gzipping


josegonzalez:
Going to get this merged into the next minor since it might be a BC break.

PR #6009, opened 2023-07-08
Properly set the plugin repository name when installing in a docker-based dokku install

josegonzalez:

PR #6022, opened 2023-07-10
Fix issue with docker plugin-list install failing boot for docker-based installations

iphoting:
Follows incomplete fix #6009 based on comment <https://github.com/dokku/dokku/pull/6009#issuecomment-1627694934>:
- Fixed `docker/etc/my_init.d/10_dokku_init`
- Added tests in `tests/ci/setup.sh`

### DIFF
--- a/tests/ci/setup.sh
+++ b/tests/ci/setup.sh
@@ -115,6 +115,7 @@ build_dokku_docker_image() {
 
 run_dokku_container() {
   echo "=====> run_dokku_container on CIRCLE_NODE_INDEX: $CIRCLE_NODE_INDEX"
+  echo "redirect: https://github.com/dokku/dokku-redirect.git" | sudo tee /var/lib/dokku/plugin-list
   docker run -d \
     --env DOKKU_HOSTNAME=dokku.me \
     --name dokku \
@@ -132,7 +133,7 @@ check_container() {
   echo "=====> check_container on CIRCLE_NODE_INDEX: $CIRCLE_NODE_INDEX"
   local is_up
   local cnt=0
-  while true; do
+  while [ $cnt -lt 100 ]; do
     echo "$(date) [count: $cnt]: waiting for dokku startup"
     is_up=$(
       docker exec dokku ps -ef | grep "/usr/sbin/sshd -D" >/dev/null 2>&1
--- a/tests/ci/setup.sh
+++ b/tests/ci/setup.sh
@@ -139,8 +139,8 @@ check_container() {
       docker exec dokku ps -ef | grep "/usr/sbin/sshd -D" >/dev/null 2>&1
       echo $?
     )
+    echo "" && docker logs dokku
     if [[ $is_up -eq 0 ]]; then
-      echo "" && docker logs dokku
       break
     fi
     sleep 2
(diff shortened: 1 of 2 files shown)


iphoting:
The CI (check-commit) failure looks like a bug in the git checkout code. It looks like it is trying to fetch the PR branch, but the correct repo remote was not added first.

josegonzalez:
Yeah almost seems like a bug in the action? I've seen that happen a few times, I wonder if there is an upstream issue open.

iphoting:
Any reason for this specific line? <https://github.com/dokku/dokku/blob/8e779e25fb527db7cc9c6c867a9f4c228930b738/.github/workflows/ci.yml#L26>

Differs from upstream documentation: https://github.com/actions/checkout/tree/main#Checkout-pull-request-HEAD-commit-instead-of-merge-commit

josegonzalez:
IIRC I got this from [some tutorial](https://monadical.com/posts/filters-github-actions.html), but maybe we can update it to use the sha instead? Not sure if it works as well, but if you want to issue a separate PR with that change, we can test it out :D

iphoting:
> IIRC I got this from [some tutorial](https://monadical.com/posts/filters-github-actions.html), but maybe we can update it to use the sha instead? Not sure if it works as well, but if you want to issue a separate PR with that change, we can test it out :D

That example/tutorial will work for PRs raised from the same repo. For external repos, it will definitely break. For what it's worth, I've created another PR incorporating this change #6035.

josegonzalez:
@iphoting that was merged, mind rebasing?

tests/ci/setup.sh
```diff
@@ -115,6 +115,7 @@ build_dokku_docker_image() {
 
 run_dokku_container() {
   echo "=====> run_dokku_container on CIRCLE_NODE_INDEX: $CIRCLE_NODE_INDEX"
+  echo "redirect: https://github.com/dokku/dokku-redirect.git" >>/var/lib/dokku/plugin-list
```
josegonzalez, line 118: I think maybe we need to create that directory prior to setup? Maybe try creating via `sudo`?

iphoting: Do you know if sudo can be used in that context?

I'm thinking of doing a `| sudo tee /var/lib/dokku/plugin-list` instead.

josegonzalez: Yeah, try that.

iphoting:
That's a strange silent failure?

josegonzalez:
Yeah I think you might need to create that directory. Docker will silently create the directory if it doesn't exist. Just `mkdir` it first and then create the file.

iphoting:
> Yeah I think you might need to create that directory. Docker will silently create the directory if it doesn't exist. Just `mkdir` it first and then create the file.

Still fails with a mysterious hash… 🤔

iphoting:
The docker instance startup is still failing:
```
*** Running /etc/my_init.d/10_dokku_init...
       Copying /etc/ssh skel into place
       Copying /home/dokku skel into place
=====> check_container on CIRCLE_NODE_INDEX: 
       Copying /var/lib/dokku/config skel into place
       Copying /var/lib/dokku/data skel into place
       Copying /var/lib/dokku/services skel into place
 !     First argument must be --core or url to plugin
 !     Valid prefixes: https:, git:, ssh:, file:
 !     Valid suffixes: .tar.gz, .tgz
 !     Please retry with valid arguments
*** /etc/my_init.d/10_dokku_init failed with status 1
```

When the image is being build, is the PR’s `/etc/my_init.d/10_dokku_init` version correctly installed?

iphoting:
Resolved. Earlier errors caused by leading space in the URL argument. 😱

iphoting:
May we have this bug fix rolled into a release soon, please?

PR #6036, opened 2023-07-13
Fix traefik vhosts compose template

l3x4:
`auth` middleware label should only be added if `TRAEFIK_BASIC_AUTH` set

Original discord question: https://discord.com/channels/952684157123837952/1128590189812518972


josegonzalez:
Now the problem is that the labels are empty. I think maybe we should add a generic label to all of the compose configs...

PR #6177, opened 2023-08-31
Update shellcheck junit integration for newer version of shellcheck

maxvisser:
Shellcheck v0.9.0 added more info to output like suggestions to improve your lines


josegonzalez:
I probably should package this into it's own tooling (maybe rewrite in golang so it can be easily distributed)...

Thanks for the PR!

PR #6400, opened 2024-01-02
docs: add missing dokku command

gtpho:
This pull request fixes the missing `dokku` command from the [docs](https://dokku.com/docs/deployment/methods/image/#docker-image-deployment). The command is missing from the SSH calls.


josegonzalez:
No, this isn't the correct command. You have a broken entry in your `/home/dokku/.ssh/authorized_keys` file. You likely directly added an ssh key there, where instead you should have used the `dokku ssh-keys:add` to add the ssh key.

Please check on your config, remove the entry that you manually added, and use the `dokku ssh-keys:add` command to add the keys in question.

gtpho:
@josegonzalez That worked. Thanks!

PR #6576, opened 2024-02-10
Allow creating a detached container with TTY enabled for attaching remote terminal

Tashows:
Developing this for my project. Opening a new PR to discuss and/or get feedback, in case it seems useful.

### DIFF
--- a/plugins/run/help-functions
+++ b/plugins/run/help-functions
@@ -28,7 +28,7 @@ fn-help-content() {
   declare desc="return help content"
   cat <<help_content
     run [-e|--env KEY=VALUE] [--no-tty] <app> <cmd>, Run a command in a new container using the current app image
-    run:detached [-e|-env KEY=VALUE] [--no-tty] <app> <cmd>, Run a command in a new detached container using the current app image
+    run:detached [-e|-env KEY=VALUE] [--no-tty|--force-tty] <app> <cmd>, Run a command in a new detached container using the current app image
     run:list [--format json|stdout] <app>, List all run containers for an app
     run:logs <app|--container CONTAINER> [-h] [-t] [-n num] [-q], Display recent log output for run containers
     run:stop <app|--container CONTAINER>, Stops all run containers for an app or a specified run container
--- a/plugins/run/internal-functions
+++ b/plugins/run/internal-functions
@@ -54,6 +54,10 @@ fn-run() {
     esac
   done
 
+  if [[ "$DOKKU_DISABLE_TTY" == "true" ]] && [[ "$DOKKU_FORCE_TTY" == "true" ]]; then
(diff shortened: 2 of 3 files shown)


plugins/run/help-functions
```diff
@@ -29,6 +29,7 @@ fn-help-content() {
   cat <<help_content
     run [-e|--env KEY=VALUE] [--no-tty] <app> <cmd>, Run a command in a new container using the current app image
     run:detached [-e|-env KEY=VALUE] [--no-tty] <app> <cmd>, Run a command in a new detached container using the current app image
+    run:attachable [-e|-env KEY=VALUE] <app> <cmd>, Run a command in a new detached container using the current app image. Container has TTY enabled and is not removed automatically.
```
josegonzalez, line 32: Could this just be a `--force-tty` flag on `run:detached` instead?

Tashows: You are right. Thought I needed to get rid of the DOKKU_CONTAINER_RM too, but it seems to work and is actually better. Will clean this up

plugins/run/help-functions
```diff
@@ -28,7 +28,7 @@ fn-help-content() {
   declare desc="return help content"
   cat <<help_content
     run [-e|--env KEY=VALUE] [--no-tty] <app> <cmd>, Run a command in a new container using the current app image
```
josegonzalez, line 30: The flag also is here right?

plugins/run/internal-functions
```diff
@@ -21,6 +21,10 @@ fn-run() {
         export DOKKU_DISABLE_TTY=true
         shift
         ;;
+      --force-tty)
+        export DOKKU_FORCE_TTY=true
+        shift
+        ;;
```
josegonzalez, line 27: Might be worth adding some error checking after this switch statement to ensure both `DOKKU_DISABLE_TTY` and `DOKKU_FORCE_TTY` aren't set.

Tashows: In has_tty() I put the force-tty check first, since it is more explicit and not used as default anywhere, so it won't matter if both are set. Maybe check the final variables and add a warning in has_tty, so it's not run-specific only?

josegonzalez: I'd rather we exit abnormally immediately in places that depend on this behavior (so, this plugin) vs a warning that then has potentially unspecified behavior.

plugins/run/help-functions
Tashows: It will work as-is (i.e. the flag will be parsed), but seems redundant and maybe it doesn't make sense to add it to the description?
You either have tty already (and it will be used) or you are executing something from a non-terminal environment. In the second case if you do need tty you will need to attach it externally, in which case you might as well go with run:detached.

PR #6577, opened 2024-02-11
fix: whitespace breaking `SIGIL_PARAMS` (NGINX)

andrewswait:
Opening a draft PR with fix for the issue as detailed in this Slack thread: https://gliderlabs.slack.com/archives/C0FP0R13J/p1707611187420929

This replicates the change I've made manually on the server experiencing the issue which resolved it.

The crux is this is what's currently produced:
```
'DOKKU_APP_RELEASE                                                                         _LISTENERS=' 
```

Instead of this:
```
'DOKKU_APP_RELEASE_LISTENERS=' 
```

Have no idea why I haven't experienced this before now as the code I've adjusted was written years ago!

Will open a proper issue tomorrow morning with details extracted from that thread and link this PR up to it 👍🏻

### DIFF
--- a/plugins/nginx-vhosts/functions
+++ b/plugins/nginx-vhosts/functions
@@ -382,7 +382,7 @@ nginx_build_config() {
         PROXY_X_FORWARDED_SSL="$PROXY_X_FORWARDED_SSL")
 
       while read -r line || [[ -n "$line" ]]; do
-        PROC_TYPE=${line%%=*}
+        PROC_TYPE=$(echo "${line%%=*}" | xargs)
         LISTENERS="$(plugn trigger network-get-listeners "$APP" "$PROC_TYPE" | xargs)"
         UPP_PROC_TYPE="${PROC_TYPE^^}"
         UPP_PROC_TYPE="${UPP_PROC_TYPE//-/_}"


dennmart:
I was just about to report this bug since it bit me this morning, causing a few of my Dokku app rebuilds to fail. I wanted to confirm that the change in this PR also fixes the problem in my case.

josegonzalez:
Similar issue here: https://github.com/dokku/dokku/discussions/6488#discussioncomment-8429952

Here is where that file is being parsed: https://github.com/dokku/dokku/blob/master/plugins/ps/functions.go#L192

I wonder if trimming spaces there would fix both setting and getting things?

josegonzalez:
Investigating further, I think the issue comes from procfile-util. I made changes recently to how the commands are structured, and I think it's outputting trailing whitespace in some places. I'll take a further look, but pretty sure thats what is causing the issues here.

We'll definitely need to fix it there, but also can bandaid over it in Dokku when parsing the Procfile on our end.

josegonzalez:
https://github.com/josegonzalez/cli-skeleton/pull/59 <-- I think this is the bug. The framework I wrote that underlies the tool always padded the string out to 80 characters (I don't remember why tbh) and so we had whitespace where it had no business being.

josegonzalez:
Rather than using this change, would it be possible for ya'll to test using procfile-util `0.17.1` with the bad file contents? I think it'll resolve the issue.

andrewswait:
> Rather than using this change, would it be possible for ya'll to test using procfile-util `0.17.1` with the bad file contents? I think it'll resolve the issue.

Sure 👍🏼 will take a look at this today

andrewswait:
> > Rather than using this change, would it be possible for ya'll to test using procfile-util `0.17.1` with the bad file contents? I think it'll resolve the issue.
> 
> Sure 👍🏼 will take a look at this today

Yep looks good @josegonzalez 

Screenshot 2024-02-11 at 12 37 04

The only thing I find weird about this is I'd expect to have seen this happening before (it looks like that padding has been there for years) 🤔 

Is it worth merging this PR as well, just in case? Can't hurt to have whitespace trimming for this. Looks like you have it already for `LISTENERS` on the line below (although that doesn't help in this specific case as the padding is no longer trailing by that point).

Happy either way and cheers for patching this upstream 🙌🏻

josegonzalez:
The procfile-util package switched to the cli-skeleton recently, hence why the output is new.

josegonzalez:
I spent some time figuring out how to best apply this everywhere. Rather than doing a one-off fix here, I patched it in the ps plugin in #6582. This should handle issues globally vs me needing to copy-paste the fix to every place where we pull in the process scale.

Regardless, thanks for the initial PR and legwork, it definitely helped figure out what was causing the issue :)

andrewswait:
> I spent some time figuring out how to best apply this everywhere. Rather than doing a one-off fix here, I patched it in the ps plugin in #6582. This should handle issues globally vs me needing to copy-paste the fix to every place where we pull in the process scale.
> 
> 
> 
> Regardless, thanks for the initial PR and legwork, it definitely helped figure out what was causing the issue :)

Ah ok that makes sense 👌🏼 thanks!

PR #6588, opened 2024-02-12
add extra tags to every image dokku push when push-extra-tags are set

taraszka:
Some corner cases require the `latest` tag (i.e, jobs or cronjob that aren't a part of `dokku` but are a part of dokku-based deployments onto the Kubernetes).
This simple change ensures that both the `tag` (`int` type) and the `latest` tags are pushed to the remote registry on the app build.

PR #6595, opened 2024-02-14
Add ability for users to specify alternative kubeconfig and kubecontext

josegonzalez:
This will provide the possibility for users to talk to existing kubernetes clusters, thereby removing one of the biggest reasons for having the old scheduler-kubernetes plugin around.

See #6588 for impetus /cc @taraszka.


taraszka:
Excellent, I will check this out!

PR #6598, opened 2024-02-14
fix dokku_client.sh to correctly handle git urls that contain port numbers

bjelline:
read both application name and port number from remote

### DIFF
--- a/contrib/dokku_client.sh
+++ b/contrib/dokku_client.sh
@@ -111,7 +111,9 @@ main() {
   if [[ -z "$APP" ]]; then
     if [[ -d .git ]] || git rev-parse --git-dir &>/dev/null; then
       set +e
-      APP=$(git remote -v 2>/dev/null | grep -Ei "^${DOKKU_GIT_REMOTE}\s" | grep -Ei "dokku@$DOKKU_REMOTE_HOST" | head -n 1 | cut -f2 -d'@' | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
+      # Read app name and port from a git url of the form "dokku ssh://dokku@dokkuhost.example.com:5412/project-name (fetch)"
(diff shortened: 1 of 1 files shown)


contrib/dokku_client.sh
```diff
@@ -111,7 +111,9 @@ main() {
   if [[ -z "$APP" ]]; then
     if [[ -d .git ]] || git rev-parse --git-dir &>/dev/null; then
       set +e
-      APP=$(git remote -v 2>/dev/null | grep -Ei "^${DOKKU_GIT_REMOTE}\s" | grep -Ei "dokku@$DOKKU_REMOTE_HOST" | head -n 1 | cut -f2 -d'@' | cut -f1 -d' ' | cut -f2 -d':' 2>/dev/null)
+      # Read app name and port from a git url of the form "dokku ssh://dokku@dokkuhost.example.com:5412/project-name (fetch)"
+      APP=$(git remote -v 2>/dev/null | grep -Ei "^${DOKKU_GIT_REMOTE}\s" | grep -Ei "dokku@$DOKKU_REMOTE_HOST" | head -n 1 | cut -f2 -d'@' | cut -f2 -d'/' | cut -f1 -d' ' 2>/dev/null)
+      DOKKU_PORT=$(git remote -v 2>/dev/null | grep -Ei "^${DOKKU_GIT_REMOTE}\s" | grep -Ei "dokku@$DOKKU_REMOTE_HOST" | head -n 1 | cut -f2 -d'@' | cut -f1 -d'/' | cut -f2 -d':' 2>/dev/null)
```
josegonzalez, line 116: I think this change breaks the default remote format of `dokku@host:app`.

josegonzalez:
@bjelline are you interested in fixing the PR to support the existing format and your new format?

The alternative here is to just specify a port override for the host in your `~/.ssh/config` like so:

```
Host dokku.me
    User dokku
    Port 2202
```

That would ensure any ssh commands going to the `dokku.me` would have the dokku user as the default and hit port `2202`. You could even change the `dokku.me` there to be a custom domain and then set the `HostName` property to the real thing:

```
Host push-dokku
    HostName dokku.me
    User dokku
    Port 2202
```

Then your dokku remote would just be `push-dokku:$APP_NAME`.

josegonzalez:
I'm going to close this. The correct way is to use a custom `.ssh/config` entry.

PR #6612, opened 2024-02-25
Rewrite run command in golang

josegonzalez:
Due to how pflags works, this also changes commands such that they need to have a `--` separating them from `run` flags. That may change if/when pflags is forked for Dokku usage, but thats a bc-breaking change for now

This won't be merged until that BC break is removed.


josegonzalez:
Closing this since the fix this could have added (force-tty/no-tty) was merged.

PR #6650, opened 2024-03-03
Traefik: not expose all containers

bastianh:
Hi,

by setting exposedByDefault to false only containers are exposed that have the `traefik.enable` label. The label is already there for the web containers, it was only missing for the dashboard.

It might be a breaking change for people who have manually exposed other containers

traefik documentation: https://doc.traefik.io/traefik/providers/docker/#exposedbydefault


josegonzalez:
Thanks for this. I'll need to add a note for the 0.34.x release, but otherwise this looks great.

PR #6652, opened 2024-03-04
Added dokku-mdns to list of community plugins

Calyhre:
I did [a small plugin](https://github.com/calyhre/dokku-mdns) to auto publish mDNS subdomains on my home network when deploying an `app`.


josegonzalez:
Nice! I built [something similar](https://github.com/josegonzalez/dokku-avahi-register) years ago, but this is cool too :)

PR #6716, opened 2024-03-14
Move app and global ENV files to consolidated config path

josegonzalez:
Refs #1558


josegonzalez:
This is going to take more work to figure out the error and the release is big enough so I'm punting on it for now.

PR #6731, opened 2024-03-17
Clarify alternate proxy support status

strugee:
AFAICT the future is now.


josegonzalez:
Thanks for this. I've merged it into the upcoming release branch so it should land on the site soon.

PR #6763, opened 2024-03-24
Add docker-compose example to installation docs

iloveitaly:


josegonzalez:
I dislike that we need to do this - someone proposed this before - but I understand because docker doesn't have a built-in way to generate this. :sigh:

PR #6769, opened 2024-03-26
fix: using correct path for `docker-container-healthchecker`

joennlae:
First of all, `dokku` is fantastic 💯 .

I had this error:
```bash
remote: sudo: /usr/bin/docker-container-healthchecker: command not found
```

Which I think is due to:
It is installed via the Makefile https://github.com/dokku/dokku/blob/e2d51bfeb308c01fa3a2ec912cde30d98058bee7/Makefile#L163 in `/usr/local/bin/docker-container-healthchecker` although it is here called directly with `/usr/bin/docker-container-healthchecker`.
But I could also be wrong :-)

Here is a quick fix if someone else has the same issue:

```bash
cp /usr/local/bin/docker-container-healthchecker /usr/bin/docker-container-healthchecker
```

### DIFF
--- a/plugins/scheduler-docker-local/check-deploy
+++ b/plugins/scheduler-docker-local/check-deploy
@@ -153,7 +153,7 @@ trigger-scheduler-docker-local-check-deploy() {
     echo "$content" >"$TMP_APP_JSON_OUTPUT"
   fi
 
-  sudo /usr/bin/docker-container-healthchecker check "$DOKKU_APP_CONTAINER_ID" "${ARG_ARRAY[@]}" || FAILEDCHECKS="$?"
+  sudo $(which docker-container-healthchecker) check "$DOKKU_APP_CONTAINER_ID" "${ARG_ARRAY[@]}" || FAILEDCHECKS="$?"
 
   if [[ $FAILEDCHECKS -gt 0 ]]; then
     "$DOCKER_BIN" container update --restart=no "$DOKKU_APP_CONTAINER_ID" &>/dev/null || true


josegonzalez:
How are you installing dokku?

joennlae:
I installed it as mentioned in the documentation here: https://dokku.com/docs/getting-started/installation/#1-install-dokku

```bash
wget -NP . https://dokku.com/install/v0.34.0/bootstrap.sh
sudo DOKKU_TAG=v0.34.0 bash bootstrap.sh
```
I am using a fresh Ubuntu 22.04 installation.

josegonzalez:
Sounds like there is something wrong here. When installing that way, the deb package for docker-container-healthchecker installs in /usr/bin.

Let me see if I can replicate this and debug what's going on before merging.

joennlae:
> Sounds like there is something wrong here. When installing that way, the deb package for docker-container-healthchecker installs in /usr/bin.
> 
> Let me see if I can replicate this and debug what's going on before merging.

It could very well be that I did something wrong. But I just followed the instructions in the documentation :-)

But yeah, I was also surprised that no one else was having this issue.

Let's see if it is reproducible, and if not, then close it. Then it must have been my mistake :-)

joennlae:
Yeah, something is strange: it is not only docker-container-healthchecker that ends up in /usr/local/bin
it is all of these:
```bash
docker-container-healthchecker  docker-image-labeler  dokku  lambda-builder  netrc  plugn  procfile-util  sigil  sshcommand
```
I see that from the deb packages, it should be in `/usr/bin/`, so it is a problem on my side...

Sorry for the panic :-)

josegonzalez:
Are you sure you didn't do a source based install by specifying a branch or something instead of a dokku tag? Mind checking the output of bash history to verify?

joennlae:
> Are you sure you didn't do a source based install by specifying a branch or something instead of a dokku tag? Mind checking the output of bash history to verify?

Yeah you are absolutely right :-)

I did master...

```bash
 112* sudo DOKKU_BRANCH=master bash bootstrap.sh
```

joennlae:
So, it did a build and did not pull the deb packages. I see.

Again, sorry for the inconvenience, but we can close this pull request :-)

josegonzalez:
No worries. The "from source" option is there but not heavily tested, so I'd expect certain things to fail. Things get weird on the border of `sudo` usage as well (which we need because we don't add the `dokku` user to `docker`).

PR #6808, opened 2024-04-07
Traefik optional custom http https entrypoints

Coffee2CodeNL:
The code changes add the capability to configure custom HTTP and HTTPS entry points for the Traefik virtual hosts plugin. New settings have been added to the list of valid and global keys. Also, function definitions for retrieving default HTTP and HTTPS entry points were included in the internal functions file. Furthermore, the 'docker-args-process-deploy' script now considers the new entry points while processing deployment arguments.

### DIFF
--- a/docs/networking/proxies/traefik.md
+++ b/docs/networking/proxies/traefik.md
@@ -67,6 +67,17 @@ dokku traefik:stop
 
 The Traefik container will be stopped and removed from the system. If the container is not running, this command will do nothing.
 
+### Changing the Traefik entrypoint names
+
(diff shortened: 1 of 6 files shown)


josegonzalez:
What is the use case here?

Coffee2CodeNL:
I'm not using Dokku's Traefik instance, my entrypoints are "web" and "websecure", so this is for when you're routing through an existing Traefik instance.

josegonzalez:
Seems legit. I wonder if we should change our defaults...

Anyways:

- this needs to be in the report output
- needs documentation
- needs a unit test

Coffee2CodeNL:
Everything should be good for merge now :)

PR #6853, opened 2024-04-26
Support Ubuntu 24.04 LTS

Reggino:
Dokku seems to install and work normally on Ubuntu 24.04 LTS


josegonzalez:
This will force the install of the 22.04 package onto 24.04. There are a number of packages we need to release for 24.04 in order to get this through.

josegonzalez:
@Reggino if you wanna take a look at [this](https://github.com/dokku/dokku/issues/6856#issuecomment-2081371732), that would be wonderful!

PR #6902, opened 2024-05-20
Let app.json define default plugin properties

frmdstryr:
This allows the repo's `app.json` to set app specific default plugin properties. It will not overwrite any properties that are explicitly set using the normal plugin property commands but it will override any defaults passed to `PropertyGetDefault`  (and `PropertyGet` which uses a blank default) in code.

For example if the `app.json` contains a `pluginproperties` key, it can set plugin properties using a key: value mapping like this:

```json
{
  "pluginproperties": {
    "nginx": {
        "client-max-body-size": "50m"
    }
  }
}
``` 

I have never worked with go before so let me know if anything could be improved. I had to move some of the app-json path functions into common to avoid a cyclical dependency error.

If this is ok to add, please let me know where I should add details in the docs. 

Any feedback is welcome, thanks.

### DIFF
--- a/plugins/app-json/appjson.go
+++ b/plugins/app-json/appjson.go
@@ -191,16 +191,16 @@ func GetAppjsonDirectory(appName string) string {
 
 // GetAppjsonPath returns the path to a given app's extracted app.json file for use by other plugins
 func GetAppjsonPath(appName string) string {
-	return getProcessSpecificAppJSONPath(appName)
+	return common.GetProcessSpecificAppJSONPath(appName)
 }
 
 // GetAppJSON returns the parsed app.json file for a given app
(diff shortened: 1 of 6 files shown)


josegonzalez:
I'm torn on this as I don't know if this has any negative impacts permissions-wise. For instance, you might not want to let users modify nginx properties via a plugin, but then this allows a back-channel into that...

frmdstryr:
In my case there are no users, but I can see how this is a problem if random people can deploy apps. 

I'll close for now as anything I can think of with that scenario would require validation on every property. It's not that big of a deal to manually update a bunch of apps.

josegonzalez:
I think allowing all properties is fine, just need validation on whether the functionality is allowed at all or not somehow...

PR #7039, opened 2024-08-11
fix: correct case of pip in builder-pack test output

josegonzalez:
The upstream buildpack recently changed this so we need to update in order to conform to the new standard.


edmorley:
The Python CNB log output will change again in the future when we switch to the new logging style/helpers. 

To avoid churn in these tests I would recommend not asserting on strings from the logs, and instead running commands against the built app image to check it works and it's built the way you expect (eg `python --version` or `python -c 'import flask`).

PR #7077, opened 2024-08-26
Add cron dependency because it's necessary to run scheduled cron jobs

bakatz:
I hit an issue where my deployment failed with this error:
`Unable to update schedule file: exec: "crontab": executable file not found in $PATH`

The issue is just that you need to have cron installed. I did a `sudo apt install cron` on my host machine and that fixed it, so I added the `cron` dependency to the control file to prevent others from making the same mistake.


josegonzalez:
Probably makes more sense to figure out what package provides crontab and then add that as a dependency?

bakatz:
> Probably makes more sense to figure out what package provides crontab and then add that as a dependency?

Yeah fair enough. I thought you had intentionally left it out because not everyone is going to want to use cron.

If enough people are using this feature it seems like a good idea to include it by default.

Let me see if I can make an edit to the install script or wherever you manage dependencies for Dokku.

josegonzalez:
There should be a `debian/control` file you can add the `cron` dependency to.

bakatz:
> There should be a `debian/control` file you can add the `cron` dependency to.

@josegonzalez ok, added to control

Issue #6312, opened 2023-10-27
All Heroku getting-started apps fail on arm64/v8


jukefr:
### Description of problem

Hello

I just installed dokku on an Oracle Ampere VM,
Its arm64/v8,

The install went okay,
I then tried following the steps on https://dokku.com/docs/deployment/application-deployment/

It was fine until the `git push dokku main`

At which point i realised i had to follow the instructions to install `pack` https://buildpacks.io/docs/tools/pack/ (might be worth mentionning that this is required in the getting started guide)

But still I get 
```
remote: exec /cnb/lifecycle/creator: exec format error
remote: ERROR: failed to build: executing lifecycle: failed with status code: 1
remote:  !     Removing invalid image tag dokku/ruby-getting-started:latest
remote:  !     App build failed
```
When running the push

I tried 3 of the heroku demo apps and all result in the same error

```
> dokku apps:list                                                                                                                      3542ms < Fri Oct 27 21:58:55 2023
=====> My Apps
java-getting-started
node-js-getting-started
ruby-getting-started
```

I would love to figure out how to get this working and would propose to add these instructions to the getting started documentation.

### dokku report $APP_NAME

<details>

```
-----> uname: Linux oracle 5.15.0-1046-oracle #52-Ubuntu SMP Mon Oct 16 09:55:50 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
-----> memory:
                      total        used        free      shared  buff/cache   available
       Mem:           23987        6432       14570           4        2984       17229
       Swap:              0           0           0
-----> docker version:
       Client: Docker Engine - Community
        Version:           24.0.7
        API version:       1.43
        Go version:        go1.20.10
        Git commit:        afdd53b
        Built:             Thu Oct 26 09:08:26 2023
        OS/Arch:           linux/arm64
        Context:           default

       Server: Docker Engine - Community
        Engine:
         Version:          24.0.7
         API version:      1.43 (minimum version 1.12)
         Go version:       go1.20.10
         Git commit:       311b9ff
         Built:            Thu Oct 26 09:08:26 2023
         OS/Arch:          linux/arm64
         Experimental:     false
        containerd:
         Version:          1.6.24
         GitCommit:        61f9fd88f79f081d64d6fa3bb1a0dc71ec870523
        runc:
         Version:          1.1.9
         GitCommit:        v1.1.9-0-gccaecfc
        docker-init:
         Version:          0.19.0
         GitCommit:        de40ad0
[...]

PR #7215, opened 2024-09-29
Clarify x64 as amd64/arch64 in readme

Lewiscowles1986:
So I started by searching after reading README.md, and came across #6312 

Maybe others assume x64 is synonymous with 64-bit, but I find the long-hand, helped me to understand that `arm64` and `aarch64`, and `amd64` as well as `x86_64` (which I thought x64 most closely resembled) would all be supported, and it's a minimum native bus width limitation, dispensing with the history of x86 and 8086, than anything else.

6312 suggests arm64 is fair game, and should work.

Unsure what this means for RISC-V, or if there are other 64-bit architectures this has not been tested on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,8 @@ Support us with a monthly donation and help us continue our activities. [[Become
 
 A fresh VM running any of the following operating systems:
 
-- Ubuntu 20.04 / 22.04 / 24.04 (64-bit) - Any currently supported release
-- Debian 11+ (64-bit)
-- Arch Linux (64-bit) *(experimental)*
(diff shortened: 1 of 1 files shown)


josegonzalez:
Going to change this to amd64/arch64 as those are the dpkg supported architectures (and more common parlance for this sort of thing among our users).

josegonzalez:
Thanks for the PR.

Lewiscowles1986:
Thanks @josegonzalez but it's aarch64 or arm64. I'm uncertain what arch64 is. 🫂 I'm sure folks will work it out.

josegonzalez:
Bah I meant to type arm64!

Issue #7219, opened 2024-09-29
docker-options:add run options not used even afer app rebuild


chrisjsimpson:
### Description of problem

`dokku docker-options:add ruby-getting-started run "--ulimit nofile=12"` is not used/enforced when the container gets ran.

### Steps to reproduce

After performing:

` dokku docker-options:add ruby-getting-started run "--ulimit nofile=12"`

Verify:
```
dokku docker-options:report 
=====> ruby-getting-started docker options information
       Docker options build:          --link dokku.postgres.railsdatabase:dokku-postgres-railsdatabase 
       Docker options deploy:         --link dokku.postgres.railsdatabase:dokku-postgres-railsdatabase --restart=on-failure:10 
       Docker options run:            --link dokku.postgres.railsdatabase:dokku-postgres-railsdatabase --ulimit nofile=12 
```

and rebuild

`dokku ps:rebuild ruby-getting-started`

Wait for rebuild and new container.

The started container contains no Ulimits:

`docker inspect ruby-getting-started.web.1`

```
...
 "Ulimits": [],
...
```

What I expect to see, is the ulimis present. For example, (on same host) running a container directly with `docker` cli:

```
docker run --ulimit nofile=12 httpd
```

And inspecting it, we see the `Ulimits` in the container metadata:

`docker inspect <containerID>`

```
...
  "Ulimits": [                                                                                                                                                                      
      {                                                                                                                                                                             
          "Name": "nofile",                                                                                                                                                         
          "Hard": 12,                                                                                                                                                               
          "Soft": 12                                                                                                                                                                
      }                                                                                                                                                                             
  ],
...
```  
         

### dokku report $APP_NAME

```
dokku report ruby-getting-started
-----> uname: Linux dokku 6.8.0-45-generic #45-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug 30 12:02:04 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
-----> memory: 
                      total        used        free      shared  buff/cache   available
       Mem:            4840         979         420          16        3766        3861
       Swap:           2969           0        2969
-----> disk utilization: 
       Filesystem                         Size  Used Avail Use% Mounted on
       /dev/mapper/ubuntu--vg-ubuntu--lv   14G  8.6G  4.4G  67% /
       /dev/sda2                          2.0G   95M  1.7G   6% /boot
-----> disk inode utilization: 
       Filesystem                        Inodes IUsed IFree IUse% Mounted on
[...]

PR #7239, opened 2024-10-05
[docs] Clarify difference between dokku run and docker run in docker-options.md

chrisjsimpson:
Hopefully clarifies the difference between dokku run and docker run in docker-options.md for other readers who may also get confused by this.

Ref https://github.com/dokku/dokku/issues/7219

How I got confused:

I read the docs: "Pass [container options](https://docs.docker.com/engine/reference/run/) to the docker run" ([here](https://dokku.com/docs/advanced-usage/docker-options/?)), and read the *docker* run reference, missing the nuanced detail different between `docker run`/`dokku run`  -> `dokku deploy` stage.

### DIFF
--- a/docs/advanced-usage/docker-options.md
+++ b/docs/advanced-usage/docker-options.md
@@ -3,7 +3,9 @@
 > [!IMPORTANT]
 > New as of 0.3.17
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Followed up with #7421. Thoughts on that change instead?

chrisjsimpson:
Closed via https://github.com/dokku/dokku/pull/7241

PR #7241, opened 2024-10-05
Clarify phase utilization in docker-options docs

josegonzalez:
Also generally update the docker-options docs to follow how documentation is written for the other plugins.

Refs #7219

### DIFF
--- a/docs/advanced-usage/docker-options.md
+++ b/docs/advanced-usage/docker-options.md
@@ -3,57 +3,89 @@
 > [!IMPORTANT]
 > New as of 0.3.17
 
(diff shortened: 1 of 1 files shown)


josegonzalez:
@chrisjsimpson does the update in this PR better document how phases work, or do you think it still needs clarification?

chrisjsimpson:
> @chrisjsimpson does the update in this PR better document how phases work, or do you think it still needs clarification?

That would is great yes. Especially https://github.com/dokku/dokku/pull/7241/files#diff-470fddcfa778710a2dd78bc57d5789593ff4e1469de9e2809dba1d875bc9efd0R30


Thanks @josegonzalez

PR #7400, opened 2024-12-10
Check if tty is actually used and support proper output when it's not

Tashows:
Calling `dokku run app_name command` via a non-tty flow (eg. via a daemon) did not return command output. Using this I could make this work, without affecting original functionality. Might not be optimal, but leaving this here for discussion/suggestions.

### DIFF
--- a/plugins/scheduler-k3s/k8s.go
+++ b/plugins/scheduler-k3s/k8s.go
@@ -339,7 +339,6 @@ func (k KubernetesClient) ExecCommand(ctx context.Context, input ExecCommandInpu
 	req.Param("stdin", "true")
 	req.Param("stdout", "true")
 	req.Param("stderr", "true")
-	req.Param("tty", "true")
 
 	if input.Entrypoint != "" {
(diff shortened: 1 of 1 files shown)


josegonzalez:
Can you give more information on how to trigger the bug?

Tashows:
I used a forked version of https://github.com/dokku/dokku-daemon. Feeding the socket `run app_name ls` for an app that has been deployed with the k3s scheduler and then reading the output would only show 
```
.
.
```
vs running the same command via a terminal would show (example output)
```
.
.
app lib storage other_directories ...
```
Figured out that the driving factor for this difference occured because sizeQueue variable ended up being nil when run via the daemon. The rest of the changes is so that the output format works for the daemon/socket environment while also not breaking for the terminal. Having said that, there might be a better way (better functions to call) to figure out if the current environment actually supports tty or not. These changes are just what I figured out by trial and error.

josegonzalez:
Feel free to make PRs to that repo :)

PR #7403, opened 2024-12-10
Properly resolve imagePullSecrets from app and deploymentID if relevant property does not exist

Tashows:
On TriggerSchedulerRun (k3s scheduler) the existing configuration only checks for image-pull-secrets scheduler-k3s property on the app. For a case of a private registry/repo, this means that the authentication configured by `dokku registry:login` is not available, so the pod cannot fetch the image. Unless you set the image-pull-secrets manually.
The pull secrets resolution is properly configured on [`func TriggerSchedulerDeploy`](https://github.com/dokku/dokku/blob/443757125d14dfb8a9755fb4ba8e704b23c60bd0/plugins/scheduler-k3s/triggers.go#L224). Not sure if it would be a better solution to update the scheduler property on each deploy instead of re-constructing a string here though; haven't dug deeper to check all use-cases. Thoughts?

### DIFF
--- a/plugins/scheduler-k3s/triggers.go
+++ b/plugins/scheduler-k3s/triggers.go
@@ -1106,10 +1106,32 @@ func TriggerSchedulerRun(scheduler string, appName string, envCount int, args []
 	}
 
 	attachToPod := os.Getenv("DOKKU_DETACH_CONTAINER") != "1"
+
+	clientset, err := NewKubernetesClient()
+	if err != nil {
+		return fmt.Errorf("Error creating kubernetes client: %w", err)
+	}
+
+	if err := clientset.Ping(); err != nil {
+		return fmt.Errorf("kubernetes api not available: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


plugins/scheduler-k3s/triggers.go
```diff
@@ -1107,6 +1107,9 @@ func TriggerSchedulerRun(scheduler string, appName string, envCount int, args []
 
 	attachToPod := os.Getenv("DOKKU_DETACH_CONTAINER") != "1"
 	imagePullSecrets := getComputedImagePullSecrets(appName)
+	if imagePullSecrets == "" {
+		imagePullSecrets = fmt.Sprintf("ims-%s.%d", appName, deploymentID)
```
josegonzalez, line 1111: Do we generate this automatically on a per-app basis?

josegonzalez:
I refactored this a bit to check if an imagePullSecret exists before referencing it. I think thats a bit safer here, since otherwise it's a potential false assumption that there _will_ be a value.

PR #7406, opened 2024-12-11
Pin packeto buildpack to older version that may not have incorrect platform

josegonzalez:
CI appears to be failing with the following error:

    # ERROR: failed to build: downloading buildpack: extracting from registry 'gcr.io/paketo-buildpacks/python': fetching image: image with reference gcr.io/paketo-buildpacks/python:latest was found but its platform (linux) does not match the specified platform (linux/amd64)

This change is being implemented to see if we can avoid that entirely, thereby showing that the error is potentially something in the release process of the buildpack.


josegonzalez:
This isn't the fix, but not sure whats going on. Might be a lifecycle change?

PR #7422, opened 2024-12-17
Fix detached tty for k3s

Tashows:
Fixes some inconsistencies for k3s when compared to executing `dokku run:detached` with docker-local scheduler:

1. Prints pod name (similarly docker-local prints container name)
2. Properly sets interactive and tty arguments/variables respecting --force-tty flag (which sets DOKKU_FORCE_TTY)

--

Other suggestion: I notice some ENV variables are being set to "true" and others to "1". Maybe update common.ToBool to also check for "1"?

### DIFF
--- a/docs/processes/one-off-tasks.md
+++ b/docs/processes/one-off-tasks.md
@@ -66,7 +66,7 @@ dokku run --no-tty node-js-app ls -lah
 > [!IMPORTANT]
 > New as of 0.25.0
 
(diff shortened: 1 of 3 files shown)


Tashows:
I noticed `DOKKU_RM_CONTAINER` as I was reading through code but there are more. Not sure if it makes sense to find all of them. Some are set in bash scripts, some using os.Setenv. Might be more cases.

Printing the docker container name is mentioned in docs as well. I think it makes sense for the `run:detached` commands, since you never directly get access to the container/pod and it might stay up and running (especially considering the use of --force-tty)

PR #7442, opened 2025-01-05
Update how domains:clear works and add domains:reset command

### LINKED ISSUE #7438: domains:clear is misleading
### Description of problem

The semantics of `domains:clear` and `domains:clear-global` are not the same and induces the user to errors.

`domains:clear-global` completely removes the global domains, by calling the function below:

```
# plugins/domains/functions
domains_clear_global() {
  declare desc="remove all domains"
  local GLOBAL_VHOST_PATH="$DOKKU_ROOT/VHOST"

  rm "$GLOBAL_VHOST_PATH"
  touch "$GLOBAL_VHOST_PATH"
}
```

On the other hand, `domains:clear` removes the app's vhost file and then calls `domains_setup "$APP"`, which copies the default vhost file to the app's directory. So it's more like a "reset" (reset to the global defaults) than a "clear" (remove all the domains).

I think `domains:clear` must really remove all the domains associated with the app. And if it's desirable to have a feature to "reset" the app domains to the global ones, I propose adding a `domains:reset <app-name>` command, which would do this action. These changes would make the command more or less backwards incompatible, but I think it's more like a bug-fix than an API change.
Another option would be to deprecate `domains:clear` and implement two new ones: `domains:reset` and `domains:remove-all`.

I'm willing to implement this if the idea is accepted. If not accepted, I think there should be a documentation change clarifying this behavior (the current one says `domains:clear` "Clear all domains for app").

### Steps to reproduce

```shell
dokku domains:clear-global
dokku domains:report --global
dokku domains:add-global test1.example.com
dokku domains:add-global test2.example.com
dokku apps:create test-app
dokku apps:domains:report test-app
# Got: test-app.test1.example.com and test-app.test2.example.com (as expected)

dokku domains:report test-app
dokku domains:clear test-app
dokku domains:report test-app 
# Got: test-app.test1.example.com and test-app.test2.example.com (instead of none)
```

### dokku report $APP_NAME

Not applicable.

### Additional information

_No response_

_No response_

_No response_

turicas:
Fixes #7438

Note: this is my first PR/change to a Dokku core plugin and I wasn't sure how to implement a proper test, so a test missing for `domains:clear` (since the old test now actually tests the `domains:reset` functionality and I just renamed it). I think this is the only thing missing in this PR. Could you help me implementing this?

### DIFF
--- a/docs/appendices/0.36.0-migration-guide.md
+++ b/docs/appendices/0.36.0-migration-guide.md
@@ -4,3 +4,4 @@
 
 - Process stop timeouts are now configured via the `ps` property `stop-timeout-seconds`. Existing `DOKKU_DOCKER_STOP_TIMEOUT` environment variables will be automatically migrated to the new value.
 - Processes now default to a `30` second stop timeout.
+- The `domains:clear` command no longer runs domain setup after clearing the domains. Use `domains:reset` or run `domains:setup` manually to have the same effect.
--- a/docs/configuration/domains.md
+++ b/docs/configuration/domains.md
@@ -6,8 +6,8 @@
 ```
 domains:add <app> <domain> [<domain> ...]      # Add domains to app
 domains:add-global <domain> [<domain> ...]     # Add global domain names
-domains:clear <app>                            # Clear all domains for app
-domains:clear-global                           # Clear global domain names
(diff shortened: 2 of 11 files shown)


turicas:
Thanks! :)

PR #7499, opened 2025-02-01
feat(nginx-vhosts): add support for custom nginx service command

nonZero:
In a custom dokku setup, we need to use a slightly complex command for restarting nginx via a custom script.  This would allow to specify this script via an env variable (instead of using a standard linux service)


josegonzalez:
I'm super hesitant to add this without more information. Can you describe this custom setup?

nonZero:
I am aware this setup is non standard and not recommended...
Some other app controls the server and dokku is the 2nd platform.
The other app is running a custom nginx with `include
/home/dokku/*/nginx.conf`.
Everything works except I need to patch this function every update.
I won't be surprised if I am the only user for this and would also be happy
with any other workaround.

Thank you!

josegonzalez:
Can you tell me what this other system is/what that script is? The contents of that script would also be great so I can figure out if there is a way to do what you're asking without needing to change dokku.

nonZero:
An example app would be gitlab which comes with an nginx service.
Restarting gitlab's nginx has different commands for different versions and the command in not guarnteed to stay the same in future versions.
A script supplied by the user would be the best adapter to handle those details.

nonZero:
Can this be merged?

josegonzalez:
Not until the next minor release (soon) and it needs docs (I can write that).

nonZero:
any news?

Issue #6814, opened 2024-04-10
k3s plugin - add PV option


taraszka:
### Description of feature

Old legacy https://github.com/dokku/dokku-scheduler-kubernetes can add PV for an App build by dokku and deployed onto Kubernetes.
The way how it works is simple:
1. it claims PV using this template: https://github.com/dokku/dokku-scheduler-kubernetes/blob/master/templates/pvc.json.sigil
2. it mounts volume here https://github.com/dokku/dokku-scheduler-kubernetes/blob/master/internal-functions#L776

`dokku` params are also quite clear:

```
# dokku scheduler-kubernetes:add-pvc ${pvc_name} ${pvc_size} --access-mode ReadWriteMany --namespace ${namespace} --storage-class-name ${storage_class}
# dokku scheduler-kubernetes:mount ${app_name} ${pvc_name} ${path}'
```

Note: `dokku-scheduler-kubernetes`, for PVC size, only accepts integer value that represents Mi. So for 2 Gi Volume, you have to say: 2000. Would be nice if k3s would accept Mi, Gi, Ti, so param would be: `(..):add-pvc my-awsome-pvc 2Gi --access-mode (...)`

amalysh:
Hi, added MR:
https://github.com/dokku/dokku/pull/7509

josegonzalez:
This is what i was thinking for an interface:

```shell
# this currently creates a `/var/lib/dokku/data/storage/node-js-storage` directory
dokku storage:ensure-directory node-js-storage

# instead, we should create it via `storage:create`
# at it's simplest, this is just a directory on the host
# each storage entry is named, and names are globally unique across a dokku installation
dokku storage:create node-js-storage /var/lib/dokku/storage/node-js-storage

# we can also specify chown options
# chown options are implementation-specific
dokku storage:create node-js-storage /var/lib/dokku/storage/node-js-storage --chown herokuish

# storage entries are specific to schedulers, so we could create a storage entry for k3s
# note that the folders could be reused across multiple storage entries
# and it is up to the user to ensure that folders are not conflicting in a manner that does not make sense
dokku storage:create node-js-storage-k3s /var/lib/dokku/storage/node-js-storage-k3s --scheduler k3s

# aside from the --chown option, there are other options that can be specified
# k3s supports an access mode, so we can specify that
dokku storage:create node-js-storage-k3s /var/lib/dokku/storage/node-js-storage-k3s --scheduler k3s --access-mode=ReadWriteMany

[...]

PR #7509, opened 2025-02-03
#6814 add PV options to scheduler-k3s

amalysh:
added scheduler-k3s:add-pvc, scheduler-k3s:remove-pvc, scheduler-k3s:mount, scheduler-k3s:unmount subcommands


josegonzalez:
Thanks for the PR. I was actually hoping we could better align this with the existing storage plugin, but maybe something completely separate is useful. Let me think about this for a bit.

amalysh:
I hoped it as well to align to storage plugin, but this plugin is not prepared to to use any of schedulers as local.

josegonzalez:
Closing this as I'm reworking the storage plugin as per my scope in the original ticket. That said, thanks for the initial attempt, it was definitely appreciated!

Issue #7600, opened 2025-03-17
Dokku overwrites some container proxy labels set with docker-options


dragonhunt02:
### Description of problem

I tried to set a traefik label ``traefik.http.routers.myapp-web-http.rule=Host(`myapp.dokku.me`) && Path(`/myapp`)`` to route path `/myapp` to an app.
Dokku ignored my traefik rule label set with `docker-options`.

### Steps to reproduce

**1.** Run Dokku with docker compose or docker run
```
docker container run -d \
  --env DOKKU_HOSTNAME=dokku.me \
  --env DOKKU_HOST_ROOT=/var/lib/dokku/home/dokku \
  --env DOKKU_LIB_HOST_ROOT=/var/lib/dokku/var/lib/dokku \
  --name dokku \
  --publish 3022:22 \
  --publish 8080:80 \
  --publish 8443:443 \
  --volume /var/lib/dokku:/mnt/dokku \
  --volume /var/run/docker.sock:/var/run/docker.sock \
  dokku/dokku:0.35.16
```

**2.** Clone repo
```
git clone https://github.com/dragonhunt02/dokku-test-dockerfile.git dokkutest
```
**3.** Setup ssh keys and configure repo push

**4.** Enter container
```
docker exec -it dokku bash
```
**5.** Setup
```
dokku proxy:set --global traefik
dokku nginx:stop
dokku traefik:start

dokku apps:create myapp

# This label will be correctly set
dokku docker-options:add myapp deploy --label 'testlabel=foobar'

# This will be overwritten
dokku docker-options:add myapp deploy --label "'traefik.http.routers.myapp-web-http.rule=Host(\`myapp.dokku.me\`) && Path(\`/myapp\`)'"

dokku docker-options:report myapp
```
**6.** Enable trace and increase Terminal max lines
```
dokku trace:on
```
**7.** Git push
```
git push dokku main
```

## Result
Run `dokku ps:inspect myapp` to see labels.
```
            ...
            "Image": "dokku/myapp:latest",
            "Labels": {
                "com.dokku.app-name": "myapp",
                "com.dokku.builder-type": "dockerfile",
                "com.dokku.container-type": "deploy",
                "com.dokku.dyno": "web.1",
                "com.dokku.image-stage": "release",
                "com.dokku.process-type": "web",
                "dokku": "",
                "org.label-schema.schema-version": "1.0",
                "org.label-schema.vendor": "dokku",
                "testlabel": "foobar",
                "traefik.enable": "true",
[...]

PR #7608, opened 2025-03-22
Implement proxy container :label feature

dragonhunt02:
## Implement proxy container label feature

Container labels can be set with:
`proxy-name:label <app> <show | set | unset> <property> (<value>)`

`proxy-name` can be: `caddy` `haproxy` `openresty` `traefik`

Labels set with this command can override Dokku-set labels (user is warned with a message on build) #7600


josegonzalez:
Apologies for the late reply.

Can you write some usage documentation for this, as well as some tests?

Issue #7840, opened 2025-08-09
Unable to set vector sink due to url


emilsedgh:
### Description of problem

My log destination requires me to input a token which has a `=` in in. But `dokku logs:set` doesn't seem to support that.

### Steps to reproduce

Imagine my token is `foobar`. I'll run:

```
dokku logs:set --global vector-sink "loki://?endpoint=https://host&encoding[codec]=text&auth[token]=foobar&auth[strategy]=bearer"
```

and it works fine.

but if my token is `foobar=`

```
dokku logs:set --global vector-sink "loki://?endpoint=https://host&encoding[codec]=text&auth[token]=foobar=&auth[strategy]=bearer"
```

This will obviously break because my vector sink file will look like this:

```
"auth": {
    "strategy": "bearer",
    "token=foobar": ""
}
```

Now, what I expected was being able to encode `=` as url and use `%3D` instead but that also fails.

### dokku report $APP_NAME

It's a global setting.

### Output of failing commands after running: dokku trace:off

```shell

```

### Output of failing commands after running: dokku trace:on

```shell

```

PR #7841, opened 2025-08-11
Fix handling of equals sign in vector sink values

leksyib14:
fixes https://github.com/dokku/dokku/issues/7840

- fixes the parsing situation where `=` is present in the value

### DIFF
--- a/go.work
+++ b/go.work
@@ -1,4 +1,4 @@
-go 1.24.0
+go 1.24.5
 
 use (
 	./plugins/app-json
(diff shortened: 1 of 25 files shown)


josegonzalez (changes requested):
This needs a bats test, mind adding one?

leksyib14:
> This needs a bats test, mind adding one?

sure, will do.

josegonzalez:
I think maybe just using this fork of qson might fix the problem? https://github.com/fastfishio/qson

josegonzalez:
Yeah that fixes it it seems...

PR #7989, opened 2025-09-21
Add support to prevent overlapping execution of cron tasks

### LINKED ISSUE #7988: Add support to prevent overlapping execution of tasks
### Description of feature

Currently, tasks executed on a schedule can overlap with each other without an internal locking mechanism in the task startup.

The simpler alternative is to run the dokku tasks directly from a cron tab entry using a wrapper like flock to avoid overlaps. However, this then requires manual cron management instead of using dokku's simple scheduling config.

I'm not sure what kind of lock would work best with dokku's plugin system or if something else needs to be added to the events for that to be possible. Users might want to implement hooks for many external storage systems, but hoping for something simple.

josegonzalez:
Closes #7988

### DIFF
--- a/docs/appendices/file-formats/app-json.md
+++ b/docs/appendices/file-formats/app-json.md
@@ -22,6 +22,7 @@
 
 - `command`: (string, required)
 - `schedule`: (string, required)
+- `concurrency_policy`: (string, optional, default: `allow`, options: `allow`, `deny`, `replace`)
 
 ## Formation
 
--- a/plugins/app-json/appjson.go
+++ b/plugins/app-json/appjson.go
@@ -57,6 +57,9 @@ type CronCommand struct {
 
 	// Schedule is the cron schedule to execute the command on
 	Schedule string `json:"schedule"`
+
+	// ConcurrencyPolicy is the concurrency policy for the cron command
+	ConcurrencyPolicy string `json:"concurrency_policy"`
 }
 
 // Formation is a struct that represents the scale for a process from an app.json file
--- a/plugins/app-json/appjson.go
+++ b/plugins/app-json/appjson.go
@@ -26,7 +26,7 @@ var (
 // AppJSON is a struct that represents an app.json file as understood by Dokku
 type AppJSON struct {
 	// Cron is a list of cron commands to execute
-	Cron []CronCommand `json:"cron"`
+	Cron []CronEntry `json:"cron"`
 
 	// Formation is a map of process types to scale
 	Formation map[string]Formation `json:"formation"`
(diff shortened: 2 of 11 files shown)


othercorey:
Should this fail instead of wait for the lock? If the task has an issue, won't the schedule create a backlog of tasks to run one after the other? I'm not very familiar with the scheduler wrapper to know if does something else.

josegonzalez:
Yeah I used an exclusive lock for docker-local. Kubernetes might start containers though if the old one finishes within the timeout (though I intend to increase the timeout to allow for better autoscaling).

othercorey:
Is there anything I can do to help with this? I don't know how to create test cases for these, but I can try if needed.

josegonzalez:
Yeah the next thing is to add a test that does something like `sleep forever` in a detached command with a madeup cron id, and then try to run that twice. The second command should fail (and of course we should clean up the first one).

othercorey:
I wrote a test for this, but I couldn't get it to fail. Noticed a few things:

- The document says `deny` instead of `forbid`
- Missing stdout format of Concurrency in CommandList()
- `trigger-scheduler-docker-local-scheduler-run` always sets concurrency_policy twice
```
    if [[ -n "$DOKKU_CONCURRENCY_POLICY" ]]; then
      concurrency_policy="$DOKKU_CONCURRENCY_POLICY"
    fi
    concurrency_policy="$DOKKU_CONCURRENCY_POLICY"
```

I don't know what's wrong or why I can't seem to affect the test.

othercorey:
I don't know what's missing in the test :(.

josegonzalez:
I haven't taken a look but we should probably `docker inspect` the generated container. When you run a detached app container, it outputs the new container id, so we can do something like

```
container_id="$output"
run /bin/bash -c "docker container inspect $container_id"
```

That would help in figuring out if the container has the expected labels.

Also, run the second app container with `--trace` mode enabled, so we can see if its even hitting the logic or if the logic is just bad :)

josegonzalez:
Looks like the concurrency policy attached is `allow`, so there is something wrong in how I retrieve the policy. I'll take a look tonight.

josegonzalez:
Ah the problem is that we release the lock basically immediately due to using `--detach` in the test. I think actually we don't need a lock and can just check if there is a running container with the same `cron_id` if the policy is `forbid`.

josegonzalez:
There were a few small bugs but I think I've squashed them and the tests should pass now :)

josegonzalez:
I also need to see if we can force the concurrency for `cron:run` when using the k3s scheduler. Might need to interact with the kubernetes api for this and manually handle `forbid` and `replace`.

PR #8004, opened 2025-09-30
Buildpacks detect

Tashows:
Utilizes [https://github.com/gliderlabs/herokuish/pull/1622](https://github.com/gliderlabs/herokuish/pull/1622) to implement buildpack detection.


josegonzalez:
Mind bumping the version [here](https://github.com/dokku/dokku/blob/master/contrib/dependencies.json#L88) and adding a test?

Tashows:
Bumped the dependency, fixed the other two notes and updated the detect command to use deploy-branch, so detection is consistent with the deployment phase. Branch to detect can be overridden with a user-defined flag.

Will try to add test(s) in the next few days.

josegonzalez:
I thought about this a bit more, and I think this probably actually belongs under `builder-herokuish`. Does `pack` have something similar though?

Tashows:
Hi @josegonzalez. Sorry for not replying earlier, got caught up with another project. Will probably be able to add tests in December.

Issue #8052, opened 2025-10-17
Using Procfile command for dokku run with exec breaks command


othercorey:
### Description of problem

When using `exec` in the Dockerfile entrypoint, the command from Procfile is passed as a string and the entire line (command + arguments) is  interpreted as the executable.

### Steps to reproduce

1. Create App dockerfile with `exec` entrypoint
2. Create Procfile entry with executable + arguments
  `my-task: start-task arg1 arg2`
4. Pass Procfile entry to `dokku run` you will get a "not found" error from `exec`.
  `dokku run my-app my-task`

If you pass the same command + arguments from the Procfile entry directly to `dokku run`, it executes normally.

### dokku report $APP_NAME

N/A

### Output of failing commands after running: dokku trace:off

```shell

```

### Output of failing commands after running: dokku trace:on

```shell

```

josegonzalez:
Can you provide the entire deploy output? With trace mode on if possible (just redact any secrets of course).

othercorey:
Here is the trace output for the Procfile command `task-test: echo hello` with all env variables removed.

```
ubuntu@ip-10-2-7-240:~$ dokku run portal-test task-test
+ export DOKKU_HOST_ROOT=/home/dokku
+ DOKKU_HOST_ROOT=/home/dokku
+ export DOKKU_DISTRO
++ . /etc/os-release
++ echo ubuntu
+ DOKKU_DISTRO=ubuntu
+ export DOCKER_BIN=docker
+ DOCKER_BIN=docker
+ export DOKKU_IMAGE=gliderlabs/herokuish:latest-24
+ DOKKU_IMAGE=gliderlabs/herokuish:latest-24
+ export DOKKU_CNB_BUILDER=heroku/builder:24
+ DOKKU_CNB_BUILDER=heroku/builder:24
+ export DOKKU_LIB_ROOT=/var/lib/dokku
+ DOKKU_LIB_ROOT=/var/lib/dokku
+ export PLUGIN_PATH=/var/lib/dokku/plugins
+ PLUGIN_PATH=/var/lib/dokku/plugins
+ export PLUGIN_AVAILABLE_PATH=/var/lib/dokku/plugins/available
+ PLUGIN_AVAILABLE_PATH=/var/lib/dokku/plugins/available
+ export PLUGIN_ENABLED_PATH=/var/lib/dokku/plugins/enabled
+ PLUGIN_ENABLED_PATH=/var/lib/dokku/plugins/enabled
[...]

PR #8061, opened 2025-10-20
Procfile exec command test

othercorey:
Test case for https://github.com/dokku/dokku/issues/8052


josegonzalez:
Can you provide a Dockerfile?

othercorey:
You want to duplicate the dockerfile in docerkfile-entrypoint?

PR #7956, opened 2025-09-08
Add a Railpack builder to Dokku

josegonzalez:
Closes #7615

PR #8103, opened 2025-11-16
Correct the release version for railpacks support

josegonzalez:
Refs #7956 (thanks @pil0u)


pil0u:
You're the man. I was in the middle of assessing other PaaS alternatives with nice UIs and buttons, ended up going back to the simplicity of Dokku (with [notes](https://gist.github.com/pil0u/c48bd756326dd58f35758a5de77d9806)). It just works. Thank _you_ for that.

I will certainly try out the new Railpacks!

PR #8114, opened 2025-11-20
Ensure we can execute run commands when exec is executed as part of an entrypoint

### LINKED ISSUE #8052: Using Procfile command for dokku run with exec breaks command

josegonzalez:
Closes #8052


othercorey:
Thanks! I had trouble figuring out tests for this when I tried.

PR #8229, opened 2025-12-19
Fix command syntax for removing docker options

kleutzinger:


josegonzalez:
Thanks for the pull request!

PR #8248, opened 2025-12-27
Add `proxy-keepalive` to Nginx properties

osbre:
Continuation of https://github.com/dokku/dokku/pull/7263. Having this as a property avoids the need to customise the whole config.

- https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#no-keepalives

### DIFF
--- a/docs/networking/proxies/nginx.md
+++ b/docs/networking/proxies/nginx.md
@@ -338,7 +338,7 @@ Changing these value globally or on a per-app basis will require rebuilding the 
 | proxy-buffers             | `8 8k`                                | string  | Number and size of the buffers used for reading the proxied server response, for a single connection |
 | proxy-busy-buffers-size   | `16k`                                 | string  | Limits the total size of buffers that can be busy sending a response to the client while the response is not yet fully read. |
 | proxy-connect-timeout     | `60s`                                 | string  | Timeout (with units) for establishing a connection to your backend server           |
-| proxy-keepalive           | empty string                          | string  | Number of idle keepalive connections to upstream servers (disabled by default)      |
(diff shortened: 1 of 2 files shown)


docs/networking/proxies/nginx.md
```diff
@@ -338,6 +338,7 @@ Changing these value globally or on a per-app basis will require rebuilding the
 | proxy-buffers             | `8 8k`                                | string  | Number and size of the buffers used for reading the proxied server response, for a single connection |
 | proxy-busy-buffers-size   | `16k`                                 | string  | Limits the total size of buffers that can be busy sending a response to the client while the response is not yet fully read. |
 | proxy-connect-timeout     | `60s`                                 | string  | Timeout (with units) for establishing a connection to your backend server           |
+| proxy-keepalive           | empty string                          | string  | Number of idle keepalive connections to upstream servers (disabled by default)      |
```
osbre: Might be a good idea to mention `integer` as a type here, similarly to `hsts-max-age`?

josegonzalez:
Shame kubernetes doesn't have an annotation for this.

osbre:
@josegonzalez It seems they only have it [as a ConfigMap](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#upstream-keepalive-connections). Potentially can be done via [configuration-snippet](https://github.com/kubernetes/ingress-nginx/issues/2441#issuecomment-419714384) instead.

Meanwhile, the F5's nginx ingress offers such an annotation, unlike Kubernetes's ingress.

docs/networking/proxies/nginx.md
osbre: Updated and added validation for the set command.

PR #8402, opened 2026-03-10
Upgrade keda to 2.19.0 and keda-add-ons-http to 0.12.2

josegonzalez:
While there are breaking changes in the addon, none of these impact how Dokku exposes the addon.


josegonzalez:
This will require also checking if the chart is an older keda-http-add-ons release, deleting the deployments, then installing our chart.

PR #8509, opened 2026-04-27
Ship default catch-all site on fresh apt install

### LINKED ISSUE #7309: Nginx Default Site
### Link to the documentation page

https://dokku.com/docs/networking/proxies/nginx/#default-site

### Documentation problem

It seems like the documentation for the default site for `nginx` is broken as following the exact steps leads to `nginx` being unable to start stating the following error: 

```
Reloading A high performance web server and a reverse proxy server...
nginx: [emerg] a duplicate default server for 0.0.0.0:80 in /etc/nginx/sites-enabled/default:22
nginx.service: Control process exited, code=exited, status=1/FAILURE
Reload failed for A high performance web server and a reverse proxy server.
```

Could you help me figure out whats wrong s.t. I can change the default site behaviour and adapt the documentation accordingly? `

### dokku report

```
dokku nginx:stop
dokku nginx:start
```

josegonzalez:
Closes #7309.

Fresh apt installs of dokku now ship a catch-all server block at `/etc/nginx/conf.d/00-default-vhost.conf` that uses `ssl_reject_handshake on` (HTTPS) and `return 444` (HTTP) to drop requests with unknown Host headers, replacing the manual workaround previously documented in the nginx docs. The behavior can be opted out at install time via the new `dokku/install_default_site` debconf flag.

Conflicting upstream nginx default vhosts (`/etc/nginx/sites-enabled/default`, `/etc/nginx/sites-available/default`, `/etc/nginx/conf.d/default.conf`) are renamed to `${path}.dokku-disabled` instead of deleted so any local customizations remain recoverable. Upgrade-in-place installs leave existing nginx config untouched.

A new `dokku nginx:reload` command is added for reloading nginx config without restarting the process; the postinst uses it after writing the catch-all.


othercorey:
Is this something we should manually patch in existing installs?

josegonzalez:
If you want it, sure. It's basically updating the default nginx site. You could do that today.

palfrey:
Please note this breaks on Debian Bullseye (which is the old, old stable, but had it's [last release in 2024](https://www.debian.org/News/2024/2024083102)) because this needs Nginx at least 1.19.4 for `ssl_reject_handshake`

josegonzalez:
Good catch @palfrey. I'm working on a fix for that now as Debian Bullseye is still LTS until August 2026.

PR #8517, opened 2026-04-27
Send SIGTERM to old containers immediately on deploy

### LINKED ISSUE #6833: Send SIGTERM immediately to old containers when deploying via docker-local
### Description of feature

We should implement this: https://devcenter.heroku.com/articles/dynos#graceful-shutdown-with-sigterm

That way we aren't waiting for the retire process to occur.

josegonzalez:
## Summary

The docker-local scheduler now sends `SIGTERM` to old containers immediately after a successful deploy via `docker container kill --signal=SIGTERM`, rather than waiting `wait-to-retire` seconds before signaling. This matches Heroku's [graceful-shutdown contract](https://devcenter.heroku.com/articles/dyno-shutdown-behavior) and lets applications begin draining in-flight work as soon as proxy traffic switches. The existing `wait-to-retire` grace period and `stop-timeout-seconds` hard-stop continue to apply unchanged as the authoritative cleanup path.

Closes #6833.


othercorey:
Just curious, why isn't wait to retire started when the first SIGTERM is sent?

josegonzalez:
What do you mean by that?

othercorey:
I think I misunderstood. You're saying the grace period now starts wit a SIGTERM instead of running until the SIGTERM from docker container stop.

josegonzalez:
Yeah we now send an immediate SIGTERM but don't wait, but then background killing the container.

PR #8563, opened 2026-05-04
Source property-functions where fn-plugin-property-* helpers are used

### LINKED ISSUE #8560: fn-plugin-property-get-default: command not found
### Description of problem

After updating to version 0.38 my rails app builds are failing (the fail is specific to rails not having a secret key base), however, I wonder if it is related to this error in the logs immediately before:

```
/var/lib/dokku/core-plugins/available/common/functions: line 640: fn-plugin-property-get-default: command not found        
/var/lib/dokku/core-plugins/available/common/functions: line 641: fn-plugin-property-get-default: command not found  
-----> Releasing app-www...        
-----> Checking for predeploy task        
       No predeploy task found, skipping    
```

I'm guessing something in here is preventing the RAILS_MASTER_KEY environment variable from getting passed which then is causing the secret key base to be missing? The RAILS_MASTER_KEY is showing as set via `dokku config:show $APP`

There have been no other changes to the apps. The last successful build did not have these errors in the logs.

### Steps to reproduce

1. Update to latest dokku via `apt update -y && apt upgrade -y`
2. dokku ps:rebuild --all
3. All rails apps have failed to deploy with the error above. Non rails apps deploy, but also include the error.

### dokku report $APP_NAME

n/a

### Additional information

Installed via Ubuntu 24.04.4 LTS packages.

dokku version 0.38.0

All apps are using Dockerfile for build.

Maybe related? https://github.com/dokku/dokku/commit/1308e219470e52a9c6d6f6d18c404af0f4ea304c

```shell

```

```shell

```

josegonzalez:
## Summary

The `release_and_deploy` function and several proxy plugin scripts call `fn-plugin-property-*` helpers without sourcing `plugins/common/property-functions`, producing `command not found` errors during deploys. Sourcing the helpers from `plugins/common/functions` and from each proxy plugin entry point that consumes them resolves issue #8560.

closes #8560

### DIFF
--- a/plugins/caddy-vhosts/command-functions
+++ b/plugins/caddy-vhosts/command-functions
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/property-functions"
 source "$PLUGIN_AVAILABLE_PATH/proxy/command-functions"
 source "$PLUGIN_AVAILABLE_PATH/caddy-vhosts/internal-functions"
 set -eo pipefail
--- a/plugins/caddy-vhosts/pre-restore
+++ b/plugins/caddy-vhosts/pre-restore
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 source "$PLUGIN_CORE_AVAILABLE_PATH/common/functions"
+source "$PLUGIN_CORE_AVAILABLE_PATH/common/property-functions"
 source "$PLUGIN_AVAILABLE_PATH/caddy-vhosts/command-functions"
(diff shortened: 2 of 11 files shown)


othercorey:
If it was being called, is there a missing test for the functionality? I see a new test to verify source works, but what about the feature that was breaking with the missing call?

josegonzalez:
The log messages were related to these two lines, which don't actually impact deploys:

```
local DOKKU_APP_SKIP_DEPLOY="$(fn-plugin-property-get-default "ps" "$APP" "skip-deploy" "")"
local DOKKU_GLOBAL_SKIP_DEPLOY="$(fn-plugin-property-get-default "ps" "--global" "skip-deploy" "")"
```

There is a test for injected environment variables, but at least these warnings should go away.

PR #8595, opened 2026-05-10
Resolve nginx upstreams via coredns-docker

### LINKED ISSUE #8313: Wrong vhost being served by nginx during startup and upgrades (2026)
### Description of problem

Hello.

On Ubuntu 24.04, with 0.36.0 packages from packagecloud, after a reboot the web content of an incorrect container is sometimes served.

Seen: Web content for application x served on vhost/domain for site y.
Expected: 500 until the correct container is up and running.

This was also reported in https://github.com/dokku/dokku/issues/6192 and https://github.com/dokku/dokku/issues/3932 .

As far as non-private use of dokku, this is quite troubling. We cannot serve data for the wrong customer during routine patch&reboot operations.

### Steps to reproduce

As far as I can tell, the flow is:

* After reboot, nginx and docker starts early (1-2s after init starts)
* Containers starts to come up, and nginx.conf (incl. /home/dokku/*/nginx.conf ) files on disk are stale. Nginx starts serving the wrong applications.
* systemd service dokku-redeploy starts running. In our instance it takes 60-65 seconds to run though the 18 deployed apps.
* after dokku-redeploy has completed, the problem is corrected.

### dokku report $APP_NAME

Please let me know what in this extensive report would be needed to debug this issue. 

### Additional information

_No response_

```shell

```

```shell

```

josegonzalez:
Stops nginx from routing to stale container IPs after host reboots, Docker upgrades without live-restore, or runtime container restarts. The bundled `nginx.conf.sigil` now uses an nginx `resolver` directive against the local `coredns-docker` socket and a variable-based `proxy_pass` keyed on `<app>.<process-type>.docker`, so the upstream is resolved at request time honoring DNS TTL. `coredns-docker` becomes a hard `Depends:` of the dokku deb and is bundled in the Docker image under runit. Operators can revert to the legacy IP-based upstream block with `dokku nginx:set --global dns-resolver off`.

Closes #8313.


othercorey:
What error is returned when the docker container isn't available?

josegonzalez:
So coredns-docker is going to be a system service running under systemd/runit. If it isn't available, don't recall what nginx does as it is resolving the dns on the fly for that proxy_pass implementation. I'll add that to the list to test/verify.

That said, if the Docker daemon goes away, coredns-docker is implemented such that you'll be served cached responses until it comes back, so we survive a daemon upgrade with live-upgrades as well as one that rebuilds the containers.

pedroborges:
My app just went because of this very issue, Docker restarted after failure on new IPs.

PR #8633, opened 2026-05-13
Correctly redirect http traffic to https when using scheduler-k3s

Mordred:
I'm setuping dokku with scheduler-k3s, but redirect from http to https is not currently working. Based on my research, you cannot have both entrypoints (web and websecure) in one IngressRoute when using `tls`.

When using `tls` on `web` entrypoint, Traefik exepcts TLS handshake. Because only plain HTTP will be received, it won't be matched and 404 will be returned.

We need two IngressRoute configs in this case.

### DIFF
--- a/plugins/scheduler-k3s/ingress_route_template_test.go
+++ b/plugins/scheduler-k3s/ingress_route_template_test.go
@@ -1,0 +1,188 @@
+package scheduler_k3s
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/engine"
+)
+
+func renderIngressRouteTemplate(t *testing.T, values map[string]interface{}) []map[string]interface{} {
+	t.Helper()
+
+	chartDir := t.TempDir()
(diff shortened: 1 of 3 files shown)


josegonzalez:
Thanks for this. Mind splitting the template into two separate files? One for web and one for websecure?

Mordred:
Sure, but in my opinion is more readable in one template. Also if you disable tls, only one config should be generated and in that case, second template will be empty file - which seems strange to me.

josegonzalez:
Can you combine the templates better then? ATM it's mostly copy-paste behind an if statement, might be easier to scope that if statement instead to change the couple places that need to change.

Mordred:
OK, I have simplified the change. Now it will generate each spec in a loop.

Currently one resource is `myapp-web-http-80-5000` and second is `myapp-web-http-80-5000-websecure`. Maybe we add `-web` to the first too? It would remove one `if` from the template :)

But it is a naming change - but I don't think it will break something.

josegonzalez:
Mind adding a bats test for this as well for integration testing?

Mordred:
I cannot do that. I have tried to run your Vagrant setup for almost 6 hours with no luck. Tried virtualbox, vmware_fusion, utm or qemu. Always ended with some errors related to arm architecture (M1 cpu), macos NFS version problem or failed provisioning.

If you guide me how to correctly setup test environment for MacBook M1, I can try to write the test, but currently I'm unable to run your test environment setup.

Screenshot 2026-05-15 at 15 00 52

josegonzalez:
I normally set it up in a devcontainer, but don't know if k3s works in that.

I'll get to adding tests for this soon.

PR #8640, opened 2026-05-13
Split report global keys into raw and computed

### LINKED ISSUE #8631: caddy/haproxy/traefik :report --global lacks raw key for global properties
The `:report --global --format json` output for `caddy`, `haproxy`, and `traefik` exposes each global property as a single `global-<property>` key (or bare `<property>` for the caddy/haproxy/traefik schemas) that returns the **default value** after the property is unset, not empty. There is no separate raw key, so external tooling cannot distinguish "set globally to the default" from "not set globally".

This is the same class of bug as #8610 (git deploy-branch/keep-git-dir), #8611 (scheduler-docker-local init-process), and #8626 (caddy per-app tls-internal) - all of which were fixed by adding a raw key alongside the computed/global value. Those fixes addressed the per-app side; the global side of caddy/haproxy/traefik still has the same shape.

### Reproduction (caddy, dokku 0.38.5)

```shell
$ dokku --version
dokku version 0.38.5

$ dokku --quiet caddy:report --global --format json | jq .
{
  "global-tls-internal": "false",
  "image": "lucaslorentz/caddy-docker-proxy:2.12",
  "letsencrypt-email": "",
  "letsencrypt-server": "https://acme-staging-v02.api.letsencrypt.org/directory",
  "log-level": "ERROR",
  "polling-interval": "5s"
}

$ dokku caddy:set --global image custom/caddy
=====> Setting image to custom/caddy

$ dokku --quiet caddy:report --global --format json | jq .image
"custom/caddy"

$ dokku caddy:set --global image
=====> Unsetting image

$ dokku --quiet caddy:report --global --format json | jq .image
"lucaslorentz/caddy-docker-proxy:2.12"
```

After unset, `image` returns `lucaslorentz/caddy-docker-proxy:2.12` (the default), not empty. Same pattern for `caddy global-tls-internal` (returns `"false"`), `caddy log-level` (returns `"ERROR"`), `caddy polling-interval` (returns `"5s"`), and the equivalent properties in `haproxy:report --global --format json` and `traefik:report --global --format json`.

### Expected

Match the convention used by `nginx:report --global`, `git:report --global`, and `caddy:report <app>` post-#8626: add a raw global key that returns empty when explicitly unset, alongside the existing default-style key. For example for caddy:

- `global-image`: raw (empty when unset)
- `computed-global-image`: effective (default when raw is empty)

Or equivalent naming consistent with the existing scheme.

This unblocks idempotency probes that need to detect "this property has been removed globally" without external state.

### Affected plugins/scopes

- `caddy:report --global` - all properties (`image`, `letsencrypt-email`, `letsencrypt-server`, `log-level`, `polling-interval`, `tls-internal`)
- `haproxy:report --global` - all properties (`image`, `letsencrypt-email`, `letsencrypt-server`, `log-level`, `refresh-conf`)
- `traefik:report --global` - all properties (verified by symmetry with caddy/haproxy)

### LINKED ISSUE #8636: git:set --global keep-git-dir works but key missing from git:report --global --format json
## Description

`dokku git:set --global keep-git-dir true` is accepted as a valid command, but the value cannot be read back from `dokku git:report --global --format json`. The companion `deploy-branch` key was added in #8610; `keep-git-dir` was not.

## Reproduction (dokku 0.38.5)

```
# dokku git:set --global keep-git-dir true
-----> Setting keep-git-dir to true

# dokku git:report --global --format json | jq .
{
  "global-archive-max-files": "30",
  "global-archive-max-size": "30",
  "global-deploy-branch": ""
}
```

`global-keep-git-dir` is not in the payload despite the property being settable globally.

## Expected

After `git:set --global keep-git-dir true`, the JSON payload exposes `global-keep-git-dir` (returns `"true"`) plus optionally `computed-global-keep-git-dir` matching the per-app pattern. Per #8610 the per-app side already exposes both `keep-git-dir` and `computed-keep-git-dir`.

## Why this matters

External tooling (e.g. dokku/docket, ansible-dokku) wants to probe whether a property has been set so apply runs can be idempotent. With `keep-git-dir` missing from the global report, the probe falls through to an unconditional `:set` on every run.

Companion to #8610. The internal-functions `cmd-git-report-single` flag map already exposes `--git-keep-git-dir` and `--git-global-keep-git-dir` for per-app reports, but the global flag map omits it.

josegonzalez:
The `<plugin>-global-<property>` keys in `:report --global --format json` returned the resolved value with the built-in default substituted in, so external tooling could not distinguish a property that had been set globally to the default from one that had never been set. After `dokku <plugin>:set --global <property>` followed by an unset, the JSON still reported the default value, masking the unset state from idempotency probes that rely solely on the report.

The bare per-app keys for caddy, haproxy, and traefik global-only properties (`caddy-image`, `haproxy-log-level`, `traefik-api-vhost`, etc.) had the same shape - they exposed the resolved value with no separate raw key.

The `global-<property>` keys now hold the raw stored value (empty when nothing has been set) and a new `computed-<property>` key holds the effective value used at runtime, falling back through the per-app value (where one exists), the global value, and the built-in default. The bare global-only keys for the three proxy plugins are removed in favor of the raw/computed pair. The same convention is applied across `caddy-vhosts`, `haproxy-vhosts`, `traefik-vhosts`, `app-json`, `builder-dockerfile`, `builder-herokuish`, `builder-lambda`, `builder-nixpacks`, `builder-pack`, `builder-railpack`, `checks`, `git`, `logs`, `network`, `proxy`, `scheduler`, and `scheduler-docker-local`.

Internal callers that previously read the `global-` helper expecting a default-resolved value (template renderers, deploy paths, the proxy type safety net) now call the `computed-` helper, so deploy-time behavior is unchanged. This is a breaking change for downstream tooling that read `<plugin>-global-<property>` and relied on the default value; consumers should switch to `<plugin>-computed-<property>`. The migration guide for 0.38.0 documents the key shape and migration path.

Closes #8631.
Closes #8636.


plugins/git/internal-functions
```diff
@@ -369,19 +370,27 @@ cmd-git-report-single() {
...
+      "--git-global-archive-max-files: $(fn-plugin-property-get "git" "--global" "archive-max-files" "")"
+      "--git-global-archive-max-size: $(fn-plugin-property-get "git" "--global" "archive-max-size" "")"
+      "--git-global-deploy-branch: $(fn-plugin-property-get "git" "--global" "deploy-branch" "")"
+      "--git-global-keep-git-dir: $(fn-plugin-property-get "git" "--global" "keep-git-dir" "")"
     )
   else
     verify_app_name "$APP"
     flag_map=(
+      "--git-computed-archive-max-files: $(fn-archive-max-files)"
+      "--git-computed-archive-max-size: $(fn-archive-max-size)"
```
josegonzalez: We should rename these to have the word computed in them - same for the other computed ones.

plugins/git/internal-functions
```diff
@@ -369,19 +370,27 @@ cmd-git-report-single() {
...
       "--git-computed-deploy-branch: $(fn-git-deploy-branch "$APP")"
       "--git-computed-keep-git-dir: $(fn-git-keep-git-dir "$APP")"
       "--git-deploy-branch: $(fn-plugin-property-get "git" "$APP" "deploy-branch" "")"
```
josegonzalez: This should have the word `computed` in the key.

plugins/logs/report.go
```diff
@@ -20,11 +20,13 @@ func ReportSingleApp(appName string, format string, infoFlag string) error {
...
-			"--logs-global-vector-sink":     reportGlobalVectorSink,
-			"--logs-vector-global-image":    reportVectorGlobalImage,
-			"--logs-vector-global-networks": reportVectorGlobalNetworks,
+			"--logs-computed-app-label-alias": reportComputedAppLabelAlias,
+			"--logs-computed-max-size":        reportComputedMaxSize,
+			"--logs-global-app-label-alias":   reportGlobalAppLabelAlias,
+			"--logs-global-max-size":          reportGlobalMaxSize,
+			"--logs-global-vector-sink":       reportGlobalVectorSink,
+			"--logs-vector-global-image":      reportVectorGlobalImage,
+			"--logs-vector-global-networks":   reportVectorGlobalNetworks,
```
josegonzalez: The word `global` should come before `vector`. Also, I think we need computed versions for each that the internal code respects?

plugins/network/report.go
```diff
@@ -17,11 +17,12 @@ func ReportSingleApp(appName string, format string, infoFlag string) error {
...
-			"--network-global-attach-post-deploy":  reportGlobalAttachPostDeploy,
-			"--network-global-bind-all-interfaces": reportGlobalBindAllInterfaces,
-			"--network-global-initial-network":     reportGlobalInitialNetwork,
-			"--network-global-tld":                 reportGlobalTld,
+			"--network-computed-bind-all-interfaces": reportComputedBindAllInterfaces,
+			"--network-global-attach-post-create":    reportGlobalAttachPostCreate,
+			"--network-global-attach-post-deploy":    reportGlobalAttachPostDeploy,
+			"--network-global-bind-all-interfaces":   reportGlobalBindAllInterfaces,
+			"--network-global-initial-network":       reportGlobalInitialNetwork,
+			"--network-global-tld":                   reportGlobalTld,
```
josegonzalez, line 29: We need computed values for each of these, and the code should use the computed values.

tests/unit/checks.bats
```diff
@@ -41,44 +41,67 @@ teardown() {
...
+  echo "status: $status"
+  assert_success
+
+  run /bin/bash -c "dokku --quiet checks:report --global --format json | jq -e ."
   echo "output: $output"
   echo "status: $status"
   assert_success
 
-  run /bin/bash -c "dokku checks:report --global --format json | jq -e ."
+  run /bin/bash -c "dokku --quiet checks:report --global --format json | jq -r '.\"global-wait-to-retire\"'"
```
josegonzalez: Why did we add quiet here?

tests/unit/checks.bats
```diff
@@ -41,44 +41,67 @@ teardown() {
...
+  assert_success
+
+  run /bin/bash -c "dokku --quiet checks:report --global --format json | jq -r '.\"global-wait-to-retire\"'"
   echo "output: $output"
   echo "status: $status"
   assert_success
   assert_output "90"
 
-  run /bin/bash -c "dokku checks:report --global --format json | jq -r 'has(\"wait-to-retire\")'"
+  run /bin/bash -c "dokku --quiet checks:report --global --format json | jq -r '.\"computed-wait-to-retire\"'"
```
josegonzalez: Same for the rest of these, why did we add `--quiet`?

PR #8664, opened 2026-05-24
Convert filesystem migration markers to plugin properties

### LINKED ISSUE #8656: Convert filesystem migration markers to plugin properties
Several plugin migrations track completion via filesystem flag files instead of plugin properties, which makes the marker invisible to `:report` and breaks per-property idempotency probes.

### Current filesystem markers

| Plugin | Marker | Path | Writer |
|---|---|---|---|
| docker-options | `DOCKER_OPTIONS_<PHASE>.migrated` | `$DOKKU_ROOT/<app>/DOCKER_OPTIONS_<PHASE>.migrated` | `plugins/docker-options/functions.go:280,317` |
| storage | `migrations/<app>` | `$DOKKU_LIB_ROOT/data/storage/migrations/<app>` | `plugins/storage/migrate.go:20-26` |

Both serve the same purpose as the property-based sentinels `config/env-migrated` (set in `plugins/config/triggers.go:113,157,179`), `docker-options/migrated-from-files` (set in `plugins/docker-options/functions.go:259,289`), and `nginx-vhosts/nginx-conf-sigil-migrated` (set in `plugins/nginx-vhosts/install:19`) - but they live on disk rather than in the property store, so they cannot be inspected via `:report` and do not survive a property-store backup/restore that ignores the flag-file paths.

### Proposed change

1. Replace each filesystem marker with a plugin property written via `common.PropertyWrite` (or the shell equivalent). Suggested names:
   - `docker-options/migrated-<phase>` (per-app), one per phase the plugin currently writes a `.migrated` file for.
   - `storage/legacy-mounts-migrated` (per-app), single boolean.
2. Only write the property when the original legacy value was actually found and migrated. The current code writes the marker unconditionally at install time; the new code should:
   - Inspect the source (legacy flat file, legacy `-v` docker-options entry, etc.).
   - Skip the property write if there is nothing to migrate.
   - Write the property only after a successful migration of a non-empty source.
3. Read the property at the top of every migration entry point and short-circuit when it is `\"true\"`, mirroring the current file-marker check.
4. Keep one upgrade cycle where the migration code still recognizes the old filesystem markers (treat their presence as equivalent to the new property being set), then delete them after writing the property. This avoids re-running the migration on installs that already advanced past it.

### Why this matters

- `:report` becomes the single source of truth for what a plugin has done at install time. External tooling currently cannot tell whether the docker-options or storage migration ran without shelling into the dokku host and stat'ing files.
- The property store is what gets dumped/restored by backup tooling. Operators who restore the property store but lose the data directory currently lose the migration markers and get a re-run on the next install, which - depending on whether the source still exists - can either be a no-op or a duplicate migration.
- Aligns the docker-options and storage plugins with the existing pattern used by `config/env-migrated`, `docker-options/migrated-from-files`, and `nginx-vhosts/nginx-conf-sigil-migrated`.

### Acceptance criteria

- `dokku docker-options:report <app>` and `dokku storage:report <app>` (or their global variants) expose the migration markers via the existing property store.
- The migration code only writes the new property if the legacy source actually existed before the migration started.
- Existing installations that have the old filesystem markers continue to short-circuit correctly on the next install (no double migration).
- bats coverage in `tests/unit/docker-options*.bats` and `tests/unit/storage*.bats` asserts the property is set after a forced migration and that a re-run of the migration with the property already set is a no-op.

josegonzalez:
The docker-options `DOCKER_OPTIONS_<PHASE>.migrated` sentinels and the storage `data/storage-registry/migrations/<app>` flag files are now stored as per-app properties (`migrated-<phase>` and `legacy-mounts-migrated` respectively), so they survive a property-store backup/restore and align with the existing `migrated-from-files`, `env-migrated`, and `nginx-conf-sigil-migrated` patterns.

The new markers are written only when a non-empty legacy source was actually migrated, so apps that never had legacy state are distinguishable from apps that did and were drained. The storage marker is per-app and written by `migrateApp` only when `-v` lines were drained; the docker-options markers are per-app per-phase and written only when the legacy file contained non-comment/whitespace content.

A one-cycle upgrade-cycle conversion drains any leftover filesystem sentinels from the previous release into the new properties and removes the files. For docker-options the conversion runs BEFORE the global `migrated-from-files` short-circuit so installs upgrading from the previous release - where the global flag is already set AND `.migrated` sentinels are on disk - still get those sentinels drained into the per-phase properties.

Both new properties are documented under the `Internal properties` table in their respective plugin docs and are not surfaced via `:report`, matching the convention used by `migrated-from-files`, `env-migrated`, and `nginx-conf-sigil-migrated`.

Closes #8656.

### DIFF
--- a/plugins/docker-options/functions.go
+++ b/plugins/docker-options/functions.go
@@ -245,20 +245,64 @@ func CommandReport(appName string, format string, infoFlag string) error {
 	return ReportSingleApp(appName, format, infoFlag)
 }
 
-// migrateLegacyDockerOptionsFiles converts pre-properties DOCKER_OPTIONS_*
-// files into property lists. It is gated by a single global marker so it never
-// re-runs - even if a user later restores a DOCKER_OPTIONS_* file by hand.
-func migrateLegacyDockerOptionsFiles() error {
-	if common.PropertyExists("docker-options", "--global", "migrated-from-files") {
+// migratedPropertyKey returns the per-app property name recording that
+// an app's DOCKER_OPTIONS_<PHASE> file was drained into the property
+// store. Surfaces in the property store as `docker-options/<app>/migrated-<phase>`.
+func migratedPropertyKey(phase string) string {
+	return "migrated-" + strings.ToLower(phase)
+}
+
+// convertLegacyMigratedMarker drains a leftover DOCKER_OPTIONS_<PHASE>.migrated
+// sentinel into the new per-phase property and removes the file. Runs
+// as part of the upgrade-cycle pass; no-op when the file is absent.
+// TODO(post-deprecation): remove this helper and its caller.
+func convertLegacyMigratedMarker(appName, phase string) error {
+	migratedPath := filepath.Join(common.AppRoot(appName), "DOCKER_OPTIONS_"+strings.ToUpper(phase)+".migrated")
+	if !common.FileExists(migratedPath) {
 		return nil
 	}
+	if err := common.PropertyWrite("docker-options", appName, migratedPropertyKey(phase), "true"); err != nil {
+		return err
+	}
+	return os.Remove(migratedPath)
+}
 
+// migrateLegacyDockerOptionsFiles converts pre-properties
+// DOCKER_OPTIONS_* files into property lists. Idempotency comes from
+// two layers: a per-phase per-app `migrated-<phase>` property recording
+// what was drained, and a global `migrated-from-files` short-circuit so
+// the steady-state install path returns immediately.
+//
+// The upgrade-cycle conversion of leftover `.migrated` filesystem
+// sentinels runs BEFORE the global gate so users upgrading from the
+// previous release - who have `migrated-from-files` set AND `.migrated`
+// files on disk - still get those files drained into properties.
+func migrateLegacyDockerOptionsFiles() error {
 	apps, err := common.DokkuApps()
 	if err != nil {
 		if errors.Is(err, common.NoAppsExist) {
-			return common.PropertyWrite("docker-options", "--global", "migrated-from-files", "true")
+			if !common.PropertyExists("docker-options", "--global", "migrated-from-files") {
+				return common.PropertyWrite("docker-options", "--global", "migrated-from-files", "true")
+			}
+			return nil
 		}
 		return err
+	}
+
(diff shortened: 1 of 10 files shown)


plugins/docker-options/functions.go
```diff
@@ -245,22 +245,66 @@ func CommandReport(appName string, format string, infoFlag string) error {
...
+			if err := convertLegacyMigratedMarker(appName, phase); err != nil {
+				return err
+			}
+		}
+	}
+
+	if common.PropertyExists("docker-options", "--global", "migrated-from-files") {
+		return nil
+	}
+
```
josegonzalez: Perform this check at the start. Also actually read the value and check if it is equal to `true`.

plugins/docker-options/functions.go
```diff
@@ -245,22 +245,66 @@ func CommandReport(appName string, format string, infoFlag string) error {
...
+// previous release - who have `migrated-from-files` set AND `.migrated`
+// files on disk - still get those files drained into properties.
+func migrateLegacyDockerOptionsFiles() error {
 	apps, err := common.DokkuApps()
 	if err != nil {
 		if errors.Is(err, common.NoAppsExist) {
-			return common.PropertyWrite("docker-options", "--global", "migrated-from-files", "true")
+			if !common.PropertyExists("docker-options", "--global", "migrated-from-files") {
+				return common.PropertyWrite("docker-options", "--global", "migrated-from-files", "true")
+			}
```
josegonzalez: Just write the file, no need to check for existence.

plugins/docker-options/functions.go
```diff
@@ -312,6 +359,7 @@ func readLegacyOptionsFile(path string) ([]string, error) {
 // removeMigratedLegacyFiles deletes any leftover DOCKER_OPTIONS_*.migrated
 // files for an app. Called from post-delete so we don't leak files into a
 // directory that is about to be torn down anyway.
+// TODO(post-deprecation): remove this helper and the call in TriggerPostDelete.
```
josegonzalez: Remove this helper and the call now, but also make sure we delete properties for the plugin in TriggerPostDelete, and also handle migrating properties in the TriggerPostRename trigger like we do other plugins.

PR #8751, opened 2026-06-10
docs: add SECURITY.md

mayank-dev-15:
Added SECURITY.md with vulnerability reporting guidelines following GitHub best practices.


josegonzalez:
We already have one that covers the entire organization available [here](https://github.com/dokku/.github/blob/master/SECURITY.md). Thanks for the contribution though.

PR #8752, opened 2026-06-10
docs: add CODE_OF_CONDUCT.md

mayank-dev-15:
Added Contributor Covenant Code of Conduct to foster a welcoming and inclusive community.


josegonzalez:
We already have one that covers the entire organization [here](https://github.com/dokku/.github/blob/master/CODE_OF_CONDUCT.md). Thanks for the contribution though.

PR #8763, opened 2026-06-18
Add centralized backup and restore mechanism

### LINKED ISSUE #5008: Add a centralized backup and restore mechanism
## Description of problem

As a dokku admin, if I want to export a backup of an entire application and all of the dependencies, I need to know how that application works, whether or not the plugins that provide the backing services offer backup commands, etc. It would be much better if there were a mechanism by which an application and all of the attached services can be told to export whatever they need to export.

## Possible Solution

I think a very lightweight solution could be put in place here: we just need six new plugn triggers:

* pre-backup
* backup
* post-backup
* pre-restore
* restore
* post-restore

The only thing Dokku needs to handle is dispatching those triggers + providing a directory that each plugin should write stuff to or read stuff from (as well as backing up/restoring core data, of course -- from what I can see, this is already handled in other commands so it would just need wired up).

Example:

If I'm running an application that was pushed to Git with a MySQL backing service and a persistent storage directory, here's what I'm imagining would happen when I trigger a backup:

* Dokku creates `/tmp/backup/app-name-[timestamp]/`
* Trigger pre-backup
* Trigger backup
	  * Dokku exports a copy of the application code from the git repo to `/tmp/backup/app-name-[timestamp]/code.tar.gz`
	  * Dokku exports a list of domains to `/tmp/backup/app-name-[timestamp]/domains.txt`
	  * Dokku exports a snapshot of the persistent storage dir to `/tmp/backup/app-name-[timestamp]/some-meaningful-name.tar.gz`
	  * (repeat for all core plugins)
	  * The MySQL plugin is asked to export a backup for app `[app-name]` into dir `/tmp/backup/app-name-[timestamp]`
	  		* Under the hood, this could use `mysqldump` or `mysqlhotcopy` or whatever. The implementation details don't matter and will vary by plugin.
* Trigger post-backup
	* (This is important for e.g. a third party plugin to be able to ship backup data off to S3 or whatever)

When I ask to restore something, the reverse should happen. If I have an S3 backups plugin or w/e, it should be responsible for downloading the backup that I've specified and extracting it into the location where Dokku expects the backup data to be available. Each plugin is responsible for reading its own data from the backup dir.

## Aside

I _feel_ like this existed at some point and that I'm reproducing it (or something like it) from memory, but maybe I'm remembering wrong? Is there some reason that the core product _should not_ provide this kind of plumbing? Backup and restore seems really important for a PaaS.

josegonzalez:
Adds backup:export and backup:import commands that archive and restore apps, datastore services, and the global configuration as a single tarball. Each plugin contributes and reapplies its own state through new backup plugin triggers, services are discovered with a new datastore-list trigger alongside the existing service-list trigger, persistent storage volume data can be bundled with --include-storage, and access to apps and services is enforced through the existing user-auth triggers.

Closes #5008.


othercorey:
Will there be a dependency order where plugins like let's encrypt will restore after the domains? Will all internal plugins restore first regardless?

josegonzalez:
Somewhat. The ordering is guaranteed by the import phases plus the deploy running last. An import reinstalls recorded plugins first, then restores global state, then services, then each app. For a single app the sequence is:

```
app-create
pre-backup-app-import
backup-app-import            # independent config: domains, config, ports, certs, storage, scale, ...
post-backup-app-import   # config that depends on another plugin's restored config
receive-app                         # the deploy: build, run, regenerate proxy, normal post-deploy hooks
```

Each phase is a separate blocking broadcast, so post-backup-app-import always runs after every backup-app-import handler (including domains). The restores for plugins generally should just write files into place, and not take any further action. And because the deploy (receive-app) is the last step, all config (including domains and certs) is restored before the app is built.

In the case of letsencrypt, once it implements the contract, it would probably write the files out/symlink them as appropriate, so you'd be back in a position where letsencrypt is available prior to the app being deployed.

othercorey:
> In the case of letsencrypt, once it implements the contract, it would probably write the files out/symlink them as appropriate, so you'd be back in a position where letsencrypt is available prior to the app being deployed.

I think those event states makes sense. it sounds like we expect lets-encrypt backup to have valid certs on restore. That probably makes sense unless you only backup when changing an app config - and the app doesn't change very often.

I suppose the user is responsible for validating the apps after restoring and manually re-enabling let's encrypt if the cert is that old.

josegonzalez:
I'll think more about how this should work. Maybe a hook that happens prior to the backup deploy...

othercorey:
@josegonzalez For documentation purposes, the advantage of this style backup over normal tarball (https://dokku.com/docs/advanced-usage/backup-recovery/?h=backup#creating-a-backup) is the ability to restore plugins like databases?

Issue #7000, opened 2024-07-19
Support for Ubuntu 24.04


dmmeteo:
### Description of feature

Currently, I receive the following message when I try to install Dokku on the new Ubuntu server:

```
Unsupported Linux distribution. Only the following versions are supported: Debian [10, 11, 12], Ubuntu [18.04, 20.04, 22.04]
```

Is there a plan to extend support to Ubuntu 24.04? This would help in maintaining up-to-date server environments and ensure compatibility with the latest Ubuntu version. Thank you for considering this feature request.

HectorMalot:
There is an ongoing issue already on Ubuntu 24.04 support: https://github.com/dokku/dokku/issues/6856 

Almost there. Herokuish needs to be released for 24.04 and then dokku can support 24.04 as well.

josegonzalez:
Closing as duplicate.

PR #8791, opened 2026-07-03
Add Ubuntu 26.04 LTS to bootstrap supported versions

### LINKED ISSUE #8768: Support for Ubuntu 26.04
### Description

The bootstrap installer (`contrib/bootstrap.sh`) does not yet recognize Ubuntu 26.04 LTS, so a fresh install fails before any package work happens:

```
Unsupported Linux distribution. Only the following versions are supported: Debian [11, 12, 13], Ubuntu [22.04, 24.04]
```

This is the same kind of whitelist gap that was handled for previous releases (e.g. #7000 for 24.04).

### Where it comes from

`install-dokku-from-deb-package()` hard-gates the allowed versions:

```bash
if ! in-array "$DOKKU_DISTRO_VERSION" "22.04" "24.04" "10" "11" "12" "13"; then
  log-fail "Unsupported Linux distribution. Only the following versions are supported: $SUPPORTED_VERSIONS"
fi
```

and the apt-repo codename list only knows `jammy`/`noble`:

```bash
OS_IDS=("jammy" "noble")
```

26.04's `VERSION_ID` (`26.04`) and its codename both miss these lists.

### Workaround

Adding `26.04` to the version check is enough to get past the gate — the existing codename fallback already pins unknown Ubuntu codenames to the `noble` (24.04) packagecloud repo, which installs cleanly on 26.04 in testing. A proper fix would also add the 26.04 codename to `OS_IDS` (and publish a matching packagecloud component when available).

### Environment

- Dokku version: v0.36.10 (also reproduces against the latest bootstrap, v0.38.19)
- OS: Ubuntu 26.04 LTS

### Request

Add Ubuntu 26.04 LTS to the supported versions in `contrib/bootstrap.sh` (version whitelist + codename list), mirroring how 24.04 was added.

akhil-naidu:
## Summary

Fixes #8768. The bootstrap installer (`bootstrap.sh`) rejected Ubuntu 26.04 LTS before any package work happened, because `install-dokku-from-deb-package()` hard-gated the allowed versions to `22.04`/`24.04` (plus Debian):

```
Unsupported Linux distribution. Only the following versions are supported: Debian [11, 12, 13], Ubuntu [22.04, 24.04]
```

This mirrors how 24.04 was added in #7000.

## Changes

- Add `26.04` to the version whitelist in `install-dokku-from-deb-package()`.
- Add `26.04` to the `SUPPORTED_VERSIONS` message.
- Update the header comment to mention 26.04.

## Notes

- Ubuntu 26.04's codename is `resolute` (Resolute Raccoon). I intentionally did **not** add it to `OS_IDS`, because there is no matching packagecloud component published yet — adding it would make 26.04 point at a nonexistent repo. The existing codename fallback already pins unknown Ubuntu codenames to the `noble` (24.04) packagecloud repo, which installs cleanly on 26.04 per the issue.
- Follow-up for maintainers: once a `resolute` packagecloud component is published, add the codename to the Ubuntu `OS_IDS` list.

## Test plan

- [ ] Confirm a fresh install on Ubuntu 26.04 LTS passes the version gate and installs via the `noble` packagecloud repo.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -3,7 +3,7 @@ set -eo pipefail
 [[ $TRACE ]] && set -x
 
 # A script to bootstrap dokku.
-# It expects to be run on Ubuntu 22.04/24.04 via 'sudo`
+# It expects to be run on Ubuntu 22.04/24.04/26.04 via 'sudo`
 # If installing a tag higher than 0.3.13, it may install dokku via a package (so long as the package is higher than 0.3.13)
 # It checks out the dokku source code from GitHub into ~/dokku and then runs 'make install' from dokku source.
 
@@ -12,7 +12,7 @@ set -eo pipefail
 # That's good because it prevents our output overlapping with wget's.
 # It also means that we can't run a partially downloaded script.
 
-SUPPORTED_VERSIONS="Debian [11, 12, 13], Ubuntu [22.04, 24.04]"
+SUPPORTED_VERSIONS="Debian [11, 12, 13], Ubuntu [22.04, 24.04, 26.04]"
 
 log-fail() {
   declare desc="log fail formatter"
@@ -181,7 +181,7 @@ install-dokku-from-deb-package() {
   local NO_INSTALL_RECOMMENDS=${DOKKU_NO_INSTALL_RECOMMENDS:=""}
   local OS_ID
 
-  if ! in-array "$DOKKU_DISTRO_VERSION" "22.04" "24.04" "10" "11" "12" "13"; then
+  if ! in-array "$DOKKU_DISTRO_VERSION" "22.04" "24.04" "26.04" "10" "11" "12" "13"; then
     log-fail "Unsupported Linux distribution. Only the following versions are supported: $SUPPORTED_VERSIONS"
   fi
 


usualdesigner:
Ran the one unchecked box in the test plan on a fresh box, plus a real deploy on top of it.

**Environment** — GCP `e2-standard-2`, Ubuntu 26.04.1 LTS (`VERSION_CODENAME=resolute`, `lsb_release -cs` → `resolute`), kernel `7.0.0-1011-gcp`, no Docker and no Dokku preinstalled. Installed with `DOKKU_TAG=v0.38.19`.

**Control run** — unpatched `bootstrap.sh` from `master` on the same image:

```
Unsupported Linux distribution. Only the following versions are supported: Debian [11, 12, 13], Ubuntu [22.04, 24.04]
exit 1
```

Nothing was installed — the gate fires before the Docker install.

**Patched run** — this PR's `bootstrap.sh` (sha256 `276828dcea8fffb88d0a141e0b164af0ff21b4e810888aacb45370d4ae820002`). Gate passed, and the two things that could have gone wrong without `OS_IDS` changes both held:

```
deb https://packagecloud.io/dokku/dokku/ubuntu/ noble main
deb [arch=amd64 signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu resolute stable
Docker version 29.8.0, build 88096ef
```

The `noble` fallback resolves as intended, and `get.docker.com` does publish a `resolute` repo, so no Docker-side change is needed either. The `noble` packages install cleanly on 26.04:

```
dokku 0.38.19   herokuish 0.11.17   plugn 0.17.1
gliderlabs-sigil 0.12.1   dokku-event-listener 0.20.1   dokku-update 0.10.0
nginx 1.28.3-2ubuntu1.10
```

**Beyond the test plan** — a passing gate isn't the same as a working install, so I deployed something real:

```
dokku ssh-keys:add ...
dokku apps:create smoke
dokku domains:set smoke <ip>.sslip.io
git push dokku main          # heroku/node-js-getting-started
```

herokuish detected the Node.js app, built it, and the deploy finished with `All checks successful (web.1)`. `curl http://<ip>.sslip.io/` returns **200** with the sample app's page. After a hard reset of the instance, `docker`, `nginx` and `dokku-event-listener` all came back active and the app container restarted on its own — `curl` still 200.

One thing worth noting so nobody else chases it: immediately after install `dokku-event-listener` reads `inactive`. That is the package's own postinst behaviour (it `enable`s `dokku-event-listener.target` without starting it) and it resolves on the first boot — confirmed by the reboot above. It reproduces independently of this PR and is not 26.04-specific.

So from my side:

- [x] Confirm a fresh install on Ubuntu 26.04 LTS passes the version gate and installs via the noble packagecloud repo.

Happy to re-run anything specific if a maintainer wants a different tag or a Debian 13 control.

Joelius300:
Would also love to see this merged, lmk if I can test anything.

PR #8805, opened 2026-07-07
Allow custom values for chown

RichardDorian:
This PR adds the ability to set custom values when creating a persistent storage.

This works by trying to parse the argument value as a a u16. If it fails, it will throw the usual error.

Context: I want to deploy rustfs and their image runs as `10001:10001`. The unability to define a custom chown value results in "Permission denied" at runtime.

### DIFF
--- a/docs/advanced-usage/persistent-storage.md
+++ b/docs/advanced-usage/persistent-storage.md
@@ -106,6 +106,8 @@ By default, permissions are set for usage with Herokuish buildpacks. These permi
 - `--chown root`: Use `0:0` as the folder permissions.
     - This is used for containers that run their processes as root, as is typical for most Dockerfile or Docker image deploys.
 - `--chown false`: Skips the `chown` call.
+- `--chown <uid>`: Use `<uid>:<uid>` as the folder permissions, where `<uid>` is a custom numeric user/group id.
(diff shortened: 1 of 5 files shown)


josegonzalez:
Mind adding a unit test for this?

RichardDorian:
Good idea, I followed the same pattern as the test just above the one I made.

josegonzalez:
@RichardDorian seems the tests fail, mind investigating?

RichardDorian:
I missed the `chown-storage-dir` script the first time. If I understand this well, this script runs with elevated privileges. Since I want to allow every ID possibility I reduced the check from the allow list to a simple positive and less than `165536`. Does this seem like a security concern to you?

RichardDorian:
The packer workflow failure seems to be unrelated:

```
* api_token for auth must be specified

  on contrib/images/digitalocean/packer.pkr.hcl line 14:
  (source code not available)
```

josegonzalez:
That seems fine. Can you also add tests for the chown being out of bounds? We should both gate that in the chown-storage-dir script as well as in the go code that calls it.

RichardDorian:
Not sure if you meant adding go tests as well so I did them. My english is a bit rusty.

PR #8856, opened 2026-07-21
Add pre-parsed port_mappings to ports:report json

### LINKED ISSUE #8847: Add pre-parsed port mappings to ports JSON output
### Description

When port mappings are requested as JSON, they are returned as a single space-delimited string rather than a structured value, so consumers have to re-parse the string to recover the individual fields.

`ports:report --format json` returns the configured map under `ports-map` (and the detected map under `ports-map-detected`) as one space-delimited string of `scheme:host-port:container-port` tokens:

```json
{
  "ports-map": "http:80:5000 https:443:5000",
  "ports-map-detected": "http:80:5000"
}
```

To use these values a consumer has to split on whitespace, split each token on `:`, and separately handle the empty-map case (`"ports-map": ""`). `ports:list` does not currently accept `--format json` at all.

### Proposal

Add a pre-parsed representation of the port mappings to the JSON output of either `ports:list --format json` or `ports:report --format json` - for example an array of objects with the scheme, host port, and container port already split out and the ports as integers:

```json
[
  { "scheme": "http", "host_port": 80, "container_port": 5000 },
  { "scheme": "https", "host_port": 443, "container_port": 5000 }
]
```

The existing string fields can remain unchanged for backwards compatibility.

### Motivation

Tools and automation built on top of the JSON output could consume the port mappings directly, without each consumer re-implementing the whitespace/colon parsing and empty-map handling.

Observed on dokku 0.38.23.

youdie006:
Adds `--ports-map-json` and `--ports-map-detected-json` to `ports:report`, returning port mappings as `[{scheme, host_port, container_port}, ...]`. Existing string flags are unchanged.
Also documents the flags and covers them in the ports bats suite.

Closes #8847

### DIFF
--- a/docs/networking/port-management.md
+++ b/docs/networking/port-management.md
@@ -221,7 +221,60 @@ You can pass flags which will output only the value of the specific information 
 dokku ports:report node-js-app --ports-map
 ```
 
+Use `--ports-map-json` or `--ports-map-detected-json` to get the same data as a JSON array of objects. Each object has the following fields:
+
+- `scheme`: the port scheme (e.g. `http`, `https`, `udp`)
+- `host_port`: the host/listener port
+- `container_port`: the container port
+
+```shell
+dokku ports:report node-js-app --ports-map-json
+```
+
+```json
+[
+  {
+    "scheme": "http",
+    "host_port": 80,
+    "container_port": 5000
+  },
+  {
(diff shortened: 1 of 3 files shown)


josegonzalez:
I updated the PR to require fewer new features and also add tests/docs.

PR #8889, opened 2026-08-04
Redirect app.json cron task output to host log file(s)

### LINKED ISSUE #8890: Feature request: redirect app.json cron task output to host log file(s)
# Feature request: redirect app.json cron task output to host log file(s)

## Summary

Today an `app.json` cron task renders in the `dokku` user's crontab as a bare
`dokku cron:run <app> <id>` line with no output redirection. The only place a
task's stdout/stderr goes is the `dokku` user's cron mail (via `MAILTO`). There
is no supported way to persist a scheduled task's output to a file on the host.

This is awkward because:

- The `command` string is tokenized and exec'd directly in the container, and
  bare shell operators (`>`, `>>`, `2>&1`, ...) are explicitly rejected at
  `app.json` validation time. So users cannot self-redirect inside `command`.
- `MAILTO` delivers *everything* as mail, which is noisy for chatty tasks and
  useless if no MTA is configured.

Notably, the internal/global cron path already supports a log file: the
`cron-entries` trigger accepts an optional 3rd `;`-separated field that populates
`CronTask.LogFile`, rendered as `<cmd> &>> <logfile>`. App-defined tasks have no
equivalent.

## Proposed feature

Add three optional per-task fields to the `app.json` cron schema:

- `logfile` — merged stdout+stderr appended to one host path (`... &>> <path>`)
- `stdout_logfile` — stdout appended to a host path (`... >> <path>`)
- `stderr_logfile` — stderr appended to a host path (`... 2>> <path>`)

Rules:

- `logfile` is mutually exclusive with `stdout_logfile`/`stderr_logfile`.
- Any subset of `stdout_logfile`/`stderr_logfile` may be set.
- Each path is validated (absolute path, conservative safe character set) so it
  cannot break the crontab line or inject commands, since the path is
  interpolated into a shell line cron runs via `/bin/bash`.

Rendered crontab examples:

| app.json fields | rendered line |
|---|---|
| `logfile: /var/log/dokku/x.log` | `<sched> dokku cron:run app id &>> /var/log/dokku/x.log` |
| `stdout_logfile: /a.log` | `<sched> dokku cron:run app id >> /a.log` |
| `stderr_logfile: /b.log` | `<sched> dokku cron:run app id 2>> /b.log` |
| both separate | `<sched> dokku cron:run app id >> /a.log 2>> /b.log` |

## Open question for maintainers — trust model / configuration surface

There are two plausible homes for this configuration, and I'd like your
preference before finalizing:

1. **`app.json` fields (what this proposal implements).** Log paths travel with
   the app's repo. This is consistent with how `command`/`schedule`/
   `concurrency_policy` are already defined, and keeps a task's log destination
   next to the task. The tradeoff: the log path is *host* state controlled by
   whoever can push the repo, so a path validator is required (implemented) to
   keep it from becoming a crontab-injection vector. The paths still can't
   escalate beyond what the `dokku` user can already write, but they are
   repo-authored host paths.

2. **A `cron:set` property (host-operator-controlled).** Log destinations would
   be set by the server operator per app rather than by the repo. This keeps
   host paths entirely out of the deployable artifact, at the cost of being less
   discoverable and not co-located with the task definition (and cron IDs are
   derived from command+schedule, so per-task targeting via `cron:set` is
   clunky).

I've implemented option (1) with strict path validation because it mirrors the
existing schema and the existing internal `LogFile` mechanism, but I'm happy to
rework it as a `cron:set` property (or support both) if that better matches the
project's trust model. What's your preference?

An implementation (with tests and docs) is up as a PR at #8889 — happy to rework
the configuration surface there based on your direction.

l3x4:
# Redirect app.json cron task output to host log file(s)

Fixes #8890

## What

Adds three optional per-task fields to the `app.json` `cron` schema so a
scheduled task can append its output to log file(s) on the host:

- `logfile` — merged stdout+stderr appended to one host path
- `stdout_logfile` — stdout appended to a host path
- `stderr_logfile` — stderr appended to a host path

## Why

Previously an `app.json` cron task rendered as a bare
`dokku cron:run <app> <id>` line with no output redirection — output only
reached the `dokku` user's cron mail (`MAILTO`). Users could not redirect it
themselves either: `command` is tokenized/exec'd directly and bare shell
operators are rejected at validation time. The internal/global cron path already
supported a log file (`CronTask.LogFile`, rendered `&>> <path>`); this brings the
same capability to app-defined tasks, with separate-stream and merged options.

## How

- `plugins/app-json/appjson.go`: adds `logfile`, `stdout_logfile`,
  `stderr_logfile` to the app.json `CronTask` schema.
- `plugins/cron/cron.go`:
  - `CronTask` gains `StdoutLogFile`/`StderrLogFile` (`json:"-"`) alongside the
    existing `LogFile`.
  - `DokkuRunCommand()` now appends redirection to the `dokku cron:run` branch
    as well as the existing `AltCommand` branch. Merged `LogFile` keeps its
    existing `&>> <path>` rendering (back-compat); separate streams render
    `>> <stdout>` and/or `2>> <stderr>`.
  - `FetchCronTasks()` wires the new fields through, enforces mutual exclusion of
    `logfile` with `stdout_logfile`/`stderr_logfile`, and validates every log
    path. As with the existing command/schedule validation, `WarnToFailure`
    turns violations into deploy errors; otherwise the task is skipped with a
    warning.
  - New exported `ValidateCronLogPath()` rejects anything that is not an
    absolute path made of a conservative safe character set (letters, digits,
    `. _ - /`). This is a security guard: the path is interpolated into a
    crontab line cron executes via `/bin/bash`, so whitespace or shell
    metacharacters (`; | & $ \` ( ) < >`, newlines, quotes, globs) must not reach
    it.

### Rendered crontab lines

| app.json | rendered |
|---|---|
| `logfile: /var/log/dokku/x.log` | `dokku cron:run app id &>> /var/log/dokku/x.log` |
| `stdout_logfile: /a.log` | `dokku cron:run app id >> /a.log` |
| `stderr_logfile: /b.log` | `dokku cron:run app id 2>> /b.log` |
| both separate | `dokku cron:run app id >> /a.log 2>> /b.log` |

## Tests

Added Go tests in `plugins/cron/cron_test.go` covering: stdout-only, stderr-only,
both-separate, and merged rendering; the preserved back-compat rendering of the
internal `AltCommand` + `LogFile` case; `ValidateCronLogPath` accept/reject
tables; and `FetchCronTasks` wiring, mutual-exclusion (both fail-hard and skip
modes), and unsafe-path rejection (both modes).

```
go test ./plugins/cron/... ./plugins/app-json/...   # ok
gofmt -l <changed files>                             # empty
git diff --check                                     # clean
```

## Docs

`docs/processes/scheduled-cron-tasks.md` documents the three new properties, a
JSON example each for the merged and separate-stream forms, the merge semantics,
the mutual-exclusion rule, and the path-validation restriction.

## Open question for maintainers — configuration surface

I implemented this as **`app.json` fields** because it mirrors the existing
`command`/`schedule`/`concurrency_policy` fields and the existing internal
`CronTask.LogFile` mechanism. The tradeoff is that the log path is *host* state
authored by whoever can push the repo — hence the strict `ValidateCronLogPath`
guard (the paths still can't write anywhere the `dokku` user couldn't already,
but they are repo-authored host paths).

The alternative is a **`cron:set` property** (host-operator-controlled), which
keeps host paths out of the deployable artifact but is less discoverable, not
co-located with the task, and awkward to target per-task (cron IDs are derived
from command+schedule). I'm happy to rework this as a property (or support both)
if that better matches the project's trust model — just let me know your
preference. I can also open a tracking issue first if you'd rather discuss there.

## Notes

- The existing internal-task rendering (`&>> <logfile>`) is unchanged, so the
  global/injected cron path and its test are unaffected.
- Log paths are host paths; parent directories must already exist and be
  writable by the `dokku` user. Output is appended.

### DIFF
--- a/docs/processes/scheduled-cron-tasks.md
+++ b/docs/processes/scheduled-cron-tasks.md
@@ -39,6 +39,51 @@ A cron task takes the following properties:
 - `maintenance`: A boolean value that decides whether the cron task is in maintenance and therefore executable or not.
 - `schedule`: A [cron-compatible](https://en.wikipedia.org/wiki/Cron#Overview) scheduling definition upon which to run the command. Seconds are generally not supported.
 - `concurrency_policy`: A string (default: `allow`), that controls whether the cron task can be run concurrently with another invocation of itself. Valid options are `allow` (allow concurrency), `forbid` (exit the new cron task if there is an existing one), `replace` (delete any existing cron task and start the new one).
+- `logfile`: A host path to which the task's **merged** standard output and standard error are appended. Mutually exclusive with `stdout_logfile`/`stderr_logfile`.
+- `stdout_logfile`: A host path to which the task's standard output is appended.
+- `stderr_logfile`: A host path to which the task's standard error is appended.
+
+#### Redirecting task output to a log file
+
+By default a scheduled cron task produces no on-host log file - its output is only delivered to the `dokku` user's cron mail (see the `MAILTO`/`MAILFROM` settings below). To persist output to a file on the host instead, set one or more of the `logfile`, `stdout_logfile`, and `stderr_logfile` properties. Paths are on the **host** (not inside the container) and output is **appended** to them; the file is created by cron on first write, but any parent directories must already exist and be writable by the `dokku` user.
+
+Use `logfile` to send merged standard output and standard error to a single file:
+
+```json
+{
+  "cron": [
+    {
+      "command": "npm run send-email",
+      "schedule": "@daily",
+      "logfile": "/var/log/dokku/send-email.log"
+    }
+  ]
+}
+```
+
+This renders in the crontab as the equivalent of `... &>> /var/log/dokku/send-email.log`.
+
+Alternatively, redirect standard output and standard error to separate files. Either may be set independently:
+
+```json
+{
+  "cron": [
+    {
+      "command": "npm run send-email",
+      "schedule": "@daily",
+      "stdout_logfile": "/var/log/dokku/send-email.out.log",
+      "stderr_logfile": "/var/log/dokku/send-email.err.log"
+    }
+  ]
+}
+```
+
+This renders as `... >> /var/log/dokku/send-email.out.log 2>> /var/log/dokku/send-email.err.log`.
+
+> [!NOTE]
+> `logfile` cannot be combined with `stdout_logfile` or `stderr_logfile`; doing so fails validation at deploy time.
+>
+> Because the log path is interpolated directly into a crontab line executed via `/bin/bash`, each log path is validated at deploy time and must be an absolute path composed only of letters, digits, and the characters `.`, `_`, `-`, and `/`. Paths containing whitespace or shell metacharacters (such as `;`, `|`, `&`, `$`, `` ` ``, `(`, `)`, `<`, `>`, quotes, or glob characters) are rejected to prevent breaking the crontab or injecting commands.
 
 
 Zero or more cron tasks can be specified per app. Cron tasks are validated after the build artifact is created but before the app is deployed, and the cron schedule is updated during the post-deploy phase.
--- a/plugins/app-json/appjson.go
+++ b/plugins/app-json/appjson.go
@@ -115,6 +115,16 @@ type CronTask struct {
 
 	// ConcurrencyPolicy is the concurrency policy for the cron command
 	ConcurrencyPolicy string `json:"concurrency_policy"`
+
+	// LogFile is a host path to which merged stdout and stderr are appended.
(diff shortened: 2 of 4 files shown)


josegonzalez:
I think this should actually be handled with Vector, not with this method. The reason we support logging paths for external plugins is that they don't necessarily run in containers - specifically, the datastore backup and letsencrypt renewal run on the host.

josegonzalez:
Closing in favor of #8914, which provides the "correct" way to do this (using vector). Thanks for the initial report though, its a good improvement!

PR #8914, opened 2026-08-09
Add vector-cron-sink for scheduled cron task output

### LINKED ISSUE #8890: Feature request: redirect app.json cron task output to host log file(s)

josegonzalez:
Scheduled cron task output previously reached only the `dokku` user's cron mail, and could not be redirected because `app.json` rejects bare shell operators in a cron `command`. Setting `vector-cron-sink` on an app or globally routes that output to a dedicated sink instead, on both the `docker-local` and `k3s` schedulers, which keeps log destinations under operator control rather than in a deployed repository. Cron events carry `dokku_app` and `dokku_cron_id` fields so a sink can give each task its own destination.

This also fixes a `k3s` bug where configuring a global `vector-sink` silently removed the vector prometheus exporter sink.

Closes #8890

### DIFF
--- a/docs/deployment/logs.md
+++ b/docs/deployment/logs.md
@@ -124,6 +124,21 @@ The `/etc/vector` mount includes the `vector.json` configuration file, but also 
 
 The final volume mount - `/var/log/dokku/apps` - may be used for users that wish to ship logs to a file on disk that may be later logrotated. This directory is owned by the `dokku` user and group, with permissions set to `0755`. At this time, log-rotation is not configured for this directory.
 
+Operators using a `file` sink are encouraged to configure rotation themselves, as Dokku will not truncate these files. A minimal `/etc/logrotate.d/dokku-app-logs` might look like:
+
+```
+/var/log/dokku/apps/*/*.log {
+  daily
+  rotate 14
+  compress
+  missingok
+  notifempty
+  copytruncate
+}
+```
+
+`copytruncate` is used because Vector holds the file open between writes.
+
 #### Stopping the Vector container
 
 Vector may be stopped via the `logs:vector-stop` command.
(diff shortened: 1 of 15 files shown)


josegonzalez:
The reason I'd prefer this method is that it is compatible with our direction around how regular app logs should be persisted, so should also apply to cron tasks.

PR #9011, opened 2026-09-03
Propagate command exit code on dokku run

### LINKED ISSUE #9002: scheduler-k3s: `dokku run` does not propagate the container's exit code
Under the k3s scheduler, `dokku run` exits 1 for any failure rather than reporting the command's own exit status. Under docker-local it reports the real one, so the same command gives a different answer depending on the app's scheduler.

`plugins/scheduler-docker-local/scheduler-run` waits for the container and returns its status:

```bash
"$DOCKER_BIN" container start --attach "$CONTAINER_ID" || EXIT_CODE=$?
DOKKU_CONTAINER_EXIT_CODE="$("$DOCKER_BIN" container wait "$CONTAINER_ID" 2>/dev/null || echo "$EXIT_CODE")"
...
return "$DOKKU_CONTAINER_EXIT_CODE"
```

`TriggerSchedulerRun` in `plugins/scheduler-k3s/triggers.go` has the number in hand and discards it. All three `PodFailed` paths reach the terminated container status and read only `.Message` (lines 1250, 1286, and 1317):

```go
if status.State.Terminated == nil {
    continue
}
return fmt.Errorf("Unable to attach as the pod has already exited with a failed exit code: %s", status.State.Terminated.Message)
```

That error reaches `common.LogFailWithError`, which exits 1 for a plain error. So `dokku run <app> sh -c 'exit 3'` exits 3 on docker-local and 1 on k3s.

This matters for anything that branches on the status — a CI step, a deploy script — and for anything that records it. A caller that captures the exit code of a one-off command can only report success or failure on k3s. The `.Message` reported in its place is usually empty, so a failure comes back with neither a code nor a reason.

The mechanism for fixing it already exists. `common.LogFailWithError` honours `common.ErrWithExitCode`:

```go
if errExit, ok := err.(ErrWithExitCode); ok {
    os.Exit(errExit.ExitCode())
}
os.Exit(1)
```

and there is precedent for implementing it in-tree: `RetireLockFailed` in `plugins/ps/ps.go` does exactly this to propagate 137. An error type in scheduler-k3s carrying `status.State.Terminated.ExitCode` at those three returns would line the two schedulers up.

Two things worth settling alongside it: whether the attached/TTY path (`enterPod` → `ExecCommand`) already propagates a status or needs the same treatment, and whether the termination `.Message` is still worth carrying in the error text once the code itself is there.

`run:detached` is unaffected — it prints the pod name and returns nil.

Noticed while adding one-off command support to Dokku Pro, where a run's exit code is recorded against the background job that started it. The value is exact on docker-local and success/failure only on k3s, which is currently a documented caveat rather than something that can be fixed on that side.

visheshgubrani:
### Problem

Under the k3s scheduler, `dokku run` exits 1 for any failing command instead of reporting the command's own exit status. `plugins/scheduler-docker-local/scheduler-run` waits on the container and returns its exit code, so the same command gives a different answer depending on the app's scheduler. `TriggerSchedulerRun` has the number in hand and discards it: all three `PodFailed` paths reach the terminated container status and read only `.Message`, which is empty for a plain command failure, so a failure comes back with neither a code nor a reason.

1. **Non-TTY runs (the common CI case).** The run pod's logs are streamed, the pod ends in `PodFailed`, and the error is built from the empty `.Message`, exiting 1.

2. **TTY runs of short commands.** The pod reaches a terminal phase while dokku waits for it to be running, the logs are streamed, and the same error path exits 1.

3. **TTY runs that complete before the attach pre-flight.** The pod is observed running and then terminal before the exec attach starts; the same error path exits 1.

The attached/TTY path loses the code separately. The exec stream returns `utilexec.CodeExitError`, which carries the code but does not satisfy `common.ErrWithExitCode`, so `common.LogFailWithError` exits 1 there too. Long-running interactive `dokku run` sessions and `dokku enter` both report 1 for any failing command, while docker-local reports the real status through `docker container exec`.

### Solution

- **Add an `ExitCodeError` type implementing `common.ErrWithExitCode`.** `common.LogFailWithError` already honors that interface and exits with `ExitCode()`, with `RetireLockFailed` in the ps plugin as the in-tree precedent. The error carries the code plus the termination message, and the message is only included when non-empty, so the failure text keeps its reason when one exists instead of printing a dangling colon.

- **Route the three `PodFailed` paths through one helper.** `failedRunContainerError` keeps the existing container-name matching logic and returns `Terminated.ExitCode`; when no terminated container with a nonzero exit code exists it falls back to the previous generic error, so `PodFailed` states without a usable code still exit 1 rather than pretending success.

- **Convert the exec-stream error at the `enterPod` boundary.** `attachExitError` maps `utilexec.CodeExitError` (the same type `storage:exec` already extracts) onto `ExitCodeError` with unchanged error text, so interactive `dokku run` sessions and `dokku enter` exit with the remote command's status. Scoping the conversion to `enterPod` leaves `storage:exec`'s own exit handling and the proxy-log exec reads untouched.

- **Tests.** Go unit tests for the error contract, the container-status helper (nonzero exit code with and without message, zero exit code, missing and mismatched containers, unrelated terminated containers), and the exec adapter (nonzero, zero, plain error, nil), plus bats tests in `tests/unit/run_2.bats` (docker-local control) and `tests/unit/scheduler-k3s-herokuish.bats` (k3s reproduction) asserting `dokku run ... sh -c 'exit 3'` reports status 3 and `exit 0` still succeeds.

Closes #9002

### DIFF
--- a/plugins/scheduler-k3s/functions.go
+++ b/plugins/scheduler-k3s/functions.go
@@ -36,6 +36,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilexec "k8s.io/client-go/util/exec"
 	"k8s.io/kubernetes/pkg/client/conditions"
 	"k8s.io/utils/ptr"
 	"mvdan.cc/sh/v3/shell"
@@ -409,6 +410,24 @@ func createKubernetesNamespace(ctx context.Context, namespaceName string) error 
 	return nil
 }
 
+// attachExitError surfaces the exit code of the remote command from a
+// kubernetes exec stream as an ExitCodeError, so that attached sessions
+// (dokku run in interactive mode, dokku enter) exit with the command's own
+// status rather than a generic failure, matching docker-local scheduler
+// behavior. Other errors pass through unchanged.
+func attachExitError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	var coded utilexec.CodeExitError
+	if errors.As(err, &coded) && coded.Code > 0 {
+		return &ExitCodeError{Code: coded.Code, Message: coded.Error()}
+	}
+
+	return err
+}
+
 func enterPod(ctx context.Context, input EnterPodInput) error {
 	labelSelector := []string{}
 	for k, v := range input.SelectedPod.Labels {
@@ -440,13 +459,13 @@ func enterPod(ctx context.Context, input EnterPodInput) error {
 		return fmt.Errorf("No container specified and no default container found")
 	}
 
-	return input.Clientset.ExecCommand(ctx, ExecCommandInput{
+	return attachExitError(input.Clientset.ExecCommand(ctx, ExecCommandInput{
 		Command:       input.Command,
 		ContainerName: input.SelectedContainerName,
 		Entrypoint:    input.Entrypoint,
 		Name:          input.SelectedPod.Name,
 		Namespace:     input.SelectedPod.Namespace,
-	})
+	}))
 }
 
 func extractStartCommand(input StartCommandInput) string {
--- a/plugins/scheduler-k3s/functions_test.go
+++ b/plugins/scheduler-k3s/functions_test.go
@@ -1,11 +1,14 @@
 package scheduler_k3s
 
 import (
+	"errors"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


josegonzalez:
Run tests fail.

visheshgubrani:
Fixed in the latest push. Well the Bats test was previously deploying the default `python` (herokuish) app, whose internal `/exec` wrapper was stripping quotes from `sh -c 'exit 3'` and turning it into `sh -c exit 3` (which exited 0). Switched the test to use `deploy_app dockerfile` so the command is passed directly to the container and rebased onto `master`

